PR #52, opened 2021-07-09
Specify cohort definition

benbc:


tests/test_end_to_end.py
```diff
@@ -1,27 +1,52 @@
 import csv
 import shutil
+from pathlib import Path
 
 import pytest
 
 
+class Study:
+    def __init__(self, study_path):
+        super().__init__()
```
inglesp, line 10: nit: I don't think there's any need for `super()` here.

cohortextractor/main.py
```diff
@@ -1,31 +1,25 @@
...
     write_output(results, workspace)
 
 
-def load_cohort(workspace):
-    sys.path.append(workspace)
-    study_definition = importlib.import_module("study_definition")
+def load_cohort(definition):
+    spec = importlib.util.spec_from_file_location("cohort_definition", definition)
+    cohort_definition = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(cohort_definition)
```
inglesp, line 19: This is using a fairly esoteric corner of the standard library -- could you add a comment to explain what's going on?

cohortextractor/main.py
```diff
@@ -1,31 +1,25 @@
...
 from .backends import BACKENDS
 from .query_utils import get_column_definitions
 
 
-def main(workspace="/workspace", backend_id=None, db_url=None):
-    if not db_url:
-        db_url = os.environ["TPP_DATABASE_URL"]
-    if not backend_id:
-        backend_id = os.environ["BACKEND"]
+def main(workspace, definition, backend_id, db_url):
```
inglesp, line 9: Would names like `workspace_path` and `definition_path` make it clearer what this function accepts?

inglesp (approved):
All fine, a couple of minor comments/suggestions.

tests/test_end_to_end.py
benbc: No, you're quite right. Boilerplate IDE-generated code. Thanks.

cohortextractor/main.py
benbc: As you can imagine, what's going on is code being copied from Stackoverflow. Clarifying function and comment added. I'm actually going to remove this in my next PR anyway.

cohortextractor/main.py
benbc: Are you including "path" there to indicate the _type_ of the thing (a `Path`), or because it's the "path of" the workspace and definition that's being passed in? I could get behind the latter (especially for the definition), but if we feel we need type annotations then I think we should be using the real thing rather than mangling the variable names.

benbc: How would you feel about `workspace_dir` and `definition_path`?

PR #53, opened 2021-07-09
Fix temporary table name

evansd:
The `#` prefix on the table name makes it a session-scoped local
temporary table:
https://stackoverflow.com/questions/2920836/local-and-global-temporary-tables-in-sql-server

For bonus fun, because of this naming scheme SQL Server folk sometimes
refer to these as "hash tables".


rebkwok (approved):
It's failing tests, but it's just the one that tests the literal sql string, and I got rid of that somewhere else already (the in/not in PR maybe)

evansd:
Going to close this as it's addressed elsewhere and fixing the tests is faffy

PR #57, opened 2021-07-12
Add directory containing cohort definition to path

benbc:
This allows researchers to factor out code from the cohort definition
into local library modules, as demonstrated by the end-to-end test.

### DIFF
--- a/cohortextractor/main.py
+++ b/cohortextractor/main.py
@@ -1,6 +1,8 @@
 import csv
 import importlib.util
 import inspect
+import sys
+from contextlib import contextmanager
 
 from .backends import BACKENDS
 from .query_utils import get_column_definitions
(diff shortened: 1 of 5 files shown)


cohortextractor/main.py
```diff
@@ -13,24 +14,26 @@ def main(definition_path, output_file, backend_id, db_url):
...
-    cohort_definition = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(cohort_definition)
-    return cohort_definition
+def load_module(definition_path):
+    definition_dir = definition_path.parent
+    module_name = definition_path.stem
+
+    # Add the directory containing the definition to the path so that the definition can import library modules from
+    # that directory
+    sys.path.append(str(definition_dir))
```
evansd, line 34: Is it worth restoring the original `sys.path` value after the import? It's unlikely to cause us trouble, but it's always possible that a user might add a local module which shadows something we later want to import.
 
Genuine non-rhetorical question! The answer might be that it's not worth worrying about.

evansd (approved):
Looks good!

cohortextractor/main.py
evansd: Oh yes, nice!

benbc: No. This doesn't work. Once something has been imported it's cached indefinitely. So removing the directory from the path doesn't have any effect (except for modules which are defined there but never loaded, which seems like a pretty extreme edge case) and prepending doesn't fix the problem that I had seen before.

I'm going to leave in the removal, because it just seems like the right thing to do from a semantic point of view, though.

PR #59, opened 2021-07-12
Add the most basic categorisation

rebkwok:


cohortextractor/query_language.py
```diff
@@ -142,3 +148,17 @@ def __init__(self, source, function, column):
...
+    return ValueFromCategory(mapping, default)
+
+
+CategoryGroup = namedtuple("CategoryGroup", ("source", "operator", "value"))
+
+
+class ValueFromCategory(Value):
+    def __init__(self, groups, default):
+        self.default = default
+        self.source = tuple({group.source for group in groups.values()})
```
evansd, line 163: Do we need `source` as a separate thing here? It feels like it's already captured in `definitions`. We could teach `walk_query_dag` to check the `definitions` dict.

Or maybe another approach would be to add a `get_query_node_references()` method which you can all on any QueryNode instance to get a list of the other query nodes it references. That keeps the logic on the QueryNode subclasses rather than in `walk_query_dag` which is probably the right place for it.

cohortextractor/query_language.py
```diff
@@ -1,7 +1,28 @@
...
+    "is_in": "in_",
+    "not_in": "not_in",
+}
+
+
 def table(name):
     return Table(name)
 
 
+def category_group(source, comparison, value):
```
evansd, line 22: I guess we can get rid of this once we add operator overloading and then the API becomes even simpler.

cohortextractor/query_engines/mssql.py
```diff
@@ -261,15 +274,34 @@ def get_population_table_query(self, population):
...
+
+    def get_case_expression(self, category_node):
+        category_definitions = category_node.definitions.copy()
+        statements = []
+        tables_used = set()
+        for label, category_definition in category_definitions.items():
+            source_col = category_definition.source
+            table = self.output_group_tables[self.get_type_and_source(source_col)]
+            tables_used.add(table)
+            column = table.c[source_col.column]
```
evansd, line 299: This section here is effectively duplicating the logic in `get_value_expression` and I think we should be calling out to that instead. That means we can support arbitrary nesting of expressions.

evansd (approved):
This does look on the right track. There was one specific bit that didn't make sense to me, but beyond that I think everything else is stuff we can iteratively improve on as we add support for more kinds of expressions and for operator overloading.

cohortextractor/query_engines/mssql.py
```diff
@@ -123,17 +124,27 @@ def walk_query_dag(self, nodes, seen=None):
             seen.update(nodes)
         parents = set()
         for node in nodes:
-            yield node
+            if self.is_output_node(node):
```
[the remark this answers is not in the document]
rebkwok: Hmm. Yes, I think so.  Actually, I think  I meant to take the conditional on line 92 out, and the fact that I didn't is probably why it still works in the tests

cohortextractor/query_language.py
rebkwok: I had put it in the query engine at one point and then I took it out to keep it with the `ValuefromCategory`, and as an attribute to make it work with query engine methods that expected a source.  But yes, a query node method would work too and be more explicit

cohortextractor/query_engines/mssql.py
rebkwok: Sorry, not sure I understand.  This is called from `get_value_expression`, and it's only the `table=` and `column=` lines that are duplicated, and they're done for each  category definition in turn.  It's only in its own method to break up the `get_value_expression` method.  I could put it all in `get_value_expression` instead, something like this:
```
    def get_value_expression(self, value):
        tables = None
        value_expr = value
        if self.is_category_node(value):
            category_definitions = value.definitions.copy()
            statements = []
            tables = set()
            for label, category_definition in category_definitions.items():
                source_col = category_definition.source
                table = self.output_group_tables[self.get_type_and_source(source_col)]
                column = table.c[source_col.column]
                tables.add(table)
                method = getattr(column, category_definition.operator)
                statements.append((method(category_definition.value), label))
            value_expr = sqlalchemy.case(*statements, else_=value.default)            
        elif self.is_output_node(value):
            table = self.output_group_tables[self.get_type_and_source(value)]
            column = self.get_output_column_name(value)
            value_expr = table.c[column]
            tables = table,
        return value_expr, tables
```
but I'm not sure where I can easily remove duplication

evansd: Ah right yeah, this is me jumping ahead slightly sorry. `category_definition` here is actually just a boolean value: there's nothing particularly category-specific about it. So the translation of `category_definition` into a SQLAlchemy expression is something that can be done inside `get_value_expression`. Then `get_case_expression` becomes a method that takes a mapping of boolean expressions to categories and builds a case expression out of it. What I'm trying to get at is that `get_case_expression` doesn't, and shouldn't, need to "know" anything about to handle a `category_definition`.

If that still doesn't make any sense (which wouldn't surprise me!) then feel free to merge as is and we can work on refactoring later.

rebkwok: Not sure I understand still, but I've refactored a bit :shrug:

rebkwok:
I'm going to merge this because I have another branch that tries to do the operator overloading and will refactor this anyway

PR #60, opened 2021-07-13
Minimal TPP backend

benbc:
This is just enough of a TPP backend to be worthy of the name.


tests/test_tpp_backend.py
```diff
@@ -0,0 +1,33 @@
...
+from lib.tpp_schema import Base, Events, Patient, RegistrationHistory
+
+from cohortextractor import table
+from cohortextractor.backends.tpp import TPPBackend
+
+
+@pytest.fixture
+def setup_tpp_database(setup_test_database):
+    def setup(data):
+        setup_test_database(data, base=Base)
```
rebkwok, line 12: Can we not just use the `setup_test_database` fixture directly?  Does this separate fixture do any more than just avoid passing `base=Base` to `setup_test_database`?

rebkwok (approved):
LGTM, just one trivial comment. Feel free to ignore.

tests/test_tpp_backend.py
benbc: Yes, that's all it does. It seems worth it to me to simplify the test code -- so that all the detail in the test relates to the specific details of the test. This seems like a difference of taste. Can I persuade you?

rebkwok: Yes, sure.  I can see how it will help when there are lots of tpp_backend tests.

PR #62, opened 2021-07-14
Unify test setup

benbc:


rebkwok (approved):
Removes a nice chunk of code from the tests :)

PR #64, opened 2021-07-14
Overload operators for category comparison

rebkwok:
This overloads the comparison operators on a Value, so you can do `foo > 100` and return a Comparator object which the query engine can turn into a case statement. (>, >=, <, <=, ==, !=)

It also overloads the `__and__` and `__or__` methods on the `Comparator` so that it can combine Comparators.  This is mostly working; it can deal with e.g. `(foo > 100) & (bar = "blah")` and with nesting comparisons with brackets e.g. `((foo > 100) & (bar = "blah")) | (foobar < 10)`.  Between shorthand also works (`50 < foo < 100`).

It doesn't deal with True/False comparisons properly yet - you can't do `(foo > 100) & bar`.   and `(foo > 100) & (bar == True)` on the results of an `exists()` query (a `ValueFromAggregate`) seemed like it should work but returned odd results so I abandoned it for the next iteration.

It also doesn't do `~` yet.


cohortextractor/query_language.py
```diff
@@ -19,8 +16,47 @@ def table(name):
     return Table(name)
 
 
-def category_group(source, comparison, value):
-    return CategoryGroup(source, _OPERATOR_MAPPING[comparison], value)
+class Comparator:
+    """A generic comparator to represent a comparison between a source object and a value"""
```
rebkwok, line 20: Based on a very simplified version of django `Q` 
https://github.com/django/django/blob/main/django/db/models/query_utils.py#L29

cohortextractor/query_language.py
```diff
@@ -19,8 +16,47 @@ def table(name):
...
+    def __init__(
+        self, children=None, connector="and_", source=None, operator=None, value=None
+    ):
+        """
+        Construct a new Comparator.
+        A single comparator will have a source, operator and value.  A tree of Compararors
+        will have at most two child Comparators, which are to be connected with self.connector.
+        Each child may itself have more child Comparators, again with a connector to indicate
+        how they should be joined
+        """
```
benbc, line 31: I feel there are multiple types trying to get out here. `Predicate`, a generic tree type, and `Comparator` and  `Conjunction` (or something) the concrete types that are the nodes and leaves in the tree. `connector` and `children` are fields only on `Conjunction`, whereas `source`, `operator` and `value` are fields only on `Comparator` (I think).

tests/test_query_engine.py
```diff
@@ -616,15 +615,189 @@ def test_categorise(database, setup_test_database):
...
+            "tall": (_height > 190) & (_code == "abc"),
+        }
+        height_group = categorise(_height_with_codes_categories, default="missing")
+
+    result = extract(Cohort, MockBackend, database)
+    assert result == [
+        dict(patient_id=1, height_group="tall"),
+        dict(patient_id=2, height_group="missing"),
+        dict(patient_id=3, height_group="short"),
+    ]
```
benbc, line 756: A general comment on testing, with this as a specific example. I think we'd benefit from trying to slim down each test case as much as possible -- reducing the test data and cohort definition to the absolute minimum. This makes the purpose of the tests easier to understand, more clearly points at the semantics of the feature and reduces the chance that we'll need to modify the test in future.

Here, for example, I think we only need a single STP and I don't think the `filter` in the code handling is necessary.

I know it's tempting to grab what we already have to hand and reuse it, but I think the extra effort will pay off in the long term.

benbc (commented):
I really like this.

cohortextractor/query_language.py
rebkwok: I think these could be two things; so a comparison on a Value returns a Comparator, and a combination of Comparators returns a Conjunction which has children, which can be either Comparators or Conjunctions.

tests/test_query_engine.py
rebkwok: I think the filter in the code is a leftover from me initially trying to use an `exists` on it.  I'll simplify it

evansd (approved):
This looks great. I think further down the line it should be possible to restructure things so as to make them arbitrarily nestable i.e. a `Comparator` becomes just another type of Value and can be used anywhere a Value can, including as the input to another Comparator. That will really unlock the full power of this. But best to start simple and get that right first.

cohortextractor/query_language.py
benbc: Yes, agreed.

evansd: Do we need all those types? I was thinking we've basically got an Expression object which has a type and then one or more arguments.

Possibly it's useful to have a BooleanExpression subclass whose type is always boolean regardless of the types of its arguments (which would cover all the comparison cases).

PR #70, opened 2021-07-16
Backend tables can be defined based on arbitrary queries

benbc:


evansd (approved):
I like!

cohortextractor/backends/base.py
```diff
@@ -23,6 +23,7 @@ def __init_subclass__(cls, **kwargs):
         # Register each Backend by its id so we can identify it from an environment variable
         register_backend(cls)
         # Make sure each Backend knows what its tables are
+        cls.tables = set()
```
evansd, line 26: If you really wanted to be belt and braces you could do:
```py
tables = set()
...
cls.tables = frozenset(tables)
```
But that's probably not necessary.

evansd (approved):
Fix looks good.

Just by the by, for some reason commented lines have appeared in your commit message.

PR #71, opened 2021-07-16
Don't record setup

benbc:


evansd (approved):
This is great, and I think it makes the fixtures stuff more readable than before rather than less.

Just copying some remarks from Slack so they've preserved along with the PR:

> I'm a tiny bit confused as to why there's still a little bit of "preamble" code appearing in the recordings though.
> 
> Ah no, now I understand. It's because we run the tests with a new session, so we're seeing the session set up code
>
> So I guess in theory we could make the `setup()` function returned by `setup_test_database()` return the session it used, and then pass that to the extract() function, instead of passing `database`.
> But then we'd have to fiddle around in `Backend` to make it accept a pre-initialised session rather than just connection parameters and it all gets a bit messy.

PR #72, opened 2021-07-16
Still more TPP

benbc:


tests/test_end_to_end.py
```diff
@@ -111,11 +111,9 @@ def test_extracts_data_from_sql_server_integration_test(
 
 def run_test(load_study, setup_tpp_database, cohort_extractor):
     setup_tpp_database(
```
rebkwok, line 113: is there a reason not to use the patient data builder in these tests too?

benbc: No. I just didn't take the time to convert it.

rebkwok (approved):

PR #73, opened 2021-07-16
Add minimum viable support for codelists

evansd:


cohortextractor/query_language.py
```diff
@@ -215,3 +215,11 @@ class ValueFromCategory(Value):
...
         self.default = default
         self.definitions = definitions
+
+
+class Codelist(QueryNode):
+    def __init__(self, codes, system, has_categories=False):
+        if has_categories:
+            raise NotImplementedError("categorised codelists are currently unsupported")
+        self.codes = list(codes)
+        self.system = system
```
rebkwok, line 225: What does `system` do?  I vaguely think I remember you saying it wasn't really necessary for it to be specified in the study definition

evansd: Clinical codes belong to a coding system, which provides the set of valid codes, their clinical definitions and their relationships. The common ones we use are SNOMED, CTV3 (also known as Read v3), and ICD-10.

Although it isn't strictly necessary for running queries, we really ought to be validating that the codes a user supplies via a codelist match the coding system of the column they're querying against.

At the moment that involves specifying the coding system inline in the study definition. I think what you might be remembering is the [proposal](https://github.com/opensafely-core/cohort-extractor-v2/issues/31) I made recently to grab that information from the opencodelists site and not require the user to specify it manually. I still think we should do that, but it's extra work so for I now I was just going to copy what we do already, as far as possible.

We're also going to need to add metadata to the schema definition so we can record the coding system used for each column that contains codes. Hopefully that's going to be fairly easy to do though.

rebkwok: OK, cool.  Sorry, I knew what the `system` was, just not what the purpose was in specifying it here.  Was just wondering if we could remove the requirement to specify it now

rebkwok (approved):

cohortextractor/query_engines/mssql.py
```diff
@@ -73,22 +76,26 @@ def __init__(self, column_definitions, backend):
             )
 
         # Walk the nodes and identify output groups
-        self.output_groups = self.get_output_groups(column_definitions)
+        all_nodes = self.get_all_query_nodes(column_definitions)
+        self.output_groups = self.get_output_groups(all_nodes)
+        self.codelists = [node for node in all_nodes if isinstance(node, Codelist)]
```
benbc, line 81: This seems to be operating at a different level of abstraction to `self.output_groups`, where the filtering gets its own method. Could we bung all this logic into the same method as `get_all_query_nodes()` and have it return multiple values?

cohortextractor/query_engines/mssql.py
```diff
@@ -222,6 +229,42 @@ def create_output_group_tables(self):
...
+            codes = codelist.codes
+            max_code_len = max(map(len, codes))
+            # TODO: Figure out the best way to get the appropriate collation
+            # here. Possibly something like:
+            #
+            #  collation = self.backend.collation_for_coding_system(codelists.system)
+            #
+            collation = "SQL_Latin1_General_CP1_CI_AS"
+            table = sqlalchemy.Table(
+                f"#codelist_{n}",
```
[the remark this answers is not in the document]
evansd: They're session scoped: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-ver15#temporary-tables

cohortextractor/query_engines/mssql.py
```diff
@@ -222,6 +229,42 @@ def create_output_group_tables(self):
...
+            # Constuct the queries needed to create and populate this table
+            self.codelist_tables_queries.append(sqlalchemy.schema.CreateTable(table))
+            # There's a limit on how many rows we can insert in one go using
+            # this method See:
+            # https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver15#limitations-and-restrictions
+            batch_size = 999
+            for i in range(0, len(codes), batch_size):
+                codes_batch = codes[i : i + batch_size]
+                insert_query = table.insert().values([(code,) for code in codes_batch])
+                self.codelist_tables_queries.append(insert_query)
```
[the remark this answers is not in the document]
evansd: Ah quite right. I copied it verbatim from the old cohortextractor which is why I forgot to add a test it for myself. I'll do so now.

cohortextractor/query_engines/mssql.py
```diff
@@ -222,6 +229,42 @@ def create_output_group_tables(self):
...
+                    nullable=False,
+                ),
+            )
+            self.codelist_tables[codelist] = table
+            # Constuct the queries needed to create and populate this table
+            self.codelist_tables_queries.append(sqlalchemy.schema.CreateTable(table))
+            # There's a limit on how many rows we can insert in one go using
+            # this method See:
+            # https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver15#limitations-and-restrictions
+            batch_size = 999
```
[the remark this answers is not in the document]
evansd: We do, sadly. It's very easy to create enormous codelists by including a few high level concepts and then saying "and all child concepts".

cohortextractor/query_engines/mssql.py
```diff
@@ -502,9 +552,15 @@ def get_sql(self):
         return "\n\n\n".join(sql)
 
     def query_expression_to_sql(self, query):
+        dialect = self.sqlalchemy_dialect.dialect()
+        # SQLAlchemy will set this for us when connecting to an actual database
+        # (it asks for the server version and then updates capabilities
+        # accordingly) but we haven't connected to a database here so we have
+        # to set it manually
+        dialect.supports_multivalues_insert = True
```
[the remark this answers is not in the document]
evansd: I think there's definitely an argument for asserting that that property is supported at the point we actually connect. It would feel a bit annoying to me though (for debugging purposes) to structure things so that you can't even see the generated queries without having a real database to connect to.

cohortextractor/query_engines/mssql.py
evansd: I don't quite follow this point here. I originally had this as a separate method for parity with `get_output_groups` but the code felt overall clearer to me having inlined the one-liner.

cohortextractor/query_engines/mssql.py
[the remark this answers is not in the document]
benbc: Yes, that would be annoying. Fair enough.

cohortextractor/query_engines/mssql.py
benbc: I mean something like

```
self.output_groups, self.codelists = self.extract_nodes(column_definitions)
```

I don't feel very strongly.

evansd: Ah, I see. I'm sure we can make the whole thing a bit clearer structurally. It might be better to see what else we end up wanting alongside codelists and output groups (there must be a better name for these!) before we do attempt that though.

benbc: As you prefer. I'm secretly working up to trying to persuade you to completely restructure this "interpretation" of the DAG, so I hope we'll be revisiting this anyway.

benbc (approved):

PR #76, opened 2021-07-19
Categorising by truthiness (non-nullness) of a value and inversion

### LINKED ISSUE #51: Categorizing for bucketed values
- [x] simple single comparisons (`x > y`)
- [x] and/or conditions (`(x > y)` & `(y == z)`)
- [x] nesting conditions (`((x > y)` & `(y == z)) | (z == a)`)
- [ ] inverting conditions (`~((x > y)` & `(y == z))`)
- [ ] True/False comparisons

rebkwok:
Allows categorisation on whether a Value (of any sort) is truthy, by using a `c` function that returns a comparator for != None.  Since every patient will have a row in any value table created from a filter or aggregation, we can use the None values to check for existence.

Fixes #51

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -7,10 +7,9 @@ build-cohort-extractor:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    CI="${CI:-0}"
-    [ "$CI" == "1" ] && echo "::group::Build cohort-extractor (click to view)" || echo "Build cohort-extractor"
+    [[ -v CI ]] && echo "::group::Build cohort-extractor (click to view)" || echo "Build cohort-extractor"
     docker build . -t cohort-extractor-v2
-    [ "$CI" == "1" ] && echo "::endgroup::" || echo ""
+    [[ -v CI ]] && echo "::endgroup::" || echo ""
 
 # tear down the persistent cohort-extractor-mssql docker container and network
 remove-persistent-database:
@@ -60,14 +59,12 @@ test-all ARGS="": build-cohort-extractor
     #!/usr/bin/env bash
     set -euo pipefail
 
-    CI="${CI:-0}"
-
-    [ "$CI" == "1" ] && echo "::group::Run tests (playback mode) (click to view)" || echo "Run tests (playback mode)"
+    [[ -v CI ]] && echo "::group::Run tests (playback mode) (click to view)" || echo "Run tests (playback mode)"
     . scripts/setup_functions
     dev_setup
 
     DATABASE_MODE="${DATABASE_MODE:-ephemeral}" RECORDING_MODE="${RECORDING_MODE:-playback}" pytest --cov=cohortextractor --cov=tests {{ ARGS }}
-    [ "$CI" == "1" ]  && echo "::endgroup::" || echo ""
+    [[ -v CI ]]  && echo "::endgroup::" || echo ""
 
 # run all tests including integration and smoke tests against a persistent database. Optional args are passed to pytest
 test-all-fast ARGS="":
@@ -86,14 +83,12 @@ test-assert-recordings-up-to-date:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    CI="${CI:-0}"
-
-    [ "$CI" == "1" ] && echo "::group::Run tests (record mode) (click to view)" || echo "Run tests (record mode)"
+    [[ -v CI ]] && echo "::group::Run tests (record mode) (click to view)" || echo "Run tests (record mode)"
     rm tests/recordings/*.recording
     just test-record
-    [ "$CI" == "1" ] && echo "::endgroup::" || echo ""
+    [[ -v CI ]] && echo "::endgroup::" || echo ""
 
-    [ "$CI" == "1" ] && echo "::group::Diff Recordings (click to view)" || echo "Diff Recordings"
+    [[ -v CI ]] && echo "::group::Diff Recordings (click to view)" || echo "Diff Recordings"
     git update-index -q --really-refresh # avoid false positives due to last modification time changing
     if ! git diff-index --quiet HEAD -- tests/recordings; then
         git status -- tests/recordings
@@ -103,7 +98,7 @@ test-assert-recordings-up-to-date:
     else
         echo "Recordings are up-to-date"
     fi
-    [ "$CI" == "1" ] && echo "::endgroup::" || echo ""
+    [[ -v CI ]] && echo "::endgroup::" || echo ""
 
 
 # runs the format (black), sort (isort) and lint (flake8) check but does not change any files
--- a/Justfile
+++ b/Justfile
@@ -4,7 +4,13 @@ default:
 
 # build the cohort-extractor docker image
 build-cohort-extractor:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    CI="${CI:-0}"
+    [ "$CI" == "1" ] && echo "::group::Build cohort-extractor (click to view)" || echo "Build cohort-extractor"
     docker build . -t cohort-extractor-v2
+    [ "$CI" == "1" ] && echo "::endgroup::" || echo ""
 
 # tear down the persistent cohort-extractor-mssql docker container and network
 remove-persistent-database:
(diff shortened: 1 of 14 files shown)


cohortextractor/query_language.py
```diff
@@ -44,6 +51,14 @@ def __or__(self, other):
...
         return len(self.children)
 
+    def __invert__(self):
+        self.negate()
+        return self
+
+    def negate(self):
+        """Negate the sense of the root comparator."""
+        self.negated = not self.negated
+
```
evansd, line 61: I'm a bit worried by the mutation here. I think we'll make life a lot easier for ourselves in the long run if we insist that everything in the query DAG is immutable and all methods like this return a new object with different properties.

evansd (approved):
This is nice.

I've got a worry about mutability though, which doesn't need to block for now but is something I think we should refactor away before too long.

I also wonder if we can replace the `c` function with `is_null`/`is_not_null` methods on `Value` which is maybe a more natural API.

cohortextractor/query_language.py
rebkwok: :+1: I can make it do that

rebkwok: Also, I did wonder about making it use `is_null`/`not_null` instead, but the `c` function seemed a bit closer to how they write it at the moment (in the long covid study there's one that does `sgss_positive OR primary_care_covid AND NOT hospital_covid`).  But it's easy enough to change later if people prefer that

evansd: Right, yes, good point. I suppose the ideal would be if we could do the Pythonic thing and implicitly convert non-boolean values to booleans if they're involved in boolean conditions. But I think that would require us tracking the types of each Value, which I'm sure we'll want to do eventually but will be quite a lot of work to add now.

rebkwok: I dithered a bit over how to make the new object in the invert method, since it can't just use the `_combine` one (it could be negating a lone comparator or one with children).  I've used `deepcopy` but it could also more explicitly do something like
```
obj = type(self)(
    connector=self.connector,
    negated=not self.negated,
    source=self.source,
    operator=self.operator,
    value=self.value,
)
for child in self.children:
    obj.add(child, self.connector)
```
Any particular opinions either way?

evansd: I can't remember exactly how `deepcopy` works, but I'd worry that it would end up duplicating the entire query tree i.e. it will see `self.source` and start copying that. I think maybe just shallow `copy.copy()` would do the right thing here. But no harm in doing in the explicit way.

evansd: I feel like it ought to be possible for negation to be a generic operation on Values, rather than something specific to Comparator. Ultimately, we just need SQLAlchemy to call it's `not_` operator  on whatever the argument is. I don't quite have a clear picture of that would all work though so it's a thing to ponder rather than a suggestion for what to do now.

rebkwok: Actually, I think we could have an `__invert__` method on a Value too. In fact, we should, so we can do `~foo` and don't have to do `~c(foo)`.

cohortextractor/query_engines/mssql.py
```diff
@@ -162,21 +162,33 @@ def get_output_column_name(node):
         else:
             raise TypeError(f"Unhandled type: {node}")
 
-    def get_sources_from_category_definitions(self, definitions, sources=None):
-        sources = sources or set()
+    def get_query_nodes_from_category_definitions(self, definitions, query_nodes=None):
```
rebkwok, line 165: Was `query_node` a bad term? I mean, I think it's better than `source`...

evansd: I don't think it's a bad term in general, and it's definitely better than "source" in this context. I think we were just saying that `get_parent_nodes()` might be a better method name than `get_query_node_references()`. Something that makes it clear that we're just interested in walking the graph and aren't fussed about exactly what sort of objects we're getting back or what sort of role they play.

rebkwok: Maybe I'll just leave it as it is for now and it can be renamed again when we do other renaming things, if we want to

evansd: Yep, sorry I didn't mean it needed changing now. I think that's part of moving some of this logic onto the DSL objects themselves, which we can do at a later point.

Justfile
```diff
@@ -4,7 +4,13 @@ default:
 
 # build the cohort-extractor docker image
 build-cohort-extractor:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    CI="${CI:-0}"
+    [ "$CI" == "1" ] && echo "::group::Build cohort-extractor (click to view)" || echo "Build cohort-extractor"
```
evansd, line 11: Github [sets](https://docs.github.com/en/actions/reference/environment-variables) `CI="true"`, so if you check for that you won't need to set it manually in the workflow yaml which would be a bit cleaner.

Or you could just test that it's non-empty rather than looking for a specific value.

cohortextractor/query_engines/mssql.py
```diff
@@ -162,21 +162,33 @@ def get_output_column_name(node):
...
-                    definition.children, sources
+                query_nodes = self.get_query_nodes_from_category_definitions(
+                    definition.children, query_nodes
                 )
-        return sources
+        return query_nodes
+
+    def list_query_nodes_from_category_definitions(self, definitions):
+        # Sort the query nodes to ensure consistent order.  We have to use a custom sort key
+        # here because query nodes are Values with overloaded lt/gt operators.
```
evansd, line 180: Arghh! I bet that was fun figuring out :)

cohortextractor/query_language.py
```diff
@@ -44,6 +56,22 @@ def __or__(self, other):
     def __len__(self):
         return len(self.children)
 
+    def __invert__(self):
+        obj = copy(self)
```
evansd, line 60: Is this `copy()` call doing anything more than calling the `__copy__()` method below? If not, could we just get rid of the `import copy`?

cohortextractor/query_language.py
```diff
@@ -20,20 +23,29 @@ class Comparator:
     """A generic comparator to represent a comparison between a source object and a value"""
 
     def __init__(
-        self, children=None, connector="and_", source=None, operator=None, value=None
+        self,
+        children=None,
```
evansd, line 27: I think I'm slightly confused about the role of `children` vs `lhs` and `rhs` here. (I know this isn't something you're adding in this PR and so it's not something I'm suggesting changing now, I just haven't quite got my head around it.)

For an operation like `__and__` we add the two nodes to the `children` list, but why can't they just be recorded as `lhs` and `rhs` attributes like others do?

Justfile
```diff
@@ -75,14 +86,25 @@ test-assert-recordings-up-to-date:
...
     rm tests/recordings/*.recording
     just test-record
+    [ "$CI" == "1" ] && echo "::endgroup::" || echo ""
+
+    [ "$CI" == "1" ] && echo "::group::Diff Recordings (click to view)" || echo "Diff Recordings"
     git update-index -q --really-refresh # avoid false positives due to last modification time changing
     if ! git diff-index --quiet HEAD -- tests/recordings; then
         git status -- tests/recordings
         echo >&2 "ERROR: Recordings are not up-to-date"
+        git diff tests/recordings
```
evansd, line 101: Having this is great

evansd (approved):
Thanks @rebkwok this looks good!

I have a few questions below but nothing blocking

Justfile
rebkwok: :+1: didn't know Github set it already

cohortextractor/query_language.py
rebkwok: It's not, no.  It just means that `copy.copy` on one of these used elsewhere would use that method.  But we don't actually use it elsewhere, so  it's probably better to be more explicit about it

cohortextractor/query_language.py
rebkwok: Maybe they can?  I may have been misled by using django's `Q` as a model, and thinking that there could be more than 2 children at some point.  I'll give it a try (in a separate PR)

PR #78, opened 2021-07-19
Update future long covid test Cohort

rebkwok:


tests/acceptance/test_long_covid_study.py
```diff
@@ -42,22 +42,20 @@ class SimplifiedCohort:
...
+    #     "80+": _age >= 80,
     # }
-    # age_group = (
-    #     table("patients")
-    #     .categorise("age", groups=_age_categories, reference=registration_date)
-    # )
-    # sex = table("patients").categorise("sex")
+    # age_group = categorise(_age_categories, default="missing")
+    #
+    # sex = table("patients").latest("sex")
```
benbc, line 58: I don't understand `latest("sex")`. Patient ID is expected to be the primary key for this table, so we should only get one value, shouldn't we?

rebkwok: I think `table("patients").get("sex")` will give you a `Column`, and we need a `ValueFromRow`.  It probably should be `table("patients").get("sex").latest()`

rebkwok: Although actually, maybe a `Column` is ok here.  It needs to be a `Value` for `categorise()`, but we're not using `categorise` for sex

benbc (approved):

PR #79, opened 2021-07-19
Update recordings

evansd:
Maybe the workflow we need is to rebase and re-run recordings before merge?


benbc (approved):
Ugh. That's really annoying. I remember thinking this would need re-recording when I reviewed the PR. Sorry.

PR #80, opened 2021-07-21
Some query engine refactoring

benbc:


cohortextractor/query_engines/mssql.py
```diff
@@ -133,7 +133,7 @@ def walk_query_dag(self, nodes, seen=None):
         for node in nodes:
             yield node
             for reference in [
-                *self.get_query_node_references(node),
+                *self.get_sources(node),
                 getattr(node, "value", None),
```
evansd, line 137: This `getattr(node, "value", None)` could/should be pulled into `get_sources` (or did we decide on `get_parent_nodes`) which would simplify this to `for reference in self.get_parent_nodes(node)`

evansd (approved):

evansd (approved):
YES!

This is absolutely and obviously the correct way to do it, now that you've done it.

PR #81, opened 2021-07-21
Add date of birth to TPP backend

benbc:


cohortextractor/backends/tpp.py
```diff
@@ -41,3 +42,10 @@ class TPPBackend(BaseBackend):
             SELECT Patient_ID as patient_id, Specimen_Date AS date, 0 AS positive_result FROM SGSS_AllTests_Negative
         """,
     )
+
+    hospitalizations = MappedTable(
+        source="APCS",
+        columns=dict(
+            date=Column("date", source="Admission_Date"),
+        ),
+    )
```
rebkwok, line 51: It will need the column to match codes on too, won't it? - I think `Der_Diagnosis_All`, from looking at the old tpp backend code.   That does a `LIKE` query on the codelists, so it might need something more?  Maybe a question for @evansd

benbc: Of course. Too hot to think. :-)

benbc: @evansd What is going on with the `LIKE` matching there? Do you already have a plan for how we should handle it?

evansd: Oh man, I somehow hadn't twigged that SUS data was part of the long covid study. This is whole new can of worms :(

We're going to need some kind of special column type which is "multiple codes stuffed in a space separated string" and then code which knows hows to handle these and transform code filters into LIKE queries. It's not going to be fun.

evansd (approved):

PR #83, opened 2021-07-21
Support `age_as_of` queries

### LINKED ISSUE #77: `age_as_of()`

evansd:
Closes #77 

Work to address #92 has been pulled into a separate branch as this is wider issue affecting more than just age calculations.


cohortextractor/query_engines/mssql.py
```diff
@@ -378,6 +390,31 @@ def get_case_expression(self, mapping, default):
...
+            raise ValueError(f"Unsupported function: {value.function}")
+        value_expression = method(*argument_expressions)
+        return value_expression, tuple(tables)
+
+    def apply_function_date_diff_in_years(self, start_date, end_date):
+        # `literal_column` doesn't seem quite the right construct here, but I
+        # need SQLAlchemy to generate the string "year" without quotes, and
+        # this seems to do the trick
+        YEAR = sqlalchemy.literal_column("year")
+        year_diff = sqlalchemy.func.datediff(YEAR, start_date, end_date)
```
rebkwok, line 412: Does datediff round up to the nearest int?

evansd: It diffs the year part but ignores month and day entirely

evansd: It diffs the year part but ignores month and day entirely

benbc:
I'm on holiday today, so I won't do a proper review. I think this is mostly fine -- I like `ValueFromFunction` and the idea of having this as a special function on the patients table.

I think I would do the dispatch of the function by having a mapping from subclasses of `ValueFromFunction` to callables, rather than by string-mangling.

And I think that for special well-known tables we should have explicit references to them and then specialized types. That way we get nice auto-complete in the IDE.

Something like

```
class Cohort:
    foo = tables.called("whoknows").get("foo")
    age = tables.patients.age_as_of(today())
```

cohortextractor/query_engines/mssql.py
rebkwok: Ah OK.  I see the reason for the -1 calculation then :+1:

rebkwok:
This looks good to me.  

I'm not entirely sure about the idea of having explicitly named tables; it feels like it diverges from the aim of having a generic DSL.  IDE completion would be nice, but if we do it for certain tables, we should do it for all possible tables - people are likely to use their IDE tab completion to look for what options they have (I know I often do), so if they see `tables.patients` in the autocomplete list, they might reasonably expect to see all the table options there.  That would mean we'd need to at least have an alias for `Table("whatever")` in the query language for every table even if it didn't have its own type.  Maybe that's not a terrible thing if it makes users' experience better though.  

I do like the idea of a `PatientTable` type so the `age_as_of` method is only available on the table it belongs to .

evansd:
> I think I would do the dispatch of the function by having a mapping from subclasses of ValueFromFunction to callables, rather than by string-mangling.

Yes, agreed. I copied a pattern which made sense in the context of the old cohortextractor but we can do better here.

> And I think that for special well-known tables we should have explicit references to them and then specialized types. That way we get nice auto-complete in the IDE.

I'd forgotten we talked already about making tables attributes of some object rather than referencing them by string. I think that's a good idea, although I agree with Becky that if we're going to do it we should do it for all tables not just some.

One thing which this forces us to address (which needs addressing anyway) is the fact that different backends will have different schemas. Here's one way we might address this:

1. Tables are attributes on an object specific to each backend e..g to write a study using the TPP backend you do something like:
   ```py
    from cohortextractor.backends import tpp

    class Cohort:
       date_of_birth = tpp.patients.get("date_of_birth")
       event_date  tpp.coded_events.latest().get("date")
    ```

2. If you want to write a study to be run on multiple backends then you can create a sort of "intersection backend" by combining two or more backends. The result has just those tables, and just those columns on those tables, which are common to all the supplied backends.
    ```py
    from cohortextractor import combine_backends
    from cohortextractor.backends import tpp, emis

   common = combine_backends(tpp, emis)

    class Cohort:
       date_of_birth = common.patients.get("date_of_birth")
       event_date = common.coded_events.latest().get("date")
    ```
    The indirection we've given ourselves with respect to column and table names (and to a certain extent, types) should allow us to make backends look identical to the extent that they're semantically equivalent.

3. We also want to give researchers the option to do different things in different backends so something like:
    ```py
    class Cohort:
      my_val = common.one_of(
        tpp.some_tpp_table.latest(),get("value"),
        emis.some_emis_table.latest().get("value"),
      )
    ```
    Or, if you were happy for a column to just be null when run against a particular backend:
    ```py
    class Cohort:
      my_val = common.one_of(
        tpp.some_tpp_table.latest(),get("value"),
        emis=None,
      )
    ```

4. It would be a error to try to directly use more than one backend object in a cohort definition e.g. this isn't allowed
    ```py
    class Cohort:
       date_of_birth = common.patients.get("date_of_birth")
       event_date = tpp.coded_events.latest().get("date")
    ```
   It would also be an error to use the `common.one_of` method and not define options for all of the backends included in `common`.

(The function names `combine_backends` and `one_of` are just the first things that occurred to me. No doubt we can do better.)

Any thoughts?

benbc:
Basically, yes -- I think this all sounds good.

I too have thought we might want the backend to be the "root" for retrieving tables. We might also want "data source" in there as well (ONS/SUS etc).

Given that I'm sure there's some way of doing cross-backend cohort definition that makes sense. Your suggestion looks perfectly reasonable.

Why do you think it should be an error to use more than one backend in a single cohort? To force explicit choices for missing stuff?

evansd:
> Why do you think it should be an error to use more than one backend in a single cohort? To force explicit choices for missing stuff?

Yes, exactly. I suppose the alternative is to say that when you run against backend X all the columns you defined from backend Y are implicitly null (and vice versa) and only ones drawn from common will appear in both. Maybe that's acceptable too. Erroring was just my initial instinct but open to other suggestions.

benbc:
I think it's not a bad idea to force a very explicit description when doing complex stuff like this.

cohortextractor/query_engines/mssql.py
```diff
@@ -404,6 +412,49 @@ def get_case_expression(self, mapping, default):
...
+        try:
+            method = class_method_map[value.__class__]
+        except KeyError:
+            raise ValueError(f"Unsupported function: {value}")
+        value_expression = method(*argument_expressions)
+
+        # TODO: Once we have consistent hashing for QueryNodes we can remove
+        # this sort. This is not required for correctness, just for consistent
+        # test recordings.
+        tables = sorted(tables, key=lambda i: i.name)
```
rebkwok, line 440: Are the tables from a `ValuefromFunction`'s arguments guaranteed to always be `BaseTable`s, with a `name`, and not a `FilteredTable`?

rebkwok: I guess they are for now, since we only have the `age_as_of` method that uses this, and it can only be called on the patient table

rebkwok (approved):
One question, which is probably fine, for now at least

cohortextractor/query_engines/mssql.py
evansd: This is a good catch, but not quite for the reason you think! In fact, my comment is totally wrong because these are instances of `sqlalchemy.Table` rather than `query_language.Table`. So they are guaranteed to have a `name` attribute, but fixing the hashing on QueryNode won't make any difference here. A good illustration of how confusing this codebase can be!

PR #84, opened 2021-07-22
Port over codelist_from_csv and combine_codelists

rebkwok:


evansd (approved):
I might be inclined to keep the `Codelist` class inside the `query_language` module. The `codelistlib` module is then basically just helper functions which hopefully we can replace later. But that way all the AST classes still live together.

Other than that all looks good.

PR #86, opened 2021-07-23
Add support for querying hospitalization codes

benbc:


cohortextractor/backends/tpp.py
```diff
@@ -2,6 +2,35 @@
...
+    """
+    assert len(char) == 1
+    assert char not in ["^", "]"]
+    return f"""
+        REVERSE(
+            SUBSTRING(
+                REVERSE({ref}),
+                PATINDEX(
+                    '%[^{char}]%',
+                    REVERSE(fully_split.Value)
```
rebkwok, line 26: Should this also be `REVERSE({ref})`?

tests/test_tpp_backend.py
```diff
@@ -124,3 +129,115 @@ class Cohort:
     assert extract(Cohort, TPPBackend, database) == [
         dict(patient_id=1, admission=date(2020, 12, 12))
     ]
+
+
+@pytest.mark.integration
+def test_hospitalization_table_returns_a_single_code(database, setup_tpp_database):
```
rebkwok, line 135: These tests look like they could easily be a single parametrized test

rebkwok: Will this just work with filtering on codelists?  I think they will, but maybe a test for it would be useful

cohortextractor/backends/tpp.py
benbc: Gah! Thank you.

tests/test_tpp_backend.py
benbc: > These tests look like they could easily be a single parametrized test

You're quite right. Good idea.

benbc: > Will this just work with filtering on codelists? I think they will, but maybe a test for it would be useful

I feel like the answer is "of course", but I can't completely convince myself it's right -- which I think is a good signal that a test is needed.

tests/test_tpp_backend.py
```diff
@@ -124,3 +129,67 @@ class Cohort:
...
+        ),
+    )
+
+    class Cohort:
+        code = table("hospitalizations").filter(code="xyz").get("code")
+
+    assert extract(Cohort, TPPBackend, database) == [
+        dict(patient_id=1, code=None),
+        dict(patient_id=2, code="xyz"),
+    ]
```
rebkwok, line 195: Do we also need/want a test with an `is_in` `Codelist` filter?  (Not sure it's necessary - I guess if we've tested that this filters ok on a single code, the other codelist tests cover the codelist scenario)

rebkwok (approved):
One comment on tests but no need to add if you feel it's unnecessary

tests/test_tpp_backend.py
benbc: What do you think? You know the filtering code much better than me. I'm very happy to add it -- or maybe replace this filter with that one since it seems on the face of it that they would be more likely to interact badly.

rebkwok: Yes, maybe make this one a Codelist filter instead

PR #91, opened 2021-07-26
Make the ephemeral test database session-based

rebkwok:
The `recording` fixture needs to remain function-scoped, otherwise it doesn't have access to the specific test data it needs for recording, as does the `setup_database` fixture, so it can set up the data for each test.  Since we can't mix fixture scopes, we can't just make the database a session-scoped fixture, so the docker client, network and database are now set on the session itself, and the fixtures just retrieve them.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,23 +13,31 @@ from lib.tpp_schema import Base
 from sqlalchemy.orm import sessionmaker
 
 
-@pytest.fixture
-def docker_client(request):
-    yield request.session.docker_client
+@pytest.fixture(scope="session")
+def docker_client():
+    yield docker.from_env()
 
 
-@pytest.fixture
-def network(request):
-    yield request.session.network_name
+@pytest.fixture(scope="session")
+def network(docker_client):
+    name = "test_network"
+    docker_client.networks.create(name)
+    yield name
+    docker_client.networks.get(name).remove()
 
 
-@pytest.fixture
-def containers(request):
-    yield Containers(request.session.docker_client)
+@pytest.fixture(scope="session")
+def containers(docker_client):
+    yield Containers(docker_client)
 
 
-def mssql_dir():
+def mssql_dir_path():
     return Path(__file__).parent.absolute() / "support/mssql"
+
+
+@pytest.fixture(scope="session")
+def mssql_dir():
+    yield mssql_dir_path()
 
 
 def is_smoke_test(request):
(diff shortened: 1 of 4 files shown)


benbc:
@rebkwok I see that it's possible to define a fixture as session-scoped like `@pytest.fixture(scope="session")`. Can't we use that here?

rebkwok:
> @rebkwok I see that it's possible to define a fixture as session-scoped like `@pytest.fixture(scope="session")`. Can't we use that here?

No, for the reason I mentioned in the description - you can't mix fixture scopes.  The `recording` fixture has to be function-scoped, otherwise you don't have info about the specific test that's running in `request.node`.  If we made the db session-scoped, any test that uses it can also only use session-scoped fixtures.

evansd:
> If we made the db session-scoped, any test that uses it can also only use session-scoped fixtures.

I feel like we must be talking about something slightly different here because you can definitely use mixed scope fixtures in a single test. Is it something to do with nested fixtures i.e. you can't have a fixture that uses fixtures of mixed scope?

evansd (approved):
I'm very happy about this!

While it's obviously slightly less than ideal to have to use the session explicitly like this, and while I feel like there _might_ be some other way of getting pytest to do what we want here, I'm quite prepared to accept that there isn't. And in any case there's nothing particularly bad about doing it this way and it's not a big invasive change that would be hard to alter later if we have a sudden lightbulb moment.

This is going to make a big difference to my dev experience I think :)

benbc:
> No, for the reason I mentioned in the description - you can't mix fixture scopes.

So sorry @rebkwok -- I didn't read the PR description

benbc:
I see examples on the web of function-scoped fixtures using session-scoped ones.

rebkwok:
> > If we made the db session-scoped, any test that uses it can also only use session-scoped fixtures.
> 
> I feel like we must be talking about something slightly different here because you can definitely use mixed scope fixtures in a single test. Is it something to do with nested fixtures i.e. you can't have a fixture that uses fixtures of mixed scope?

Yes, I think you're right.  Maybe I misunderstood the reason for the can't-mix-fixture-scopes error message I was getting before.  I'll see if I can figure out what caused it

rebkwok:
Found a way to make the databases session-scoped but the database fixture still function-scoped so the checks for the test type still work.  The only issue is that, by using the `record_db` fixture in the `database` fixture it will always create the `record_db` database, even if you're running `just test-integration` in playback mode.  Since it only creates it once, it's only a couple of seconds unnecessary time, and I think this is better.  

I tried using a session-based fixture that takes a database type (record/playback) and returns a function that creates the database if necessary, but of course then it's the function is session-based, and it gets called each time it's used, so it will try to create the database each time.

PR #94, opened 2021-07-27
Yet more TPP

benbc:


cohortextractor/query_engines/mssql.py
```diff
@@ -246,12 +246,7 @@ def create_codelist_tables(self):
         for n, codelist in enumerate(self.codelists):
             codes = codelist.codes
             max_code_len = max(map(len, codes))
-            # TODO: Figure out the best way to get the appropriate collation
-            # here. Possibly something like:
-            #
-            #  collation = self.backend.collation_for_coding_system(codelists.system)
-            #
-            collation = "SQL_Latin1_General_CP1_CI_AS"
+            collation = "Latin1_General_BIN"
```
rebkwok, line 249: I don't really know what this does - what's the reason for the change?

rebkwok (approved):

cohortextractor/query_engines/mssql.py
benbc: It's roughly MSSQL-speak for text encoding. I believe (based on the old CE) that this is the correct value for TPP, so I think we can safely hard-code it everywhere for now.

PR #103, opened 2021-07-28
Add a test for the full long covid study 

rebkwok:


cohortextractor/query_engines/mssql.py
```diff
@@ -188,7 +188,7 @@ def list_query_nodes_from_category_definitions(self, definitions):
         # here because query nodes are Values with overloaded lt/gt operators.
         return sorted(
             self.get_query_nodes_from_category_definitions(definitions),
-            key=lambda x: (x.column, x.source),
+            key=lambda x: (x.column, repr(x.source)),
```
rebkwok, line 191: This fixes a bug that came up in the previous_covid categorisation for the real study.  If x.source is a `Row`, it can't be used as a key in the sort.  This should be a temporary fix until we change the types of the Value/Comparator things, and/or implement proper hashes

tests/test_query_engine.py
```diff
@@ -1023,3 +1023,46 @@ class Cohort:
         dict(patient_id=3, has_code="yes"),
         dict(patient_id=4, has_code="yes"),
     ]
+
+
+@pytest.mark.integration
+def test_categorise_multiple_truthiness_categories(database, setup_test_database):
```
rebkwok, line 1029: This tests for the sorting bug above

cohortextractor/query_language.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         self.rhs = rhs
 
     def __repr__(self):
-        return f"Comparator(children={self.children}, connector={self.connector}, source={self.source}, operator={self.operator}, value={self.value})"
+        return f"Comparator(connector={self.connector}, negated={self.negated}, lhs={self.lhs}, operator={self.operator}, rhs={self.rhs})"
```
benbc, line 45: I just fixed this too: :-)

cohortextractor/query_engines/mssql.py
benbc: I think that deserves a code comment rather that just a comment on the PR. I certainly would be surprised by that code if I came to it in a few weeks.

tests/test_query_engine.py
benbc: We could really do with unit tests for this kind of case, couldn't we? It's very hard to see what's really being tested for. Not asking you to add one now, but we need to work out how to do it at some point.

benbc (approved):
Cool. We're really not that far off, are we?

cohortextractor/query_engines/mssql.py
rebkwok: Good point, will add one

tests/test_query_engine.py
rebkwok: For now I'll add a comment about what it's testing

PR #106, opened 2021-07-28
Publish the Docker image

benbc:


.github/workflows/main.yml
```diff
@@ -39,3 +39,20 @@ jobs:
       - uses: hadolint/hadolint-action@v1.5.0
         with:
           failure-threshold: error
+
+  docker-publish:
+    if: github.ref == 'refs/heads/main'
```
evansd, line 44: I think this will work fine, but the standard approach would be to make a new workflow file and specify that this only runs on the `main` branch c.f.
https://github.com/opensafely-core/python-docker/blob/main/.github/workflows/build_and_publish.yaml

(Edit: unless this is deliberate because the `needs` dependencies don't work across files, which is quite possible)

.github/workflows/main.yml
```diff
@@ -39,3 +39,20 @@ jobs:
...
+    env:
+      image: ghcr.io/opensafely-core/cohortextractor-v2:latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Pull image
+        run: docker pull ${{ env.image }}
+      - name: Build image
+        run: docker build . --file Dockerfile --tag ${{ env.image }}
+      - name: Log into GitHub Container Registry
+        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
```
evansd, line 56: I'll need to enable this secret for this repo, will do so now

evansd (approved):

.github/workflows/main.yml
benbc: Yes -- I wanted dependencies on the other stuff.

.github/workflows/main.yml
benbc: Oh, thanks.

evansd: Done

PR #108, opened 2021-07-29
Add Organisation table data

### LINKED ISSUE #107: Add region and pseudo_id from Organisation table

rebkwok:
Fixes #107


cohortextractor/backends/tpp.py
```diff
@@ -52,12 +52,22 @@ class TPPBackend(BaseBackend):
         ),
     )
 
-    practice_registrations = MappedTable(
-        source="RegistrationHistory",
+    practice_registrations = QueryTable(
```
rebkwok, line 55: I initially added a QueryTable for `organisations`, but since it needs to know about the start date and end date as well as the patient id, it made more sense to add the relevant Organisation data to the practice_registrations table instead.  Otherwise we essentially build a copy of the RegistrationHistory and have to do the `date_in_range` type selections on each of them.

evansd: Yeah, I think that's the right approach

tests/acceptance/test_full_long_covid_study.py
```diff
@@ -34,7 +35,7 @@ class Cohort:
     # Patients registered on 2020-11-01
     _registrations = table("practice_registrations").date_in_range(index_date)
```
evansd, line 36: Does this need something to make it a single row? i.e. some form of `first_by` or `last_by` operation. I guess it works as is at the moment because the data happens to have only one matching row, but at some point we'll want to start enforcing this in the cohort definition.

evansd (approved):
Looks good to me!

tests/acceptance/test_full_long_covid_study.py
rebkwok: Yes - but I think I've already fixed it in one of the other PRs

rebkwok: No, actually I'd fixed it on this branch but forgot to push.  I need to fix some conflicts anyway.

PR #109, opened 2021-07-29
Implement Index of Multiple Deprivation

### LINKED ISSUE #96: Implement Index of Multiple Deprivation
IMD queries in the old cohort extractor are calculated by finding the patients with addresses registered on the index_date, and then ordering them by 
```
ORDER BY
              StartDate DESC,
              EndDate DESC,
              IIF(MSOACode = 'NPC', 1, 0),
              PatientAddress_ID
```
This orders by most recent start date, then latest end date, then prefers addresses which are not marked "NPC" (for "No Postcode") and finally uses the address ID as a tie-breaker.

The IMD query in the old cohort extractor could almost be implemented as a `last_by` method in the new version, like:
```
imd_value = (
        table("patient_address")
        .date_in_range(index_date)
        .last_by("date_start", "date_end", "has_postcode", "patientaddress_id")
        .get("index_of_multiple_deprivation")
    )
``` 

Except that `has_postcode` is not a field, it's the IIF statement on MSOACode.

A `Row` could accept an expression as a column (with relevant DSL wrapping for defining the expression).   We probably still want a custom method of some sort so that researchers don't have to type the column ordering explicitly though.

rebkwok:
Fixes #96 

Adds the patient address table, with a custom `has_postcode` boolean field.  For now, there is a `imd_rounded_as_of` method on a query language `Table`, which is just a nicer interface to calling
```
table("patient_address")
            .date_in_range(date)
            .last_by("date_start", "date_end", "has_postcode", "patientaddress_id")
            .get("index_of_multiple_deprivation_rounded")
```

We discussed having named tables with their own methods, but this was the simplest solution for now which doesn't require researchers to write the sorting columns out, and it should be easily changed later.


cohortextractor/query_language.py
```diff
@@ -170,6 +170,27 @@ def sum(self, column):  # noqa: A003
     def aggregate(self, function, column):
         return ValueFromAggregate(self, function, column)
 
+    def imd_rounded_as_of(self, reference_date):
```
rebkwok, line 173: This adds a table-specifc method in the same way that Dave's age_as_of does.  I don't really like it, but it's the simplest thing for now, pending decisions on adding named tables etc.

evansd: I agree about not liking it and about it being the best thing to do for now!

evansd (approved):

PR #112, opened 2021-07-30
Use custom MSSQL Date/DateTime types

### LINKED ISSUE #92: Correctly format dates for MSSQL
For reasons best known to itself, MSSQL will change it's interpretation of ISO format dates (YYYY-MM-DD) according to it's localisation settings and so will sometimes treat them as a bizarre (and seen nowhere else) hybrid of ISO and US formats: YYYY-DD-MM. (See https://github.com/opensafely-core/cohort-extractor/pull/80)

However, it _will_ consistently and correctly interpret dates in YYYYMMDD format (i.e ISO without the hyphens). 

We obviously don't want to leak this monstrous detail out to study authors and so our approach thus far has been to use ISO format everywhere and then convert to MSSQL-format at the last moment when building the SQL.

We do this by passing every string though this function, which we reformat it if it looks like an ISO date and leave it alone otherwise:
https://github.com/opensafely-core/cohort-extractor/blob/bb5d2e9faa8e3e5aa35824edcfe2e48d088be59c/cohortextractor/tpp_backend.py#L2421-L2434

This has the advantage of being comprehensive. Accidentally passing an ISO date through would lead to very subtle bugs because it's unlikely to trigger an error, just a subtlety wrong date condition. So we really, really don't want to do that. On the other hand, I think there have been one or two cases where this has bitten us because someone has wanted to use date-like-strings as categories and then got very confused.

Theoretically, the new cohortextractor has access to enough type information to know if a value is supposed to be a date or not and format it accordingly. However, propagating this type information through is something we're not yet doing and this problem needs solving now so I suggest we adopt the same approach as the old cohortextractor but hopefully implemented in a such a way that we can surgically swap it out for a "proper" solution when the time comes.

It would also be good to work out how to put the test database into the right localisation mode to trigger this behaviour so we can increase our chances of catching problems early.

evansd:
This PR makes use of SQLAlchemy's type-tracking features to address the date format issue in #92 and hopefully make typing issues in general easier to address.

Note that this also moves us away from interpolating escaped parameters directly into SQL strings (the"literal binds" feature of SQLAlchemy) and instead uses placeholders and parameter passing. This is almost certainly what we want, but it does mean that every database call in the test suite changes its signature.

Also note that I'm not wild about the passing around of the `type_map` dictionary as it further complicates the already unclear (who calls who?) relationship between QueryEngine and Backend. However sorting that out feels like it needs a larger piece of thinking and I'm reasonably confident that this approach won't make that job any harder than it was already.

Closes #92


cohortextractor/backends/base.py
```diff
@@ -43,6 +53,19 @@ class SQLTable:
...
         raise NotImplementedError()
 
+    def _make_columns(self):
+        return [
+            self._make_column(name, column) for name, column in self._columns.items()
+        ]
+
+    def _make_column(self, name, column):
+        source = column.source or name
+        type_ = DEFAULT_SQLALCHEMY_TYPES[column.type]
```
benbc: I've not been happy with any of my solutions to the problem of _really_ wanting to use a name that shadows something from the stdlib like this. Shall we go for underscore suffix as a general solution? It's fine by me.

cohortextractor/query_engines/mssql.py
```diff
@@ -601,44 +656,42 @@ def generate_results_query(self):
...
-            sql.append(f"SELECT * INTO {table.name} FROM (\n{query_sql}\n) t")
+            # This is a bit of a hack but works OK. We want to be able to take
+            # an arbirary select query and generate the SQL:
+            #
+            #   SELECT * INTO some_temporary_table FROM (some_select_query) AS some_alias
+            #
+            # which is the MSSQL-specific syntax for writing the results of a
+            # query directly into a temporary table. We can trick SQLAlchemy
+            # into generating this for us by giving it a literal column named
+            # "* INTO table_name".
```
bloodearnest, line 677: SLOW CLAP

cohortextractor/query_engines/mssql.py
```diff
@@ -49,10 +50,65 @@ def get_primary_table(query):
...
+
+    def process_literal_param(self, value, dialect):
+        """
+        Convert a Python value into an escaped string suitable for
+        interpolating directly into an SQL string
+        """
+        value = self.process_bind_param(value, dialect)
+        # Use the Text literal processor to handle quoting and esacping
+        literal_processor = self.text_type.literal_processor(dialect)
+        return literal_processor(value)
```
benbc, line 85: Why don't we have to do any conversion or checking here?

benbc (approved):
Seems good to me.

> I'm not wild about the passing around of the type_map dictionary

Since the backend holds a reference to the engine, could the custom type mapping be a field on the engine, which the backend grabs when it needs it in `get_table_expression()`?

cohortextractor/backends/base.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, source, columns):
         self._columns = columns
 
     def learn_patient_join(self, source):
-        self._columns["patient_id"] = Column("int", source)
+        self._columns["patient_id"] = Column("integer", source)
```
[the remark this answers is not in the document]
evansd: I don't think these type strings were ever passed to SQLAlchemy -- or indeed anywhere at all.

cohortextractor/backends/base.py
evansd: I'm just following SQLAlchemy here as it's the name it uses whenever you pass type as a keyword argument. I agree that none of the conventions are nice but following SQLAlchemy seems like a reasonable choice.

cohortextractor/query_engines/mssql.py
evansd: That all happens in this line which calls the method defined above:
```
value = self.process_bind_param(value, dialect)
```
So we're doing all the things we normally do to prepare a parameter to be passed to `cursor.execute()` and then we're invoking the Text type to convert the resulting string into something that can be safely interpolated.

I agree I could definitely comment this better!

evansd:
> Since the backend holds a reference to the engine, could the custom type mapping be a field on the engine, which the backend grabs when it needs it

That's a neat suggestion. It would certainly avoid threading that variable through so many function calls. I think ultimately though we're going to want to get rid of that direct reference between the backend and query engine so that we can split out the DSL and query execution into separate packages. The backend will still need to specify its corresponding query engine somehow, but it won't be by importing it. And so I worry that change will be harder to unpick later.

cohortextractor/query_engines/mssql.py
benbc: Oh, I just read through that first line.

benbc:
> we're going to want to get rid of that direct reference between the backend and query engine so that we can split out the DSL and query execution into separate packages

Oh, interesting. Do you anticipate including all the backend-specific stuff in the AST? I haven't thought about this before, but I think I would guess that the full backend definition would be available at runtime.

(On this point in particular, I don't have any problem with the design as you have it -- just suggesting an alternative since you expressed dissatisfaction.)

evansd:
> Do you anticipate including all the backend-specific stuff in the AST?

I wouldn't say I've got a clear plan here exactly. But I think that if we want nice autocompletion and helpful error messages then the DSL package is going to have to know what tables/columns/types are available for each backend.

It's possible that we could provide that by having the Backend classes generate some kind of declarative schema definition (e.g. a JSON file) which the DSL package consumes. Or alternatively, by having the Backends in the DSL package and the Query Engines elsewhere. I haven't thought through the details enough here to have a strong opinion.


> just suggesting an alternative since you expressed dissatisfaction

I think my dissatisfaction is more semantic than syntactic. It's the fact that Backend has a reference to QueryEngine via a class variable, and QueryEngine has a reference to Backend via an instance variable. And now QueryEngine calls a method on Backend passing it a reference to one of its own class variables. I'm sure we can sort it out, I just didn't want to have to think about right now.

PR #113, opened 2021-07-30
Dummy data from file

rebkwok:
This is largely ported over from the existing cohort extractor, and modified since we don't have known types from the Cohort definition, and we only support csvs

Also I noticed a few differences between the real long covid study variable names and our V2 one, so have modified it to match (so that the post-extraction analysis code in the study repo should hopefully just work).  I realised also that BMI isn't actually included in the real study, but I've left it in the tests because we've implemented the simple version of it anyway.


cohortextractor/main.py
```diff
@@ -1,18 +1,33 @@
...
 import importlib.util
 import inspect
+import shutil
 import sys
 from contextlib import contextmanager
+from pathlib import Path
 
-from .backends import BACKENDS
-from .query_utils import get_column_definitions
+from cohortextractor.backends import BACKENDS
```
benbc, line 9: What's the reason for changing from relative to absolute imports here?

cohortextractor/main.py
```diff
@@ -1,18 +1,33 @@
...
-    backend = BACKENDS[backend_id](db_url)
+def main(definition_path, output_file, backend_id, db_url, dummy_data_file=None):
     cohort = load_cohort(definition_path)
-    results = extract(cohort, backend)
-    write_output(results, output_file)
+
+    if dummy_data_file:
+        results = None
+        if Path(output_file).suffixes != dummy_data_file.suffixes:
+            expected_extension = "".join(output_file.suffixes)
```
benbc, line 23: It looks odd to have `Path(output_file).suffixes` in the line above and `output_file.suffixes` here. Is it intentional?

cohortextractor/main.py
```diff
@@ -1,18 +1,33 @@
...
-def main(definition_path, output_file, backend_id, db_url):
-    backend = BACKENDS[backend_id](db_url)
+def main(definition_path, output_file, backend_id, db_url, dummy_data_file=None):
     cohort = load_cohort(definition_path)
-    results = extract(cohort, backend)
-    write_output(results, output_file)
+
+    if dummy_data_file:
+        results = None
+        if Path(output_file).suffixes != dummy_data_file.suffixes:
```
benbc, line 22: Could we include this validation in`validate_dummy_data`? That way we can keep everything in `main()` at a similar, high level of abstraction and all the details hidden in the individual methods.

cohortextractor/main.py
```diff
@@ -54,15 +69,18 @@ def extract(cohort, backend):
             yield dict(row)
 
 
-def write_output(results, output_file):
+def write_output(results, output_file, dummy_data_file=None):
```
benbc, line 72: I don't think that we gain anything from having the two different ways of producing output files in this same function. It would be clearer to keep the conditionals all in one place (in `main()`). Something like this, in pseudo-code:

```
cohort = load_cohort()

if dummy_data:
    validate(dummy_data, cohort)
    copy(dummy_data, output_file)
else:
    results = extract(cohort)
    write(results, output_file)
```

cohortextractor/main.py
```diff
@@ -1,18 +1,33 @@
...
-    results = extract(cohort, backend)
-    write_output(results, output_file)
+
+    if dummy_data_file:
+        results = None
+        if Path(output_file).suffixes != dummy_data_file.suffixes:
+            expected_extension = "".join(output_file.suffixes)
+            msg = f"Expected dummy data file with extension {expected_extension}; got {dummy_data_file}"
+            raise DummyDataValidationError(msg)
+        validate_dummy_data(get_column_definitions(cohort), dummy_data_file)
```
benbc, line 26: This is a tiny thing, but I would move the call to `get_column_definitions()` into the validation function. Again to keep the stuff here at the same level of abstraction.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+            raise DummyDataValidationError(
+                f"Invalid value `{value!r}` for patient_id in row {ix + 2}"
+            )
+
+
+def read_into_dataframe(path):
+    """Read data from path into a Pandas DataFrame."""
+
+    try:
+        if path.suffixes in [[".csv"], [".csv", ".gz"]]:
```
benbc, line 53: Could we remove the duplication between this and `SUPPORTED_FILE_FORMATS`?

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+    Raises DummyDataValidationError if dummy data is not valid.
+    """
+
+    df = read_into_dataframe(dummy_data_file)
+    validate_expected_columns(df, column_definitions)
+
+    for col_name, query_node in column_definitions.items():
+        if col_name == "population":
+            # Ignore the population definition, since it is not used as a column in the
+            # output
```
benbc, line 25: We do this ignoring in two places. Could we just strip it out of the column definitions before validating?

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+            except ValueError:
+                raise DummyDataValidationError(
+                    f"Invalid value `{value!r}` for {col_name} in row {ix + 2}"
+                )
+
+    for ix, value in enumerate(df["patient_id"]):
+        if not isinstance(value, int):
+            raise DummyDataValidationError(
+                f"Invalid value `{value!r}` for patient_id in row {ix + 2}"
+            )
```
benbc, line 46: I think that this and the other place where we add `patient_id` to the expected columns could be collapsed into a single thing by adding it to the column definitions (a bit like my comment about removing population). Then we get rid of a special case.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+def validate_dummy_data(column_definitions, dummy_data_file):
+    """Validate that dummy data provided by user matches expected structure and format.
+
+    Raises DummyDataValidationError if dummy data is not valid.
+    """
+
+    df = read_into_dataframe(dummy_data_file)
+    validate_expected_columns(df, column_definitions)
+
+    for col_name, query_node in column_definitions.items():
```
benbc, line 22: I like the fact that the validation of which columns are present is pulled into its own function. I would pull this loop into its own value-validation function too.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+    """Return function that validates that value is valid to appear in column.
+
+    A validator is a single-argument function that raises a ValueError on invalid input.
+    """
+
+    def bool_validator(value):
+        if value not in [True, False]:
+            raise ValueError
+
+    def date_validator_date_or_datetime(value):
```
benbc, line 98: This name is super-hard to read. Could it be `date_validator`, or `date_or_datetime_validator`?

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+    missing_columns = expected_columns - set(df.columns)
+    if missing_columns:
+        if len(missing_columns) == 1:
+            msg = f"Missing column in dummy data: {list(missing_columns)[0]}"
+        else:
+            msg = f"Missing columns in dummy data: {', '.join(missing_columns)}"
+        raise DummyDataValidationError(msg)
+
+
+def get_csv_validator(col_name, query_node):
```
benbc, line 88: `col_name` isn't used.

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+    for col_name, query_node in column_definitions.items():
+        if col_name == "population":
+            # Ignore the population definition, since it is not used as a column in the
+            # output
+            continue
+
+        validator = get_csv_validator(col_name, query_node)
+
+        for ix, value in enumerate(df[col_name]):
+            if pd.isna(value) or not value:
```
benbc, line 31: I think this will allow `0` as a value for something that should be a date (for example).

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+        col_name
+        for col_name in columns_definitions.keys()
+        if not col_name == "population"
+    }
+    expected_columns.add("patient_id")
+
+    extra_columns = set(df.columns) - expected_columns
+    if extra_columns:
+        if len(extra_columns) == 1:
+            msg = f"Unexpected column in dummy data: {list(extra_columns)[0]}"
```
benbc, line 74: What is `list(extra_columns)[0]` doing here?

cohortextractor/validate_dummy_data.py
```diff
@@ -0,0 +1,131 @@
...
+        else:
+            msg = f"Unexpected columns in dummy data: {', '.join(extra_columns)}"
+        raise DummyDataValidationError(msg)
```
benbc, line 77: I'd just `raise DummyDataValidationError(f"Unexpected column(s) in dummy data: {', '.join(extra_columns)}")`, rather than going to the trouble of this special-case error message. (Or, if you want to keep it, factor out the duplication with the logic below.)

benbc (commented):
Sorry, Becky, there are an awful lot of comments here. I expect that some of them relate to stuff that you're just copying across, but I think it's worth cleaning up as much as possible as we go. Happy to pair on some of this stuff with you tomorrow, if you like, so I can explain what I mean a bit better in person.

rebkwok:
> Sorry, Becky, there are an awful lot of comments here. I expect that some of them relate to stuff that you're just copying across, but I think it's worth cleaning up as much as possible as we go. Happy to pair on some of this stuff with you tomorrow, if you like, so I can explain what I mean a bit better in person.

I think most of these things are fairly subjective, and I don't mind changing to your preference if you feel strongly about them.  Given that the code is largely ported over from the old cohort-extractor, and it was added quite recently, we might want Peter's input on a couple of the comments, especially if there are things that we think might potentially be bugs in the old version.

cohortextractor/csv_utils.py
```diff
@@ -0,0 +1,25 @@
+import csv
+import gzip
+
+
```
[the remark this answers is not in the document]
rebkwok: It is, but in the old cohort extractor, which has different output file formats, it's used in `validate_dummy_data`.  At least, `is_csv_filename` is, and `write_rows_to_csv` is used for writing output files.  Presumably we'll need some version of this later when we support gzip csvs, but I can move it to the tests for now if you prefer.

cohortextractor/main.py
rebkwok: At one point the tests refused to run with the relative imports (even using `just` which should be setting the PYTHONPATH).  I'll see if they'll work if I change them back now

cohortextractor/main.py
rebkwok: no, just a copy/paste from the old cohort extractor (which also has this discrepancy, and I assume also isn't intentional). Both `output_file` and `dummy_data_file` should always be a `Path` (assuming they've been called via __main__.py

cohortextractor/validate_dummy_data.py
rebkwok: I made it this because the old CE has `date_validator_day`, `date_validator_month` and `date_validator_year`, and I hadn't noticed that made it a weird name.  WIll change to `date_validator`, since there's only one now

cohortextractor/validate_dummy_data.py
rebkwok: It's returning the name of the extra column it found

cohortextractor/validate_dummy_data.py
rebkwok: Also taken directly from the old version, but I agree, it can be simplified

cohortextractor/main.py
rebkwok: I was trying to keep `main` doing fewer things, but sure, this is fine too

cohortextractor/validate_dummy_data.py
rebkwok: It doesn't (but I had to write a test to check).  At this point, the `0` in the csv has been read in as a string, so `not "0"` is False

cohortextractor/validate_dummy_data.py
rebkwok: Eventually there will be more supported file formats, and it won't look so duplicated.  It's like this in the old CE:
```
        if path.suffixes in [[".csv"], [".csv", ".gz"]]:
            return pd.read_csv(path)
        elif path.suffixes in [[".dta"], [".dta", ".gz"]]:
            return pd.read_stata(path)
        elif path.suffix == ".feather":
            return pd.read_feather(path)
        else:
            msg = f"Dummy data must be in one of the following formats: {', '.join(SUPPORTED_FILE_FORMATS)}"
            raise DummyDataValidationError(msg)
```

I can change it to 
```
suffixes = ".".join([suffix.strip(".") for suffix in path.suffixes])
if suffixes in SUPPORTED_FILE_FORMATS:
    return pd.read_csv(path)
```

But presumably later we'll have different file extensions to deal with and we'll need something more like this again

cohortextractor/main.py
benbc: I don't mind at all -- just wanted to improve my understanding for this stuff.

cohortextractor/validate_dummy_data.py
benbc: Ah! Good to have the test anyway. Thanks.

cohortextractor/validate_dummy_data.py
benbc: I mean -- the `list()`, can't be necessary, can it?

cohortextractor/validate_dummy_data.py
benbc: Even then, I think we'd want to somehow pull the list of extensions together with the handling code to help us check they're not inconsistent. Just putting the definitions next to each other would help.

cohortextractor/validate_dummy_data.py
rebkwok: It is, because `extra_columns` is a set

rebkwok:
@benbc I think I've addressed all the comments now.  In the end, the check for null values is the only one that I thought could impact the old cohort extractor, and that's actually fine

cohortextractor/validate_dummy_data.py
benbc: Ah! Thank you.

benbc (approved):
That looks great. Thank you.

PR #115, opened 2021-08-02
Categorisation on ValueFromFunction values

rebkwok:


tests/acceptance/test_simplified_long_covid_study.py
```diff
@@ -99,12 +99,24 @@ def test_simplified_cohort(database, setup_tpp_database):
...
+            "1990-1-1",
+            registration(start_date="2001-01-01", end_date="2002-02-02"),
+        ),
     )
+    # try:
+    #     res = extract(SimplifiedCohort, TPPBackend, database)
+    # except Exception as e:
+    #     import ipdb; ipdb.set_trace()
+    #     pass
+
```
benbc, line 114: I think this was probably left in by mistake.

cohortextractor/query_engines/mssql.py
```diff
@@ -171,12 +170,21 @@ def get_query_nodes_from_category_definitions(self, definitions, query_nodes=Non
...
         definition (Comparator) has a LHS which is not itself a Comparator,
-        it will be a query node
+        it will be a query node or a function-generated node
         """
         query_nodes = query_nodes or set()
         for definition in definitions:
             if isinstance(definition.lhs, Value):
-                query_nodes.add(definition.lhs)
+                # A ValueFromFunction is not itself a QueryNode, so we look for referenced
+                # nodes in its arguments
```
benbc, line 179: I think this is just another signal that our type hierarchy is messed up, but looking at the code I see that a `ValueFromFunction`_is_ a `QueryNode` (it's a subclass of `Value`, which is a subclass of `QueryNode`).

cohortextractor/query_engines/mssql.py
```diff
@@ -171,12 +170,21 @@ def get_query_nodes_from_category_definitions(self, definitions, query_nodes=Non
...
+                parent_nodes = (
+                    definition.lhs.arguments
```
benbc, line 181: Why do we never need to unwrap the arguments for a RHS?

cohortextractor/query_engines/mssql.py
```diff
@@ -355,10 +364,8 @@ def build_condition_statement(self, comparator, tables):
             connector = getattr(sqlalchemy, comparator.connector)
             condition_statement = connector(left_conditions, right_conditions)
         else:
-            query_node = comparator.lhs
-            table = tables[query_node]
```
benbc, line 359: This `tables` argument is no longer used, I think.

tests/acceptance/test_simplified_long_covid_study.py
rebkwok: ah, yes :facepalm:

cohortextractor/query_engines/mssql.py
rebkwok: You're right, it is a QueryNode, but it's not a of a thing-we-need-to-add-temp-tables-for.  I'll update the comment

cohortextractor/query_engines/mssql.py
rebkwok: Because it's never a Value.  It's always either a simple str/int/date/None or a Comparator. If it's a Comparator, it gets fed back into this to get the parents from its LHS.

cohortextractor/query_engines/mssql.py
rebkwok: :+1:

cohortextractor/query_engines/mssql.py
benbc: I see. Thanks.

benbc (approved):

PR #116, opened 2021-08-03
Don't write cohort definition bytecode

### LINKED ISSUE #101: Don't write bytecode files to the analysis directory, so that the researcher doesn't have to wonder what they are and whether they should be commited.

benbc:
Fixes #101.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # hadolint ignore=DL3007
 FROM ghcr.io/opensafely-core/base-docker:latest
 
+# hadolit ignore=DL3008
 RUN \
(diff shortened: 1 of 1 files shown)


evansd (approved):
:+1: 

That's a nice, and unexpected, win from moving the entrypoint.

By the way, something odd happened with the commit message here (I feel like I might have seen this before):
https://github.com/opensafely-core/cohort-extractor-v2/pull/116/commits/f8cd354a15365ff681e737a56292c2ba8b6c2161

benbc:
@evansd Thanks. Fixed the commit message. The speed-up is just from the overhead of copying images to create the last few trivial layers.

PR #118, opened 2021-08-03
Ignore dummy-data-file arg on real backend

rebkwok:
This means that the `run` command in a project.yaml can specify --dummy-data-file and it will only be used in a local run.  On a real backend (where `DATABASE_URL` is set), it will be ignored.

### DIFF
--- a/cohortextractor/main.py
+++ b/cohortextractor/main.py
@@ -14,7 +14,7 @@ def main(definition_path, output_file, backend_id, db_url, dummy_data_file=None)
     cohort = load_cohort(definition_path)
 
     output_file.parent.mkdir(parents=True, exist_ok=True)
-    if dummy_data_file:
+    if dummy_data_file and not db_url:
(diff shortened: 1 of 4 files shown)


benbc (approved):

cohortextractor/main.py
```diff
@@ -14,7 +14,7 @@ def main(definition_path, output_file, backend_id, db_url, dummy_data_file=None)
     cohort = load_cohort(definition_path)
 
     output_file.parent.mkdir(parents=True, exist_ok=True)
-    if dummy_data_file:
+    if dummy_data_file and not db_url:
```
evansd, line 17: This looks good.

One thing it might be worth doing is logging the fact that a dummy data file was supplied but it's not being used because database url was set. And by logging I think I just mean `print()` at this stage :) Anything written to stdout or stdin will end up in the job logs.

PR #120, opened 2021-08-04
Fix SQL Server compatibility level

benbc:
Fixes #88.

PR #121, opened 2021-08-04
Fix "hadolit" > "hadolint" in Dockerfile

StevenMaude:


StevenMaude:
(Spotted while looking to see what the fuss of #120 was about :smile:)

inglesp (approved):

PR #132, opened 2021-08-06
Ensure that cohorts contain only one row per patient

### LINKED ISSUE #129: Enforce patient uniqueness in a cohort

benbc:
We do this in the simplest way possible -- by asserting that every
cohort variable is a Value, rather than a Column. This is what
researchers will want, but unfortunately a whole bunch of our tests have
taken advantage of the fact that we could previously return Columns.

Mostly the tests are fixed by a judicious sprinkling of uniqueifying
operators, but there are a couple that needed significant
restructuring.

Working on the filter tests, I found it hard to reason about the tests
with the test data separated from the results, so I've restructured
those tests to bring the two together (and minimized the data while I
was at it).

Fixes #129.

### DIFF
--- a/cohortextractor/query_utils.py
+++ b/cohortextractor/query_utils.py
@@ -1,11 +1,17 @@
+from .query_language import Value
+
+
 def get_class_vars(cls):
     default_vars = set(dir(type("ArbitraryEmptyClass", (), {})))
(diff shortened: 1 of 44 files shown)


evansd (approved):
:+1: Thanks for doing this. It must've been pretty tedious, but I appreciated the drive-by testing improvements you did along the way :)

PR #133, opened 2021-08-06
Work around EOF errors from database server

evansd:
This PR attempts to work around the intermittent problems we have downloading large results sets from the database. It adopts largely the same approach to this as we did in cohortextractor v1 (see [#286](https://github.com/opensafely-core/cohort-extractor/pull/286) and [#322](https://github.com/opensafely-core/cohort-extractor/pull/322) for the full inglorious history) with two main changes:
 1. We try to package up the gnarly logic into a single function call so the effect on the rest of the codebase is as minimal as possible.
 2. We use the SQLAlchemy API, rather than previous raw SQL string approach.

The basic approach is, instead of fetching a queries results directly, we write them to a temporary table and then download them in small batches (paginating on the `patient_id` column), sleeping briefly and retrying if we encounter an error.

Sometimes retrying a query using the same connection works, however we have hit cases where the connection needs to be reestablished in order to continue. This means we can't use session-scoped temporary tables which disappear when the connection dies and therefore that we have to write our results to persistent tables which we then delete once the download is complete.

A further wrinkle this creates is that in order to test whether a temporary table exists we have to attempt to read from it and catch the exception (the privileges granted to our production database user aren't sufficient to use other methods). But the database record/playback framework used to make local testing fast can't record and replay exceptions. So we have to mark such tests as xfail in playback mode, though they still run fully in CI.


evansd:
@benbc Unfortunately this PR has run into a limitation of the database recording system. Because we don't have sufficient permission to check whether a table exists in any of the "[normal](https://stackoverflow.com/questions/167576/check-if-table-exists-in-sql-server)" ways we're forced to do so by trying to query it and catching the exception:
https://github.com/opensafely-core/cohort-extractor-v2/blob/2bf0a20f4a27da52d103e2de208c5cbd5e7e619c/cohortextractor/query_engines/mssql_lib.py#L283-L291

However the playback/record library doesn't (for understandable reasons) capture and reply exceptions. So it doesn't behave equivalently in playback mode.

I wonder if we need some form of escape hatch which we can use to exempt certain tests from the playback mechanism and say that they either run against a real database or get skipped.

It hit another less serious issue with test playback here, and although the workaround is fairly simple it is not a thing of beauty:
https://github.com/opensafely-core/cohort-extractor-v2/blob/2bf0a20f4a27da52d103e2de208c5cbd5e7e619c/cohortextractor/query_engines/mssql_lib.py#L308-L316

cohortextractor/query_engines/mssql_lib.py
```diff
@@ -0,0 +1,339 @@
...
+    to workaround flakiness in our connection to the database.
+
+    Args:
+        engine: sqlalchemy.Engine instance
+
+        queries: list of sqlalchemy queries, with the final being a Select
+
+        temp_table_prefix: prefix to use when generating temporary table names
+            use a leading `#` character for session-scoped temporary tables,
+            use the `DatabaseName..table_prefx` syntax to store tables in
```
inglesp, line 54: Is the `..` a typo?

cohortextractor/query_engines/mssql_lib.py
```diff
@@ -0,0 +1,339 @@
...
+    key = sqlalchemy.Column(key_column)
+    retries = 0
+    batch_count = 0
+    total_rows = 0
+    min_key = None
+
+    while True:
+        query = (
+            sqlalchemy.select("*").select_from(table).order_by(key).limit(batch_size)
+        )
```
inglesp, line 165: Unless there's an index on `key`, I think this might be accidentally quadratic.  Not sure how much of a problem this is.

tests/query_engines/test_mssql_lib.py
```diff
@@ -0,0 +1,215 @@
...
+    engine = database.engine()
+    with engine.connect() as conn:
+        _populate_table(conn, table, test_data)
+
+    query = sqlalchemy.select("*").select_from(table)
+    with fetch_results_in_batches(
+        engine,
+        [query],
+        # We have to use a persistent table to write results to because we
+        # going to close and reopen the db connection
```
inglesp, line 75: *are going

cohortextractor/query_engines/mssql_lib.py
inglesp: Looks like it's not.  Might be worth a brief comment for the uninitiated?

inglesp (approved):
Some serious (and seriously good) software engineering going on here.

cohortextractor/query_engines/mssql_lib.py
evansd: something, something mssql something ...

I'll explicitly refer to it as the double-dot syntax in the comment so it's clear it's not a typo.

cohortextractor/query_engines/mssql_lib.py
evansd: You're quite right. We do create the appropriate index on the temporary table before passing it through here. But I'll add a note to the docstring in case we end up using this function in other contexts.

PR #144, opened 2021-09-01
feat: use Github cacheing of pip's cache

madwort:
* only a marginal improvement to CI times (a few seconds), but keen to have this best practice be widespread around our repos

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,12 @@ jobs:
       - uses: "actions/setup-python@v2"
         with:
           python-version: "3.9"
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
       - uses: extractions/setup-just@v1
       - name: Check formatting, linting and import sorting
         run: ./scripts/do check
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):

.github/workflows/main.yml
```diff
@@ -13,6 +13,12 @@ jobs:
       - uses: "actions/setup-python@v2"
         with:
           python-version: "3.9"
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
ghickman, line 19: This should probably be hashing both `requirements.prod.txt` and `requirements.dev.txt` (it's not hashing anything atm!).

ghickman (changes requested):

.github/workflows/main.yml
madwort: doh! you're right - it was using the fallback key

iaindillingham: I missed that -- how embarassing. Sorry!

ghickman: I just did exactly the same thing in another repo, only reason I caught it!

madwort:
closing this PR, because I'm not convinced it's worth it - it doesn't actually speed up pip very much - if we can actually cache venvs that would make a much bigger difference

PR #150, opened 2021-09-08
Make tests backend-agnostic 

rebkwok:
This PR mostly just moves the end-to-end test files around and modifies fixtures to allow a custom backend to be specified rather than always defaulting to `tpp`

I had been planning to just update the tests to allow for different backends, but as part of checking that worked I added a very minimal placeholder graphnet backend to test that it was working (obviously just a placeholder, since I don't know the graphnet schema).  This is probably useful as a starting point for graphnet backend tests, so I've left it in.


tests/end_to_end/test_end_to_end_tpp.py
```diff
@@ -0,0 +1,58 @@
...
+    assert_results_equivalent(actual_results, study.expected_results())
+
+
+@mark_xfail_in_playback_mode
+@pytest.mark.integration
+def test_extracts_data_from_sql_server_ignores_dummy_data_file(
+    load_study, setup_backend_database, cohort_extractor_in_process
+):
+    # A dummy data file is ignored if running in a real backend (i.e. DATABASE_URL is set)
+    # THis provides an invalid dummy data file, but it is ignored so no errors are raised
```
madwort, line 52: typo `THis`

benbc (approved):

madwort (approved):

PR #152, opened 2021-09-16
Adding the Measures framework, part 1

rebkwok:
This ports over the Measures framework from V1, but currently doesn't implement the index-date-range part (which involves changes to the cohort generation as well as running the measures).

We discussed previously in V2 that filepaths should be specified explicitly rather than implied, so the `generate-measures` command also requires that you explicitly state where the input files come from.  Also means that measures can now be based on a single input file rather than a date series (that may or may not be a useful thing).


cohortextractor/log_utils.py
```diff
@@ -0,0 +1,63 @@
+import logging.config
```
rebkwok, line 1: copied over from V1

cohortextractor/measure.py
```diff
@@ -0,0 +1,236 @@
+import typing
```
rebkwok, line 1: This file is mostly just copied over from V1; the `MeasuresManager` is new, but basically does what V1 does outside of `measures.py` to load an input dataframe and calculate each measure from it

cohortextractor/main.py
```diff
@@ -29,30 +35,37 @@ def generate_cohort(
...
+        results.to_csv(measure_output_file, index=False)
+        log.info("Created measure output", output=output_file)
+
+
+def calculate_measures_results(cohort, input_file, patient_dataframe=None):
+    measures = get_measures(cohort)
+    measures_manager = MeasuresManager(measures, input_file, patient_dataframe)
+    assert (
+        measures_manager.patient_dataframe is not None
+    ), "No matching input file found. You may need to first run:\n  cohortextractor generate_cohort ..."
```
benbc, line 53: It seems a shame to expose this `patient_dataframe` property just to do the assertion, when the rest of the interface to that object is so simple. Could we do the assertion inside `MeasureManager` instead?

benbc (approved):
This looks great.

I think I'd be tempted to ask @evansd to cast an eye over it to see if there is anything that you're porting over which he really wishes we could dispose of or replace. If there is anything then it could always go into another issue.

cohortextractor/main.py
rebkwok: Yes - in fact it more or less does already, I'll move it over

rebkwok: Although I vaguely think that the MeasuresManager only takes a `patient_dataframe` parameter in order to allow it to be passed in tests; I'll have another look at that too

cohortextractor/log_utils.py
evansd: I can't remember who wrote this, though I'm pretty sure it wasn't me. It all looks pretty sensible though so I don't see a problem with copying it verbatim for now.

evansd (approved):
This looks like a sensible port of the previous code. I've mentioned one small feature in the old code that doesn't actually get used, but doesn't do any harm either.

> the generate-measures command also requires that you explicitly state where the input files come from. Also means that measures can now be based on a single input file rather than a date series (that may or may not be a useful thing).

I really like this.

cohortextractor/__main__.py
```diff
@@ -15,34 +15,81 @@ def existing_python_file(value):
...
     type=Path,
 )
-parser.add_argument(
+generate_cohort_parser.add_argument(
     "--dummy-data-file",
     help="Provide dummy data from a file to be validated and used as output",
     type=Path,
 )
+generate_cohort_parser.add_argument(
+    "--skip-existing",
```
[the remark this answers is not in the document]
rebkwok: Ah, actually I meant to ask you about this - I couldn't find where it was used in a real study and I'd more or less come to the conclusion it wasn't.  I haven't implemented it here, I'd just left the arg in - in the next PR I took it out and I was going to check if it was still a thing

PR #154, opened 2021-09-22
OS-38: Implement index-date-range for generate_cohort

rebkwok:
Implements the index-date-range option for generating cohort (measure to come later).

I dithered a bit about how to implement this. Instead of specifying "index date" as a string everywhere, the index-date-range requires you to define a `BASE_INDEX_DATE` which can be None or a normal date string.  Then you can use the `BASE_INDEX_DATE` in value definitions, e.g.

```
class Cohort:

    BASE_INDEX_DATE = "2021-08-01"
    population = table("practice_registrations").date_in_range(BASE_INDEX_DATE).exists()

````
This can then be run without `--index-date-range`, in which case it will use the date defined in the Cohort class, or with e.g. `--index-date-range "2021-01-01 to 2021-03-01"`, which will create 3 cohorts, using the dates 2021-01-01, 2021-02-01 and 2021-03-01.

It does this by making a copy of the study definition file, replacing the `BASE_INDEX_DATE` definition, and then loading that module instead of the original study definition file.  That avoids having to do the recursive walking the query nodes to find and replace all instances of the "index_date" string.

I'm not 100% sure about this approach, but it feels like it simplifies things; it means that researchers can do date arithmetic in the study definition file if they want; and/or we can define date functions that they can import and use directly with an index date (e.g. `last_day_of_month(BASE_INDEX_DATE)` rather than the string `"last_day_of_month(index_date)"` string that has to be parsed and converted to a date function.  I think it should simplify the date expressions.

### CODE REVIEW

File: cohortextractor/date_utils.py [hunk 1]
```diff
@@ -0,0 +1,71 @@
+import calendar
+import datetime
+
+
+def generate_date_range(date_range_str):
```
rebkwok, line 5: This is ported from V1 almost entirely, with the addition of a check for valid end of month dates on line 61

cohortextractor/main.py
```diff
@@ -32,18 +37,102 @@ def generate_cohort(
...
+    else:
+        start = end = date_range_str
+    try:
+        start = _parse_date(start)
+        end = _parse_date(end)
+    except ValueError:
+        raise ValueError(
+            f"Invalid date range '{date_range_str}': Dates must be in YYYY-MM-DD "
+            f"format or 'today' and ranges must be in the form "
+            f"'DATE to DATE by (week|month)'"
```
benbc: I think if I got this error I would interpret it to mean that the `by` clause was mandatory.

cohortextractor/main.py
```diff
@@ -32,18 +37,102 @@ def generate_cohort(
...
+        raise ValueError(
+            f"Invalid date range '{date_range_str}': end cannot be earlier than start"
+        )
+    dates = []
+    while start <= end:
+        dates.append(start.isoformat())
+        start = _increment_date(start, period)
+    # The latest data is generally more interesting/useful so we may as well
+    # extract that first
+    dates.reverse()
```
benbc: This seems _so_ unlikely to be genuinely useful. I bet it originates from some earlier, unreliable version of the system where just getting _something_ out was considered a wild succcess. @evansd?

cohortextractor/main.py
```diff
@@ -74,19 +163,42 @@ def load_cohort(definition_path):
...
+                # with the current index date and load this modified module
+                contents = orig.read()
+                pattern = r"(BASE_INDEX_DATE\s*=\s*.+)([\s|\n].*)"
+                contents = re.sub(
+                    pattern, rf'BASE_INDEX_DATE = "{index_date}"\2', contents
+                )
+                new.write(contents)
+            module = _load_module(temp_path)
+        return module
+    return _load_module(definition_path)
```
benbc, line 127: I found this a bit hard to follow with the nesting. What if the `None` case was handled as a guard clause? (If it is `None` then return immediately.)

cohortextractor/main.py
```diff
@@ -74,19 +163,42 @@ def load_cohort(definition_path):
...
+        with temp_definition_path(definition_path, index_date) as temp_path:
+            with open(definition_path, "r") as orig, open(temp_path, "w") as new:
+                # Copy the cohort definition file, replace the BASE_INDEX_DATE definition
+                # with the current index date and load this modified module
+                contents = orig.read()
+                pattern = r"(BASE_INDEX_DATE\s*=\s*.+)([\s|\n].*)"
+                contents = re.sub(
+                    pattern, rf'BASE_INDEX_DATE = "{index_date}"\2', contents
+                )
+                new.write(contents)
```
benbc, line 124: It seems such a shame to do this code-mangling when we've put so much effort into doing everything else "in code". I don't have an immediate alternative suggestion that I love, though.

It seems as though it would require either a super-class for `Cohort`s or reference to the index date as a function rather than a variable (or both). We're paying a cost here of having the `Cohort` be a class rather than a function that we could easily parameterize.

Oh -- what if we created subclasses of the `Cohort` at runtime with different values for the index date class field? Could that work?

This may be worth a design discussion with @evansd since it's so fundamental to the cohort extractor design.

benbc (commented):
The tests for the date range generation seem very heavy-weight since they're running the actual cohort extraction (if I've understood it correctly). Could we instead thoroughly unit test the date range generation in isolation and then have a single test that checks that it's all wired together correctly and results in multiple outputs?

I expect that a few of my comments are on things you've just pulled over directly from the old implementation. I've tried to keep those to a minimum.

cohortextractor/main.py
evansd: I think that, as with the `--keep-existing-files` option, this originates from pre-job-runner days where output was going to a researcher's working directory on TPP. Under those circumstances you could start inspecting output before the extraction had finished running and so this made more sense. As Ben says, it's pointless now.

cohortextractor/main.py
rebkwok: This is how it currently is in V1, but I agree.  The error message could be more specific to what's actually idenitified as wrong

cohortextractor/main.py
rebkwok: OK, will ditch it

cohortextractor/main.py
rebkwok: Yes, I struggled for a while with how to get the new class.  We could have a Cohort class that users have to inherit from; that would also avoid the issue of expecting a single class in the study definition.   Referencing the index date as a function doesn't work because it still gets executed when the module is loaded.  I think that might have similar issues with creating subclasses of the Cohort with different index dates.  I kept wanting an __init__ method

evansd: There's a tension here I suppose. There's a lot that's wrong about the existing measures system, both in API and implementation. Some of it made sense given the way things were previously done but now doesn't. Other bits never really made sense, but were the result of decisions made very quickly and then never revisited.

So we definitely don't want to preserve it all in the new world. But equally, we don't want to be forced to design the perfect One True measures system as part of getting a proof-of-concept running in Graphnet. Particularly as it involves thinking through a whole set of wider issues like whether we should be using classes as the language construct for defining studies.

If we accept that whatever API we implement is going to be temporary (even if it's an exact match for the existing API) then maybe we should focus on simplicity of implementation, even if the result isn't particularly user friendly. Wrapping the whole `Cohort` definition in a function which takes an `index_date` argument might be the simplest thing from this point of view.

It also worth saying that the whole idea of supplying a date range as command line arguments, rather than having that defined in the study_definition file along with the measures is a historical artefact that I can't quite remember the reason for. I think it makes more sense to have everything defined together. I only mention that in case it makes implementation any easier; there's no particular need to change it now.

rebkwok: Interesting - it probably does make it simpler to put the date range in the study definition (and also clearer to researchers looking at the code).  The main reason I can think of for having it as a command line argument is that you can run a study on different ranges without having to commit code changes first.  Although, I suppose if you're running it through job-server you still have to commit changes to the project.yaml, so maybe that's irrelevant.

I'll have a look at some other options and see if I can find the simplest thing

benbc: Oh, _please_ can we have the date range specified in just the cohort definition? If there isn't a strong use case for doing it at the command line then that's a huge improvement -- both for the implementation and for researchers (because they only have to put it in one place and because we can have a nicely-typed API for it rather than a string they can easily mis-spell.

rebkwok:
@evansd @benbc I've pushed a wip commit that modifies the study definition to include the date range, and be a function that takes the index date and returns the Cohort class (which is very much along the lines of what Dave has also come up with in https://github.com/opensafely-core/cohort-extractor-v2/issues/155). 

Can you take a look before I go any further (it'll need lots of tests updating)?  I've updated fixtures and run it with these tests so far to check it's working:

- `tests/end_to_end/test_end_to_end_tpp.py::test_extracts_data_from_sql_server_integration_test`
- `tests/end_to_end/test_end_to_end_index_date_range.py::test_extracts_data_with_index_date_range_integration_test`

cohortextractor/main.py
```diff
@@ -24,19 +25,55 @@ def generate_cohort(
...
+                db_url, temporary_database=temporary_database
+            )
+            results = extract(cohort, backend)
+            write_output(results, output_file_with_date)
 
 
 def generate_measures(definition_path, input_file, output_file):
-    cohort = load_cohort(definition_path)
+    definition_module = load_module(definition_path)
+    cohort_generator, _ = load_cohort_generator(definition_module)
```
rebkwok, line 75: This will need to deal with index dates too, but that's the next PR

cohortextractor/main.py
```diff
@@ -51,28 +88,56 @@ def calculate_measures_results(cohort, input_file):
...
 
-def load_module(definition_path):
-    definition_dir = definition_path.parent
-    module_name = definition_path.stem
+def load_cohort_functions(definition_module):
+    return [
+        obj
+        for name, obj in inspect.getmembers(definition_module)
+        if inspect.isfunction(obj) and obj.__name__ == "cohort"
+    ]
```
rebkwok, line 104: These now require that you name your class/function `Cohort`/`cohort`; previously when we just had the class, we allowed people to name the class anything, and we just looked for a class in the module.  That did mean that we had to explicitly exclude certain imported classes (e.g. `Measure`) and could mean that a cohort definition breaks if a research imports something unexpected.  I think it's probably acceptable to retrict the class/function name.

benbc: I think it may even be an improvement to have a fixed name.

benbc (approved):
Great.

PR #158, opened 2021-09-27
OS-31

mikedferris:
Completed graphnet backend and tests


cohortextractor/backends/graphnet.py
```diff
@@ -5,11 +5,69 @@
...
+    )
+
+    clinical_events = MappedTable(
+        source="TRE.ClinicalEvents",
+        columns=dict(
+            code=Column("varchar", source="CTV3Code"),
+            date=Column("datetime", source="ConsultationDate"),
+            numeric_value=Column("float", source="NumericValue"),
+        ),
+    )
```
rebkwok, line 26: This makes me realise that we're treating the `clinical_events` table as a single table, however in v2 so far we've only defined the ctv3 part of it.  TPP is set up similarly, with 2 different tables for ctv3 and snomed, so we'll need to address how to deal with those for both backends.  I think we can merge this PR with the current setup and deal with that on another ticket

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,59 @@
+import pytest
```
rebkwok, line 1: This test file seems almost the same as the one in `tests/end_to_end/test_end_to_end_graphnet.py`.  I think what you had here before was a subset of the type of tests in `tests/test_tpp_backend.py`, which test just the data extraction from the database using minimal `Cohort` definitions.  

We want to be sure that the data in the graphnet database can be extracted properly, so we want to set up test data that's in the same format as we expect in the graphnet database, and check that the cohort-extractor can retrieve it.  E.g. I know that TPP never has null dates, so if there's a patient in the `practice_registrations` table with no end date, it'll be entered as "9999-12-31", however Graphnet would have a null value there, so we want to test something like:

```
@pytest.mark.integration
def test_registration_dates(database, setup_backend_database):
    setup_backend_database(
        Patient(Patient_ID=1),
        RegistrationHistory(Patient_ID=1, StartDate="2001-01-01", EndDate=None),
        backend="graphnet"
    )

    class Cohort:
        _registrations = table("practice_registrations").first_by("patient_id")
        arrived = _registrations.get("date_start")
        left = _registrations.get("date_end")

    assert extract(Cohort, GraphnetBackend, database) == [
        dict(patient_id=1, arrived=datetime(2001, 1, 1), left=None)
    ]
```

and also (since I think the `date_in_range` method is likely to break with null end dates):
```
@pytest.mark.integration
def test_registration_dates(database, setup_backend_database):
    setup_backend_database(
        Patient(Patient_ID=1),
        RegistrationHistory(Patient_ID=1, StartDate="2001-01-01", EndDate=None),
        Patient(Patient_ID=2),
        RegistrationHistory(Patient_ID=2, StartDate="2001-01-01", EndDate="2020-01-01),
        backend="graphnet"
    )

    class Cohort:
        _registrations = table("practice_registrations").date_in_range("2021-01-01")
       population = _registrations.exists()

    assert extract(Cohort, GraphnetBackend, database) == [
        dict(patient_id=1)
    ]
```

rebkwok (changes requested):
The backend setup looks good to me - I have one comment but it's about another cohort-extractor ticket we'll need, rather than the graphnet backend itself. 
I think the two test files have somehow become duplicates of the end-to-end tests, and it'll be important to make sure we've got unit tests to test the expected format of the graphnet backend data

cohortextractor/backends/graphnet.py
mikedferris: In our database CTV3 and Snomed are all in the same table, but have been implemented in the backend separately just for consistency with TPP, but my preference would definitely be to implement it as a single entity

rebkwok: :+1: Let's leave it as it is for now and we'll bear that in mind when adding the feature; I think it makes sense to treat them as one table from a user perspective

tests/test_graphnet_backend.py
mikedferris: Just looked through this now - think there must have been some copy and paste problems happening somewhere as I definitely had a different set of tests. Let me look into that...

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
...
+        Patients(Patient_ID=1),
+        PracticeRegistrations(
+            Patient_ID=1, StartDate="2001-01-01", EndDate="2012-12-12"
+        ),
+        PracticeRegistrations(Patient_ID=1, StartDate="2013-01-01"),
+        backend="graphnet",
+    )
+
+    class Cohort:
+        _registrations = table("practice_registrations").first_by("patient_id")
```
rebkwok, line 49: Could we also add a test that uses a Cohort like
```
    class Cohort:
        _registrations = table("practice_registrations").date_in_range("2014-01-01")
        arrived = _registrations.first_by("patient_id").get("date_start")
        left = _registrations.first_by("patient_id").get("date_end")
```
?

The `date_in_range` bit should select only registrations where a patient was registered on 2014-01-01 (i.e. the entry with a null end date).

It _should_ return
`dict(patient_id=1, arrived=datetime(2013, 1, 1), left=None)`

but I am fairly certain it'll fail and it won't select the entry with no end date (and that's something we need to fix in the query language/engine).

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
+from datetime import date, datetime
+
+import pytest
+from lib.graphnet_schema import (  # ClinicalEventsSnomed,; Hospitalizations,; PatientAddress,; hospitalization,
```
rebkwok, line 4: could you clean up the commented out imports?

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
...
+    Patients,
+    PracticeRegistrations,
+    patient,
+    patient_address,
+    registration,
+)
+from lib.util import extract
+
+from cohortextractor import table
+#  from cohortextractor import codelist, table
```
rebkwok, line 16: here too

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
...
+def test_covid_test_positive_result(database, setup_backend_database):
+    setup_backend_database(
+        Patients(Patient_ID=1),
+        PracticeRegistrations(
+            Patient_ID=1, StartDate="2001-01-01", EndDate="2026-06-26"
+        ),
+        CovidTestResults(
+            Patient_ID=1,
+            #  Remove the next line - not relevant to graphnet schema
+            # Organism_Species_Name="SARS-CoV-2",
```
rebkwok, line 68: more comments to delete

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
...
+def test_covid_test_negative_result(database, setup_backend_database):
+    setup_backend_database(
+        Patients(Patient_ID=1),
+        PracticeRegistrations(
+            Patient_ID=1, StartDate="2001-01-01", EndDate="2026-06-26"
+        ),
+        CovidTestResults(
+            Patient_ID=1,
+            #  Remove the next line - not relevant to graphnet schema
+            # Organism_Species_Name="SARS-CoV-2",
```
rebkwok, line 98: "

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,423 @@
...
+        _patients = table("patients").first_by("patient_id")
+        sex = _patients.get("sex")
+        dob = _patients.get("date_of_birth")
+
+    assert extract(Cohort, GraphnetBackend, database) == [
+        dict(patient_id=1, sex="F", dob=date(1950, 1, 1))
+    ]
+
+
+""" Removing hospitalization tests - code parsing doesn't occur in the graphnet backend, so these won't work
```
rebkwok, line 138: it's probably worth having one test that tests the graphnet hospitalization tests, although it won't need to check code parsing like these ones do

tests/test_graphnet_backend.py
rebkwok: If it fails as I think it will, we can either leave the test for it commented out and make a note of it in the new ticket to fix it, or mark it as an xfail (with `@pytest.mark.xfail`) for now

rebkwok (changes requested):
This looks good - just a few comments cleaning up commented out things, and a couple of additional test suggestions

tests/test_graphnet_backend.py
mikedferris: Added test_hospitalization_table_returns_admission_date_and_code

tests/test_graphnet_backend.py
```diff
@@ -0,0 +1,442 @@
...
+        _hospitalization = table("hospitalizations").first_by("patient_id")
+        admission = _hospitalization.get("date")
+        code = _hospitalization.get("code")
+
+    assert extract(Cohort, GraphnetBackend, database) == [
+        dict(patient_id=1, admission=date(2020, 12, 12), code="xyz")
+    ]
+
+
+""" Removing hospitalization tests - code parsing doesn't occur in the graphnet backend, so these won't work
```
rebkwok, line 182: Could you remove these commented out tests that we don't need? Otherwise it all LGTM now :tada:

rebkwok (approved):

PR #161, opened 2021-09-30
OS-39: SRO measures acceptance tests for TPP backend

rebkwok:
When the graphnet backend is merged, the test can be run with a graphnet backend setup as well


cohortextractor/measure.py
```diff
@@ -160,7 +160,8 @@ def _suppress_column(self, column, data, reporter):
         reporter(f"Additional suppression in column {column} in measure {self.id}")
 
     def _calculate_results(self, data):
-        data["value"] = data[self.numerator] / data[self.denominator]
+        if self.numerator in data:
```
rebkwok, line 163: This will almost certainly only happen in tests

benbc (approved):
This is very exciting. Maybe you could show it to one of the researchers who works on SRO?

PR #162, opened 2021-09-30
Fix bug in csv-combining

rebkwok:
Make sure csv-combining is only done for datestamped files that match the measure id exactly


benbc (approved):
Any chance of a unit test to cover this case?

rebkwok:
> Any chance of a unit test to cover this case?

Done, thanks

PR #165, opened 2021-10-01
OS-65: Support for Snomed codes in the clinical events table

rebkwok:
There are a lot of files in this PR, but mostly they're in tests because there were a bunch that needed updated now that a code filter has to be on a codelist.  
Main changes:
- TPP backend: `clinical_events` table is now a QueryTable that combines the ctv3 and snomed tables, and adds a `system` column
- mssql query engine: Update to the `apply_filter` method to add the filter on system as well as code for codelists
- query language: Added a validation for code filters


cohortextractor/backends/tpp.py
```diff
@@ -108,9 +113,10 @@ class TPPBackend(BaseBackend):
         columns=dict(
             date=Column("date"),
             code=Column("varchar"),
+            system=Column("varchar"),
         ),
         query=f"""
-            SELECT Patient_ID as patient_id, Admission_Date as date, {rtrim("fully_split.Value", "X")} as code
+            SELECT Patient_ID as patient_id, Admission_Date as date, {rtrim("fully_split.Value", "X")} as code, 'icd10' as system
```
benbc, line 119: Well spotted.

tests/lib/tpp_schema.py
```diff
@@ -56,6 +56,22 @@ def event(code, date, numeric_value=None):
     return Events(CTV3Code=code, ConsultationDate=date, NumericValue=numeric_value)
```
benbc, line 56: I know it will blow the diff up, but maybe `Events` and `event` need to be renamed for clarity and consistency?

tests/fixtures/end_to_end_tests_measures/measures_cohort.py
```diff
@@ -1,12 +1,14 @@
...
+from cohortextractor import Measure, codelist, table
 
 
 class Cohort:
     _clinical_events = table("clinical_events")
     _registrations = table("practice_registrations")
     population = _registrations.exists()
     practice = _registrations.first_by("patient_id").get("pseudo_id")
-    has_event = _clinical_events.filter(code="abc").first_by("patient_id")
+    has_event = _clinical_events.filter(code=codelist(["abc"], system="ctv3")).first_by(
```
benbc, line 9: I wonder if we should provide `ctv3(*codes)` and `snomed(*codes)` functions as sugar here? Maybe it can wait until we iterate on the DSL.

tests/test_query_engine.py
```diff
@@ -1132,14 +1085,8 @@ class Cohort:
...
-        # Patient 1
-        Patients(PatientId=1, DateOfBirth="1990-8-10"),
-        RegistrationHistory(PatientId=1),
-        Events(PatientId=1, Date="2020-10-01"),
-        # Patient 2
-        Patients(PatientId=2, DateOfBirth="2000-03-20"),
-        RegistrationHistory(PatientId=2),
-        Events(PatientId=2, Date="2018-02-01"),
+        *patient(1, event("abc", "2020-10-01"), dob="1990-8-10"),
+        *patient(2, event("abc", "2018-02-01"), dob="2000-03-20"),
```
benbc, line 1089: Wow. That was quite an effort. Thank you.

benbc (approved):
Great.

PR #168, opened 2021-10-05
OS 69 Adapt Graphnet backend to utilies a single Clinical Events source

mikedferris:


benbc (approved):
Looks good. Is this the same as https://github.com/opensafely-core/cohort-extractor-v2/pull/164?

PR #174, opened 2021-10-11
Deal with null registration end dates 

rebkwok:
This adds an option to include nulls in table filters, so while `table("foo").filter("bar", greater_than=3)` will exclude null values, `table("foo").filter("bar", greater_than=3, include_null=True)` will now return rows where bar>3  or bar==None.

The special `date_in_range` method includes null dates by default in the end date range, so when we're looking for patients registered on a certain date, we include those where the registration end date is None, which corresponds to a current registration in graphnet.  For any other filters, it defaults to False.


tests/test_query_engine.py
```diff
@@ -389,6 +389,48 @@ class Cohort:
     assert extract(Cohort, MockBackend, database) == expected
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "filtered_table,expected",
+    [
+        (
+            table("clinical_events").filter(
+                "result", greater_than=15, include_null=True
```
evansd, line 398: I think this is fine for now, but longer term I think we want some way of composing arbitrary filter expressions using something like Django's [Q object](https://docs.djangoproject.com/en/3.2/topics/db/queries/#complex-lookups-with-q-objects). We could still present an `include_null` option to the user if we think that's a common enough need to deserve its own option. But under the hood it would be better if the AST could represent this as an OR over two filter expressions as that keeps things at a higher level of generality.

PR #180, opened 2021-10-14
Initial Databricks backend

evansd:
This implements a basic Databricks backend which runs against a Spark-compatible database. For the context see: https://github.com/opensafely-core/cohort-extractor/issues/570

The narrative structure of this PR (which is definitely best approached commit-by-commit) is roughly:
1. Basic groundwork in the form of fixes, tweaks and removing some unnecessary MSSQL-isms.
2. Factor out a `BaseSQLQueryEngine` from the existing `MssqlQueryEngine`.
3. Add a Databricks backend which has the necessary schema but runs against MSSQL.
4. Add all the parts required to provide a `SparkQueryEngine`.
5. Switch the Databricks backend to run against Spark.

I got all this right first time and have never heard of `git rebase`.

Note: we have some suggested (and some required) SQL modifications from the backend provider. This PR doesn't yet take these into account so further work will be needed here.


tests/conftest.py
```diff
@@ -148,9 +148,8 @@ def setup(input_data, base=mock_backend.Base):
             Session.configure(bind=engine)
             session = Session()
             # Load test data
-            for entity in iter_flatten(input_data):
-                session.add(entity)
-                session.commit()
+            session.bulk_save_objects(iter_flatten(input_data))
+            session.commit()
```
benbc, line 158: Great. I tried optimizing this at one point and couldn't get a good speed-up. I didn't find `bulk_save_objects()`, though.

tests/conftest.py
```diff
@@ -147,7 +148,7 @@ def setup(input_data, base=mock_backend.Base):
             Session.configure(bind=engine)
             session = Session()
             # Load test data
-            for entity in input_data:
+            for entity in iter_flatten(input_data):
```
benbc: This is a great idea. That `*` nonsense was infuriating.

cohortextractor/query_engines/mssql.py
```diff
@@ -1,8 +1,66 @@
...
+    # See https://stackoverflow.com/a/25548626/559140
+    format_str = "%Y-%m-%dT%H:%M:%S"
 
 
 class MssqlQueryEngine(BaseSQLQueryEngine):
-    pass
+    type_map = {
+        "date": MSSQLDate,
+        "datetime": MSSQLDateTime,
+    }
```
benbc, line 70: Do we actually use the `datetime` element anywhere?

cohortextractor/backends/databricks.py
```diff
@@ -0,0 +1,66 @@
...
+            JOIN
+                HES_AHAS_MPS_hes_apc_{year} AS mps
+            ON
+                apc.EPIKEY = mps.EPIKEY
+            LEFT JOIN
+                HES_AHAS_hes_apc_otr_{year} AS apc_otr
+            ON
+                apc.EPIKEY = apc_otr.EPIKEY
+            """
+            for year in ["1920"]
```
benbc, line 63: Only one year? Why the `UNION`?

benbc (approved):
It's a pleasure watching you work, Mr Evans.

cohortextractor/backends/databricks.py
evansd: Sorry yes, just updated with more comments.

evansd:
> It's a pleasure watching you work, Mr Evans.
:smile: Well, you're not really watching _me_ at work; you're watching idealised omnisci-Dave. The actual development process didn't look an awful lot like this and involved a lot more randomly commenting things out and swearing. And editing installed code in `site-packages`.

cohortextractor/query_engines/mssql.py
evansd: I'm not actually sure we do to be honest. We have columns in the TPP data which are datetimes but I think we just treat them all as dates. I think I wanted to create the `MSSQLDateTime` class just so we had a record of the correct way to do it, should we ever need to.

cohortextractor/query_engines/base_sql.py
```diff
@@ -0,0 +1,687 @@
...
+                    nullable=False,
+                ),
+            )
+            self.codelist_tables[codelist] = table
+            # Constuct the queries needed to create and populate this table
+            self.codelist_tables_queries.append(sqlalchemy.schema.CreateTable(table))
+            # There's a limit of 999 on how many rows we can insert in one go using
+            # this method See:
+            # https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver15#limitations-and-restrictions
+            for codes_batch in split_list_into_batches(codes, size=999):
```
benbc, line 299: Did you consider moving this into the MSSQL engine? I can see that it'll work on other dbs, but it seems a bit odd here.

cohortextractor/backends/databricks.py
```diff
@@ -0,0 +1,81 @@
...
+                HES_AHAS_MPS_hes_apc_{year} AS mps
+            ON
+                apc.EPIKEY = mps.EPIKEY
+            LEFT JOIN
+                HES_AHAS_hes_apc_otr_{year} AS apc_otr
+            ON
+                apc.EPIKEY = apc_otr.EPIKEY
+            """
+            # "1920" means "Financial year 2019/2020". We only have the one
+            # year in the sample data for now but we're anticipating more.
```
benbc, line 78: Whoever came up with this encoding needs a special award. True genius.

cohortextractor/backends/databricks.py
```diff
@@ -0,0 +1,81 @@
...
+        # HES data comes in distinct tables for each financial year which we
+        # union together to pretend its a single table.
+        query="\nUNION ALL\n".join(
+            # We join the APC (Admitted Patient Care) table with the MPS
+            # (Master Person Service) table so we can get the associated
+            # "person ID" which functions as our unique patient identifier that
+            # can be matched with the prescriptions data. We also join with the
+            # APC-OTR table (not sure what this stands for) so we can get a
+            # "spell ID" which allows us to see when multiple admissions
+            # correspond to the same hospitalisation "spell".
```
benbc, line 57: Documenting the meaning of the abbreviations is _really_ helpful. Is `EPIKEY` a contraction for "episode"? If so, what's the difference between "episode" and "spell"?

evansd: I didn't know, but your question prompted me to ask in Slack where I got a very helpful reply from Helen which I'll incorporate into the comments.

cohortextractor/query_engines/base_sql.py
evansd: Oh that's a good point. We should at least move the specific size limit to the MSSQL class. I'm also sure we're not done with the refactor here and there are still other MSSQL-isms hiding in the base class that we'll discover later.

PR #197, opened 2021-11-08
fix: databricks: rename ADMINMETH colum to ADMIMETH, as per docs

bloodearnest:


iaindillingham (approved):
This looks painful.

evansd (approved):
Ah yes, my bad. In my defence, it comes in a slightly confusing sequence of column names: `ADMIMETH, ADMINCAT, ADMINCATST`. Thanks for fixing.

bloodearnest:
> Ah yes, my bad. In my defence, it comes in a slightly confusing sequence of column names: `ADMIMETH, ADMINCAT, ADMINCATST`. Thanks for fixing.

I read the documentation of the columns and I did not blame you at all :D

PR #198, opened 2021-11-09
fix: databricks: support the use of db schemas

bloodearnest:
The PoC databricks dataset uses db schemas, so support that.

This involves referencing them with . notation from queries, and also
adding support to create the schemas as part of the test db set up.


evansd (approved):
:+1: Nice

PR #199, opened 2021-11-09
Fix warnings

inglesp:
Docs for `get_final_froms()`:
https://docs.sqlalchemy.org/en/14/core/selectable.html#sqlalchemy.sql.expression.Select.get_final_froms

Docs for `scalar_subquery()`:
https://docs.sqlalchemy.org/en/14/core/selectable.html#sqlalchemy.sql.expression.SelectBase.scalar_subquery


evansd (approved):
Thanks for doing this! I learned useful things from this.

PR #202, opened 2021-11-09
justfile update

bloodearnest:
- chore: update justfile to use latest conventions from the repo-template
- chore: do not include prod deps in dev, rather constrain to pinned prod


ghickman (approved):

madwort:
this PR changes the user interface for just, but does not update the `DEVELOPERS.md` docs to match

PR #204, opened 2021-11-09
Fill in the blank software name in LICENSE

StevenMaude:


StevenMaude:
I've followed the current cohort-extractor's LICENSE here, and used the existing software name as from README and the repository name.

This raises a question: does v2 in the name cause confusion if we ever start tagging release with actual version numbers?

StevenMaude:
NB: I think this value didn't get templated in on repository creation because the [repo-template repository only substitutes in `README.md`](https://github.com/opensafely-core/repo-template/blob/5477d3ad61db64243a23c9aa162834b806a19a08/.github/workflows/setup.yml#L12-L19).

iaindillingham (approved):

StevenMaude:
The name might be changed in future. However, merging because the current changes improves on what there is now.

PR #205, opened 2021-11-10
Add minimal new DSL

rebkwok:
Implements the minimal cohort definition defined in https://app.shortcut.com/ebm-datalab/story/62/minimal-introduction-of-new-dsl and tests that it produces the same query language graph as the equivalent old-style cohort definition.

`query_interface.py` is intended to hold the things that translate the new DSL to the query_language.

The second commit adds a minimal end-to-end test, which was mainly for my benefit to check that the cohort registration worked - it adds a  small amount of code to `main.py` so it may be beyond the scope of this ticket, and can be removed if it's in the way.

### DIFF
--- a/cohortextractor/definition/__init__.py
+++ b/cohortextractor/definition/__init__.py
@@ -1,0 +1,8 @@
+from .definition import Cohort, pick_first_value, register
+
+
+__all__ = [
+    "Cohort",
+    "pick_first_value",
+    "register",
+]
--- a/cohortextractor/definition/base.py
+++ b/cohortextractor/definition/base.py
@@ -1,4 +1,12 @@
-from weakref import WeakSet
+class CohortRegistry:
+    def __init__(self):
+        self.cohorts = set()
+
+    def add(self, cohort):
+        self.cohorts.add(cohort)
+
+    def reset(self):
+        self.cohorts = set()
 
 
-registered_cohorts = WeakSet()
+cohort_registry = CohortRegistry()
--- a/cohortextractor/definition/base.py
+++ b/cohortextractor/definition/base.py
@@ -1,0 +1,4 @@
+from weakref import WeakSet
+
+
+registered_cohorts = WeakSet()
--- a/cohortextractor/main.py
+++ b/cohortextractor/main.py
@@ -9,6 +9,7 @@ from pathlib import Path
 import structlog
 
 from .backends import BACKENDS
+from .definition.base import registered_cohorts
 from .measure import MeasuresManager, combine_csv_files_with_dates
 from .query_utils import get_column_definitions, get_measures
 from .validate_dummy_data import validate_dummy_data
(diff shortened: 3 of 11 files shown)


cohortextractor/definition/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .definition import Cohort, pick_first_value, register
```
rebkwok, line 1: I'm assuming that we'll have more of these definition things later and will probably want to split them out in some sensible way, but for now they're all in one `definition.py` file

cohortextractor/definition/base.py
```diff
@@ -0,0 +1,4 @@
+from weakref import WeakSet
```
rebkwok, line 1: I'm not really sure if this is necessary/sensible or not; I was basing it on how Django registers admin sites, but maybe we don't care too much about it hanging around when everything's run in docker containers.

cohortextractor/main.py
```diff
@@ -38,11 +39,14 @@ def run_cohort_action(
         else:
             date_suffix = ""
 
-        cohort = (
-            cohort_class_generator(index_date)
-            if index_date
-            else cohort_class_generator()
-        )
+        if registered_cohorts:
+            (cohort,) = registered_cohorts
```
rebkwok, line 43: For now at least, this assumes there's only ever at most one registered cohort

cohortextractor/definition/base.py
benbc: My best guess is that this is _probably_ wrong -- couldn't the cohort objects could get garbage collected between the module being loaded (triggering the call to `register()`) and the cohort extractor actually accessing the definition?

benbc: (And, as you suggest, I don't think we need to worry about these things hanging around since the whole process is fairly shortlived.)

tests/test_dsl.py
```diff
@@ -0,0 +1,29 @@
+from cohortextractor.concepts import tables
+from cohortextractor.definition import Cohort, pick_first_value, register
+from cohortextractor.definition.base import registered_cohorts
+from cohortextractor.query_language import table
+from cohortextractor.query_utils import get_column_definitions
+
+
+def test_minimal_cohort_definition():
```
benbc, line 8: This test is perfect. Right now it's a good check that we're really replicating the old DSL, but we can hopefully unit test all the DSL construction functionality like this by asserting against the semantic model that's generated.

benbc (approved):
The end-to-end test and the switching in `main` are perfect. Just one comment about the weakref stuff.

cohortextractor/definition/base.py
rebkwok: :+1:  I am a bit hazy on garbage collection, so definitely not committed to it.  I'll take it out

rebkwok: While I was doing that, I realised that we can (and did, in the tests) have more than one cohort registered.  Last commit has a slight modification of the registry to allow tests to reset it, and to assert (for now, at least) that there's only one cohort registered.

benbc (approved):

PR #207, opened 2021-11-10
Improve and test CLI

inglesp:


CarolineMorton (approved):
Thanks

CLI tool `__main__.py` is much easier to read. 

I have 97% coverage of tests now showing when I run.

PR #208, opened 2021-11-10
Use the default docker network in all tests

rebkwok:
Remove the network fixture and use the default network in tests; as a result, we need to specify the database container IP in the database url passed to the cohort-extractor container, since the default network doesn't allow DNS lookup by container name


evansd (approved):
Great to see this stuff getting simplified!

PR #209, opened 2021-11-10
Use dataclasses for query language objects

inglesp:


evansd (approved):
:+1: Thanks for doing this

rebkwok:
Could there be a benefit to using attrs instead of dataclasses?  I think you can change the value of a field after instantiation with attrs, and it also allows type validation, which might be useful later for validating cohort definitions

cohortextractor/query_language.py
```diff
@@ -1,3 +1,7 @@
+from dataclasses import dataclass
+from typing import Any
```
ghickman, line 2: We don't appear to be running any type checking on this project, so I'm not sure using Any (which effectively disables type checking if we _were_ using a checker) is useful here.  Do we need to set types on the dataclass variables?

inglesp:
> Could there be a benefit to using attrs instead of dataclasses?

Worth considering, but I'm keen not to add extra dependencies unless we need to.

> I think you can change the value of a field after instantiation with attrs

You can do this with dataclasses as well, but then you can't hash instances without losing the ability to compare instances for equality.  As far as I can tell, attrs has the same restriction.  [(docs)](https://www.attrs.org/en/stable/hashing.html)

> it also allows type validation

So does [dataclasses](https://mypy.readthedocs.io/en/stable/additional_features.html).

rebkwok:
> > it also allows type validation
> 
> So does [dataclasses](https://mypy.readthedocs.io/en/stable/additional_features.html).

Do they (in the validators-at-runtime sense that attrs has?)?  Is that new(ish)?  I came across a bunch of references to that being one of the main things attrs had that dataclasses didn't.

inglesp:
> Do they (in the validators-at-runtime sense that attrs has?)?

Ooh, I don't think they do at runtime.  That might be something worth considering in future, thanks.

cohortextractor/query_language.py
inglesp: >  Do we need to set types on the dataclass variables?

As far as I understand (and I may be very wrong) the syntax that dataclasses uses require you to specify a type.  For most of these dataclasses, I'm not sure what the range of valid types is.  It would certainly be good to pin these down, but I don't see much point in doing so until we're actually checking the types.  (This module feels like a good candidate for proper type checking.)

PR #211, opened 2021-11-10
Improve test coverage

inglesp:
We're up to 98%, and most of the low-hanging fruit have been picked.  I expect somebody who's been around the project longer could pick a few more!

### DIFF
--- a/cohortextractor/query_engines/base_sql.py
+++ b/cohortextractor/query_engines/base_sql.py
@@ -175,7 +175,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         elif isinstance(node, (ValueFromRow, Column)):
             return node.column
         else:
-            raise TypeError(f"Unhandled type: {node}")
+            assert False, f"Unhandled type: {node}"
 
     def get_parent_nodes_from_category_definitions(
         self, definitions, parent_nodes=None
@@ -472,7 +472,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         try:
             method = class_method_map[value.__class__]
         except KeyError:
-            raise ValueError(f"Unsupported function: {value}")
+            assert False, f"Unsupported function: {value}"
         value_expression = method(*argument_expressions)
 
         return value_expression, tuple(tables)
--- a/cohortextractor/query_utils.py
+++ b/cohortextractor/query_utils.py
@@ -26,8 +26,7 @@ def get_column_definitions(cohort_class):
     for name, value in variables:
         if name.startswith("_") or name in ignored_names:
             continue
-        if not isinstance(value, Value):
-            raise ValueError(f"Cohort variable '{name}' is not a Value")
+        assert isinstance(value, Value), f"Cohort variable '{name}' is not a Value"
         columns[name] = value
     return columns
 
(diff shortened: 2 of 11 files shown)


tests/query_engines/test_mssql_lib.py
```diff
@@ -213,10 +209,3 @@ def list_all(self):
             return sorted(
                 row[0] for row in conn.execute(query) if row[0].startswith(self._prefix)
             )
-
-    def drop_all(self):
-        tables = self.list_all()
-        with self.engine.connect() as conn:
-            for table in tables:
-                conn.execute(sqlalchemy.text(f"DROP TABLE {self.database}..{table}"))
-            conn.commit()
```
CarolineMorton, line 222: why don't we need to drop the db tables?

CarolineMorton (approved):
Looks good. One comment for clarity

tests/query_engines/test_mssql_lib.py
rebkwok: I *think* because the tests all do `_populate_table` which drops the table if it exists and recreates it

rebkwok (approved):
LGTM - one comment based on my own confusion

cohortextractor/query_engines/base_sql.py
```diff
@@ -175,7 +175,7 @@ def get_output_column_name(node):
         elif isinstance(node, (ValueFromRow, Column)):
             return node.column
         else:
-            raise TypeError(f"Unhandled type: {node}")
+            assert False, f"Unhandled type: {node}"
```
benbc, line 178: @inglesp And you claim that the codebase is well typed? :rofl:

cohortextractor/query_utils.py
```diff
@@ -26,8 +26,7 @@ def get_column_definitions(cohort_class):
     for name, value in variables:
         if name.startswith("_") or name in ignored_names:
             continue
-        if not isinstance(value, Value):
-            raise ValueError(f"Cohort variable '{name}' is not a Value")
+        assert isinstance(value, Value), f"Cohort variable '{name}' is not a Value"
```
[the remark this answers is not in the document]
benbc: This is a great case to raise. We should aim to make type errors like this literally impossible in the improved DSL.

cohortextractor/query_engines/base_sql.py
inglesp: I stand by that!  There's just an undefined interface that `ValueFromAggregate` and `ValueFromRow` implement.

What I meant is that this kind of instance checking seems very self-contained and doesn't bleed through the codebase.

tests/query_engines/test_mssql_lib.py
evansd: We don't need the drop tables if all the code under test works correctly and cleans up after itself. But if the tests fail then we'll leave detritus around. This is a general issue, I think, with attempting to enforce coverage of code which is only there to clean up if something goes wrong in a test: under normal circumstances it won't get executed, but we still want it there.

cohortextractor/query_utils.py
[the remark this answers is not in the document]
evansd: Yeah, I think agree with Becky here: this is a user error and therefore warrants an exception rather than an assertion. I think the fix for the coverage issue here is to add a tiny test case which triggers this condition.

evansd (commented):
Two small comments that I've made inline, but overall this is really great and thanks for doing it. 100% coverage is one of those things that it's an awful lot easier to maintain than achieve and I think it's definitely worth the effort in these relatively early days.

PR #212, opened 2021-11-11
Set Up Interrogate to Track Docstrings Usage

ghickman:
This adds [interrogate](https://interrogate.readthedocs.io/en/latest/) to our dev tooling stack to track how many objects have docstrings.

[sc-159]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,12 @@ exclude_lines = [
 
 [tool.coverage.html]
 
+[tool.interrogate]
+fail-under = 26
+ignore-init-module = true
+omit-covered-files = true
+verbose = 1
+
 [tool.isort]
 force_grid_wrap = 0
 include_trailing_comma = true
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -28,6 +28,12 @@ exclude_lines = [
 
 [tool.coverage.html]
 
+[tool.interrogate]
+fail-under = 26
```
CarolineMorton, line 32: What does this mean?

CarolineMorton (approved):
Looks good to me

pyproject.toml
ghickman: Interrogate will return a failure code on the command line if coverage is below 26% (mirroring how test coverage works).  We're currently just over that, so this is base level to start with!

sebbacon:
Given we use the google python style guide's docstring conventions, is there some linter / checker option for that?

ghickman:
This tool only checks that objects have docstrings, not which style we're using. There are other tools for that, @StevenMaude has collated a [selection of them](https://app.shortcut.com/ebm-datalab/story/162/use-a-standard-for-docstrings).

PR #214, opened 2021-11-11
feat: document requirements for using macOS

madwort:


StevenMaude (approved):

DEVELOPERS.md
```diff
@@ -74,3 +75,11 @@ just remove-persistent-database
 
 Set the environment variable `LOG_SQL=1` (or anything non-empty) to get
 all SQL queries logged to the console.
+
+## macOS / Bash
+
+Starting with version 4.0, Bash is licenced under GPLv3. Because of this, macOS still ships with version 3.2, which is incompatible with some scripts in this repository. We recommend using [homebrew](https://brew.sh/) to install a more recent version, and then ensure that it is on your `$PATH`.
+
+```bash
+brew install bash
```
CarolineMorton, line 84: I also had to do `brew upgrade bash` and then restart terminal. Could we add?

madwort: hmm, interesting, did you already have a homebrew version of bash installed, then? And to get this repo working you did `brew install bash` and then also `brew upgrade bash`? I want to point people in the right direction here, but I'm not sure we necessarily want give detailed instructions for all possible scenarios.

CarolineMorton: I did have to do both but I am not entirely sure why.

madwort: weird, `install` on its own was enough for me. Do you mind if we treat it as an anomaly for now & revisit it if/when it comes up again?

PR #219, opened 2021-11-11
Add a filter operation to the DSL

inglesp:
This does the minimum to make the test case in the ticket pass.

### DIFF
--- a/cohortextractor/concepts/tables.py
+++ b/cohortextractor/concepts/tables.py
@@ -5,5 +5,8 @@ class ClinicalEvents(QueryBuilder):
     code = "code"
     date = "date"
 
-    def __init__(table_name):
+    def __init__(self):
         super().__init__("clinical_events")
+
+
+clinical_events = ClinicalEvents()
--- a/cohortextractor/concepts/tables.py
+++ b/cohortextractor/concepts/tables.py
@@ -7,6 +7,3 @@ class ClinicalEvents(QueryBuilder):
 
     def __init__(table_name):
         super().__init__("clinical_events")
-
-
-clinical_events = ClinicalEvents()
--- a/cohortextractor/query_interface.py
+++ b/cohortextractor/query_interface.py
@@ -7,11 +7,11 @@ from cohortextractor import query_language
 class QueryBuilder:
     def __init__(self, table_name):
         self.table_name = table_name
-        self.filter_details = None
+        self.filter_details = []
 
     def filter(self, *args, **kwargs):  # noqa: A003
-        filtered = copy.copy(self)
-        filtered.filter_details = (args, kwargs)
+        filtered = copy.deepcopy(self)
+        filtered.filter_details.append((args, kwargs))
         return filtered
 
     def select_column(self, column_name):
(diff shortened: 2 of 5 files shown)


benbc (approved):

benbc (commented):
Obviously we should let The King of the DSL decide these things, but I do think it's simpler for study authors to be able to grab hold of the table directly rather than having to construct it. You're right that we don't now yet how the DSL will end up, but betting on immutable objects doesn't seem unreasonable -- especially assuming we're going to keep the chaining.

cohortextractor/concepts/tables.py
```diff
@@ -7,6 +7,3 @@ class ClinicalEvents(QueryBuilder):
 
     def __init__(table_name):
         super().__init__("clinical_events")
-
-
-clinical_events = ClinicalEvents()
```
rebkwok, line 12: I think maybe this was one of those places I mentioned earlier today, where I was trying too hard to implement exactly what was written int he code snippets in the original ticket.  I thought perhaps we were trying not to make users have to instantiate the table, but I realise we were already making them instantiate a `Cohort()` so it's probably better to be explicit about it anyway

cohortextractor/query_interface.py
```diff
@@ -4,30 +4,33 @@
 
 class QueryBuilder:
     def __init__(self, table_name):
-        self.table_name = table_name
+        self.table = query_language.table(table_name)
```
rebkwok, line 7: I think I'm confused again about how we're separating the DSL and the semantic model.  I had thought the idea was to have the tables in `concepts` (or whatever) be separate to the semantic model, which is why it only had a table name before, not a `query_language.table` (which I *think* is a semantic model thing).  

I had been thinking about this one in my head though, and I hasn't come up with an obvious way to get a Variable with enough only non-query_language things which could do all the compiling to query_language things itself

cohortextractor/query_utils.py
```diff
@@ -16,11 +16,11 @@ def get_cohort_variables(cohort):
     ]
 
 
-def get_column_definitions(cohort_class):
-    if isinstance(cohort_class, CohortConcept):
-        variables = get_cohort_variables(cohort_class)
+def get_column_definitions(cohort):
+    if isinstance(cohort, CohortConcept):
```
rebkwok, line 20: `CohortConcept` will be a particularly bad name if/when we rename the `concepts` module

cohortextractor/concepts/tables.py
benbc: I think there's a semantic difference. Their purpose is to create a cohort, so it seems reasonable for them to construct a cohort object. But they're not creating the tables, they're accessing tables that already exist.

inglesp: I think I prefer the original from a usability perspective.  I think it's quite likely that users will be confused about what things are classes and what things are instances, and the fewer things they have to instantiate, the better.

cohortextractor/query_interface.py
inglesp: I think it makes sense for the objects in `query_interface.py` to not know about semantic model objects until we come to compilation.  So storing a table here is wrong, and we should store the table name.

In particular, this will make serialisation of these DSL objects very easy.

cohortextractor/query_utils.py
inglesp: ho ho

inglesp:
Some open questions here that I'm not around today to help resolve.  I'd suggest merging, to allow progress with implementing other DSL features such as sorting.

benbc:
@rebkwok since Peter isn't around today, I took the liberty of making the changes that I think he's agreed should be made. Please could you take another quick look?

rebkwok (approved):
LGTM.  For now, we can only do a single filter, right?  Should/do we have a ticket for chained filters?

benbc:
@rebkwok No, that's totally a bug! Thank you -- I'll add a test.

rebkwok (approved):

PR #222, opened 2021-11-11
Set `fail-under=0` for interrogate

StevenMaude:
This is a temporary measure. Adding docstrings, particularly to the
domain specific language parts of the code, is a task to be completed
fairly soon, but not yet started.

For now, allow this check to run, but always pass. This avoids
inconveniencing developers who are doing other things right now. We
should enforce this again once docstrings are added to the relevant
parts of the code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,11 +20,6 @@ repos:
     - id: isort
       language_version: python3.9
 
-  - repo: https://github.com/econchick/interrogate
-    rev: 1.4.0
-    hooks:
-    - id: interrogate
-
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
     hooks:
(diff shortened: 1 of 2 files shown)


ghickman (changes requested):
Can you also remove it from pre-commit, after some thought that's almost the bigger issue here.  Although doing this is still relevent so lets do both 🙂

StevenMaude:
@ghickman: to clarify, you mean both of:

* `interrogate` shouldn't run in pre-commit at all for now, as opposed to what it does now: run and not complain
* we should leave the configuration existing in pyproject.toml

ghickman:
@StevenMaude – yep, the more I think about it the more I don't think we want it in pre-commit at all because we don't plan to enforce docstrings everywhere (eg tests!), so I think it's just going to annoy us in pre-commit.

ghickman (approved):

StevenMaude:
Updated!

PR #225, opened 2021-11-11
Add minimal backend "concepts"

evansd:
Add basic walking skeleton code which allows tables to declare that they implement certain "concepts" i.e. they have a certain set of tables and those tables have a certain type. This also involves the assertion that those columns have a certain _meaning_ in the context of clinical research and can play approximately the same role across other backends which share that "concept". But of course we can't enforce that programmatically.

You'll note the word "contract" has slipped into the codebase here. I seem to remember the word floating around in earlier discussions and it seemed a natural way of characterising the assertion being made here.

In contrast to the initial pseudocode I've made the `implements` declaration be part of the table definition, rather than a method call we make somewhere else as it seemed more natural to keep the two together. It's easy enough to change if there's some other reason we'd like to keep them separate though.

**Note**: this doesn't yet do any rudimentary documentation generation because, although that's part of the below story, I felt it was better tackled in a separate PR.

Relates to:
[sc-39]
https://app.shortcut.com/ebm-datalab/story/39/minimal-end-to-end-backend-definition

### DIFF
--- a/cohortextractor/concepts/tables.py
+++ b/cohortextractor/concepts/tables.py
@@ -1,5 +1,8 @@
 from cohortextractor.query_interface import QueryBuilder
 
+from . import types
+from .table_contract import Column, TableContract
+
 
 class ClinicalEvents(QueryBuilder):
     code = "code"
@@ -10,3 +13,25 @@ class ClinicalEvents(QueryBuilder):
 
 
 clinical_events = ClinicalEvents()
+
+
+class Patients(TableContract):
+    """
+    The Patients table holds personal details about the patient.
+    """
+
+    patient_id = Column(
+        type=types.PseudoPatientId(),
+        help=(
+            "Patient's pseudonymous identifier, for linkage. You should not normally "
+            "output or operate on this column"
+        ),
+    )
+    date_of_birth = Column(
+        type=types.Date(),
+        help="The patient's date of birth",
+    )
+    sex = Column(
+        type=types.Choice("F", "M"),
+        help="The patient's sex",
+    )
--- a/tests/concepts/test_table_contract.py
+++ b/tests/concepts/test_table_contract.py
@@ -1,0 +1,49 @@
+import pytest
+
+from cohortextractor.backends.base import BaseBackend, Column, MappedTable
+from cohortextractor.concepts import types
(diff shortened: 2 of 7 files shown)


sebbacon:
Not got time for a code review today, but dropping by to say this looks like a great PR to add a glossary entry alongside.

I thought your descriptions in the PR text are at least a good start e.g.

> the assertion that those columns have a certain meaning in the context of clinical research and can play approximately the same role across other backends which share that "concept"

Though we may want to break out a thread in slack to discuss more

cohortextractor/concepts/tables.py
```diff
@@ -10,3 +13,25 @@ def __init__(table_name):
...
+class Patients(TableContract):
+    """
+    The Patients table holds personal details about the patient.
+    """
+
+    patient_id = Column(
+        type=types.PseudoPatientId(),
+        help=(
+            "Patient's pseudonymous identifier, for linkage. You should not normally "
+            "output or operate this column"
```
benbc, line 27: "operate _on_"?

tests/concepts/test_table_contract.py
```diff
@@ -0,0 +1,23 @@
...
+            query_engine_class = BaseSQLQueryEngine
+            patient_join_column = "Patient_ID"
+
+            patients = MappedTable(
+                implements=tables.Patients,
+                source="Patient",
+                columns=dict(
+                    date_of_birth=Column("date", source="DateOfBirth"),
+                ),
+            )
```
benbc, line 23: I suppose that in an ideal world I'd prefer to see a happy-path test case as well.

benbc (commented):
Some small comments only, most of which I could be argued out of.

I too considered moving the implementation declaration into the constructor. I don't think there's any strong reason to have it as a method -- although @sebbacon may feel otherwise.

And getting this in without the doc gen would be great since it will unblock some other stuff.

benbc (approved):
Agh! I meant to hit "approve" since the changes are tiny and I don't need to see this again.

evansd:
@benbc Thanks I agree with all those. There is a sort of implicit happy path test when we test the TPPBackend which is why I only thought to test the unhappy path. But it would be _far_ better to do this explicitly, wouldn't it?

PR #227, opened 2021-11-12
Add mypy

ghickman:
This adds mypy, with some settings which looked useful, verbosely linked to the docs because we don't have a lot of team knowledge around the tool yet.

Mypy seems to be quite permissive by default and there are many settings we can turn on to add further strictness.  I've opted for the defaults with a few extra checks to start with.  But I'd like to aim for getting the following added over time: 

* [check_untyped_defs](https://mypy.readthedocs.io/en/stable/config_file.html#confval-check_untyped_defs)
* [disallow_untyped_calls](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_calls)
* [disallow_untyped_decorators](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_decorators)
* [disallow_untyped_defs](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_defs)

### DIFF
--- a/cohortextractor/backends/base.py
+++ b/cohortextractor/backends/base.py
@@ -1,5 +1,7 @@
 import sqlalchemy
 
+from ..query_engines.base_sql import BaseSQLQueryEngine
+
 
 # Mutable global for storing registered backends
 BACKENDS = {}
@@ -10,16 +12,16 @@ def register_backend(backend_class):
 
 
 class BaseBackend:
-    backend_id = NotImplemented
-    query_engine_class = NotImplemented
-    patient_join_column = NotImplemented
+    backend_id: str
+    query_engine_class: type[BaseSQLQueryEngine]
+    patient_join_column: str
 
     tables = None
 
     def __init_subclass__(cls, **kwargs):
-        assert cls.backend_id != NotImplemented
-        assert cls.query_engine_class != NotImplemented
-        assert cls.patient_join_column != NotImplemented
+        assert cls.backend_id is not None
+        assert cls.query_engine_class is not None
+        assert cls.patient_join_column is not None
 
         # Register each Backend by its id so we can identify it from an environment variable
         register_backend(cls)
--- a/cohortextractor/query_engines/base_sql.py
+++ b/cohortextractor/query_engines/base_sql.py
@@ -1,7 +1,7 @@
 import contextlib
 from collections import defaultdict
 from types import ModuleType
-from typing import Union
+from typing import Optional, Union
 
 import sqlalchemy
 import sqlalchemy.dialects.mssql
@@ -73,7 +73,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     type_map = None
 
     # No limit by default although some DBMSs may impose one
-    max_rows_per_insert = None
+    max_rows_per_insert: Optional[int] = None
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
--- a/cohortextractor/query_engines/base_sql.py
+++ b/cohortextractor/query_engines/base_sql.py
@@ -69,7 +69,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
 
     sqlalchemy_dialect: Union[type[Dialect], ModuleType]
 
-    custom_types = {}
+    custom_types: dict[str, type[sqlalchemy.types.TypeDecorator]] = {}
     type_map = None
 
     # No limit by default although some DBMSs may impose one
--- a/cohortextractor/query_engines/base_sql.py
+++ b/cohortextractor/query_engines/base_sql.py
@@ -1,10 +1,13 @@
 import contextlib
 from collections import defaultdict
+from types import ModuleType
+from typing import Union
 
 import sqlalchemy
 import sqlalchemy.dialects.mssql
 import sqlalchemy.schema
 import sqlalchemy.types
+from sqlalchemy.engine.interfaces import Dialect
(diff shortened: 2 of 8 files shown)


cohortextractor/backends/base.py
```diff
@@ -10,17 +16,13 @@ def register_backend(backend_class):
 
 
 class BaseBackend:
-    backend_id = NotImplemented
-    query_engine_class = NotImplemented
-    patient_join_column = NotImplemented
+    backend_id: str
+    query_engine_class: type["BaseSQLQueryEngine"]
```
benbc, line 20: I think you can just use the class name here rather than a string, can't you? (And therefore probably do away with the `TYPE_CHECKING` guard on the import.

cohortextractor/backends/base.py
```diff
@@ -10,17 +16,13 @@ def register_backend(backend_class):
...
+    patient_join_column: str
 
     tables = None
 
     def __init_subclass__(cls, **kwargs):
-        assert cls.backend_id != NotImplemented
-        assert cls.query_engine_class != NotImplemented
-        assert cls.patient_join_column != NotImplemented
```
benbc, line 22: I think we've lost something useful here. Can we assert that they're not `None` as a check instead?

cohortextractor/query_engines/base_sql.py
```diff
@@ -64,17 +67,16 @@ def get_primary_table(query):
 
 class BaseSQLQueryEngine(BaseQueryEngine):
 
-    sqlalchemy_dialect = NotImplemented
+    sqlalchemy_dialect: Union[type[Dialect], ModuleType]
```
benbc, line 70: Gross! I've looked into why this type is so weird and don't understand what's going on. Nothing for you to do here, but this is a good example of why clarifying the types is helpful.

cohortextractor/query_engines/base_sql.py
```diff
@@ -64,17 +67,16 @@ def get_primary_table(query):
...
+    custom_types: dict[str, type[sqlalchemy.types.TypeDecorator]] = {}
     type_map = None
 
     # No limit by default although some DBMSs may impose one
-    max_rows_per_insert = None
+    max_rows_per_insert: Optional[int] = None
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        assert cls.sqlalchemy_dialect != NotImplemented
```
benbc, line 77: Could this be a truthiness check instead?

cohortextractor/query_engines/spark_lib.py
```diff
@@ -7,7 +7,7 @@
 otherwise be about how fragile they are.
 """
 import sqlalchemy.types
-from pyhive.sqlalchemy_hive import HiveHTTPDialect, HiveTypeCompiler
+from pyhive.sqlalchemy_hive import HiveHTTPDialect, HiveTypeCompiler  # type: ignore
```
benbc, line 10: Is this ignore stopping us from checkout our use of these types (in which case, why?), or stopping mypy from descending into these modules (in which case could we constrain it in the config instead?).

benbc (commented):
I'm really happy to see this going in.

cohortextractor/query_engines/base_sql.py
ghickman: I thought originally mypy would be checking it for us, but yes a runtime check just to be extra sure can't hurt.

benbc: No! If mypy _can_ check it then please don't duplicate that at runtime. I'm not sure that it can though, can it?

cohortextractor/backends/base.py
ghickman: I was trying to avoid potential cyclic imports but we're not hitting that yet so probably YAGNI.

cohortextractor/query_engines/spark_lib.py
ghickman: It's the latter, but to ignore specific lines you have to do inline comments from what I can tell.

benbc: @ghickman I see this:

```
[mypy-foobar.*]
ignore_missing_imports = True
```

https://mypy.readthedocs.io/en/stable/running_mypy.html

Maybe that's a bit less annoying than having to add it to every import?

cohortextractor/query_engines/base_sql.py
ghickman: Clearly I made an assumption here! I removed SparkQueryEngine.sqlalchemy_dialect and mypy didn't throw an error, which I really didn't expect…

I'll reinstate this assert until I can work out why mypy isn't catching that.

ghickman: If we make BaseQueryEngine a Protocol we can do this but I'm keen to get these changes in as-is so mypy is configured then tackle the Protocol separately.  I've done a quick spike on switching to a Protocol and it's more work that I expected, so I'll tackle that in a separate PR.

cohortextractor/query_engines/spark_lib.py
ghickman: Ooh brilliant. That's much nicer! I also discovered how we can do per-module overrides like this in pyproject land:

https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file

cohortextractor/query_engines/base_sql.py
benbc: :+1:

benbc (approved):

PR #228, opened 2021-11-12
Add pyupgrade

ghickman:
This adds pyupgrade to tell us about modern versions of syntax.  In particular since we're checking typing in #227 this will keep our typing syntax up to date moving forwards.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,12 @@ repos:
     - id: isort
       language_version: python3.9
 
+
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.29.0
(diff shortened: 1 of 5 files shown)


benbc (approved):
This is a handy thing. Thanks.

Only one fix to make is a bit of a surprise.

I wonder if there's some way we can make it more likely that we'll remember to update the `--py39-plus` as and when we upgrade Python? I'm not sure what that would be. Possibly a comment in `_virtualenv` in the `Justfile`?

ghickman:
I've not found a way to notify about it yet, but I suspect we'll not upgrade python versions _that_ often so 🤞 it won't be much of a burden to remember to do it.

PR #229, opened 2021-11-12
Fix validate_cohort for databricks backend

rebkwok:
And also make the default population query customisable by backend

### DIFF
--- a/cohortextractor/query_engines/spark.py
+++ b/cohortextractor/query_engines/spark.py
@@ -14,6 +14,9 @@ class CreateViewAs(Executable, ClauseElement):
         self.name = name
         self.query = query
 
+    def __str__(self):
+        return str(self.query)
+
 
 @compiles(CreateViewAs, "spark")
 def _create_table_as(element, compiler, **kw):
--- a/tests/end_to_end/test_end_to_end_databricks.py
+++ b/tests/end_to_end/test_end_to_end_databricks.py
@@ -1,0 +1,12 @@
+from cohortextractor.main import validate_cohort
+
+
+def test_validate_cohort(load_study, cohort_extractor_in_process_no_database):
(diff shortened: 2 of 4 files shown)


evansd (approved):
All makes sense.

Longer term I'd like to move away from providing a default population at all. From memory we only did it to get some of the early test cases to work; in v1 we require a population to be specified and I think we should do the same here too, at some point.

We'll need something like the equivalent of `patients.all()` though for when you just want the population to be the entire patients table.

rebkwok:
> All makes sense.
> 
> Longer term I'd like to move away from providing a default population at all. From memory we only did it to get some of the early test cases to work; in v1 we require a population to be specified and I think we should do the same here too, at some point.
> 
> We'll need something like the equivalent of `patients.all()` though for when you just want the population to be the entire patients table.

If we're going to require a population to be specified in future, does it make sense to include this change at all then?  If I back out that part, the existing v2 study defs will continue to work, and so will the tests, but any new study defs using the databricks backend will require it to be specified, which is fine, I think.  In fact, I think all v2 study defs that have been written so far specify a population, so I could just remove the default population altogether?

evansd:
Ah, maybe not then. @benbc Do you have thoughts? I think we both agreed that explicit was better than implicit here, so maybe the sooner we move to that the better.

benbc:
Yes, I had a look the other day and all the v2 study defs have populations specified. I put a story on the backlog to remove the default, but maybe you're right and we should just do it now. It'll just mean some extra work to make the tests happy -- if you can figure out a way of doing that that doesn't involve explicitly adding it in every test, that would be great.

rebkwok:
:+1:  I'll remove those commits from this PR so Lisa can run the validate_cohort on her study, and I'll look at removing the default altogether separately

PR #231, opened 2021-11-12
Improve test coverage some more

rebkwok:
This gets coverage to 99.75%, with all but 3 files fully covered.  

Mostly minor things, removing/fixing some unused code, adding the odd missing tests.


cohortextractor/validate_dummy_data.py
```diff
@@ -39,20 +39,16 @@ def validate_file_extension(output_file, dummy_data_file):
...
     """Read data from path into a Pandas DataFrame."""
-
     try:
         suffixes = ".".join([suffix.strip(".") for suffix in path.suffixes])
-        if suffixes in SUPPORTED_FILE_FORMATS:
-            return pd.read_csv(path)
-        else:
-            msg = f"Dummy data must be in one of the following formats: {', '.join(SUPPORTED_FILE_FORMATS)}"
-            raise DummyDataValidationError(msg)
+        assert suffixes in SUPPORTED_FILE_FORMATS
```
benbc, line 50: Do we not want the human-friendly error message here?

tests/lib/docker.py
```diff
@@ -23,8 +23,8 @@ def get_mapped_port_for_host(self, name, container_port):
         Given a port on a container return the port on the host to which it is
         mapped
         """
-        if isinstance(container_port, int):
-            container_port = f"{container_port}/tcp"
+        assert isinstance(container_port, int)
```
benbc, line 26: We can just skip this assertion, can't we? We'll soon discover if it's being passed nonsense.

tests/test_query_engine.py
```diff
@@ -31,7 +31,11 @@ def test_run_generated_sql_get_single_column_default_population(
     database, setup_test_database
 ):
     input_data = [
-        *patient(1, ctv3_event("Code1")),
+        *patient(
+            1,
+            RegistrationHistory(StartDate="2000-01-01", EndDate="2999-12-31"),
```
benbc, line 36: Why add this?

tests/test_validate_backend.py
```diff
@@ -1,72 +0,0 @@
...
-    assert "SELECT clinical_events.code AS code" in query1
-    query2 = str(results[1])
-    assert query2.startswith("SELECT * INTO #group_table_1")
-    assert "SELECT practice_registrations.patient_id AS patient_id" in query2
-
-    # 3rd query selects the results from the temp tables
-    query3 = str(results[2])
-    assert query3.startswith(
-        'SELECT "#group_table_1".patient_id AS patient_id, "#group_table_0".code AS code'
-    )
```
benbc: There seems to be a lot lost here. Is this related to the earlier rename commit that I commented on?

benbc (commented):
This is a good imrprovement. Thank you.

cohortextractor/date_utils.py
```diff
@@ -50,5 +60,3 @@ def _increment_date(date, period):
                 return date.replace(day=last_day_of_month)
         else:
             return date.replace(month=1, year=date.year + 1)
-    else:
-        raise ValueError(f"Unknown time period '{period}'")
```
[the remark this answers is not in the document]
rebkwok: This one was removed because the time period check is done earlier in the code; the assert isn't actually necessary and I only put it in in case we had other periods in future, to ensure we didn't else with anything other than "week"

cohortextractor/query_utils.py
```diff
@@ -35,6 +35,7 @@ def get_column_definitions(cohort_class):
 
 
 def get_measures(cohort_cls):
-    for name, value in get_class_vars(cohort_cls):
-        if name == "measures":
-            return value
+    return next(
+        (value for name, value in get_class_vars(cohort_cls) if name == "measures"),
+        [],
+    )
```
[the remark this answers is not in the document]
rebkwok: yes, makes sense

cohortextractor/validate_dummy_data.py
rebkwok: It's already checked earlier in the measures calculation, so an invalid file format never gets here

cohortextractor/validate_dummy_data.py
```diff
@@ -124,15 +124,14 @@ def date_validator(value):
...
     if (
         hasattr(query_node, "function")
         and query_node.function in functions_to_validator_mapping
     ):
         return functions_to_validator_mapping[query_node.function]
+    if (
+        hasattr(query_node, "column")
+        and query_node.column in columns_to_validator_mapping
+    ):
+        return columns_to_validator_mapping[query_node.column]
```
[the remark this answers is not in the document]
rebkwok: A query node always has a column; if we check for column first, it uses the validator for the column rather than the one for the function.  An `exists` function should have a boolean validator, for example

tests/lib/docker.py
rebkwok: I guess so

tests/test_query_engine.py
rebkwok: Just for coverage; the mock backend schema's `patient` function adds a default RegistrationHistory if one isn't included in the entities, but no test did that.  I left it in and added a test that used it in case it ewas useful later, but I guess it makes more sense to remove the `if` from the `patient` function instead

cohortextractor/validate_dummy_data.py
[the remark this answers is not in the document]
benbc: Oh, so you found an actual bug by increasing code coverage? That's great.

tests/test_validate_cohort.py
```diff
@@ -0,0 +1,63 @@
+import pytest
```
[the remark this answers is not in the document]
rebkwok: yes, it's a renaming of the `test_validate_backend`, because it's testing the validate_cohort command.  I'll check those files aren't different

tests/test_validate_backend.py
rebkwok: I think I noticed this and pushed another commit after you commented (but before I'd read the comment)

tests/test_validate_cohort.py
```diff
@@ -0,0 +1,63 @@
...
+    assert "SELECT clinical_events.code AS code" in query1
+    query2 = str(results[1])
+    assert query2.startswith("SELECT * INTO #group_table_1")
+    assert "SELECT practice_registrations.patient_id AS patient_id" in query2
+
+    # 3rd query selects the results from the temp tables
+    query3 = str(results[2])
+    assert query3.startswith(
+        'SELECT "#group_table_1".patient_id AS patient_id, "#group_table_0".code AS code'
+    )
```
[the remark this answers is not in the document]
rebkwok: It's really only checking that it's dumped SQL, not what it is. It's not writing the actual SQL anyway, and it may or may not be useful, so I don't think we care much what it is. I could make it just check that the output looks like SQL

benbc (approved):
Great

PR #235, opened 2021-11-15
Change approach to SQLAlchemy dialect configuration

evansd:
Query Engines already specified an associated dialect, but this attribute wasn't actually used and we relied on the global SQLAlchemy "registry" being configured as expected. We now force the connection to use the specified dialect.

This also means we have to separately configure the connections used from test fixture creation. This feels a bit duplicative because at the moment we use the same dialect for both use cases, but we may well want to diverge here: for test setup we need compatibility with the full API exercised by the ORM, but we don't care much about performance; for production we use only a more limited API, but we're very concerned about performance.


cohortextractor/query_engines/mssql.py
```diff
@@ -62,7 +62,7 @@ class MSSQLDateTime(_MSSQLDateTimeBase, sqlalchemy.types.TypeDecorator):
 
 
 class MssqlQueryEngine(BaseSQLQueryEngine):
-    sqlalchemy_dialect = sqlalchemy.dialects.mssql
+    sqlalchemy_dialect = sqlalchemy.dialects.mssql.pymssql.dialect
```
ghickman, line 65: I think this means we can tighten BaseSQLQuery.sqlalchemy_dialect type to `type[Dialect]` now?

ghickman (approved):
This makes much more sense, thanks for tidying up 👍

tests/lib/databases.py
```diff
@@ -17,6 +22,7 @@ def __init__(
         username="",
         password="",
         db_name="",
+        driver=None,
```
benbc, line 25: Is this really optional?

cohortextractor/query_engines/base_sql.py
```diff
@@ -247,7 +245,10 @@ def get_node_list(self, node):
     def engine(self):
         if self._engine is None:
             engine_url = sqlalchemy.engine.make_url(self.backend.database_url)
-            engine_url = set_driver(engine_url)
+            # Hardcode the specific SQLAlchemy dialect we want to use: this is the
+            # dialect the query engine will have been written for and tested with and we
+            # don't want to allow global config changes to alter this
+            engine_url._get_entrypoint = lambda: self.sqlalchemy_dialect
```
benbc, line 251: I'm a bit uncomfortable with this since it's so reliant on SQLAlchemy implementation details. But the global registry plus magic string parsing are really not helping us here.

So, what if we go with the flow? Could we make up our own driver names _per backend_? 1) register the correct driver for each backend name (even if that means registering the same driver twice with different names) and 2) insert the backend name into the URLs?

benbc:
```py
registry.register("databricks", "cohortextractor.query_engines.spark_lib", "SparkDialect")
```

and

```py
engine_url.set(drivername="databricks")
```

evansd: Yeah, I did think about that approach. We'd have to do something like this though:
```py
registry.register(
  "<unique-name-from-somewhere>",
  self.sqlalchemy_dialect.__module__,
  self.sqlalchemy_dialect.__name__,
)
```
And that didn't really feel like an improvement to me.

I wonder if the best thing to do here is to add an assert that we really have constructed an engine with the right dialect:
```py
assert isinstance(engine.dialect, self.sqlalchemy_dialect)
```
Then if SQLAlchemy change their internals here we'll know about it straight away.

benbc: Right. The insistence on separating the module and attribute name is unhelpful, isn't it? (And using strings not objects.)

I think your suggestion of adding an assert is a good one.

benbc (approved):

PR #237, opened 2021-11-15
Reduce to one row per patient based on cardinality

rebkwok:
Add the aggregation reduce functions (`exists` and `count`)

### DIFF
--- a/tests/test_dsl.py
+++ b/tests/test_dsl.py
@@ -71,7 +71,7 @@ def test_multiple_filters():
     assert_cohorts_equivalent(cohort, OldCohort)
 
 
-def test_exists_aggregation():
+def test_count_aggregation():
     class OldCohort:
         # Define tables of interest, filtered to relevant values
         num_events = table("clinical_events").count("code")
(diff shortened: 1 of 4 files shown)


tests/test_dsl.py
```diff
@@ -71,6 +71,34 @@ class OldCohort:
     assert_cohorts_equivalent(cohort, OldCohort)
 
 
+def test_exists_aggregation():
```
benbc, line 91: These test names are frack-to-bunt.

benbc (changes requested):
I _think_ the need to specify a column to pick with is obviated by sorting, which we plan to decouple from patient reduction in the new DSL (see [here](https://app.shortcut.com/ebm-datalab/story/16/add-a-sorting-operation-to-the-new-dsl)).

tests/test_dsl.py
rebkwok: oops :facepalm:

benbc (approved):
Thanks, Becky. All good if you're happy to remove that.

PR #241, opened 2021-11-17
Make test setup a method on database fixture

evansd:
This avoids the need to have two separate fixtures which is both less typing and also (I think) conceptually correct as it makes it impossible to have mismatched database and setup fixtures.


inglesp (approved):
placet

evansd:
gratias tibi ago

Issue #184, opened 2021-10-19
Before and after examples


sebbacon:
Examples illustrating the ideas in #181, comparing v1 with proposed v2

sebbacon:
Before:

```py
region = (
    patients.registered_practice_as_of(
        index_date,
        returning="nuts1_region_name",
    ),
)
```

After:

```py
table = backend.PatientRegistrations()
region = (
    t.filter((t.start_date >= index_date) & (t.end_date <= index_date))
    .select_column(t.nuts1_region_name)
    .make_one_row_per_patient(pick_first_value)
)
```

sebbacon:
Before:

```py
ethnicity = (
    patients.categorised_as(
        {
            "0": "DEFAULT",
            "1": "eth='1' OR (NOT eth AND ethnicity_sus='1')",
            "2": "eth='2' OR (NOT eth AND ethnicity_sus='2')",
            "3": "eth='3' OR (NOT eth AND ethnicity_sus='3')",
            "4": "eth='4' OR (NOT eth AND ethnicity_sus='4')",
            "5": "eth='5' OR (NOT eth AND ethnicity_sus='5')",
        },
        eth=patients.with_these_clinical_events(
            ethnicity_codes,
            returning="category",
            find_last_match_in_period=True,
            include_date_of_match=False,
        ),
        # fill missing ethnicity from SUS
        ethnicity_sus=patients.with_ethnicity_from_sus(
            returning="group_6",
            use_most_frequent_code=True,
        ),
    ),
)
```

After:

```py
clinical_events = backend.ClinicalEvents()
eth = (
    clinical_events.filter(clinical_events.code in ethnicity_codes)
    .sort_by(clinical_events.date)
    .make_one_row_per_patient(pick_last_value)
)

sus = backend.SUSEvents()
ethnicity_sus = sus.select_column(sus.group_6).make_one_row_per_patient(
    pick_modal_value
)

ethnicity_categories = {
    "1": (eth == "1") | (eth == "" & ethnicity_sus == "1"),
    "2": (eth == "2") | (eth == "" & ethnicity_sus == "2"),
    "3": (eth == "3") | (eth == "" & ethnicity_sus == "3"),
    "4": (eth == "4") | (eth == "" & ethnicity_sus == "4"),
    "5": (eth == "5") | (eth == "" & ethnicity_sus == "5"),
}

ethnicity = categorise(ethnicity_categories, default="0")
```

sebbacon:
[...]

PR #242, opened 2021-11-17
Add skeleton DSL and example

inglesp:
This draft PR contains a proposal for DSL classes and methods that constrain how users interact with the DSL, in order to provide as much help as possible with autocompletion, and to support providing good error messages when users try something invalid.

This proposal supports most of the constructs in #184.  However, I have chosen not to implement `make_one_row_per_patient`.  Instead of:

```
events.filter(events.code in codes)
    .sort_by(events.date)
    .select_column(events.code)
    .make_one_row_per_patient(pick_first_value)
```

I'm suggesting:

```
events.filter(events.code in codes)
    .sort_by(events.date)
    .first()
    .select_column(events.code)
```

This is because `.first()` and `.last()` can be suggested by autocompletion (since they are methods on a `SortedEventFrame`), while `pick_first_value` and `pick_last_value` can't (since they are names of arguments to a function).

I've also opted to support identifying columns by either their name as a string, or as attributes.  So `.sort_by(events.date)` and `.sort_by("date")` are equivalent, as are `.select_column(events.date)` and `.select_column("date")`.  There may be situations where this leads to ambiguity, in which case we'll have to insist on the attribute form.

There are docstrings, but at the moment they are designed to help developers understand the model.  Documentation for users will need to be friendlier!

I'm not intending to ever merge this PR!  But I would appreciate input from the whole team:

* Developers:
  * Is the code comprehensible?
  * Do the type definitions help or hinder?
  * Do you think any of the methods are unimplementable?
* Researchers:
  * Does the example make sense?

### DIFF
--- a/cohortextractor/dsl.py
+++ b/cohortextractor/dsl.py
@@ -1,0 +1,178 @@
+"""Skeleton classes and types for the DSL.
+
+This module contains classes and types for building a cohort.
+
+An example of the DSL in use is in cohortextractor/dsl_example.py.
+
+Data lives in database tables, which are represented by PatientTable (a table with one
+row per patient) and EventTable (a table with multiple rows per patient).
+
+Through filtering, sorting, aggregating, and selecting columns, we transform instances
+of PatientTable/EventTable into instances of PatientSeries.
+
+A PatientSeries represents a mapping between patients and values, and can be assigned to
+a Cohort. A PatientSeries can also be combined with another PatientSeries or a scalar
+value to produce a new PatientSeries.
+
+All classes except Cohort are intended to be immutable.
+
+Methods are designed so that users can only perform actions that are semantically
+meaningful.  This means that the order of operations is restricted.  In a terrible ASCII
+railway diagram:
+
+             (  filter                )*  select_column
+PatientTable ( --------> PatientFrame )  ---------------> PatientSeries
+                            ^                                   ^
+                            |                                   |
+                            | first, last                       |
+                            |                                   |
+                            |                                   |
+                    SortedEventFrame                            |
+                            ^                                   |
+                            |                                   |
+                            | sort_by                           |
+                            |                                   |
+                            |                                   |
+           (  filter              )*        count, exists       |
+EventTable ( --------> EventFrame )  ---------------------------+
+
+To support providing helpful error messages, we can implement __getattr__ on each class.
+This will intercept any lookup of a missing attribute, so that if eg a user tries to
+select a column from a SortedEventFrame, we can tell them they need to aggregate the
+SortedEventFrame first.
+"""
+
+from __future__ import annotations
+
+from typing import NoReturn, Union
+
+
+class Cohort:
+    """Represents the cohort of patients in a study."""
+
+    def set_population(self, population: PatientSeries) -> None:
+        """Set the population variable for this cohort."""
+
+    def add_variable(self, name: str, variable: PatientSeries) -> None:
+        """Add a variable to this cohort by name."""
+
+    def __setattr__(self, name: str, variable: PatientSeries) -> None:
+        return self.add_variable(name, variable)
+
+
+class EventFrame:
+    """Represents a collection of records, with multiple rows per patient.
+
+    Either an EventTable, or the result of filtering an EventTable.
+    """
+
+    def filter(self, filter: Expression) -> EventFrame:  # noqa: A002, A003
+        """Return a new EventFrame with given filter."""
+
+    def sort_by(self, *columns: ColumnOrName) -> SortedEventFrame:
+        """Return a SortedEventFrame with given sort column."""
+
+    def count(self) -> PatientSeries:
+        """Return a PatientSeries with count of matching events per patient."""
+
+    def exists(self) -> PatientSeries:
+        """Return a PatientSeries indicating whether each patient has a matching event."""
+
+    def __getattr__(self, name: str) -> NoReturn:
+        ...
+
+
+class SortedEventFrame:
+    """Represents an EventFrame that has been sorted."""
+
+    def first(self) -> PatientFrame:
+        """Return a PatientFrame with the first event for each patient."""
+
+    def last(self) -> PatientFrame:
+        """Return a PatientFrame with the last event for each patient."""
+
+    def __getattr__(self, name: str) -> NoReturn:
+        ...
+
+
+class PatientFrame:
+    """Represents a collection of records, with one row per patient.
+
+    Either a PatientTable, or the result of filtering a PatientTable.
+    """
+
+    def filter(self, filter: Expression) -> PatientFrame:  # noqa: A002, A003
+        """Return a new PatientFrame with given filter.
+
+        >>> filtered = table.filter(table.code in codes)
+        """
+
+    def select_column(self, column: ColumnOrName) -> PatientSeries:
+        """Return a PatientSeries containing given column.
+
+        >>> column = table.select_column(table.date)
+        >>> column = table.select_column("date")
+        """
+
+    def __getattr__(self, name: str) -> NoReturn:
+        ...
+
+
+class PatientSeries:
+    """Represents a column indexed by patient.
+
+    Can be used as a variable in a Cohort, or as an input when computing another
+    variable.
+    """
+
+    def __eq__(self, other: Expression) -> PatientSeries:  # type: ignore
+        """Return Expression indicating whether self is equal to other."""
+
+    def __ne__(self, other: Expression) -> PatientSeries:  # type: ignore
+        ...
+
+    def __add__(self, other: Expression) -> PatientSeries:
+        ...
+
+    def __gt__(self, other: Expression) -> PatientSeries:
+        ...
+
+    # etc
+
+    def __getattr__(self, name: str) -> NoReturn:
+        ...
+
+
+class BaseTable:
+    """A base class for database tables."""
+
+
+class EventTable(BaseTable, EventFrame):
+    """A base class for database tables with multiple rows per patient."""
+
+
+class PatientTable(BaseTable, PatientFrame):
+    """A base class for database tables with one row per patient."""
+
+
+class Column(PatientSeries):
+    """Represents a column in a database table."""
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+
+ColumnOrName = Union[Column, str]
+Expression = Union[PatientSeries, bool, int, str]
+
+
+class Codelist:
+    def __init__(self, codes: list[str]) -> None:
+        self.codes = codes
+
+    def __contains__(self, code: ColumnOrName) -> Expression:
+        """Return Expression indicating whether code in this codelist."""
+
+
+def categorise(mapping: dict[str, Expression]) -> PatientSeries:
+    """Represents a switch statement."""
--- a/cohortextractor/dsl_example.py
+++ b/cohortextractor/dsl_example.py
@@ -1,0 +1,51 @@
+from cohortextractor.dsl import (
+    Codelist,
+    Cohort,
+    Column,
+    EventTable,
+    PatientTable,
+    categorise,
+)
+
+
+# These table definitions will be imported from elsewhere, and researchers will not have
+# to write them.
+
+
+class Patients(PatientTable):
+    patient_id = Column("patient_id")
+    sex = Column("sex")
+
+
+class Immunisations(EventTable):
+    patient_id = Column("patient_id")
+    date = Column("date")
+    code = Column("code")
+
+
+patients = Patients()
+imms = Immunisations()
+
+# This is a cohort definition, of the kind that researchers will have to implement.
+
+cohort = Cohort()
+cohort.sex = patients.select_column(patients.sex)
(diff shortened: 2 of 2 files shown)


rebkwok:
In the very first new DSL ticket, I'd wondered about making `select_column` the last thing in the chain, and @sebbacon thought he had a reason for it being earlier.

I think `first()` and `last()` make sense, but I wonder if we're getting back to the old DSL issue of not being clear about what you're getting back from those methods, whereas something like `make_one_row_per_patient` is clearer?  We'd also need to implement `exists`, `count` and any other methods that reduce a table to one row per patient (not an objection, just an observation).  If the input to `make_one_row_per_patient` was an Enum representing the various reduce functions that are available, would we be able to get IDE autocomplete that way?

cohortextractor/dsl_example.py
```diff
@@ -0,0 +1,51 @@
...
+    code = Column("code")
+
+
+patients = Patients()
+imms = Immunisations()
+
+# This is a cohort definition, of the kind that researchers will have to implement.
+
+cohort = Cohort()
+cohort.sex = patients.select_column(patients.sex)
```
rebkwok, line 32: Is it confusing that we'll presumably import `Patients` and `Immunisations` (or whatever) from the same place, but e.e. `select_column` will return a different thing from each of them?  I guess it will return a `PatientSeries` from `patients`, and error if you tried to call it on `imms` without sorting/limiting first.

Maybe it's not, and it's obvious that the patients table doesn't need reducing and others do

inglesp:
> In the very first new DSL ticket, I'd wondered about making `select_column` the last thing in the chain, and @sebbacon thought he had a reason for it being earlier.

Do you have a link to that?

> I think `first()` and `last()` make sense, but I wonder if we're getting back to the old DSL issue of not being clear about what you're getting back from those methods, whereas something like `make_one_row_per_patient` is clearer?

Yeah, I've been pondering this.  I've never found `make_one_row_per_patient` particularly clear, because it's not making a row -- it's building a query.  It occurs to me that `.first_for_patient()` / `.last_for_patient()` / `.exists_for_patient()` / `.count_for_patient()` is a kind of middle ground -- what do you think?

> If the input to make_one_row_per_patient was an Enum representing the various reduce functions that are available, would we be able to get IDE autocomplete that way?

Definitely worth checking!

cohortextractor/dsl_example.py
inglesp: It might be confusing, but we can mitigate against it by catching the lookup in `.__getattr__` and providing a good error message.

rebkwok:
> Do you have a link to that?

No, because Seb couldn't remember what his reason for doing it that way was :)   See comments on sc-62.  He did say he wasn't commited to it necessarily though

rebkwok:
> Yeah, I've been pondering this. I've never found `make_one_row_per_patient` particularly clear, because it's not making a row -- it's building a query. It occurs to me that `.first_for_patient()` / `.last_for_patient()` / `.exists_for_patient()` / `.count_for_patient()` is a kind of middle ground -- what do you think?

I've always thought of it as building a query that selects a single row per patient from a table :shrug: 
`.._patient` is better, I think.  `count()` etc does imply that you might be counting everything, not just per patient

cohortextractor/dsl_example.py
```diff
@@ -0,0 +1,51 @@
...
+)
+cohort.second_date = second.select_column("date")
+cohort.second_code = second.select_column("code")
+
+cohort.codes_match = categorise(
+    {
+        "yes": cohort.first_code == cohort.second_code,
+        "no": True,
+    }
+)
```
madwort, line 51: thinking out loud - ordering is important in this hash, right? i.e. maps to a switch like:
```python
if cohort.first_code == cohort.second_code
  "yes"
elif True
  "no"
```
in which case would it be clearer to use an explicitly ordered structure? ottomh
```python
cohort.codes_match = categorise(
    [  
        ("yes", cohort.first_code == cohort.second_code),
        ("no", True),
    ]
)
```

inglesp: Dicts are now ordered in Python (although relying on that still gives me the willies) so the version in the example is safe.  But it might be better to be more explicit about this and use a list as you propose.

benbc:
> .first() and .last() can be suggested by autocompletion (since they are methods on a SortedEventFrame), while pick_first_value and pick_last_value can't

I don't think that's correct. If `pick_...` were the only objects of a type that was specified for the argument to `make_one_...()`, then the IDE will suggest autocompletion (at least mine will).

benbc:
> I've also opted to support identifying columns by either their name as a string, or as attributes.

There are two downsides to allowing strings: they can be misspelled and there's no way to offer the writer a palette of valid options. Obviously you _could_ use the references in which case you get the palette -- but if people see examples using strings then they'll copy them and may never become aware of the other option.

I _think_ we can probably even use parameterized types to ensure that you can only pass in references to columns on the right table, rather than any old one.

cohortextractor/dsl_example.py
madwort: thinking of things like this (which looks like a switch) https://github.com/opensafely/nhs-covid-vaccination-coverage/blob/b8111f531960d9ce8c36926f945eb54bdad9dda9/lib/data_processing.py#L138-L152
which would translate to the quite un-aesthetic
```python
cohort.codes_match = categorise(
    {
        "3": (df["care_home"]==1) & (df["age"]>=65),
        "1": (df["age"]>=80),
        "2": (df["age"]>=70),
        "4": (df["shielded"]==1),
        "5": (df["age"]>=65),
        "6": (df["LD"]==1),
        "7": (df["age"]>=60),
        "8": (df["age"]>=55),
        "9": (df["age"]>=50),
        "10": (df["age"]>=40),
        "11": (df["age"]>=30),
        "12": (df["age"]>=18)],
    }
)
```

madwort: alternatively, we could be encouraging researchers to write explicitly exclusive categories? (sorry if I'm heading off on a tangent here...)

benbc: I think the list of tuples version is a bit brutal syntactically for people who don't write code all the time.

benbc, line 51: Can we do this to make the syntax even simpler?

```py
cohort.codes_match = categorise(
    yes=cohort.first_code == cohort.second_code,
    no=True,
)
```

inglesp: I agree that that's quite un-aesthetic, but it looks a bit better if we drop the `df["..."]`s, which are unneeded here:

```
cohort.codes_match = categorise(
    {
        "3": (care_home == 1) & (age >= 65),
        "1": (age >= 80),
        "2": (age >= 70),
        "4": (shielded == 1),
        "5": (age >= 65),
        "6": (LD == 1),
        "7": (age >= 60),
        "8": (age >= 55),
        "9": (age >= 50),
        "10": (age >= 40),
        "11": (age >= 30),
        "12": (age >= 18),
    }
)
```

Rewritten as a list of `condition`, `value` pairs:

```
cohort.codes_match = categorise(
    [(care_home == 1) & (age >= 65), 3],
    [(age >= 80), 1],
    [(age >= 70), 2],
    [(shielded == 1), 4],
    [(age >= 65), 5],
    [(LD == 1), 6],
    [(age >= 60), 7],
    [(age >= 55), 8],
    [(age >= 50), 9],
    [(age >= 40), 10],
    [(age >= 30), 11],
    [(age >= 18), 12],
)
```

I find that harder to read.

It's easier to read with spacing, but ugly.

```
cohort.codes_match = categorise(
    [(care_home == 1) & (age >= 65),  3],
    [(age >= 80),                     1],
    [(age >= 70),                     2],
    [(shielded == 1),                 4],
    [(age >= 65),                     5],
    [(LD == 1),                       6],
    [(age >= 60),                     7],
    [(age >= 55),                     8],
    [(age >= 50),                     9],
    [(age >= 40),                    10],
    [(age >= 30),                    11],
    [(age >= 18),                    12],
)
```

I'm not sure what the right answer is here, but for now, I don't think it matters too much.

inglesp:
> > .first() and .last() can be suggested by autocompletion (since they are methods on a SortedEventFrame), while pick_first_value and pick_last_value can't
> 
> I don't think that's correct. If `pick_...` were the only objects of a type that was specified for the argument to `make_one_...()`, then the IDE will suggest autocompletion (at least mine will).

That's good to know.

inglesp:
> > I've also opted to support identifying columns by either their name as a string, or as attributes.
> 
> There are two downsides to allowing strings: they can be misspelled and there's no way to offer the writer a palette of valid options. Obviously you _could_ use the references in which case you get the palette -- but if people see examples using strings then they'll copy them and may never become aware of the other option.
> 
> I _think_ we can probably even use parameterized types to ensure that you can only pass in references to columns on the right table, rather than any old one.

Yeah, there are certainly downsides to using strings.  But did you know about [`typing.Literal`(https://adamj.eu/tech/2021/07/09/python-type-hints-how-to-use-typing-literal/)?  If we can parameterize the functions correctly to only accept valid columns (I'm not sure whether this is possible) then we could presumably do the same with strings.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,178 @@
...
+                            |                                   |
+                            |                                   |
+           (  filter              )*        count, exists       |
+EventTable ( --------> EventFrame )  ---------------------------+
+
+To support providing helpful error messages, we can implement __getattr__ on each class.
+This will intercept any lookup of a missing attribute, so that if eg a user tries to
+select a column from a SortedEventFrame, we can tell them they need to aggregate the
+SortedEventFrame first.
+"""
```
benbc, line 43: This class structure is great. Really clarifying.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,178 @@
...
+    def __add__(self, other: Expression) -> PatientSeries:
+        ...
+
+    def __gt__(self, other: Expression) -> PatientSeries:
+        ...
+
+    # etc
+
+    def __getattr__(self, name: str) -> NoReturn:
+        ...
```
benbc, line 143: Can we type-parameterize this like

```py
class PatientSeries(Generic[T]):
    def __eq__(self, other: Expression) -> PatientSeries[bool]:
        ...

    def __add__(self, other: Expression) -> PatientSeries[T]:
        ...
```

cohortextractor/dsl.py
```diff
@@ -0,0 +1,178 @@
...
+        """Set the population variable for this cohort."""
+
+    def add_variable(self, name: str, variable: PatientSeries) -> None:
+        """Add a variable to this cohort by name."""
+
+    def __setattr__(self, name: str, variable: PatientSeries) -> None:
+        return self.add_variable(name, variable)
+
+
+class EventFrame:
```
benbc, line 63: Do you anticipate having to write parallel table/frame/sorted-frame classes for each table? Or can we work some type-parameterization magic? Or something else?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,178 @@
...
+
+class EventTable(BaseTable, EventFrame):
+    """A base class for database tables with multiple rows per patient."""
+
+
+class PatientTable(BaseTable, PatientFrame):
+    """A base class for database tables with one row per patient."""
+
+
+class Column(PatientSeries):
```
benbc, line 158: Subclassing here seems odd. There clearly has to be a relationship between a column on a table and a patient series that has been specified on that column, but it's not clear that one _is_ an example of the other.

inglesp: Agree.  I think that `PatientSeries` / `PatientFrame` / `EventFrame` might be interfaces, which `Column` / `PatientTable` / `EventTable` implement.

cohortextractor/dsl.py
inglesp: Yes, I wondered about that.  I'm not sure how we do things like ensure that you can only do `s1 + s2` when both `s1` and `s2` are `PatientSeries[Num => T]`.  Possibly with `overload` or `TypeGuard`?  Would need to investigate.

cohortextractor/dsl.py
inglesp: I'm not sure this would necessary, but I need to do some implementing.

cohortextractor/dsl_example.py
inglesp: Yes, that's nice.  We lose the possibility of having non-string values though.

benbc:
> But did you know about typing.Literal?

No, I didn't!

cohortextractor/dsl_example.py
benbc: Yeah.

inglesp:
Some notes from a conversation with @sebbacon yesterday:

* The railroad diagram is helpful, and will need to be part of our public documentation.
* The intended purpose of the name `make_one_row_per_patient` was to make clear that access to data is being minimised.  Seb's happy with the `XXX_for_patient` names.
* We should consider whether the class names are adequate for building a mental model of what's going on.

inglesp:
Closing now, since implementation has begun in #244.

PR #243, opened 2021-11-18
Specify the population on a Cohort

rebkwok:
- Set the population variable on a Cohort
- Check that variables added to a Cohort are `Variable`s
- Check that the population variable is a boolean (the result of an `exists` query)*
- Require the population variable to be explicitly specified on a Cohort

Uses the `set_population` and `add_variable` layout from #242 

*When a table used in the DSL has knowledge of the types of its columns, I think the `make_one_row_per_patient` method (if we keep it) could record the variable type on the `Variable` (`PatientSeries`?) that it returns (which would need to be a combination of the column and the reduce function it used), and we can check that rather than the way I've done it here

### DIFF
--- a/cohortextractor/definition/definition.py
+++ b/cohortextractor/definition/definition.py
@@ -1,8 +1,30 @@
+from ..query_interface import Variable
 from .base import cohort_registry
 
 
 class Cohort:
-    ...
+    """Represents the cohort of patients in a study."""
+
+    population = NotImplemented
+
+    def set_population(self, population: Variable) -> None:
+        """Set the population variable for this cohort."""
+        self.add_variable("population", population)
+        if not self.population.reduce_function.__name__ == "exists":
+            raise ValueError(
+                "Population variable must return a boolean. Did you mean to use `make_one_row_per_patient(exists)`?"
+            )
+
+    def add_variable(self, name: str, variable: Variable) -> None:
+        """Add a variable to this cohort by name."""
+        if not isinstance(variable, Variable):
+            raise TypeError(
+                f"{name} must be a single value per patient (got '{variable.__class__.__name__}')"
+            )
+        self.__dict__[name] = variable
+
+    def __setattr__(self, name: str, variable: Variable) -> None:
+        return self.add_variable(name, variable)
 
 
 def register(cohort):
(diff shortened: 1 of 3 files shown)


tests/test_dsl.py
```diff
@@ -1,34 +1,37 @@
...
+
 from cohortextractor.concepts import tables
 from cohortextractor.definition import Cohort, count, exists, pick_first_value, register
 from cohortextractor.definition.base import cohort_registry
 from cohortextractor.query_language import table
 from cohortextractor.query_utils import get_column_definitions
 
 
-def test_minimal_cohort_definition():
+def test_minimal_cohort_definition(OldCohortWithPopulation, cohort_with_population):
```
rebkwok, line 12: I've so far only updated this test file with the default-population fixtures

cohortextractor/definition/definition.py
```diff
@@ -1,8 +1,31 @@
...
+    def __init__(self) -> None:
+        self.population = NotImplemented
+
+    def set_population(self, population: Variable) -> None:
+        """Set the population variable for this cohort."""
+        self.add_variable("population", population)
+        if not self.population.reduce_function.__name__ == "exists":
+            raise ValueError(
+                "Population variable must return a boolean. Did you mean to use `make_one_row_per_patient(exists)`?"
+            )
```
inglesp, line 17: I think it would make sense to do this check before the call to `add_variable`.

inglesp (approved):
All looks good -- one suggested tweak.

inglesp:
> When a table used in the DSL has knowledge of the types of its columns, I think the `make_one_row_per_patient` method (if we keep it) could record the variable type on the `Variable` (`PatientSeries`?) that it returns (which would need to be a combination of the column and the reduce function it used), and we can check that rather than the way I've done it here

Agreed.  @benbc's mentioned using parameterised types for this.

cohortextractor/definition/definition.py
rebkwok: I couldn't decide whether to do it before or after, but I'd thought it maybe made sense to raise the exception if it wasn't a Variable at all first (which happens in `add_variable`?

inglesp: Ah yes, good point!

tests/test_dsl.py
rebkwok: Moving this to a separate PR

PR #244, opened 2021-11-18

inglesp:
The main question for the reviewers is whether the implementation is understandable, and if it is, whether it is sensible.

I've tried to document what's going on in the module docstring, but if it's not clear I can talk you through the code.


cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+        *args: Any,
+        **kwargs: Any,
+    ):
+        self._path: list[PathItem] = parent._path + [
+            (parent, method_name, args, kwargs)
+        ]
+
+
+PathItem = tuple[Node, str, Any, Any]
+State = dict[str, Any]
```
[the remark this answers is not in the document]
ghickman: With the current use of State, it looks like it could be `dict[str, str]`, are you planning for there to be any other types of value in extra_state?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+Other notes...
+
+To support providing helpful error messages, we can implement __getattr__ on each class.
+This will intercept any lookup of a missing attribute, so that if eg a user tries to
+select a column from a SortedEventFrame, we can tell them they need to aggregate the
+SortedEventFrame first_for_patient.
+
+This docstring, and the function docstrings in this module are not currently intended
+for end users.
+"""
```
rebkwok, line 87: :clap:

benbc (approved):
I'm a _bit_ doubtful because we're producing an intermediate representation (the `Node` graph) just in order to allow us to build another intermediate representation (the query model). Do you think that's a long-term situation, or do you think that once we've got rid of the old DSL we might modify the query model to be directly constructable from the DSL?

ghickman (commented):
Coming to this with no knowledge of the DSL I've focussed on the description and use of types making sense to me.

The description for the architecture is absolutely great, I feel like I really understand what you're aiming for here now.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def set_population(self, population: PatientSeries) -> None:
+        """Set the population variable for this cohort."""
+
+    def add_variable(self, name: str, variable: PatientSeries) -> None:
+        """Add a variable to this cohort by name."""
+        object.__setattr__(self, name, variable)
+
+    def __setattr__(self, name: str, variable: PatientSeries) -> None:
+        return self.add_variable(name, variable)
```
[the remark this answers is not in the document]
inglesp: Yeah, this is very backwards.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+        state: State,
+        column: str,
+        **kwargs: str,
+    ) -> tuple[QMTable, State]:
+        if kwargs.get("not_equals", object()) is None:
+            assert "not_null_filter" not in state
+            extra_state = {"not_null_filter": column}
+        else:
+            extra_state = {}
+        return table.filter(column, **kwargs), extra_state
```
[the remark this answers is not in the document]
inglesp: Yes, good idea.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+    def _filter(
+        self,
+        table: QMTable,
+        state: State,
+        column: str,
+        **kwargs: str,
+    ) -> tuple[QMTable, State]:
+        if kwargs.get("not_equals", object()) is None:
+            assert "not_null_filter" not in state
+            extra_state = {"not_null_filter": column}
```
rebkwok, line 172: I'm finding all the negatives here hard to read.  I *think* it's checking if this filter operation is anything other than a != filter, asserting that we don't already have a `not_null_filter` column in the state, and adding this column to the state that gets passed on?  What happens with chained filters?  Wouldn't you have a `not_null_filter` column already in the state at the second filter operation?

cohortextractor/dsl.py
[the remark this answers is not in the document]
inglesp: We want `add_variable` to allow people to create variables dynamically without them having to use `setattr()`.

Ben says the current implementation is "[comically obtuse](https://github.com/opensafely-core/cohort-extractor-v2/pull/244#discussion_r752321284)", which I'll address.

cohortextractor/dsl.py
rebkwok: Also, what happens with a != filter?  Why don't we need to add that to the state?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+        *columns: str,
+    ) -> tuple[QMTable, State]:
+        return table, {"sort_columns": columns}
+
+    def _exists_for_patient(
+        self,
+        table: QMTable,
+        state: State,
+    ) -> tuple[ValueFromAggregate, State]:
+        return table.exists(state["not_null_filter"]), {}
```
rebkwok, line 190: Can't we call `exists` without filtering first (so there'd be no `not_null_filter` in state?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+def compile(cohort: Cohort) -> dict[str, Value]:  # noqa: A001
+    """Return a dict mapping variable names to Query Model values."""
+
+    return {k: compile_variable(v._path) for k, v in vars(cohort).items()}
+
+
+def compile_variable(path: list[PathItem]) -> Value:
+    """Return the Query Model value for this variable."""
+
+    obj: QueryNode = Table(path[0][0].name)
```
rebkwok, line 300: Could `path[0][0]` be a named variable to be clearer?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+
+class Node:
+    name: str
+
+    def __init__(
+        self,
+        parent: Node,
+        method_name: str,
+        *args: Any,
+        **kwargs: Any,
```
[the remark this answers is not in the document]
inglesp: For now they're all strings, but I anticipate that we'll introduce `Column` and `Expression` types.  I'm going to leave this as is for now.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+    return {k: compile_variable(v._path) for k, v in vars(cohort).items()}
+
+
+def compile_variable(path: list[PathItem]) -> Value:
+    """Return the Query Model value for this variable."""
+
+    obj: QueryNode = Table(path[0][0].name)
+    state: State = {}
+
+    for node, method_name, args, kwargs in path:
```
rebkwok, line 303: Would it be clearer if  PathItem was a dataclass/namedtuple so we can use dotted attributes?  Because it's a named thing in the type annotation, I was expecting it to be used like a class instance and had to remind myself it was defined as a tuple

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
+"""This module provides classes for building a cohort using the DSL and compiling it.
+
+Data lives in database tables, which are represented by PatientTable (a table with one
+row per patient) and EventTable (a table with multiple rows per patient).
+
+Through filtering, sorting, aggregating, and selecting columns, we transform instances
+of PatientTable/EventTable into instances of PatientSeries.
+
+A PatientSeries represents a mapping between patients and values, and can be assigned to
```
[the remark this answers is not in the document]
inglesp: No, it could be any value derived from values in the database.

cohortextractor/dsl.py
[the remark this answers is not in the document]
inglesp: > Could these Anys be tightened at all? Even just to object?

I don't think so -- if we use object, then mypy will object if we try to do anything with the objects.

There's an invariant here that is going to be very hard to encode with types: the compilation functions only pull things out of state if we know that the values are going to be there.

While this could be `dict[str, str]` for now, I think I'm going to leave this as is too.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,309 @@
...
+a Cohort.  In the future, a PatientSeries will be able to be combined with another
+PatientSeries or a scalar value to produce a new PatientSeries.
```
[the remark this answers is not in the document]
inglesp: Single value works too, I'll update.

cohortextractor/dsl.py
inglesp: This could all do with a comment!

cohortextractor/dsl.py
inglesp: Yes, good point.

cohortextractor/dsl.py
inglesp: Good idea.

cohortextractor/dsl.py
inglesp: Another good idea.

inglesp:
> I'm a _bit_ doubtful because we're producing an intermediate representation (the `Node` graph) just in order to allow us to build another intermediate representation (the query model). Do you think that's a long-term situation, or do you think that once we've got rid of the old DSL we might modify the query model to be directly constructable from the DSL?

I'm expecting to modify the query model to be directly constructable from the DSL.  In fact an earlier draft tried this, but came unstuck -- this is why the compiler has to keep track of some state.

cohortextractor/dsl.py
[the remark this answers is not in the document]
benbc: "almost comically obtuse" :grinning:

PR #245, opened 2021-11-18
Make population a required variable and remove the default query

rebkwok:


tests/conftest.py
```diff
@@ -36,3 +39,24 @@ def spark_database(containers):
...
+
+
+@pytest.fixture
+def OldCohortWithPopulation():
+    class OldCohort:
+        def __init_subclass__(cls):
+            if not hasattr(cls, "population"):
+                cls.population = table("practice_registrations").exists()
+
+    yield OldCohort
```
evansd, line 51: How come this needs to be a fixture? Couldn't you get the same effect by just importing this class into each of the test modules and then you wouldn't need to include it in every single test function signature.

rebkwok: It could; in fact I did it like that before and can't remember why I changed it.

rebkwok: I think it may have just been that I was doing the new version as a fixture; I'll change it back

evansd (approved):
Great, looks good - thanks!

PR #246, opened 2021-11-18
test: Parameterise tests over all query engines

evansd:
This PR creates an `engine()` pytest fixture which returns a `QueryEngineFixture()` object. This object provides the machinery required to write a test against a QueryEngine which is agnostic as to which particular query engine it runs against. We then update all the existing tests for which it makes sense to do so to use this new fixture. And finally we parameterise the fixture so the tests end up running over all our query engines.

There are a few caveats to this glorious conclusion:

1. Not all the tests against Spark actually pass. I have marked all the failing ones as xfail because I don't think fixing all the Spark issues belongs in this PR and have created separate tickets to track the three classes of failure that we're seeing:
    - #249
    - #250
    - #251

2. The Spark tests are annoyingly slow. Previously the entire test suite took 40s to run 236 tests. Now it takes 2m 54s to run 292 tests. It's easy enough to skip them locally using `-k 'not spark'` but that's obviously not ideal. I've created a separate ticket to track this and record the things I've tried already:
    - #248

3. I haven't done anything in the way of adding types to this yet, though it does pass the `mypy` check.

### DIFF
--- a/tests/lib/mock_backend.py
+++ b/tests/lib/mock_backend.py
@@ -47,26 +47,44 @@ class MockBackend(BaseBackend):
     )
 
 
+# Generate an integer sequence to use as default IDs. Normally you'd rely on the DBMS to
+# provide these, but we need to support DBMSs like Spark which don't have this feature.
+next_id = iter(range(1, 2 ** 63)).__next__
+
+
+# We need each NULL-able column to have an explicit default of NULL. Without this,
+# SQLAlchemy will just omit empty columns from the INSERT. That's fine for most DBMSs
+# but Spark needs every column in the table to be specified, even if it just has a NULL
+# value. Note: we have to use a callable returning `None` here because if we use `None`
+# directly SQLAlchemy interprets this is "there is no default".
+def null():
+    return None
+
+
 Base = sqlalchemy.orm.declarative_base()
 
 
 class RegistrationHistory(Base):
     __tablename__ = "practice_registrations"
-    RegistrationId = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
-    PatientId = sqlalchemy.Column(sqlalchemy.Integer)
-    StpId = sqlalchemy.Column(sqlalchemy.String)
-    StartDate = sqlalchemy.Column(sqlalchemy.Date)
-    EndDate = sqlalchemy.Column(sqlalchemy.Date)
+    RegistrationId = sqlalchemy.Column(
+        sqlalchemy.Integer, primary_key=True, default=next_id
+    )
+    PatientId = sqlalchemy.Column(sqlalchemy.Integer, default=null)
+    StpId = sqlalchemy.Column(sqlalchemy.String, default=null)
+    StartDate = sqlalchemy.Column(sqlalchemy.Date, default=null)
+    EndDate = sqlalchemy.Column(sqlalchemy.Date, default=null)
 
 
 class CTV3Events(Base):
     __tablename__ = "events"
-    EventId = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
-    PatientId = sqlalchemy.Column(sqlalchemy.Integer)
-    EventCode = sqlalchemy.Column(sqlalchemy.String(collation="Latin1_General_BIN"))
-    System = sqlalchemy.Column(sqlalchemy.String)
-    Date = sqlalchemy.Column(sqlalchemy.Date)
-    ResultValue = sqlalchemy.Column(sqlalchemy.Float)
+    EventId = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True, default=next_id)
+    PatientId = sqlalchemy.Column(sqlalchemy.Integer, default=null)
+    EventCode = sqlalchemy.Column(
+        sqlalchemy.String(collation="Latin1_General_BIN"), default=null
+    )
+    System = sqlalchemy.Column(sqlalchemy.String, default=null)
+    Date = sqlalchemy.Column(sqlalchemy.Date, default=null)
+    ResultValue = sqlalchemy.Column(sqlalchemy.Float, default=null)
 
 
 def ctv3_event(code, date=None, value=None, system="ctv3"):
@@ -75,11 +93,11 @@ def ctv3_event(code, date=None, value=None, system="ctv3"):
 
 class AllTests(Base):
     __tablename__ = "all_tests"
-    Id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
-    PatientId = sqlalchemy.Column(sqlalchemy.Integer)
-    PositiveResult = sqlalchemy.Column(sqlalchemy.Boolean)
-    NegativeResult = sqlalchemy.Column(sqlalchemy.Boolean)
-    TestDate = sqlalchemy.Column(sqlalchemy.Date)
+    Id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True, default=next_id)
+    PatientId = sqlalchemy.Column(sqlalchemy.Integer, default=null)
+    PositiveResult = sqlalchemy.Column(sqlalchemy.Boolean, default=null)
+    NegativeResult = sqlalchemy.Column(sqlalchemy.Boolean, default=null)
+    TestDate = sqlalchemy.Column(sqlalchemy.Date, default=null)
 
 
 def positive_test(result, test_date=None):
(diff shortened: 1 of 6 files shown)


tests/lib/lazy_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        # Keep a record of defined columns
+        cls._columns = {}
+        for key, value in vars(cls).items():
+            if isinstance(value, LazyColumn):
+                cls._columns[key] = value
+                # If we don't do this, instances of the class will have the column
+                # definition as their default column value, which is not what we want
+                setattr(cls, key, None)
+            elif isinstance(value, sqlalchemy.Column):
+                assert False, f"'{key}' must be a LazyColumn not a sqlalchemy.Column"
```
benbc, line 41: Could this be an `assert not instance...` guard clause for the loop?

tests/lib/lazy_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        self.args = args
+        self.kwargs = kwargs
+
+
+class HydrateLazyObjects:
+    def __init__(self):
+        self.orm_base_class = sqlalchemy.orm.declarative_base()
+        # Use a cache to make sure we get the same corresponding ORM class for each
+        # LazyBase subclass instance we see
+        self.get_orm_class = functools.cache(self.make_orm_class)
```
benbc, line 73: Why isn't this a decorator on `make_orm_class()`?

tests/lib/lazy_orm.py
```diff
@@ -0,0 +1,90 @@
...
+with the appropriate column types.
+
+This is, obviously, a terrible thing. But it's at least terrible in a sort of generic
+Python way rather than reaching deep into the internals of SQLAlchemy. So I felt on
+balance that it was the least-worst of the many terrible solutions I tried.
+
+It also has the advantage of being reasonably self-contained and easy to back out: the
+lazy ORM classes are just a find-and-replace away from being regular ORM classes; and
+the hydration class is used in only a single place.
+"""
```
benbc, line 24: Oh, Dave.

It's definitely short and it's definitely self-contained. Very clever of you leaving a cliff-hanger here, making me turn the page to see it actually used.

tests/lib/lazy_orm.py
```diff
@@ -82,9 +87,17 @@ def __call__(self, item):
...
             attrs[name] = column
         # Construct an ORM class using the supplied base class and the attributes
         # dictionary we've just built
         orm_class = type(lazy_class.__name__, (self.orm_base_class,), attrs)
         return orm_class
+
+    def transform_column_constructor_args(self, type_, *args, **kwargs):
+        # Modify the column type according to entries in the type_map
+        type_ = self.type_map.get(type_, type_)
+        return (type_,) + args, kwargs
```
benbc, line 103: Could you outline a few alternatives that you tried/thought through?

In particular, did you try just reaching in and dynamically modifying the objects?

benbc, line 103: Have you/would you consider raising an issue with SQLAlchemy about the lack of this type-mapping facility?

tests/lib/mock_backend.py
```diff
@@ -7,47 +7,72 @@
...
+    # We have to give this variable a new name because Python's scoping rules won't let
+    # you write:
+    #
+    #    def test():
+    #       foo = "bar"
+    #       class Test:
+    #           foo = foo
+    #
+    # Although this works outside of a function for reasons I don't fully understand.
+    query_engine_class_ = query_engine_class
```
benbc, line 23: Could you give the function argument the alternative name instead of aliasing via a local variable?

tests/test_codelists.py
```diff
@@ -58,6 +58,9 @@ class Cohort(OldCohortWithPopulation):
 
 @pytest.mark.integration
 def test_codelist_equals_query(engine):
+    if engine.name == "spark":
+        pytest.xfail()
```
benbc, line 62: There are _juuuuust_ enough of these for me to wonder if it could be factored out into a decorator.

benbc (commented):
As always, I'm wildly grateful for the care with which you have put this PR together. I think it's worth the complexity to get this working -- and we'll be laughing about it round the fire once there are twelve backends.

tests/lib/lazy_orm.py
evansd: Yes, that would be clearer, wouldn't it?

tests/lib/lazy_orm.py
evansd: Ah, I should have explained this better in the comment. We have to decorate the method on the instance rather than the class otherwise we get a single cache shared between all instances and the point of the code is to create _different_ classes per-instance.

tests/lib/mock_backend.py
evansd: I did do that originally, but I got irrationally annoyed by the fact that even positional argument names are arguably part of the public interface to a function (you can always call them as keywords if you want to) and now I'm forced to give my function a weirdly named argument just because of some Python scoping silliness. The grown up thing to do would be to acknowledge that this doesn't matter and call it `query_engine_cls` or something like that but I wasn't feeling very grown up at the time.

tests/test_codelists.py
evansd: I'm hoping they'll be short-lived enough for it not to be worth it...

tests/lib/lazy_orm.py
benbc: Huh. So `functools.cache` holds a single cache for all instances of the class, when it's used as a decorator? That seems surprising.

tests/lib/mock_backend.py
benbc: As you wish. Code needs a _bit_ of character.

tests/lib/lazy_orm.py
evansd: So I just tried this and was very surprised because it works exactly as you suggested (each instance gets its own cache) but I couldn't for the life of me figure out how because we only decorate the function once at definition time and `functools.cache` has no special handling for instance methods.

Then I eventually realised that they _do_ all share the same cache, it's just that the key is the full list of arguments which includes `self`; so this makes the single shared cache effectively partitioned by instance. That's probably obvious in retrospect, but it certainly confused me for a while.

I'll update the code and add a note.

tests/lib/lazy_orm.py
evansd: I probably should have discussed with the group earlier: there may well be better solutions floating around. I'll write up my current thoughts in a separate comment on this PR rather than burying them in a reply here.

tests/lib/mock_backend.py
evansd: I'll change it; I've calmed down a bit now.

evansd: I still don't understand why this works:
```console
python3.9 -c '

foo = "bar"
class Test:
  foo = foo
print(Test.foo)
'
```

And this doesn't:
```console
python3.9 -c '

def test():
  foo = "bar"
  class Test:
    foo = foo
  print(Test.foo)
  
test()
'
```

ghickman (commented):
Really clear set of changes 👍

tests/lib/mock_backend.py
```diff
@@ -47,26 +47,44 @@ class MockBackend(BaseBackend):
     )
 
 
+# Generate an integer sequence to use as default IDs. Normally you'd rely on the DBMS to
+# provide these, but we need to support DBMSs like Spark which don't have this feature.
+next_id = iter(range(1, 2 ** 63)).__next__
```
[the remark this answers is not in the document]
evansd: They're the same in one sense (`next(obj)` just calls `obj.__next__()`) but in this case I don't want to call the method, I just want a reference to it. This way `next_id` becomes a function which returns a new integer each time you call it.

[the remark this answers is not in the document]
ghickman: Oh, yes of course, very clever 👍

tests/lib/lazy_orm.py
benbc: I too read the `functools` code and I missed the use of `self` in the key.

tests/lib/lazy_orm.py
```diff
@@ -0,0 +1,107 @@
...
+        self.orm_base_class = sqlalchemy.orm.declarative_base()
+
+    def __call__(self, item):
+        """
+        Takes an arbitrary object and, if it is an instance of LazyBase, will transform
+        it into an instance of an appropriate ORM class. Otherwise returns its argument
+        unchanged.
+        """
+        if isinstance(item, LazyBase):
+            orm_class = self.get_orm_class(item.__class__)
```
benbc, line 81: This appears to be the wrong method name (should be `make_orm_class`).

benbc (approved):

tests/lib/lazy_orm.py
evansd: Yep! I did a quick "fix" just before leaving on Friday and then saw the test fail after it was too late to do anything about it.

evansd:
For those following along at home, I rebased against #260 and just dropped the commit which added `lazy_orm` and the follow up commit which added type customisation. And everything works!

PR #247, opened 2021-11-19

inglesp:
Here's another attempt at implementing dsl.py.  It's quite a bit simpler than #244 -- in particular, we don't build an intermediate representation in order to build a second intermediate representation, but build Query Model objects directly.


cohortextractor/dsl.py
```diff
@@ -0,0 +1,168 @@
...
+        self.name = name
+        self.qm_table: QMBaseTable = QMTable(name)
+
+
+class EventTable(BaseTable, EventFrame):
+    """A base class for database tables with multiple rows per patient."""
+
+
+class PatientTable(BaseTable, PatientFrame):
+    """A base class for database tables with one row per patient."""
```
ghickman, line 168: I'm wary of multiple inheritance here.  A developer needs to know which order the `__init__`s are resolved in to work out which of those methods will be executed (as neither is calling `super()`).

I think at least BaseTable is a good candidate for a Protocol once the team understands them better.

Nothing to do for this PR though, just wanted to flag the potential oddities around MRO resolution!

ghickman (commented):
Looks a lot simpler than the last iteration 👍

I'll leave approval to someone who is more familiar with the DSL!

rebkwok:
I'm not sure about this, but I don't feel it's separatating the  DSL and query model in the way we were trying to before (but I'll defer to @benbc because I still don't feel I've entirely got that separation clear in my head).  That was why the table classes that inherited from QueryBuilder before weren't instantiated with a query_language Table, but with a table name, and the DSL things didn't do any query building until they were compiled.  In this version, if you do `.exists()` etc, it'll be evaluated immediately into a query_language `Value`.

benbc:
This is, indeed, simpler and if we can do it this way then we should (because it's simpler).

I don't have any objection to the DSL machinery directly creating the query model. The query model will eventually be an immutable value type, just a data structure with no behaviour so doing arbitrary operations on it is really problematic (and easy to test).

PR #259, opened 2021-11-22
Automate versioning

ghickman:
This sets up automated versioning for the project.  There are a few pieces to the puzzle:

* remove the version from pyproject.toml since we're not publishing as a package to any indexes
* configure automated tagging with an action for commits to main
* a new workflow to detect new tags to main and build the docker image with that [git] tag

This means we can pip install from the git tag to get the desired version.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -1,0 +1,32 @@
+name: Tag repo; build and publish assets
+on:
+  create:
+    tags: ['v*']
+  workspace_dispatch:
+
+jobs:
+  build-and-publish-docker-image:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    env:
+      image: ghcr.io/opensafely-core/cohortextractor-v2:latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Get current git tag
+        id: tags
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+
+      - name: Build image
+        run: docker build . --file Dockerfile --tag ${{ env.image }}
+
+      - name: Log into GitHub Container Registry
+        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
+
+      - name: Push image to GitHub Container Registry
+        env:
+          VERSION: ${{ steps.tags.outputs.tag }}
+        run: |
+          docker tag ${{ env.image }} ${{ env.image }}:latest
+          docker tag ${{ env.image }} ${{ env.image }}:$VERSION
+          docker push ${{ env.image }}:$VERSION
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,17 +43,27 @@ jobs:
         with:
           failure-threshold: error
 
-  docker-publish:
-    if: github.ref == 'refs/heads/main'
+  tag-new-version:
+    # This uses `conventional commits` to generate tags.  A full list
+    # of valid prefixes is here:
+    # https://github.com/commitizen/conventional-commit-types/blob/master/index.json
+    #
+    # fix, perf -> patch release
+    # feat -> minor release
+    # BREAKING CHANGE in footer -> major release
+    #
+    # anything else (docs, refactor, etc) does not create a release
     runs-on: ubuntu-latest
-    needs: [test, lint-dockerfile]
-    env:
-      image: ghcr.io/opensafely-core/cohortextractor-v2:latest
+    outputs:
+      tag: ${{ steps.tag.outputs.new_version }}
     steps:
       - uses: actions/checkout@v2
-      - name: Build image
-        run: docker build . --file Dockerfile --tag ${{ env.image }}
-      - name: Log into GitHub Container Registry
-        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
-      - name: Push image to GitHub Container Registry
-        run: docker push ${{ env.image }}
+        with:
+          fetch-depth: 0
+      - name: Bump version and push tag
+        id: tag
+        uses: mathieudutour/github-tag-action@981ffb2cc3f2b684b2bfd8ee17bc8d781368ba60
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          default_bump: false
+          release_branches: main
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "flit_core.buildapi"
 [project]
 name = "cohortextractor"
 description = ""
-version = "2.0.0"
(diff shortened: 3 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -43,17 +43,27 @@ jobs:
         with:
           failure-threshold: error
 
-  docker-publish:
-    if: github.ref == 'refs/heads/main'
+  tag-new-version:
+    # This uses `conventional commits` to generate tags.  A full list
+    # of valid prefixes is here:
+    # https://github.com/commitizen/conventional-commit-types/blob/master/index.json)
```
iaindillingham, line 49: Is the trailing parenthesis intentional?

.github/workflows/main.yml
```diff
@@ -43,17 +43,27 @@ jobs:
...
-        run: docker build . --file Dockerfile --tag ${{ env.image }}
-      - name: Log into GitHub Container Registry
-        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
-      - name: Push image to GitHub Container Registry
-        run: docker push ${{ env.image }}
+        with:
+          fetch-depth: 0
+      - name: Bump version and push tag
+        id: tag
+        uses: mathieudutour/github-tag-action@981ffb2cc3f2b684b2bfd8ee17bc8d781368ba60
```
iaindillingham, line 65: I'm considering using `github-tag-action` for reusable actions. Why use this commit, specifically? And not, for example, the corresponding tag? (v4.5)

ghickman: Tags are mutable so this ensures we're pointing at the expected commit should they update v4.5 to point at something else.

.github/workflows/main.yml
ghickman: No, copypasta fail!

.github/workflows/main.yml
bloodearnest: 3rd party actions should be pinned to sha, not tag, so an attacker can't change the tag: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

1st part actions do not have that restriction, as we control the provenance of the tag.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,34 @@
+name: Tag repo; build and publish assets
+on:
+  create:
```
bloodearnest, line 3: Also want `workspace_dispatch:` here, as I guarantee you it won't work first time! :)

.github/workflows/main.yml
iaindillingham: Thanks for the comment and the  link, @bloodearnest.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+    tags: ['v*']
+
+env:
+  IMAGE_NAME: cohortextractor
```
bloodearnest, line 7: This is not needed as you are using a different variable, `image`

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: docker build . --file Dockerfile --tag ${{ env.image }}
+
+      - name: Log into GitHub Container Registry
+        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
+
+      - name: Push image to GitHub Container Registry
+        env:
+          VERSION: ${{ steps.tags.outputs.tag }}
+        run: |
+          docker tag $IMAGE_NAME $IMAGE_ID:latest
```
bloodearnest, line 32: This first line is not needed anymore with your explicit `image` variable.

FTR, the reason we used `IMAGE_NAME` is that we don't explicitly tag the build in other scenarios, as we have a multi-stage build Docker file, which tags each separate built image with specific names. Hence this retagging the prod version to publish in these two lines.

Currently, the Dockerfile for cohortextractor-v2 is not multi-stage, but it should be in the future. And therefore we may need to rethink the explicit tagging with `--tag ${{ env.image }}` above. But this is fine for now.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+          docker tag $IMAGE_NAME $IMAGE_ID:$VERSION
```
bloodearnest, line 33: This needs to be applied to your {{ env.image }}, not $IMAGE_NAME.

so

`docker tag ${{ env.image }} ${{ env.image }}:$VERSION`

I think?

And the publish should also have:

`docker push ${{ env.image }}:$VERSION` to push the tag

bloodearnest (commented):
A few comments.

One downside to this approach is that we are rebuilding the docker image here.

Ideally, we'd publish the same docker image that we built and tested in the the other action. But I'm not sure how to do that.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  create:
+    tags: ['v*']
+
+env:
+  IMAGE_NAME: cohortextractor
+
+jobs:
+  build-and-publish-docker-image:
+    if: github.ref == 'refs/heads/main'
```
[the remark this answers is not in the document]
ghickman: yeah :(

.github/workflows/build-and-deploy.yml
ghickman: So should I remove the first `docker tag …` here? or keep both but with `evn.image`?

ghickman: Oh, do you mean that `--tag ${{ env.image }}` in the `Build image` step is equivalent to the first line here?

bloodearnest: Yes, there's no need to tag with the prod tag as you've already done it via the expliciti --tag to `docker build`

ghickman: Gotcha, done.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Log into GitHub Container Registry
+        run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
+
+      - name: Push image to GitHub Container Registry
+        env:
+          VERSION: ${{ steps.tags.outputs.tag }}
+        run: |
+          docker tag ${{ env.image }} ${{ env.image }}:latest
+          docker tag ${{ env.image }} ${{ env.image }}:$VERSION
+          docker push ${{ env.image }}:$VERSION
```
bloodearnest, line 32: Does this automatically change `:latest`? on the registry? I guess it should, but we don't know until we try!

bloodearnest (approved):

.github/workflows/build-and-deploy.yml
ghickman: Time to find out! 🤞

PR #260, opened 2021-11-22
refactor: Define custom types on Dialect rather than Query Engine

evansd:
This moves the custom date and datetime types we need to implement for different databases from being defined on the Query Engine to being defined on the SQLAlchemy Dialect. This is the more natural place for them and, most importantly, it means they get automatically used by the ORM so we don't need to do anything silly to get cross-database ORM models to use as test fixtures.

This change wasn't quite as straightforward as it first seemed as there turn out to be some subtle differences in the way SQLAlchemy handles user-supplied custom types vs dialect-supplied customisations of built-in types. But these seem to be reasonably easy to work around once you find the relevant trick.

The `colspecs` attribute we use is documented briefly here:
https://docs.sqlalchemy.org/en/14/core/internals.html#sqlalchemy.engine.default.DefaultDialect.type_descriptor


cohortextractor/sqlalchemy_types.py
```diff
@@ -1,12 +1,52 @@
...
 
 TYPES_BY_NAME = {
-    "boolean": sqlalchemy.types.Boolean,
-    "date": sqlalchemy.types.Date,
-    "datetime": sqlalchemy.types.DateTime,
-    "float": sqlalchemy.types.Float,
-    "integer": sqlalchemy.types.Integer,
-    "varchar": sqlalchemy.types.Text,
-    "code": sqlalchemy.types.Text,
+    "boolean": Boolean,
```
benbc, line 45: I wonder about leaving these "nude" SQLAlchemy types qualified by the module name rather than importing them into this module, so that the casual reader can see that there's something special about the ones that we've wrapped?

cohortextractor/query_engines/mssql_dialect.py
```diff
@@ -46,6 +46,15 @@ class MSSQLDate(_MSSQLDateTimeBase, sqlalchemy.types.TypeDecorator):
...
 
+    def process_result_value(self, value, dialect):
+        # If the underlying database type is a DATETIME then that's what we'll get back,
+        # even if we've cast it to a DATE in our schema. So we make sure we always
+        # return the expected type here.
+        if isinstance(value, datetime.datetime):
+            return value.date()
+        else:
+            return value
+
```
benbc, line 57: Is there a failing test if this is missing?

benbc (approved):

cohortextractor/query_engines/base_sql.py
```diff
@@ -59,10 +59,6 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     # No limit by default although some DBMSs may impose one
     max_rows_per_insert: Optional[int] = None
 
-    def __init_subclass__(cls, **kwargs):
-        super().__init_subclass__(**kwargs)
-        assert cls.sqlalchemy_dialect is not None
```
ghickman, line 78: Did you test this is still enforced? I found with the initial work I did we need to make use of a Protocol to enforce a value on the subclass is defined, but we're currently using class types which don't enforce this at instantiation time.

cohortextractor/sqlalchemy_types.py
```diff
@@ -1,12 +1,52 @@
...
+    "Text",
+    "TYPES_BY_NAME",
+]
+
+# SQLAlchemy has different coercion rules for custom types vs built-in types. For
+# instance, if you compare a date and a string (using the standard types for both)
+# SQLAlchemy will decide how the type coercion should be done without ever checking
+# whether the current dialect supplies its own date type with its own coercion rules.
+# For our use case this is undesirable as we want to be able to control type coercion
+# behaviour. In particular we want to make sure that if a date is every compared to a
```
ghickman, line 21: s/every/ever/

ghickman (approved):
This looks like a nicer way to handle the custom types 👍

cohortextractor/sqlalchemy_types.py
evansd: I don't have a particularly firm view here. In my head the idea was that everywhere else in the codebase that touched SQLAlchemy types should import from here and it should be an alarm bell (although not necessarily an error) if we ever see them imported directly. I feel like the casual reader shouldn't have to care too much about the detail here beyond "always import types from this place". But open to arguments otherwise.

cohortextractor/query_engines/mssql_dialect.py
evansd: There is, which is how I realised there was a problem here, but not one that was explicitly designed to catch this problem. And now that you point that out it's obvious that I should add one :)

cohortextractor/query_engines/base_sql.py
evansd: I didn't, but I sort of felt I didn't care that much! My feeling is that if we need to add `__init_subclass` for other reasons then it's reasonable to assert that things are as we expect them to be before we do the initialisation. But adding an `__init_subclass` method _just_ for this purpose seems overkill. Especially as, in this case, it's not like it's preventing a subtle error that you might not otherwise notice :)

cohortextractor/sqlalchemy_types.py
benbc: I think you've accidentally explained why I'm wrong.

I was thinking of the casual reader of _this_ module. I had overlooked the fact that all the stuff from here is being imported elsewhere (and I think you're absolutely right that it's better to alias all the types via this module for that purpose). I suppose that in principle you _could_ use the qualified names here and yet still export the names from this module, but it would be weird. The reader elsewhere in the codebase is much more important.

cohortextractor/query_engines/base_sql.py
benbc: I think we've been going _slightly_ overboard with asserts on things like this, so I have no problem seeing it going. Asserts can be really helpful for more complex constraints or those that would cause weird problems if they are broken -- but as Dave suggests this will be really obvious if it happens.

PR #261, opened 2021-11-22
Rename Package

ghickman:
This renames the package to `cohortextractor2` so it can be installed in the documentation virtualenv alongside the v1 project.

We expect this to be a temporary solution until we break up the project into the user-facing and backend portions.


benbc (commented):
Does this break existing studies? There are only three of them, but we need to keep them working.

ghickman:
@benbc – arg, yes.  I'm not familiar with how studies invoke this, is it it all via the docker image?

benbc:
@ghickman Yes, it all runs in the Docker image, but the study definition imports stuff from `cohortextractor` explicitly. For example here: https://github.com/opensafely/SRO-Measures/blob/v2/analysis/study_definition.py.

ghickman:
Oh no… alright, I'll add a shim with the old name.

benbc:
I think we can deal with this. There are only three studies and I'm pretty sure that only Lisa's is under active development. But we'll need to provide fixes for the other two (the SRO one I linked to and https://github.com/opensafely/long-covid/tree/v2-with-validation); and help Lisa with it.

benbc:
We should try to let ourselves evolve and break stuff for now, rather than putting in lots of backwards compat stuff.

It may be that temporary shim is good just to help do the migration, but we should pull it out again straight away.

ghickman:
I _think_ a cohortextractor.py with `from cohorextractor2 import *` should be enough to satisfy the imports in existing studies.  I'll see if I can get those studies running with my branch locally.

StevenMaude:
question:

> […so it can be installed in the documentation virtualenv ](https://github.com/opensafely-core/cohort-extractor-v2/pull/261#issue-1060361267)

This being the [main OpenSAFELY documentation](https://github.com/opensafely/documentation) or something else?

If it' is the main OpenSAFELY documentation repository, then we should probably upgrade the documentation repository to Python 3.9, maybe? We're only testing this repository with Python 3.9.

(I made an issue anyway as it's probably worth doing regardless: opensafely/documentation#493 — whether it's worth me doing right now depends on whether it's relevant to cohortextractor2.)

inglesp (approved):
Looks reasonable, and easy to review.  Thanks for taking this on.

PR #268, opened 2021-11-23
Static contract checks

CarolineMorton:
Addresses: https://app.shortcut.com/ebm-datalab/story/202/allow-backend-implementers-to-check-that-they-pass-static-checks-of-their-contracts

Particularly:
1. Move the contract validation out of __init_subclass__ and into a public class method on BaseBackend.
2. Add a test (or tests) which enumerate all of the concrete backends and validates each of them; make sure that this produces a helpful failure message when there is a problem. 
3. Add a just target which runs just this one test (or set of tests), so that we can direct backend implementers to use it. (It's expected that this test will also run as part of the normal unit tests, so that developers working on the machinery get immediate feedback.) - **Question: Is this ok that you can't specify one backend you want to run this against?** 
4. Modify the cohort extractor entrypoint so that contracts on the specific backend being used are also checked at runtime.- **Question: Where in the code do you want this to execute? **


Justfile
```diff
@@ -147,3 +147,7 @@ test-all *ARGS: devenv build-cohort-extractor
     [[ -v CI ]] && echo "::group::Run tests (click to view)" || echo "Run tests"
     $BIN/python -m pytest --cov=cohortextractor --cov=tests {{ ARGS }}
     [[ -v CI ]]  && echo "::endgroup::" || echo ""
+
+# run the validate contract test only for all backends
+test-contracts *ARGS: devenv
```
benbc, line 152: `ARGS` isn't being used. It's probably less useful here than elsewhere, because the most frequent use is to specify a single test, but it could be useful, so there's no harm in having it as long as it's passed on to `pytest`.

tests/backends/test_base.py
```diff
@@ -0,0 +1,87 @@
...
+    good_backend.validate_all_contracts()
+
+
+def test_validate_all_backends():
+    """
+    Loops through all the backends, excluding test ones,
+    and validates they meet any contract that they claim to
+    """
+    real_backends = [cls.__name__ for cls in BaseBackend.__subclasses__()]
+    dummy_backends = ["BadBackend", "GoodBackend"]
```
benbc, line 71: Thinking about this again, I have doubts about my suggestion that it's better to do this by exclusion. If someone adds a new test backend there's no reason for them to think of excluding it here. How about we get all the subclasses and then only check the ones that are defined in `cohortextractor.backends`? We might add an assertion that there were at least three of them, to make sure we don't run into a problem with package restructuring.

tests/backends/test_base.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for backend in real_backends:
+        # Exclude test backends
+        if backend not in dummy_backends:
```
benbc, line 75: Tiny code style suggestion. I would write this loop like:

```py
for backend in real_backends:
    if backend in dummy_backends:
        continue
        
    # rest of loop follows
```

That "guard clause" makes it very clear to the reader that those cases cause the body of the loop to be skipped. They don't have to jump down to the bottom and eyeball indentation to work out if there is a matching `else` branch or some code following the `if`.

tests/backends/test_base.py
```diff
@@ -0,0 +1,87 @@
...
+    good_backend = GoodBackend(database_url="test")
+    good_backend.validate_all_contracts()
+
+
+def test_validate_all_backends():
+    """
+    Loops through all the backends, excluding test ones,
+    and validates they meet any contract that they claim to
+    """
+    real_backends = [cls.__name__ for cls in BaseBackend.__subclasses__()]
```
benbc, line 70: If you get a list of classes here rather than class names then you don't need to to do the import dance below, you can instantiate them directly.

tests/backends/test_base.py
```diff
@@ -0,0 +1,87 @@
...
+            backend_class = getattr(
+                importlib.import_module(
+                    f"cohortextractor.backends.{backend[:-7].lower()}"
+                ),
+                backend,
+            )
+
+            # Instantiate the class
+            backend_instance = backend_class(database_url="test")
+            backend_instance.validate_all_contracts()
```
benbc, line 87: `validate_all_contracts()` is a class method, so you don't need to create an instance here, you can call it directly on the class.

benbc (commented):
Please could you rebase this on `main` now that Dave has merged his recent changes, since the diff is a bit hard to read at the moment?

benbc:
> Question: Is this ok that you can't specify one backend you want to run this against?

Yes, absolutely. The backend implementer will only be working on a single backend, so the tests for the others will always pass; and these tests will be very fast. So it's just simpler for them not to have to specify the backend.

benbc:
> Question: Where in the code do you want this to execute?

Good question. I think in `cohortextractor.main.extract()`.

Justfile
CarolineMorton: thanks - I have removed ARGS for clarity.

tests/backends/test_base.py
CarolineMorton: Great - I have done this by removing reference to the dummy backends. The `getattr()` takes care of finding the real backends.

tests/backends/test_base.py
CarolineMorton: thanks - this has been removed

tests/backends/test_base.py
CarolineMorton: good point - remvoed.

benbc:
@CarolineMorton Please could you rebase this to make it easier to review?

Justfile
```diff
@@ -93,9 +93,9 @@ check: devenv
     $BIN/black --check .
     $BIN/isort --check-only --diff .
     $BIN/flake8
-    $BIN/interrogate cohortextractor2
+    $BIN/interrogate cohortextractor
     $BIN/pyupgrade --py39-plus --keep-percent-format \
-        $(find cohortextractor2 -name "*.py" -type f) \
+        $(find cohortextractor -name "*.py" -type f) \
```
benbc, line 98: These name changes are unexpected. Is this a mistake?

cohortextractor2/backends/base.py
```diff
@@ -25,48 +43,78 @@ def __init_subclass__(cls, **kwargs):
...
         self.database_url = database_url
         self.temporary_database = temporary_database
 
+    @classmethod
+    def _init_table(cls, table: SQLTable) -> None:
+        """
+        Initialises the table
+
+        Args:
+            name: Name of Table
```
benbc, line 64: This has got out of date.

cohortextractor2/backends/base.py
```diff
@@ -25,48 +43,78 @@ def __init_subclass__(cls, **kwargs):
...
+            table: SQLTable Object
+
+        Returns:
+            None
+        """
```
benbc, line 69: I'd argue that the type annotations do a better job of documenting this method than the docstring -- and they can't get out of date.

cohortextractor2/main.py
```diff
@@ -215,7 +219,20 @@ def added_to_path(directory):
...
 
-def extract(cohort_class, backend):
+def extract(
+    cohort_class: Union[Cohort, type], backend: BaseBackend
+) -> Generator[dict[str, str], None, None]:
+    """
+    Extracts the cohort from the backend specified
+
+    Args:
+        cohort_class: The definition of the Cohort
```
benbc, line 229: Since you've discovered that this name doesn't match the actual usage (it's _either_ a class, for the old DSL, or a `Cohort` instance, for the new DSL), I suggest that you rename this variable to reflect that (and the annotated type). I think `cohort_definition` would be better.

tests/backends/test_base.py
```diff
@@ -29,3 +50,83 @@ class BasicBackend(BaseBackend):
...
+                    f"cohortextractor2.backends.{file_of_backend_class}"
+                ),
+                backend,
+            )
+
+            backend_class.validate_all_contracts()
+            backend_no += 1
+
+    # Checks at least 3 backends
+    assert backend_no >= 3
```
benbc, line 132: I have a few comments on this test. Can I suggest that we pair on refactoring it a bit?

Justfile
CarolineMorton: oh no, definitely not. let me fix that now.

cohortextractor2/backends/base.py
CarolineMorton: agree- esp with the doc work that George is doing already.

cohortextractor2/main.py
CarolineMorton: I have named this `cohort_definition`

cohortextractor2/main.py
```diff
@@ -5,10 +5,14 @@
 import sys
 from contextlib import contextmanager
 from pathlib import Path
+from typing import Generator, Union
 
 import structlog
 
+from cohortextractor2.dsl import Cohort
```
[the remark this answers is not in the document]
CarolineMorton: done.

tests/backends/test_base.py
CarolineMorton: yes - sounds good.

benbc (approved):

PR #270, opened 2021-11-23
Simple column type-checking for backend contracts

rebkwok:
Adds type-checking for columns in a table contract against a backend, allowing for more than one acceptable type if necessary.

It does this simply by adding an `allowed_backend_types`  attribute to each type in `concepts.types`, which is a tuple of one or more keys from `sqlalchemy_types.TYPES_BY_NAME`.   `sqlalchemy_types.TYPES_BY_NAME` maps type names in a backend to sqlalchemy types.  The individual query engine dialects deal with custom handling of built in types, so we  should only need to check that the name mapping is correct here.

### DIFF
--- a/cohortextractor/concepts/table_contract.py
+++ b/cohortextractor/concepts/table_contract.py
@@ -46,6 +46,6 @@ class TableContract:
                 raise BackendContractError(
                     f"\n'{backend.__name__}.{table_name}' does not correctly implement the"
                     f" contract for '{cls.__name__}'\n\n"
-                    f"Column {column} is defined with an invalid type.\n\n"
+                    f"Column {column} is defined with an invalid type '{backend_column_type}'.\n\n"
                     f"Allowed types are: {', '.join(allowed_types)}"
                 )
--- a/cohortextractor/concepts/table_contract.py
+++ b/cohortextractor/concepts/table_contract.py
@@ -33,3 +33,15 @@ class TableContract:
                 f" contract for '{cls.__name__}'\n\n"
                 f"Missing columns: {', '.join(missing_columns)}"
             )
+        backend_table = getattr(backend, table_name)
+        for column in cls.columns:
+            backend_column_type = backend_table.columns[column].type
+            contract_column_type = cls.columns[column].type
+
+            if backend_column_type not in contract_column_type.allowed_backend_types:
+                raise BackendContractError(
+                    f"\n'{backend.__name__}.{table_name}' does not correctly implement the"
+                    f" contract for '{cls.__name__}'\n\n"
+                    f"Column {column} is defined with an invalid type.\n\n"
+                    f"Allowed types are: {', '.join( contract_column_type.allowed_backend_types)}"
+                )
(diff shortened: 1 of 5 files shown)


cohortextractor/concepts/types.py
```diff
@@ -1,15 +1,31 @@
 class BaseType:
-    pass
+    """
+    Base class for Column types defined in a Contract
+    """
+
+    # Subclasses must specify the backend Column types that they allow, as
+    # tuples of one or more keys from cohortextractor.sqlalchemy_types.TYPES_BY_NAME
+    allowed_backend_types: tuple = NotImplemented
```
benbc, line 8: In the commit message you note that this is "a tuple of one or more keys from `sqlalchemy_types.TYPES_BY_NAME`", but we don't assert that anywhere. Maybe we could check it in `__init_subclass_()`? I fear that otherwise spelling mistakes in this field will cause hard-to-debug errors.

Or, maybe even better, we could make the keys of `TYPES_BY_NAME` an enumeration and use that here (and in the backend definitions).

benbc (commented):
Looks correct. A couple of suggests about making bugs less likely or easier to debug.

cohortextractor/concepts/types.py
rebkwok: Yes, I wondered about both of these things.  I had an Enum initially and then simplified it for the first implementation, but I'll look at it again

cohortextractor/concepts/table_contract.py
```diff
@@ -33,3 +33,15 @@ def validate_implementation(cls, backend, table_name, table):
...
+        backend_table = getattr(backend, table_name)
+        for column in cls.columns:
+            backend_column_type = backend_table.columns[column].type
+            contract_column_type = cls.columns[column].type
+
+            if backend_column_type not in contract_column_type.allowed_backend_types:
+                raise BackendContractError(
+                    f"\n'{backend.__name__}.{table_name}' does not correctly implement the"
+                    f" contract for '{cls.__name__}'\n\n"
+                    f"Column {column} is defined with an invalid type.\n\n"
```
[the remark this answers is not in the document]
rebkwok: Good idea, will add

cohortextractor/concepts/types.py
```diff
@@ -1,15 +1,34 @@
...
+
 class BaseType:
-    pass
+    """
+    Base class for Column types defined in a Contract
+    """
+
+    # Subclasses must specify the backend Column types that they allow, as
+    # tuples of one or more keys from cohortextractor.sqlalchemy_types.TYPES_BY_NAME
+    allowed_backend_types: tuple[TYPES_BY_NAME, ...] = NotImplemented
```
ghickman, line 11: I think if you make this a subclass of Protocol then you can drop the ` = NotImplemented` and have the type system enforce it for you?

ghickman (commented):
Looks good, but I'm probably not the best person to sign it off!

cohortextractor/concepts/types.py
rebkwok: brb off to read more about Protocol

benbc (approved):

cohortextractor/concepts/types.py
ghickman: It worked 🎉

rebkwok: Oops forgot to uncomment the line I used to test it :)

PR #271, opened 2021-11-23

inglesp:
This PR supersedes #242, #244, and #247.  It implements enough of the DSL to make existing tests pass.  It is fairly similar to #247, except that instead of eg `SortedEventFrame.first_for_patient` returning a `PatientFrame`, it now returns an `AggregatedEventFrame`.  I expect this to make the implementation of `PatientFrame` easier, since an instance no longer has to know whether it was created with eg `SortedEventFrame.first_for_patient` or with eg `PatientFrame.filter`.

Note that I have simplified some of the populations on cohorts in `test_dsl.py`.  This is because I think there's some more design work to do in order to support `PatientFrame`s with boolean values.


cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
...
+        |   |        |                            |
+        +---+        | select_column              |
+                     V                            |
+               PatientSeries <--------------------+
+
+
+To support providing helpful error messages, we can implement __getattr__ on each class.
+This will intercept any lookup of a missing attribute, so that if eg a user tries to
+select a column from a SortedEventFrame, we can tell them they need to aggregate the
+SortedEventFrame first_for_patient.
```
benbc, line 49: I love this.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
+"""This module provides classes for building a cohort using the DSL.
+
+Data lives in database tables, which are represented by PatientTable (a table with one
+row per patient) and EventTable (a table with multiple rows per patient).
+
+Through filtering, sorting, aggregating, and selecting columns, we transform instances
+of PatientTable/EventTable into instances of PatientSeries.
```
benbc, line 7: This rewrite of the comment and diagram has lost the explicit statement that a `*Table` _is a_ `*Series`.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
...
+
+A PatientSeries represents a mapping between patients and values, and can be assigned to
```
benbc, line 9: > a mapping between patients and values

Maybe we can make this clearer? It's a one-to-one mapping, isn't it? And it's to a _single_ value (not a tuple of values like `*Frame`).

benbc: (Both of those are implicit elsewhere, but I think it's worth being _super_ clear.)

cohortextractor/query_language.py
```diff
@@ -212,14 +211,6 @@ class FilteredTable(BaseTable):
     value: Any
     or_null: bool = False
 
-    def __post_init__(self):
```
rebkwok, line 215: Do we want to allow filters on just codes?  This was added because the backend `clinical_events` table (for TPP and graphnet at least) has both `code` and `system`, and the query engine builds the codelist temporary table by querying both.  At the time I think Dave said we should require users to specify a system.  Maybe this should be a validation check somewhere else though

tests/acceptance/test_simplified_long_covid_study.py
```diff
@@ -17,7 +17,6 @@
 from ..lib.util import extract
```
rebkwok, line 17: This file should probably go - I don't think it gives us anything useful over the full long covid one, it's just a subset of that.

cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
...
+        """Add a variable to this cohort by name."""
+
+        self.__setattr__(name, variable)
+
+    def __setattr__(self, name: str, variable: PatientSeries) -> None:
+        if not isinstance(variable, PatientSeries):
+            raise TypeError(
+                f"{name} must be a single value per patient (got '{variable.__class__.__name__}')"
+            )
+        self.__dict__[name] = variable
```
benbc, line 87: Is there a reason for not spelling this `super().__setattr__(name, variable)`? To me that seems less invasive.

tests/test_dsl.py
```diff
@@ -79,80 +82,55 @@ class OldCohort(OldCohortWithPopulation):
...
         # Define tables of interest, filtered to relevant values
-        num_events = table("clinical_events").count("code")
+        num_events = table("clinical_events").filter("code", not_equals=None).count()
 
     cohort = cohort_with_population
     events = tables.clinical_events
-    cohort.num_events = events.select_column(events.code).make_one_row_per_patient(
-        count
-    )
+    cohort.num_events = events.filter(events.code, not_equals=None).count_for_patient()
```
rebkwok, line 89: why can't these just be 
```table("clinical_events").count()```
and
```events.count_for_patient()```

File: cohortextractor/dsl.py [hunk 8]
```diff
@@ -0,0 +1,185 @@
...
+
+
+class PatientFrame:
+    """Represents a collection of records, with one row per patient.
+
+    Either a PatientTable, or the result of filtering a PatientTable.
+    """
+
+    # TODO
```
benbc, line 96: Is this just not needed for the current set of tests?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
...
+class BaseTable:
+    """A base class for database tables."""
+
+    def __init__(self, name: str):
+        self.name = name
+        self.qm_table: QMBaseTable = QMTable(name)
+
+
+class EventTable(BaseTable, EventFrame):
+    """A base class for database tables with multiple rows per patient."""
```
benbc, line 181: I think this is deliberately skipping the call to `EventFrame.__init__()` in the knowledge that `BaseTable.__init__()` provides the same `self.qm_table`. That seems a bit fragile.

rebkwok (commented):
Looks good - I much prefer this simplified version.  Just a comment on the codelist validation, which may be unnecessary or deferred till later, and I'm not sure why we need the 
```table("clinical_events").filter("code", not_equals=None)``` filters in tests.

File: cohortextractor/dsl.py [hunk 10]
```diff
@@ -0,0 +1,185 @@
...
+
+    def __init__(self, value: Value):
+        self.value = value
+
+
+class BaseTable:
+    """A base class for database tables."""
+
+    def __init__(self, name: str):
+        self.name = name
```
benbc, line 176: Is this used for anything?

cohortextractor/dsl.py
```diff
@@ -0,0 +1,185 @@
...
+                     V                            |
+             SortedEventFrame                     |
+                     |                            |
+        +---+        | (first/last)_for_patient   | (count/exists)_for_patient
+        |   V        V                            |
+ filter |  AggregatedEventFrame                   |
+        |   |        |                            |
+        +---+        | select_column              |
+                     V                            |
+               PatientSeries <--------------------+
```
benbc, line 43: I've just had a thought about this sequencing, which I wish I hadn't had. There are other (hypothesised) aggregations like `mean_for_patient` or `max_for_patient`, for which we would want to do `select_column` before the aggregation. Sadly.

tests/fixtures/end_to_end_tests_tpp/tpp_cohort_new_dsl.py
```diff
@@ -1,19 +1,13 @@
...
-cohort.set_population(population)
-
-events = tables.clinical_events
-cohort.date = events.select_column(events.date).make_one_row_per_patient(
-    pick_first_value
-)
-cohort.event = events.select_column(events.code).make_one_row_per_patient(
-    pick_first_value
+cohort.set_population(tables.registrations.exists_for_patient())
+events = tables.ClinicalEvents()
```
benbc, line 7: Are you deliberately calling the constructor here rather than using `tables.clinical_events`?

tests/test_dsl.py
```diff
@@ -23,8 +24,8 @@ class OldCohort(OldCohortWithPopulation):
     # new DSL
     cohort = cohort_with_population
     events = tables.clinical_events
-    cohort.code = events.select_column(events.code).make_one_row_per_patient(
-        pick_first_value
+    cohort.code = (
+        events.sort_by(events.date).first_for_patient().select_column(events.code)
```
benbc, line 28: This is such a big improvement.

tests/test_dsl.py
```diff
@@ -79,80 +82,55 @@ class OldCohort(OldCohortWithPopulation):
 def test_count_aggregation(cohort_with_population):
     class OldCohort(OldCohortWithPopulation):
         # Define tables of interest, filtered to relevant values
-        num_events = table("clinical_events").count("code")
+        num_events = table("clinical_events").filter("code", not_equals=None).count()
```
benbc, line 85: Why do you have to change the old version?

benbc (commented):
This is getting better every time you write it. :-)

cohortextractor/dsl.py
rebkwok: I did this, and it was because `__setattr__` used to call `add_variable`.  I think `super` should be fine now

cohortextractor/dsl.py
inglesp: Given that we don't need `BaseTable.name`, we can drop `BaseTable` completely.

In tables.py, we can change

```
class ClinicalEvents(EventTable):
    code = "code"
    date = "date"

    def __init__(self):
        super().__init__("clinical_events")
```

to

```
class ClinicalEvents(EventTable):
    code = "code"
    date = "date"

    def __init__(self):
        super().__init__(QMTable("clinical_events"))
```

and then let `EventFrame.__init__` set `self.qm_table`.

cohortextractor/dsl.py
inglesp: No, let's ditch it.

tests/fixtures/end_to_end_tests_tpp/tpp_cohort_new_dsl.py
inglesp: Nope, that's a mistake, will fix.

tests/test_dsl.py
inglesp: Yes, agree.

tests/test_dsl.py
inglesp: Both `table("clinical_events").count("code")` and `table("clinical_events").filter("code", not_equals=None).count()` should return the same data, but they have different representations in the query model.

It might be possible to get `table("clinical_events").filter("code", not_equals=None).count()` to produce the simpler query model representation, but I'm going to go with [@rebkwok's suggestion to simplify this](https://github.com/opensafely-core/cohort-extractor-v2/pull/271/commits/a97955e73303d91536a08ced9640a69b26f9ae8f#r755904449).

cohortextractor/dsl.py
inglesp: I think you mean `*Frame` -- but yes, will address.

cohortextractor/dsl.py
inglesp: Will clarify.

cohortextractor/dsl.py
inglesp: Correct.

cohortextractor/query_language.py
inglesp: I think the solution to this will be in the implementation that supports doing `t.filter(t.code in codelist)`.

I hope we'll know then that `t.code` is contains codes.

cohortextractor/dsl.py
inglesp: I'll change.

tests/test_dsl.py
inglesp: Good point.

cohortextractor/dsl.py
inglesp: Don't worry, this is important to get right now, and it'll only take me a couple of hours to redraw the diagram.

`t.max_for_patient(t.value)` is equivalent to `t.sort_by(t.value).last_for_patient().select_column(t.value)`.  But this is likely to be common enough that we'd want to provide some sugar.

However, there's no equivalent of `mean_for_patient` or any other SQL aggregate functions like `SUM`.

Related, the DSL doesn't currently support passing columns to `count_for_patient` and `exists_for_patient`, so to find whether a patient has any records with a non-null value, you have to do something like `t.filter("value", include_null=True).exists_for_patient()`.  We may want to provide sugar here too.

So my proposal is to tweak/add methods to `EventFrame`:

* We tweak `(count|exists)_for_patient` so they now take an optional column (defaulting to `"patient_id"`).
* We add `(min|max|mean|sum)_for_patient`, which take a required column.
* We add `(first|last)_for_patient`, which take a required column, and which are sugar as described able.  We don't remove `SortedEventFrame.(first|last)_for_column`, since we need to support eg finding the code of the latest event.

sebbacon: I like this proposal. One possible counter is to do something like `aggregate_for_patient(agg.first)` but I think I prefer your suggestion. We're already going a bit domain-specific by talking about patients, so let's keep being domain-specific with our aggregation options....?

rebkwok: Do we need the `(first|last)_for_patient` syntactic sugar for min/max?  Those are sqlalchemy functions, can't they all just be passed to `qm_table.aggregate()`?

Is `SortedEventFrame.(first|last)_for_column` a typo (of `for_patient`)?

tests/test_dsl.py
inglesp: Addressed in 418e812e2bb1b71d28a7f9a2a628b3f5e4034d06.

cohortextractor/dsl.py
inglesp: Addressed in 33af403bb3b7560b4abd43449fb407488160ad2b.

cohortextractor/dsl.py
inglesp: Addressed in ffee5b6ce57594ee7ea10f7c2c2a090c4499dfd6.

cohortextractor/dsl.py
inglesp: Addressed in ffee5b6ce57594ee7ea10f7c2c2a090c4499dfd6.

cohortextractor/dsl.py
inglesp: Addressed in ffee5b6ce57594ee7ea10f7c2c2a090c4499dfd6.

cohortextractor/dsl.py
inglesp: Partially addressed in 11ef1d41afdc547b0269232fcc39251c7f6516f0.  I think that this will become clearer with some good documentation.

tests/acceptance/test_simplified_long_covid_study.py
inglesp: Addressed in 73f2b1c21f18d5d9b7ec5e814b2bc65025d3d5d6.

cohortextractor/query_language.py
inglesp: @benbc is there a ticket for error handling?  I think this belongs there.

cohortextractor/dsl.py
inglesp: > Do we need the `(first|last)_for_patient` syntactic sugar for min/max?

Good point!  I hadn't realised that `min_for_patient` would be equivalent to `first_for_patient`.  We should probably make `min`/`max` the defaults, and then offer `first`/`last` as aliases.

> Is SortedEventFrame.(first|last)_for_column a typo (of for_patient)?

Yes, fixed.

inglesp:
I think I've addressed all feedback, apart from https://github.com/opensafely-core/cohort-extractor-v2/pull/271#discussion_r755914891 which doesn't need to be addressed right now.

@benbc @rebkwok please could you re-review?

cohortextractor/__init__.py
```diff
@@ -4,7 +4,6 @@
 from .measure import Measure
 from .query_language import categorise, table
 
-
```
ghickman, line 7: This is a bit odd, isort is configured with `lines_after_imports = 2`, but it passed in CI so I don't think there's anything for you to do here, just flagging it!

benbc (approved):

cohortextractor/__init__.py
inglesp: Have you seen that I've changed the isort config to be compatible with black?  (They were fighting.)

PR #272, opened 2021-11-24
chore: Revert PRs related to versioning as the first one causes a problem

benbc:
This PR temporarily reverts a series of PRs so that we can take our time over working out what's causing an error. I'm reverting all of them although the problem seems to be caused by one of them because 1) it's the earliest one that causes the problem and 2) I don't know without picking through them how much of a dependency there is between them.

```
git revert -m1 6aa4c15
git revert -m1 4f51bc5
git revert -m1 43f1767
git revert -m1 806c67b
git revert -m1 7f98ce3
```


inglesp (approved):
Can confirm this fixes `just devenv`.

PR #273, opened 2021-11-24
Re-revert Packaging with Fixed Version Number

ghickman:
Something deep inside the packaging stack (in pep517 specifically) was very unhappy with a non-pep440 compliant version string, so this flips it to a minimally useful [local version identifier](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers) along with reverting the removal of all the related bits.


rebkwok (approved):
All looks good locally for me now

PR #275, opened 2021-11-25
Update test_full_long_covid_study_dsl

inglesp:


cohortextractor2/dsl.py
```diff
@@ -158,3 +160,10 @@ class PatientSeries:
 
     def __init__(self, value: Value):
         self.value = value
+
+
+Expression = Union[PatientSeries, bool, int, str]
```
ghickman, line 165: Would `Expression = PatientSeries | bool | int | str` work here? And maybe a better question… would it be more readable??

What's an Expression in this case? Could we add a short comment to make it more obvious?

ghickman (approved):
I'm probably not the best person to say if the tests are good, but they're very readable! 👍

benbc:
@inglesp I don't understand the purpose of this PR. Lots of this stuff doesn't work yet, does it?

inglesp:
> Lots of this stuff doesn't work yet, does it?

It doesn't work!  (So the tests are marked with `xfail`.)

But I wanted to update the test so that it uses our current DSL ideas, so that we've got something to aim for.

cohortextractor2/dsl.py
inglesp: This gives me an error:

```
cohortextractor2/dsl.py:163: error: Unsupported left operand type for | ("Type[PatientSeries]")  [operator]
cohortextractor2/dsl.py:166: error: Variable "cohortextractor2.dsl.Expression" is not valid as a type  [valid-type]
cohortextractor2/dsl.py:166: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

ghickman: Ooh, pipe unions are 3.10, that's probably it?

PR #276, opened 2021-11-26
Implement the patient demographics contract

rebkwok:
This PR updates the Patient table contract to reflect the real contract, with column constraints, and adds a `patient_demographics` table for the TPP backend which implements this contract.

Also adds a Constraint class, with the beginnings of a method for validating the data in backend databases based on those constraints.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -11,6 +11,8 @@ export COMPILE := BIN + "/pip-compile --allow-unsafe --generate-hashes"
 
 alias help := list
 
+test_args := "tests"
+
 # list available commands
 list:
     @just --list
@@ -124,7 +126,7 @@ connect-to-persistent-database:
     docker exec -it cohort-extractor-mssql /opt/mssql-tools/bin/sqlcmd -S localhost -U SA -P 'Your_password123!'
 
 # Full set of tests run by CI
-test *ARGS:
+test *ARGS=test_args:
     just test-all {{ ARGS }}
 
 # run the unit tests only. Optional args are passed to pytest
@@ -140,7 +142,7 @@ test-smoke *ARGS: devenv build-cohort-extractor
     $BIN/python -m pytest -m smoke {{ ARGS }}
 
 # run all tests including integration and smoke tests. Optional args are passed to pytest
-test-all *ARGS: devenv build-cohort-extractor
+test-all *ARGS=test_args: devenv build-cohort-extractor
     #!/usr/bin/env bash
     set -euo pipefail
 
--- a/Justfile
+++ b/Justfile
@@ -150,6 +150,5 @@ test-all *ARGS: devenv build-cohort-extractor
         --cov=tests \
         --cov-report=html \
         --cov-report=term-missing:skip-covered \
-        tests \
         {{ ARGS }}
     [[ -v CI ]]  && echo "::endgroup::" || echo ""
--- a/cohortextractor2/backends/tpp.py
+++ b/cohortextractor2/backends/tpp.py
@@ -71,10 +71,12 @@ class TPPBackend(BaseBackend):
         columns=dict(
             sex=Column("varchar"),
             date_of_birth=Column("date"),
+            date_of_death=Column("date"),
         ),
         query="""
             SELECT  Patient_ID as patient_id,
                 datefromparts(year(DateOfBirth), month(DateOfBirth), 1) as date_of_birth,
+                datefromparts(year(DateOfDeath), month(DateOfDeath), 1) as date_of_death,
                 CASE
                     WHEN Sex = 'M' THEN 'male'
                     WHEN Sex = 'F' THEN 'female'
--- a/cohortextractor2/backends/tpp.py
+++ b/cohortextractor2/backends/tpp.py
@@ -76,7 +76,11 @@ class TPPBackend(BaseBackend):
         query="""
             SELECT  Patient_ID as patient_id,
                 DateOfBirth as date_of_birth,
-                datefromparts(year(DateOfDeath), month(DateOfDeath), 1) as date_of_death,
+                CASE
+                    WHEN DateOfDeath = '9999-12-31' THEN NULL
+                    ELSE
+                        datefromparts(year(DateOfDeath), month(DateOfDeath), 1)
+                    END as date_of_death,
                 CASE
                     WHEN Sex = 'M' THEN 'male'
                     WHEN Sex = 'F' THEN 'female'
--- a/cohortextractor2/backends/tpp.py
+++ b/cohortextractor2/backends/tpp.py
@@ -59,12 +59,30 @@ class TPPBackend(BaseBackend):
     patient_join_column = "Patient_ID"
 
     patients = MappedTable(
-        implements=tables.Patients,
         source="Patient",
         columns=dict(
             sex=Column("varchar", source="Sex"),
             date_of_birth=Column("date", source="DateOfBirth"),
         ),
+    )
+
+    patient_demographics = QueryTable(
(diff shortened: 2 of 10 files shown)


cohortextractor2/backends/tpp.py
```diff
@@ -59,14 +59,32 @@ class TPPBackend(BaseBackend):
...
 
+    patient_demographics = QueryTable(
+        implements=tables.PatientDemographics,
+        columns=dict(
+            sex=Column("varchar"),
+            date_of_birth=Column("date"),
+        ),
+        query="""
+            SELECT  Patient_ID as patient_id,
+                datefromparts(year(DateOfBirth), month(DateOfBirth), 1) as date_of_birth,
```
rebkwok, line 77: This may not be necessary if TPP dates are already in Y-M-01 format

cohortextractor2/concepts/table_contract.py
```diff
@@ -1,13 +1,16 @@
...
 
+from .constraints import BaseConstraint
 from .types import BaseType
 
 
 @dataclasses.dataclass
 class Column:
 
     type: BaseType  # noqa: A003
+    description: str  # noqa: A003
```
rebkwok, line 11: I've added description, and kept the help field for notes about constraints.

cohortextractor2/concepts/tables.py
```diff
@@ -21,30 +22,40 @@ def __init__(self):
...
         ),
+        help="",
+        constraints=[],
     )
     date_of_birth = Column(
         type=types.Date(),
-        help="The patient's date of birth",
+        description="Patient's year and month of birth.",
+        help="The day will always be the first of the month. Must be present.",
+        constraints=[DateConstraint(match_format=["%Y-%m-01"])],
```
rebkwok, line 43: Where do we enforce required fields? Should we have a` RequiredConstraint` that checks there are no nulls in a column?

cohortextractor2/concepts/tables.py
```diff
@@ -21,30 +22,40 @@ def __init__(self):
...
-        type=types.Choice("F", "M"),
-        help="The patient's sex",
+        type=types.Choice("female", "male", "intersex", "unknown"),
+        description="Patient's sex.",
+        help=(
+            "One of male, female, intersex or unknown (the last covers all other options,"
+            "including but not limited to 'rather not say' and empty/missing values). "
+            "Must be present."
+        ),
+        constraints=[ChoiceConstraint()],
```
benbc, line 53: Can this be implied by the type, rather than having to duplicate it?

cohortextractor2/concepts/tables.py
benbc: Yes, I think that's another kind of constraint.

cohortextractor2/concepts/constraints.py
```diff
@@ -0,0 +1,42 @@
...
+    Defines a constraint on a date Column to ensure it conforms to required date format(s)
+    """
+
+    base_format = "%Y-%m-%d"
+
+    def __init__(self, match_format: str):
+        """
+        match_format: str.  A constraint format in the form '%Y-%m-%d' to compare with the base format.
+        e.g. '%Y-%m-01' ensures that each date in the backend column repesents the 1st of a month
+        """
```
benbc, line 35: I'm doubtful that the generality you're providing for here will _really_ prove to be useful. Maybe better to have a date-is-first-of-month constraint, which might also make generating documentation for it easier, them generalize later if there are other similar constraints that we need to support.

cohortextractor2/backends/tpp.py
```diff
@@ -59,14 +59,32 @@ class TPPBackend(BaseBackend):
...
+                CASE
+                    WHEN Sex = 'M' THEN 'male'
+                    WHEN Sex = 'F' THEN 'female'
+                    WHEN Sex = 'I' THEN 'intersex'
+                    ELSE 'unknown'
+                END as sex
+            FROM Patient
+        """,
```
benbc, line 85: It's great that this is fairly simple and that the existing `QueryTable` supports what we need.

benbc (commented):
This looks like a good start. We've got a story coming later to give backend implementers a way to run validation of their actual data, which I guess will hook into that stuff you're adding. Might be good to give a little thought to that now, as you are doing, but don't worry that it's not actually used in practice.

cohortextractor2/concepts/constraints.py
rebkwok: OK, sure, that makes sense

cohortextractor2/concepts/tables.py
rebkwok: yes, will do that

cohortextractor2/concepts/tables.py
```diff
@@ -21,30 +22,40 @@ def __init__(self):
...
 
     patient_id = Column(
         type=types.PseudoPatientId(),
-        help=(
+        description=(
             "Patient's pseudonymous identifier, for linkage. You should not normally "
             "output or operate on this column"
         ),
+        help="",
+        constraints=[NotNullConstraint(), UniqueConstraint()],
```
rebkwok, line 37: @benbc since patient_id isn't necessarily the primary key in a backend table, I think we probably do want to check this?

benbc: :+1:

rebkwok:
@benbc OK, I think I am done with this PR for now.  I'll leave it in draft atm while the patient demographics contract is still being finalised.  I've added DateOfDeath, and replaced ceiling dates with nulls. We can easily remove that if we decide not to put that in the contract, and I think there will be some help/description wording to update if we do.  If we don't I assume we'll want to have backends normalise their placeholder-for-null value to a consistent one, so there will still be some help/description wording to update.

rebkwok:
Making this non-draft now, since it seems there are a bunch of contract questions that will be pending for a while yet

cohortextractor2/backends/tpp.py
```diff
@@ -59,14 +59,38 @@ class TPPBackend(BaseBackend):
...
+        columns=dict(
+            sex=Column("varchar"),
+            date_of_birth=Column("date"),
+            date_of_death=Column("date"),
+        ),
+        query="""
+            SELECT  Patient_ID as patient_id,
+                DateOfBirth as date_of_birth,
+                CASE
+                    WHEN DateOfDeath = '9999-12-31' THEN NULL
```
rebkwok, line 80: This will need to be addressed again later; we may want to keep ceiling dates (in which case we need to not replace the day, and validation needs to deal with them), and/or we may need to replace nulls with ceiling dates

benbc (approved):
Looks great.

Justfile
```diff
@@ -150,6 +150,5 @@ test-all *ARGS: devenv build-cohort-extractor
         --cov=tests \
         --cov-report=html \
         --cov-report=term-missing:skip-covered \
-        tests \
```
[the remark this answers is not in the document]
rebkwok: I've changed it now so that "tests" is the default ARGS value - so it'll do what it did before by default, but still let you run just specific test(s) if you want to.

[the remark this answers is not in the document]
ghickman: Simon just hit the same thing in https://github.com/opensafely-core/cohort-extractor-v2/pull/278 and solved with adding this path to the config in pyproject.toml.  I think I'm very gently leaning towards the config option, only because it fits with our general ethos of default config in pyproject.toml, happy to be persuaded otherwise though!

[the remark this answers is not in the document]
rebkwok: I don't have a strong feeling either way, as long as I can run individual tests locally :)

PR #278, opened 2021-11-29
Support for connecting to real Databricks

bloodearnest:
This adds support for using the recommended databricks-sql-connector DBAPI when talking to an actual databricks instance.

It extends the test tooling to be able to run against an actual Databricks instance, locally and in CI.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -73,3 +73,53 @@ Starting with version 4.0, Bash is licenced under GPLv3. Because of this, macOS 
 ```bash
 brew install bash
 ```
+
+
+## Running tests against Databricks
+
+The test suite for Databricks/Spark backend by default runs tests against
+a local spark db in a container, for reliablility and speed.
+
+Open source Spark and Databricks' Spark are very similar, and this provides
+a good enough tests for the SQL parts of a backend.
+
+However, we still need to run tests against an actual Databricks instance, as
+the way the initial connection is made is different, and potentially other
+things also. We can run the test manually, using a helper script in
+`scripts/dbx` to manage our Databricks instance.
+
+
+### Running locally against Databricks
+
+First you need to set up your Databricks auth.
+
+1. Register for a free account at https://community.cloud.databricks.com/
+
+2. Log in to the databricks CLI tool (which is installed in the venv):
+
+    databricks configure --host https://community.cloud.databricks.com
+
+3. Test it's working with: `databricks clusters list`. If that doesn't error,
+   you are set up.
+
+
+You should then be able to run tests against databricks with:
+
+    just databricks-test [tests/backends/test_databricks.py]
+
+Warning: running the full test suite takes a long time.
+
+Note: This command will ensure there is an active Databricks cluster, and then
+run the tests against it.  Cluster creation is idempotent - if a cluster is
+alread up and running, it will use that. If it has terminated (after 2 hours of
+inactivity), it will delete the old one and create a new one.
+
+For more information about your Databricks cluster, you can use the dbx tool:
+
+    just dbx
+
+    unset DATABRICKS_URL
+
+5. Run the tests you want to run, e.g.
+
+    just test tests/backend/test_databricks.py
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -81,12 +81,17 @@ The test suite for Databricks/Spark backend by default runs tests against
 a local spark db in a container, for reliablility and speed.
 
 Open source Spark and Databricks' Spark are very similar, and this provides
-a good enough tests for the SQL parts of a backend.
+a good enough test basis for the SQL parts of a backend.
 
 However, we still need to run tests against an actual Databricks instance, as
-the way the initial connection is made is different, and potentially other
-things also. We can run the test manually, using a helper script in
-`scripts/dbx` to manage our Databricks instance.
+the way connections are made is different, and potentially more things down the
+line.
+
+Databricks is only only available in SaaS form, and  we use the free Community
+Edition version, but it is limited, slow, and not 100% reliable, so we
+do not run it by default, it needs to be manually run. We use some `just`
+commands and our helper script in `scripts/dbx` to manage a test Databricks
+instance to run the tests against.
 
 
 ### Running locally against Databricks
@@ -118,8 +123,12 @@ For more information about your Databricks cluster, you can use the dbx tool:
 
     just dbx
 
+or
+    ./scripts/dbx
+
+
 ### Running Databricks test in Github CI
 
 You can manually run the tests in github by triggering the "Databricks CI"
-action.  default it will just run the `tests/backends/test_databricks.py`, but
-you can specify different arguments when you trigger it.
+action. By default it will just run the `tests/backends/test_databricks.py`,
+but you can specify different arguments when you trigger it.
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -118,8 +118,8 @@ For more information about your Databricks cluster, you can use the dbx tool:
 
     just dbx
 
-    unset DATABRICKS_URL
+### Running Databricks test in Github CI
 
-5. Run the tests you want to run, e.g.
-
-    just test tests/backend/test_databricks.py
+You can manually run the tests in github by triggering the "Databricks CI"
+action.  default it will just run the `tests/backends/test_databricks.py`, but
+you can specify different arguments when you trigger it.
--- a/cohortextractor2/query_engines/spark_dialect.py
+++ b/cohortextractor2/query_engines/spark_dialect.py
@@ -30,7 +30,7 @@ class SparkDate(sqlalchemy.types.TypeDecorator):
         # return the expected type here.
         if isinstance(value, datetime.datetime):
             return value.date()
-        # databricks sql dbapi returns actual date
+        # databricks sql DBAPI returns actual date
         elif isinstance(value, datetime.date):  # pragma: no cover
             return value
         else:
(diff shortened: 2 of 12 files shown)


cohortextractor2/query_engines/spark_dialect.py
```diff
@@ -106,6 +115,39 @@ class SparkDialect(HiveHTTPDialect):
         sqlalchemy.types.DateTime: SparkDateTime,
     }
 
+    # This function is only excercised when manuall running the tests against databricks
```
sebbacon, line 118: *manually

cohortextractor2/query_engines/spark_dialect.py
```diff
@@ -95,6 +98,12 @@ def __init__(self, dialect):
 
 
 class SparkDialect(HiveHTTPDialect):
+    """Customisation of the base Hive dialect.
+
+    It is customised to add support for use with sqlalchemy ORM DDL and
+    connecting connecting to a Databricks hosted Spark.
```
sebbacon, line 104: *connecting connecting

sebbacon: As a naive drive-by reader, I don't understand the docstring here, because I'm wondering why a _sqlalchemy dialect_ needs to be customised for use with the _sqlalchemy ORM_?

sebbacon: As it's optionally customised for Databricks, should it be called `SparkOrDatabricksDialect`, or something like that?

sebbacon: And perhaps the docstring should explain the switching logic?

cohortextractor2/query_engines/spark_dialect.py
```diff
@@ -106,6 +115,39 @@ class SparkDialect(HiveHTTPDialect):
...
+        # databricks connection
+        #
+        # Switch to use databricks pyhive based-dbapi reather than the default
+        # This is not very clean, but its simple and it seems to work.
+        self.dbapi = databricks.sql
+
+        server_hostname = url.host
+        http_path = url.query["http_path"]
+
+        # TODO: handle access_token
```
sebbacon, line 142: Should this TODO still be here?

cohortextractor2/query_engines/spark_dialect.py
```diff
@@ -128,6 +170,48 @@ def _get_table_columns(self, connection, table_name, schema):
             else:
                 raise
 
+    # We do not currently use this function in our code, but we might in future.
```
sebbacon, line 173: What might we use it for? Why aren't we using it now? 

Is there an alternative to leaving in code which will probably rot anyway? A comment referring the reader to the _partition information_ weirdness & a historic commit?

DEVELOPERS.md
```diff
@@ -73,3 +73,53 @@ Starting with version 4.0, Bash is licenced under GPLv3. Because of this, macOS
...
 ```
+
+
+## Running tests against Databricks
+
+The test suite for Databricks/Spark backend by default runs tests against
+a local spark db in a container, for reliablility and speed.
+
+Open source Spark and Databricks' Spark are very similar, and this provides
+a good enough tests for the SQL parts of a backend.
```
sebbacon, line 84: s/a good/good

sebbacon: Mention that the Databricks CE is too flakey and not similar enough to production to use in tests locally, too?

DEVELOPERS.md
```diff
@@ -73,3 +73,53 @@ Starting with version 4.0, Bash is licenced under GPLv3. Because of this, macOS
...
+
+However, we still need to run tests against an actual Databricks instance, as
```
sebbacon, line 86: "...which given the limitations of Databricks CE will have to be a hosted SaaS edition..." ?

DEVELOPERS.md
```diff
@@ -73,3 +73,53 @@ Starting with version 4.0, Bash is licenced under GPLv3. Because of this, macOS
...
+    just databricks-test [tests/backends/test_databricks.py]
+
+Warning: running the full test suite takes a long time.
+
+Note: This command will ensure there is an active Databricks cluster, and then
+run the tests against it.  Cluster creation is idempotent - if a cluster is
+alread up and running, it will use that. If it has terminated (after 2 hours of
+inactivity), it will delete the old one and create a new one.
+
+For more information about your Databricks cluster, you can use the dbx tool:
```
sebbacon, line 117: maybe "_our_ dbx tool"? At first I assumed this was a Databricks thing.

sebbacon (changes requested):

cohortextractor2/query_engines/spark_dialect.py
bloodearnest: > As a naive drive-by reader, I don't understand the docstring here, because I'm wondering why a sqlalchemy dialect 
> needs to be customised for use with the sqlalchemy ORM?

Sqlalchemy does a lot more that just SQL, so its dialects provide the nitty gritty support multiple things about a DB, basically:
a) query SQL (which is usually fairly close to ANSI)
b) DDL SQL (which often drifts a bit more from ANSI IME)
c) Metadata inspection commands (usually very DB dependant)
d) Connection set up (which depends on the DBAPI used by the dialect)

The customisations @dave previously added are about type mapping between python and SQL (the date stuff), and pretending to support primary keys (which Spark does not support, but Sqlalchemy's ORM requires). He also handled some differences in error strings in order to raise the right exception for Sqlalchemy ORM's `create_all()` to understand.

The customisations I added were about switching which DBAPI we use, and handling it's the difference in strings output for partition tables.

Ultimately, this is domain knowledge about how Sqlalchemy works. We should probably have a primer like what I've written above, but not in this classes docstring I think.

bloodearnest: > As it's optionally customised for Databricks, should it be called SparkOrDatabricksDialect, or something like that?

Maybe, but there are there are two reasons its just called SparkDialect at the minute:

1) Databricks **is** Spark. It's SQL is virtually identical to the open source spark. Some of the customisations we've done to the dialect (date stuff, exception wrapping) apply to both the opensource spark we're using **and** databricks (as basically pyhive is a lagging a bit it seems).

2) We currently only support a single dialect hardcoded for a backend, for good reasons. So, for the "Databricks" backend (not its real name), we can only have one dialect. And as that dialect needs to be used to both run tests locally against a container and against a real databricks, it has to support both.

The only other option would be rehauling our backend model to support multiple dialects, or monkeypatching a spark only dialect in when when re run tests.  @dave and I felt that a 'smart' SparkDialect that can switch between connecting to databricks or regular spark (the only significant difference) was the better course of action.

bloodearnest: I do think the docstring could be re-worded, it's still valid but originally written before the full picture had settled, I will update it and you can see if it makes more sense.

bloodearnest: I do think the docstring could be re-worded, it's still valid but originally written before the full picture had settled, I will update it and you can see if it makes more sense.

cohortextractor2/query_engines/spark_dialect.py
bloodearnest: Yes. In community edition, we cannot use tokens. In production, they will almost certainly use tokens, and the kwargs will be different.

I can add that in this PR if you'd prefer, but I was going to do it in a later one, as this one's already larger than I wanted

cohortextractor2/query_engines/spark_dialect.py
bloodearnest: So, this is a copy of some databricks specific support from the only pre-exisiting sqlalchemy dialect I could find:

https://github.com/crflynn/databricks-dbapi/blob/master/databricks_dbapi/sqlalchemy_dialects/base.py

We don't want to use that module, as we want to use the recommended `databricks-sql-connector` DBAPI not pyhive's DBAPI. And the *only* other thing it brings in is this function. It is used by Sqlalchemy's machinery, but not machinery that we are using currently, but we may in future.

I didn't want to lose the knowledge that this is a databricks/OSS spark difference that could bite us in future, so I included the function. It's supports both OSS Spark and Databricks, so I saw know harm, and it might Just Work when we need it to and prevent an issue.

I'am +0 on keeping it in. If you would prefer to remove it, how can we capture the knowledge somewhere that this might be an issue in future, and if so, here's a fix?

bloodearnest: Maybe another solutions is getting the change upstreamed into pyhive? It's a ~10 character change, but no idea about testing.

https://github.com/dropbox/PyHive/blob/master/pyhive/sqlalchemy_hive.py#L306

DEVELOPERS.md
bloodearnest: I don't understand this. Databricks is a SaaS, regardless of CE or paid-for? We're definitely not going to try self-host the enterprise SaaS like NHSD do (given they are trying to avoid that!)

How about:
```
However, we still need to run tests against an actual Databricks instance, which is only available in SaaS form. We use the free Community Edition version to do so, but it is limited, slow, and not 100% reliable, so we do not run it by default, it needs to be manually run.
```

DEVELOPERS.md
bloodearnest: Have rolled this into to my suggested improvement below.

evansd (approved):
This is epic work! I had a couple of small comments, but nothing blocking.

As someone who's been deep in the weeds here as well the comments all made sense to me, but I think Seb's a better guide as to how comprehensible they'll be to others.

tests/lib/databricks_schema.py
```diff
@@ -14,10 +17,10 @@ class PCareMeds(Base):
     __tablename__ = "pcaremeds"
     __table_args__ = {"schema": "PCAREMEDS"}
 
-    Person_ID = Column(Integer, primary_key=True)
+    Person_ID = Column(Integer, primary_key=True, default=next_id)
```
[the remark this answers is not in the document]
bloodearnest: Aha, that makes sense will try it.

tests/lib/databricks_schema.py
```diff
@@ -50,12 +50,28 @@ class HESApcOtr(Base):
     SUSSPELLID = Column(Integer)
 
 
+def _get_schemas(base):
```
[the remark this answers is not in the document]
bloodearnest: I did actually start to write a generic form of this, but I was chasing a blind alley at the time, so I dropped it.

The only reason I added the drop schemas was that Databricks seemed to need it to prompt a clean up of the database files. However, I think I was now mistaken in that, but I think dropping and recreating schemas each time is cleaner anyway, so I left it in.

I'll add a comment about making this generic if its needed elsewhere.

sebbacon (approved):

DEVELOPERS.md
bloodearnest: I've updated this section with something along the line above, let me know if it makes more sense.

PR #280, opened 2021-11-30
Introduce categorization for new DSL

rebkwok:
This introduces the categorisation to the new DSL, along with some validatation of the category mapping.  It also requires adding the operator overrides for `PatientFrame`.


cohortextractor2/dsl.py
```diff
@@ -161,9 +169,109 @@ class PatientSeries:
     def __init__(self, value: Value):
         self.value = value
 
+    @staticmethod
+    def _get_other(other: Expression) -> Comparator | Value:
```
benbc, line 173: This pipe syntax is new in 3.10. Are we happy to only support that? It would be good if we can, but I suspect at least it means updating some dev environments.

cohortextractor2/dsl.py
```diff
@@ -161,9 +169,109 @@ class PatientSeries:
...
 
-def categorise(mapping: dict[str, Expression]) -> PatientSeries:
-    """Represents a switch statement."""
+    def __lt__(self, other: Expression) -> Comparator:
+        return self.value < self._get_other(other)
+
+    def __le__(self, other: Expression) -> Comparator:
+        return self.value <= self._get_other(other)
+
+    def __eq__(self, other: Expression) -> Comparator:  # type: ignore[override]
```
benbc, line 190: What's the reason for the ignore here? I feel we ought to be able to type this correctly.

cohortextractor2/dsl.py
```diff
@@ -161,9 +169,109 @@ class PatientSeries:
...
+        return ~self.value
+
+    def __repr__(self) -> str:
+        return f"PatientSeries(value={self.value})"
+
+
+Expression = Union[PatientSeries, Comparator]
+
+
+def categorise(mapping: dict[str, Expression], default: bool | None) -> PatientSeries:
```
benbc, line 212: I'd probably spell `bool | None` as `Optional[bool]`, which I think expresses the semantics more clearly.

benbc, line 212: The possibility of passing `None` for `default` isn't exercised in the tests. Do we mean to support it? If so, shouldn't we provide a default value for the argument of `None`, rather than expecting it to be passed explicitly?

cohortextractor2/dsl.py
```diff
@@ -161,9 +169,109 @@ class PatientSeries:
...
+    seen_values = set()
+    duplicates = set()
+    for key, value in mapping.items():
+        if not isinstance(value, (Comparator, PatientSeries)):
+            raise TypeError(
+                "Category values must be either a PatientSeries, or a "
+                "comparison expression involving a PatientSeries. "
+                f"Got '{value}' ({type(value)}) for category key '{key}'"
+            )
+        if repr(value) in seen_values:
```
benbc, line 264: Oh, ouch. I see why you've had to do this. Could we may be add a method for testing equality instead?

cohortextractor2/dsl.py
```diff
@@ -161,9 +169,109 @@ class PatientSeries:
...
+            duplicates.add(key)
+        seen_values.add(repr(value))
```
benbc, line 266: It seems a bit surprising to do one kind of checking as you go through and the other by iterating the results at the end.

benbc:
@inglesp Did you give any thought to including the default value in the mapping? It feels like that would make it easier to grok the definition -- but I do see that it is hard to have a really clean syntax for it. I guess I'd propose either the string `"default"` or a constant `Category.DEFAULT` as a value; neither ideal.

cohortextractor2/dsl.py
rebkwok: That's odd - `black` (I think) autoformatted it, but my local venv is 3.9

cohortextractor2/dsl.py
rebkwok: mypy complains because `__eq__` should return a bool in the thing it's overriding (object), and it says you shouldn't return a narrower return type than the thing being overridden

cohortextractor2/dsl.py
rebkwok: I did that, `black` changed it

rebkwok: Yes, that's an error,  it should default to `None`

cohortextractor2/dsl.py
rebkwok: I was thinking to report on all the duplicated values, but I guess it is more consistent with the other checks if it raises the error as soon as it encounters a duplicate

inglesp:
> @inglesp Did you give any thought to including the default value in the mapping?

We already support implicitly defaults, if the last value is `True`:

```
    age_categories = {
        "0-17": age < 18,
        "18-24": (age >= 18) & (age < 25),
        "25-34": (age >= 25) & (age < 35),
        "35-44": (age >= 35) & (age < 45),
        "45-54": (age >= 45) & (age < 55),
        "55-69": (age >= 55) & (age < 70),
        "70-79": (age >= 70) & (age < 80),
        "80+": age >= 80,
        "missing": True,
    }
```

But I thought it would be clearer if the default was explicitly provided as an argument to `.categorise()`.

However!  It occurs to me that we might want to make these dicts a first-class object (for instance, a user might want to categorise a number of variables in the same way) in which case the default belongs with the dict.  Using `Category.DEFAULT` (as an alias for `True`) makes sense to me here.

cohortextractor2/dsl.py
```diff
@@ -161,9 +169,73 @@ class PatientSeries:
...
         self.value = value
 
+    @staticmethod
+    def _get_other(other: Expression) -> bool | str | int | Value:
+        if isinstance(other, PatientSeries):
+            return other.value
+        return other
+
+    def __gt__(self, other: Expression) -> Comparator:
+        return self.value > self._get_other(other)
```
inglesp, line 179: I had expected that we would be returning a new `PatientSeries` here:

```
    def __gt__(self, other: Expression) -> PatientSeries:
        return type(self)(self.value > self._get_other(other))
```

This would allow us to do eg `(age >= 18) & (age < 25)` entirely with DSL objects, so that users never see QM objects.

cohortextractor2/dsl.py
benbc: Ooooh, I suppose it's because _all_ objects have `__eq__` defined. So we can get away with typing the other dunder methods "incorrectly" because we're not overriding anything. Python's really not build for being accurately typed, is it? Thanks. Any chance of a comment to explain?

cohortextractor2/dsl.py
benbc: Same as above, I suppose. I think that's a defect in Black, even if we were using 3.10.

cohortextractor2/dsl.py
benbc: :-) I think I'd actually do it the other way and defer both checks.

cohortextractor2/dsl.py
benbc: Oh! I missed that. Yes, I'm sure that's right. The ultimate query model shouldn't have support for syntactic niceties like operator overloading (and should be invisible to the study author).

cohortextractor2/dsl.py
rebkwok: Actually, it's `pyupgrade` that's changing it

cohortextractor2/dsl.py
rebkwok: Wasn't black, it was pyupgrade, and it's doing it because this file has `from __future__ import annotations` in it

cohortextractor2/dsl.py
rebkwok: Done :)

rebkwok:
@inglesp
> We already support implicitly defaults, if the last value is `True`
We don't now, because I implemented the validation in the ticket ("all values must be `PatientSeries`"), so it now has to be passed as an explicit argument to `categorise` - which is how `categorise` in the old DSL worked anyway.  It hadn't occurred to me that you could make the last value True to make an implicit default.  I think it's probably better for it to have to be explicit?

> However! It occurs to me that we might want to make these dicts a first-class object (for instance, a user might want to categorise a number of variables in the same way) in which case the default belongs with the dict. Using `Category.DEFAULT` (as an alias for `True`) makes sense to me here.

How would you envisage a first-class Category object being used to categorise different variables in the same way?  We could have a Catgeory that takes a PatientSeries variable and a mapping, but the mapping will reference the variable.  I think I'm missing something about what you mean here.

inglesp:
> We could have a Catgeory that takes a PatientSeries variable and a mapping, but the mapping will reference the variable. I think I'm missing something about what you mean here.

You're not missing anything, you're completely right, and I was talking rubbish.

cohortextractor2/dsl.py
```diff
@@ -158,12 +165,149 @@ class PatientSeries:
     variable.
     """
 
-    def __init__(self, value: Value):
+    def __init__(self, value: Value | Comparator):
         self.value = value
 
+    def is_comparator(self) -> bool:
```
benbc, line 171: It would be good to mark this as private by prefixing its name with an underscore. We need to be really careful about the surface we expose to study authors -- they shouldn't need to know this exists.

cohortextractor2/dsl.py
```diff
@@ -158,12 +165,149 @@ class PatientSeries:
...
+
+def raise_category_errors(errors):
+    """Recursively raise any category errors found"""
+    if not errors:
+        return
+    try:
+        next_category_error = errors.pop()
+        raise next_category_error
+    finally:
+        raise_category_errors(errors)
```
benbc, line 313: Is this a standard pattern for raising multiple exceptions in Python?

cohortextractor2/query_language.py
```diff
@@ -79,10 +79,31 @@ def __invert__(self):
...
+        return self._compare("__lt__", other)
+
+    def __le__(self, other):
+        return self._compare("__le__", other)
+
+    def __eq__(self, other):
+        return self._compare("__eq__", other)
+
+    def __ne__(self, other):
+        return self._compare("__ne__", other)
```
benbc, line 98: I don't understand what it means to carry about comparison operations on a comparator object. In my head a comparator effectively has a boolean type, so the logical operations implemented previously make sense, but these comparisons are surprising.

cohortextractor2/dsl.py
rebkwok: will do

cohortextractor2/query_language.py
rebkwok: You're right, it doesn't make sense to compare comparators (although you can compare boolean values with things, not that that makes any more sense really). 

I put this in to appease mypy, because if a `PatientSeries.value` can be either a `Value` or a `Comparator`, it complains if you try to do one of these operations on a `PatientSeries`.  I tried making a `ValueFromComparator` instead, so that a `PatientSeries` just takes a `Value` but then I just ran into the same issue further down the line, and it got more complicated.

An option might be to raise an exception if you try to use any of these methods?  So if you tried to do `(age == 20) < (age == 30)` in a mapping, it'll raise an exception because that's nonsense?

benbc: I really hope we don't have to add un-callable methods just to appease mypy. If it thinks they need to be there but we never call them then we must be lying to it somewhere. I don't know what else to suggest though, without having seen the details. Maybe we could look at this together tomorrow?

rebkwok: I added another commit that raises exceptions, which I think is better than letting them do nonsense things, but not ideal.  We don't call those methods, but we could.  We don't call them only because it doesn't make logical sense to call them, in the same way that it doesn't really make sense to say `(1 == 2) > 3` but python allows it (it's False, apparently).  And it's perfectly possible that a researcher might write something like that in error in a data definition, and it'd make more sense to tell them they couldn't do that than to let it pass without comment.

benbc (approved):

PR #281, opened 2021-11-30
Refactor DAG traversal code

evansd:
This is a small refactor to make QueryNodes responsible for knowing which other QueryNodes they reference (via a `_get_referenced_nodes()` method rather than the QueryEngine having to poke at their attributes to work it out.

This simplifies some of the DAG traversal code, particularly that dealing with category definitions.

### DIFF
--- a/cohortextractor2/query_engines/base_sql.py
+++ b/cohortextractor2/query_engines/base_sql.py
@@ -137,53 +137,6 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         else:
             assert False, f"Unhandled type: {node}"
 
-    def get_parent_nodes_from_category_definitions(
-        self, definitions, parent_nodes=None
-    ):
-        """
-        Get all the referenced parent nodes for category definitions.  If a category
-        definition (Comparator) has a LHS which is not itself a Comparator,
-        it will be a query node or a function-generated node
-        """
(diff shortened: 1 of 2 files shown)


ghickman (approved):

rebkwok (approved):
This makes sense to me, and simplifies things a lot.  I _think_ the query language module is the right place for the `_get_referenced_nodes` method to live, and it moves some of the query model stuff out of the query engine

cohortextractor2/query_engines/base_sql.py
```diff
@@ -397,25 +340,18 @@ def get_value_expression(self, value):
...
-            category_definitions = value.definitions.copy()
-            all_category_referenced_nodes = (
-                self.list_parent_nodes_from_category_definitions(
-                    category_definitions.values()
-                )
-            )
-            tables = tuple(
-                self.output_group_tables[self.get_output_group(query_node)]
-                for query_node in all_category_referenced_nodes
-            )
```
benbc, line 409: This seems like the first step on the way to getting rid of the two parallel data structures, is that right?

benbc (approved):
Great improvement.

cohortextractor2/query_engines/base_sql.py
evansd: It is, yes!

PR #282, opened 2021-11-30
Extract Contract Docs

ghickman:
This sets up pydocstyle and configures where to run it in the justfile so that our backends and contracts have docstrings in the correct format.

It also adds a command, `generate_docs`, to produce a JSON representation of the backends and contracts for the [mkdocs-opensafely-backend-contracts](https://github.com/opensafely-core/mkdocs-opensafely-backend-contracts/) plugin to consume.  This command only provides names and docstrings for contracts, and names and tables for backends currently, but is intended to provide the extra information required for the rest of the docs page the plugin will render later.

### DIFF
--- a/cohortextractor2/backends/tpp.py
+++ b/cohortextractor2/backends/tpp.py
@@ -5,14 +5,21 @@ from .base import BaseBackend, Column, MappedTable, QueryTable
 
 def rtrim(ref, char):
     """
-    MSSQL's {L,R}TRIM() functions (unlike TRIM()) don't allow trimming of arbitrary characters, only spaces. So we've
-    implemented it here. Note that this won't work as-is for '^' or ']', but they could be supported with a bit of
-    escaping.
-        * Since we're using PATINDEX(), which only works from the front of the string, we REVERSE() the
-          string every time we refer to it and then re-REVERSE() the result once we're done.
-        * We search through the string to find the index of the first character that isn't the thing we're stripping.
-        * Then we work out the length of the string remaining from that point to the end.
-        * Finally we take a substring from the index to the end of the string.
+    Right trim arbitrary characters.
+
+    MSSQL's {L,R}TRIM() functions (unlike TRIM()) don't allow trimming of
+    arbitrary characters, only spaces. So we've implemented it here. Note that
+    this won't work as-is for '^' or ']', but they could be supported with a
+    bit of escaping.
+
+      * Since we're using PATINDEX(), which only works from the front of the
+        string, we REVERSE() the string every time we refer to it and then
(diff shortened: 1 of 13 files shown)


cohortextractor2/docs.py
```diff
@@ -0,0 +1,62 @@
...
+            "contract_support": [],
+        }
+
+
+def _reformat_docstring(d):
+    """Reformat docstring to make it easier to use in a markdown/HTML document."""
+    docstring = d.strip()
+
+    lines = (line.strip() for line in docstring.split("\n"))
+    lines = (line for line in lines if line)  # remove empty items
```
benbc, line 48: Isn't there an argument for allowing paragraphs here?

benbc (approved):

cohortextractor2/backends/tpp.py
```diff
@@ -5,14 +5,21 @@
...
+    bit of escaping.
+
+      * Since we're using PATINDEX(), which only works from the front of the
+        string, we REVERSE() the string every time we refer to it and then
+        re-REVERSE() the result once we're done.
+      * We search through the string to find the index of the first character
+        that isn't the thing we're stripping.
+      * Then we work out the length of the string remaining from that point to
+        the end.
+      * Finally we take a substring from the index to the end of the string.
```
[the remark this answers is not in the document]
ghickman: I hadn't realised this wasn't public facing (although it's obvious in hindsight!), I'll unformat and confirm pydocstyle to only care about classes.

cohortextractor2/docs.py
ghickman: I'm not against them!  It just looked odd to me to maintain the formatting of the docstrings for the docs, do you have a preference?

benbc: Of course we can't really know until we've got a bunch of examples. But if the person writing the docstring feels the need for multiple paragraphs then I think it seems likely we'll want them in the final output.

You could make it a list of strings rather than relying on semantic empty lines.

cohortextractor2/backends/tpp.py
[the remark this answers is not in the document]
ghickman: Frustratingly I can't limit pydocstyle to just classes so we'll have to follow the format for any docstrings in the backend modules.

cohortextractor2/docs.py
ghickman: Yep that would work too.  This was to avoid

```
First line.

Second line.
```

getting turned into

```First line.  Second line.```

Note the double space between sentences!

ghickman: Fixed and rebased.

PR #285, opened 2021-12-02
Validate categories in dummy data validation

rebkwok:
Validates that category columns in dummy data contain valid categories

### DIFF
--- a/cohortextractor2/validate_dummy_data.py
+++ b/cohortextractor2/validate_dummy_data.py
@@ -80,7 +80,9 @@ def validate_column_values(df, column_definitions):
 
             try:
                 validator(value)
-            except ValueError:
+            except (ValueError, TypeError):
+                # Catch TypeError as well, as there's a possibility that calling the validator
+                # could raise a TypeError, depending on the value passed
                 raise DummyDataValidationError(
                     f"Invalid value `{value!r}` for {col_name} in row {ix + 2}"
                 )
@@ -104,14 +106,10 @@ def get_csv_validator(query_node):
 
     def category_validator(value, categories, default_category):
         """Ensure that a category value is one of the expected categories, or the default"""
-        # The default can be a different type (e.g. None, or the string "missing")
+        # The default must be either None, or of the same type as the categories
         category_type = type(categories[0])
-        try:
+        if value is not None:
             value = category_type(value)
-        except ValueError:
-            if default_category is not None:
-                default_type = type(default_category)
-                value = default_type(value)
 
         if not (value == default_category or value in categories):
             raise ValueError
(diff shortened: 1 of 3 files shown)


tests/test_validate_dummy_data.py
```diff
@@ -78,3 +78,28 @@ def test_validate_dummy_data_missing_data_file(file_format):
...
+@pytest.mark.parametrize("default_value", ["missing", None])
+def test_validate_dummy_data_with_categories(tmpdir, default_value):
+    class CohortWithCategories:
+        population = table("practice_registations").exists()
+        _code = table("clinical_events").filter(code__in=cl)
+        event_date = _code.latest().get("date")
+        _categories = {
+            1: event_date == "2021-01-01",
+        }
+        category = categorise(_categories, default=default_value)
```
rebkwok, line 92: I did old-style DSL here so as not to mix up old and new, but maybe it should be new-style instead :shrug:

cohortextractor2/validate_dummy_data.py
```diff
@@ -102,9 +102,18 @@ def date_validator(value):
...
-    def category_validator(value, categories):
-        category_type = type(list(categories)[0])
-        if category_type(value) not in categories:
+    def category_validator(value, categories, default_category):
+        """Ensure that a category value is one of the expected categories, or the default"""
+        # The default can be a different type (e.g. None, or the string "missing")
+        category_type = type(categories[0])
+        try:
+            value = category_type(value)
+        except ValueError:
```
evansd, line 111: I think this can raise `TypeError` as well as `ValueError` e.g. `int(object())` is a `TypeError`

cohortextractor2/validate_dummy_data.py
```diff
@@ -102,9 +102,18 @@ def date_validator(value):
         except ValueError:
             datetime.strptime(value, "%Y-%m-%d %H:%M:%S")
 
-    def category_validator(value, categories):
-        category_type = type(list(categories)[0])
-        if category_type(value) not in categories:
+    def category_validator(value, categories, default_category):
+        """Ensure that a category value is one of the expected categories, or the default"""
+        # The default can be a different type (e.g. None, or the string "missing")
```
evansd, line 107: Isn't the _only_ case where the default should be able to differ in type from the other values the case when it's None (i.e. it becomes SQL NULL)? I don't think we want to allow e.g. integer categories and a default of `"missing"`, do we?

evansd (approved):
Happy to approve this, but have a couple of questions below.

cohortextractor2/validate_dummy_data.py
rebkwok: Didn't think of that - yes, will catch both

cohortextractor2/validate_dummy_data.py
rebkwok: We currently do allow it, but we probably shouldn't.  I'll make it validate that here, and add the validation of the default to my other PR

evansd (approved):

PR #286, opened 2021-12-02
chore: Initial package rename

StevenMaude:
This PR renames:

* the Python project ("artefact") to `opensafely-databuilder`
* the entrypoint script to `databuilder`
* the directory ("package") to `databuilder`


StevenMaude:
question:
* do we want to ditch the underscore in the package name and go for `databuilder` or `osdatabuilder` (`os` is unfortunately overloaded with meaning) as the actual package/directory name?
* do we want to change the version number? With the rename, it's not really version 2 any longer.

ghickman:
I'm going to propose some terms here to avoid ambiguity around "packages" which is a little overloaded in Python land!

* `Artefact`: the thing which pip installs, can come from an index (like PyPI), a git repo, a local dir, etc.
* `Package`: a folder with a `__init__.py`

---

I would say we should go with:

* package name: `databuilder`
* entrypoint: `databuilder`
* artefact name: `opensafely-databuilder`

We don't need the opensafely prefix for the repo or project, the repo is implied by the org and lets save some characters on the imports, they're not really doing any branding for us.

The entrypoint and artefact name are sort of superfluous currently.  IIRC there's still a question mark over publishing an artefact to PyPI (side note: we use the [`opensafely-` prefix](https://pypi.org/search/?q=opensafely) there typically).  I added the entrypoint to match v1 but I have a feeling it might end up being in the same position as a potential artefact name, not that useful.

(Implementation side note: if we do publish this to PyPI we'll need to configure flit to have a local folder name for the package, with a different name for the artefact, there's an example of this in the [mkdocs-opensafely-backend-contracts pyproject.toml](https://github.com/opensafely-core/mkdocs-opensafely-backend-contracts/blob/e422c3c8d8dfdd5e1749c83cc408a584f9cb89f3/pyproject.toml#L52-L53).)

---

> do we want to change the version number? With the rename, it's not really version 2 any longer.

The version in pyproject.toml is effectively unused and has a [local version identifier as per PEP440](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers).  We started at `2` to be inline with this being version 2, but there's nothing which updates it since we don't publish an artefact to any indexes.  We just wanted an installable artefact which means setting `version` in pyproject.toml.  However we _are_ tagging the git repo and building Docker images from those versions since all current uses are done via those images.  So I don't think we need to worry about that version until we have a need for it!

StevenMaude:
Thanks @ghickman; coming back to this, that's really helpful. 
I agree that sticking to shorter names where possible is good.

I could only find one other [`databuilder` Python package](https://github.com/amundsen-io/amundsen/tree/2480cfbd5150576a33b6949f164844f31b58f08e/databuilder). I think it's very unlikely to have any overlap.

ghickman (changes requested):
One small change to pyproject.toml left to do and then I think this is good to go!

StevenMaude:
Based on the discussion above, I added a check that the package builds, similar to as in the [original cohort-extractor](https://github.com/opensafely-core/cohort-extractor/blob/38584aa8cffc39bca8cb8fa5780ee45c31ec058f/.github/workflows/test_runner.yaml#L51-L63).

ghickman (approved):
🎉

benbc (approved):

PR #288, opened 2021-12-03
Support codelist tables being written to temp db

bloodearnest:
- feat: support use of temporary database for codelist tables
- ci: fix dbx script to work inside and outside venv

### DIFF
--- a/cohortextractor2/query_engines/spark.py
+++ b/cohortextractor2/query_engines/spark.py
@@ -57,6 +57,9 @@ class SparkQueryEngine(BaseSQLQueryEngine):
         self._temp_table_names.add(temp_table_name)
         return temp_table_name
 
+    def get_temp_database(self):
+        return self.backend.temporary_database
+
     def post_execute_cleanup(self, cursor):
         """
         Called after results have been fetched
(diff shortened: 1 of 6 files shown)


cohortextractor2/query_engines/spark.py
```diff
@@ -57,6 +57,9 @@ def get_temp_table_name(self, table_name):
         self._temp_table_names.add(temp_table_name)
         return temp_table_name
 
+    def get_temp_database(self):
+        return self.backend.temporary_database
```
evansd, line 61: Does this attribute exist on the backend? I can't see it being created anywhere.

I'm also not sure whether it's a property of the backend or a property of the connection.I can see arguments either way. But as things are currently conceived the backend defines the schema but it doesn't know where that schema lives; that's defined by the environment at runtime. I feel like the temporary database name belongs in the same bucket.

evansd (approved):
Looks good apart from my question below.

cohortextractor2/query_engines/spark.py
bloodearnest: It's set in the base backend: https://github.com/opensafely-core/cohort-extractor-v2/blob/main/cohortextractor2/backends/base.py#L23

bloodearnest: I just used the existing config infrastructure for the TPP db's use of a temporary database, basically

evansd: Ahhh, fine yes, sorry! It's because I'd forgotten how the config gets passed in to the backend constructor. That could all do with refactoring at some point, but not now.

PR #293, opened 2021-12-06
Column types

benbc:


cohortextractor2/dsl.py
```diff
@@ -243,7 +243,7 @@ def not_null_patient_series(patient_series: PatientSeries) -> PatientSeries:
 
 
 def categorise(
-    mapping: dict[Expression, PatientSeries], default: Expression | None = None
+    mapping: Mapping[Expression, PatientSeries], default: Expression | None = None
```
inglesp, line 260: What does `Mapping` give us?

inglesp (approved):
Looks good.

cohortextractor2/dsl.py
```diff
@@ -55,7 +55,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Union
+from typing import Mapping, Union
```
ghickman, line 58: typing.Mapping was deprecated in 3.9, shall we use collections.abc.Mapping directly to avoid future churn?

cohortextractor2/dsl.py
```diff
@@ -237,13 +246,18 @@ def __hash__(self) -> int:
...
+
+
 def not_null_patient_series(patient_series: PatientSeries) -> PatientSeries:
     comparator_value = Comparator(lhs=patient_series.value, operator="__ne__", rhs=None)
     return PatientSeries(value=comparator_value)
 
 
 def categorise(
-    mapping: dict[Expression, PatientSeries], default: Expression | None = None
+    mapping: Mapping[Expression, PatientSeries], default: Expression | None = None
```
ghickman, line 260: What does the switch to Mapping do here?

ghickman (approved):
Looks great, just one question for help me understand the change.

rebkwok (commented):
Everything here makes sense, just a comment about naming; we now have a `Column` in `dsl.py`, another `Column` in `query_language.py`, another in the `backends/base.py` and a `DSLColumn` which is actually a type.  It's starting to get difficult to remember which one you're dealing with.  Is there something else we could call a  Column in either the DSL or query language?  And maybe types should be explictly named `ColumnType` etc?

cohortextractor2/dsl.py
benbc: `Mapping` is covariant in the value type, `dict` isn't. I actually _don't_ need that for this PR but used it in the spike this is based on when we need to pass in different types of `PatientSeries`.

Actually I think we might need a bit of rework for the typing of categories, so I'll pull that out now and we can reintroduce it later if necessary.

cohortextractor2/dsl.py
benbc: Oh yes, thank you. I've pulled this change out, actually.

cohortextractor2/dsl.py
benbc: See my answer to Peter's question.

benbc:
@rebkwok Yes, it's getting a bit messy. We need to sort out the relationship between the tables and contracts anyway -- and we discussed doing that as part of George's current story. Are you happy to merge this as is and fix up the column naming at the same time?

rebkwok:
> Are you happy to merge this as is and fix up the column naming at the same time?

yes, sure, as long as it's on the radar for later

rebkwok (approved):

PR #295, opened 2021-12-07
Operator overloading for filter syntax

benbc:


benbc:
@inglesp A few things about the surface here that deserve discussion.

1. `flake8` doesn't like us doing equality comparisons with `True`. I've experimented here with allowing `BoolColumn`s to be used directly as predicates to cope with that case. Let me know what you think of that.
2. `flake8` _also_ doesn't like doing equality comparisons against `None`. I've silenced it in that instance. I think there's a case for allowing _any_ column to be passed "naked" in order express a null check (like `events.filter(events.code)` instead of `events.filter(events.code != None)`). Not sure it's clear, though. I'd like to understand the use case for these checks.
3. I think maybe we should invert the codelist thing, so that we do `events.filter(code.is_in(codelist))` rather than `events.filter(codelist.contains(code))`. Seems a bit more consistent with the operator overloading elsewhere and it's a possibility now that we have these column types where before it was not possible.

benbc:
@inglesp I've added a commit which inverts the codelist membership test. Easy to pull it out if you don't like it.

benbc:
@inglesp And another commit with an alternative approach for checking nulls.

cohortextractor2/dsl.py
```diff
@@ -221,10 +213,51 @@ def __hash__(self) -> int:
...
+    def is_true(self) -> Predicate:
+        return Predicate(self, "equals", True)
+
+
+class DateColumn(Column):
+    def __gt__(self, other: str) -> Predicate:
+        return Predicate(self, "greater_than", other)
+
+    def __lt__(self, other: str) -> Predicate:
+        return Predicate(self, "less_than", other)
```
rebkwok, line 250: Should this have other operators too?  It's reasonable to expect to be able do <=, >=, ==, != on dates.  Possibly it'd make sense for there to be a base Column for shared operator methods - I'd expect `DateColumn` and `IntColumn` at least to share some of these methods.

cohortextractor2/query_language.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
rebkwok, line 1: There was some discussion about whether this was a bad idea or not;  I'm not sure we came to any conclusions, but maybe it should be discussed some more

tests/test_dsl.py
```diff
@@ -67,7 +68,7 @@ class OldCohort(OldCohortWithPopulation):
     cohort = cohort_with_population
     events = tables.clinical_events
     cohort.code = (
-        events.filter(codelist(["Code1"], "ctv3").contains(events.code))
+        events.filter(events.code.is_in(codelist(["Code1"], "ctv3")))
```
rebkwok, line 71: :+1: I prefer this

cohortextractor2/dsl.py
rebkwok: Also a date-specific `betweeen` method?

rebkwok (approved):
LGTM.  A few comments/questions but nothing major

cohortextractor2/dsl.py
```diff
@@ -91,11 +90,7 @@ class EventFrame:
...
 
-    def filter(  # noqa: A003
-        self,
-        column_or_expr: Column | CodelistFilterExpr,
-        **kwargs: str | Codelist | None | bool,
-    ) -> EventFrame:
+    def filter(self, predicate: Predicate | BoolColumn) -> EventFrame:  # noqa: A003
         """Return a new EventFrame with given filter.
 
         Note that while we are building the DSL, this method takes either an expression
```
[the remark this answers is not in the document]
benbc: Oh yes, I meant to come back to that. Thanks.

cohortextractor2/dsl.py
benbc: Oh yes, absolutely. I just added enough to satisfy the existing tests. You're right that I should chuck in a bunch more while I'm at it.

benbc: > Possibly it'd make sense for there to be a base Column for shared operator methods

Yes, I wondered about that. Or about having mixins for groups of operators. Or something. I'll try filling out the set of operators and then see if it makes sense to refactor.

cohortextractor2/dsl.py
```diff
@@ -222,6 +214,18 @@ def __hash__(self) -> int:
         return hash(repr(self.value))
 
 
+class Predicate:
+    def __init__(
+        self, column: Column, operator: str, other: str | bool | Codelist | None
```
inglesp, line 218: Must the LHS of the predicate be a `Column`?

This means we can do `.filter(col > 123)` but can't do `.filter(123 < col)`.

tests/test_dsl.py
inglesp: Me too.

Also, we can return to using `make_codelist` in these tests.

cohortextractor2/dsl.py
inglesp: Also: we're going to want to use `PatientSeries` in predicates.

For instance, in the example I documented we've got `covid_imms.filter(imms.date > (cohort.first_date + 28))`.  `imms.date` is a `Column`, but `cohort.first_date` and `cohort.first_date + 28` are `PatientSeries`.

inglesp (approved):
All looks good and easy to follow.

To respond to your questions:

> I've experimented here with allowing `BoolColumns` to be used directly as predicates to cope with that case...

Makes good sense.  We'd want to catch users doing `col == True` and tell them what they should do instead.  Also: will we need a way to negate a column to provide an alternative to `.filter(col == False)`?

> I think there's a case for allowing any column to be passed "naked" in order express a null check (like `events.filter(events.code)` instead of `events.filter(events.code != None)`)...

I like the explicit `isnull` that you've added in 6b2cb5a.

> I think maybe we should invert the codelist thing...

Agree.

---

Happy for this to be merged soon, but I'd like to revisit using `PatientSeries` in predicates.

benbc:
@inglesp 

> > I've experimented here with allowing BoolColumns to be used directly as predicates to cope with that case...

> Makes good sense. We'd want to catch users doing col == True and tell them what they should do instead.

Or, it occurs to me as I implement it, we could just allow them to do the `== True` check -- but provide the alternative `.is_true()`, so that they have a way of doing it that doesn't cause any warnings. WDYT?

inglesp:
> @inglesp
> 
> > > I've experimented here with allowing BoolColumns to be used directly as predicates to cope with that case...
> 
> > Makes good sense. We'd want to catch users doing col == True and tell them what they should do instead.
> 
> Or, it occurs to me as I implement it, we could just allow them to do the `== True` check -- but provide the alternative `.is_true()`, so that they have a way of doing it that doesn't cause any warnings. WDYT?

I don't have a strong opinion.  I'm happy with supporting multiple ways of doing something, as long as they're properly equivalent.

cohortextractor2/dsl.py
benbc: It seems that writing the comparison the other way round just works (TM).

benbc: I've added a separate story for using patient series in predicates.

PR #296, opened 2021-12-08
Add Rounding Methods

ghickman:
This adds the rounding methods `round_to_first_of_month` and `round_to_first_of_year` to PatientSeries, with implementations for both mssql and spark query engines.

We went with query engine specific implementations because we couldn't see a way to make the changes in a generic way.

### DIFF
--- a/databuilder/query_engines/mssql.py
+++ b/databuilder/query_engines/mssql.py
@@ -1,5 +1,9 @@
 import contextlib
 
+import sqlalchemy
+from sqlalchemy.sql.expression import type_coerce
+
+from .. import sqlalchemy_types
 from .base_sql import BaseSQLQueryEngine
 from .mssql_dialect import MSSQLDialect
 from .mssql_lib import fetch_results_in_batches, write_query_to_table
@@ -61,3 +65,14 @@ class MssqlQueryEngine(BaseSQLQueryEngine):
             # the normal manner
             with super().execute_query() as results:
                 yield results
+
+    def round_to_first_of_month(self, date):
+        date = type_coerce(date, sqlalchemy_types.Date())
+
+        first_of_month = sqlalchemy.func.datefromparts(
+            sqlalchemy.func.year(date),
+            sqlalchemy.func.month(date),
+            1,
+        )
+
+        return type_coerce(first_of_month, sqlalchemy_types.Date())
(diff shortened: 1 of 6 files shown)


benbc (approved):
Looks good, thank you.

Please could you add some more test cases to try to hit edge cases. I'm thinking in particular of first/last day of month and first/last month of year.

ghickman:
@benbc – updated with some edge cases, any others I could add?

benbc:
Since the first day of month calculation includes going back to the _previous_ month (and then forward again), I think we should include cases where that previous month is special. So: a date in March (to hit February weirdness, with and without it being a leap year) and a date in January (to check that it goes okay over the year boundary).

databuilder/query_engines/mssql.py
```diff
@@ -61,3 +65,33 @@ def execute_query(self):
...
+        )
+
+        # set month to first month of the year
+        first_of_year = sqlalchemy.func.dateadd(
+            sqlalchemy.literal_column("MONTH"),
+            -(sqlalchemy.func.month(date) - 1),
+            first_of_month,
+        )
+
+        return type_coerce(first_of_year, sqlalchemy_types.Date())
```
inglesp, line 97: We do something simpler [in v1](
https://github.com/opensafely-core/cohort-extractor/blob/main/cohortextractor/date_expressions.py#L325-L326
): 

```
    def date_function_first_day_of_year(self, date):
        return f"DATEFROMPARTS(YEAR({date}), 1, 1)"
```

In fact, seeing this I wonder whether we couldn't do the same for `round_to_first_of_month` too.

@evansd might know why we used `f"DATEADD(DAY, 1, EOMONTH({date}, -1))"` in v1.

evansd: I have no idea. My only thought is that in the `DATEFROMPARTS` solution the `{date}` expression needs to occur twice and it may itself be a complex expression with a bunch of different nested CASE statements etc. But presumably the database is quite capable of optimising this away, so other than making the SQL hard to read it probably doesn't matter.

ghickman: Rewrote both the MSSQL methods to use `DATEFROMPARTS`, seems a lot cleaner in the code 👍

PR #298, opened 2021-12-09
refactor: Refactor temp table creation and cleanup

evansd:
This means we now automatically do the right thing when our temp tables live in a different database from our data tables (as they do in Spark) and, in general, this moves more of the logic to the base class rather than relying on individual query engines to implement things.

Some of the changes here might not immediately make sense on their own, but they turn out to simplify other refactors further down the road.

### DIFF
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -55,6 +55,14 @@ def get_primary_table(query):
     return get_joined_tables(query)[0]
 
 
+class MissingString(str):
+    def __init__(self, message):
+        self.message = message
+
+    def __str__(self):
+        raise NotImplementedError(self.message)
+
+
 class BaseSQLQueryEngine(BaseQueryEngine):
 
     sqlalchemy_dialect: type[Dialect]
@@ -64,6 +72,9 @@ class BaseSQLQueryEngine(BaseQueryEngine):
 
     # Per-instance cache for SQLAlchemy Engine
     _engine: Optional[sqlalchemy.engine.Engine] = None
+
+    # Force subclasses to define this
+    temp_table_prefix: str = MissingString("'temp_table_prefix' is undefined")
 
     def get_queries(self):
         """Build the list of SQL queries to execute"""
@@ -271,7 +282,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         within this session; it's this function's responsibility to ensure it
         doesn't clash with any concurrent extracts
         """
-        raise NotImplementedError()
+        return f"{self.temp_table_prefix}{table_name}"
 
     def get_temp_database(self):
         """Which schema/database should we write temporary tables to."""
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -76,6 +76,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     # Force subclasses to define this
     temp_table_prefix: str = MissingString("'temp_table_prefix' is undefined")
 
+    temp_table_count: int = 0
+
     def get_queries(self):
         """Build the list of SQL queries to execute"""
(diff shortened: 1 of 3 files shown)


databuilder/query_engines/base_sql.py
```diff
@@ -63,6 +71,9 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     # Per-instance cache for SQLAlchemy Engine
     _engine: Optional[sqlalchemy.engine.Engine] = None
 
+    # Force subclasses to define this
+    temp_table_prefix: str = MissingString("'temp_table_prefix' is undefined")
```
ghickman, line 75: Nothing to do for this PR, but I'm hoping we can have mypy do this for us at some point!

bloodearnest (commented):
This makes sense to me.

One question though - in the spark backend, group tables are actually created via CREATE TEMPORARY VIEW, not CREATE TABLE, via `.write_query_to_table`.

As such, they are not actually the `sqlachemy.Table` objects that you might suppose.

Could this mean that the default DROP TABLE wouldn't work in that case?

I think they are like MSSQL's temp tables, in that we don't need to clean them up as they are session scoped? Docs would seem to suggest so:

https://docs.databricks.com/spark/latest/spark-sql/language-manual/sql-ref-syntax-ddl-create-view.html#parameters

So, we'd want to skip explicitly deleting group tables, possibly.

AIUI, this change tracks both codelist and group tables in the same list, so it's tricky for the backend to know which are group tables and which are codelists?

evansd:
> Could this mean that the default DROP TABLE wouldn't work in that case?

Well we know the answer to this I think, because the tests pass. I guess `if_exists=True` is doing the work here?

But more generally, I guess it would be up to the subclass to introspect the kind of thing it's been given in `drop_temp_table()` and decide what to do with it. The MSSQL class checks whether the name starts with `#` and then ignores it if so. I guess the Spark class could check whether it's a view object rather than a regular table and act accordingly.

I don't think the clean up code should care about the _purpose_ of the temp table (was it a codelist or a reified query or something else?) it should just care what kind of database entity it is.

bloodearnest:
> > Could this mean that the default DROP TABLE wouldn't work in that case?
> 
> Well we know the answer to this I think, because the tests pass. I guess `if_exists=True` is doing the work here?

Yep. Works but is a bit messy.

> But more generally, I guess it would be up to the subclass to introspect the kind of thing it's been given in `drop_temp_table()` and decide what to do with it. The MSSQL class checks whether the name starts with `#` and then ignores it if so. I guess the Spark class could check whether it's a view object rather than a regular table and act accordingly.
> 
> I don't think the clean up code should care about the _purpose_ of the temp table (was it a codelist or a reified query or something else?) it should just care what kind of database entity it is.

Yes, but currently, it will always be a `sqlalchemy.Table`, not a `sqlachemy.View`. The `queries` dict value will have a CreateViewAs thing, I guess? But the dict key will be a table.

evansd:
Ooohh, sorry I see what you mean. The trouble is there is no `sqlalchemy.View` class (or anything like it). Maybe we need separate methods for "create me a temporary thing I can insert into" and "create me a temporary thing I can write a query to".

In any case, the behaviour in this PR is no worse that it was before so I wouldn't want to block on figuring that out. And at least this now cleans up the codelists tables correctly if there in a different schema.

bloodearnest:
> Ooohh, sorry I see what you mean. The trouble is there is no sqlalchemy.View class (or anything like it). Maybe we need separate methods for "create me a temporary thing I can insert into" and "create me a temporary thing I can write a query to".

Right, that's what I was wondering.

>In any case, the behaviour in this PR is no worse that it was before so I wouldn't want to block on figuring that out. And at least this now cleans up the codelists tables correctly if there in a different schema.

Agreed, +1

bloodearnest (approved):

PR #299, opened 2021-12-10
Type Patient Series

ghickman:
This breaks down PatientSeries into specialised Series classes, tells Column subclasses to use the correct Series, and tells the type system to enforce it all.

### DIFF
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -299,7 +299,7 @@ class PatientSeries:
         return PatientSeries(value=comparator_value)
 
     def __repr__(self) -> str:
-        return f"PatientSeries(value={self.value})"
+        return f"{self.__class__.__name__}(value={self.value})"
 
     def __hash__(self) -> int:
         return hash(repr(self.value))
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -280,12 +280,6 @@ class PatientSeries:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(value={self.value})"
 
-    def round_to_first_of_month(self):
-        return PatientSeries(RoundToFirstOfMonth(self.value))
-
-    def round_to_first_of_year(self):
-        return PatientSeries(RoundToFirstOfYear(self.value))
-
 
 class BoolSeries(PatientSeries):
     def __and__(self, other: BoolSeries) -> BoolSeries:
@@ -327,6 +321,12 @@ class DateSeries(PatientSeries):
     def __ne__(self, other: DateSeries | str) -> BoolSeries:  # type: ignore[override]
         other_value = other.value if isinstance(other, DateSeries) else other
         return BoolSeries(value=self.value != other_value)
+
+    def round_to_first_of_month(self) -> DateSeries:
+        return DateSeries(RoundToFirstOfMonth(self.value))
+
+    def round_to_first_of_year(self) -> DateSeries:
+        return DateSeries(RoundToFirstOfYear(self.value))
 
 
 class IdSeries(PatientSeries):
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -229,14 +229,14 @@ class PatientFrame:
         """
         self.row = row
 
-    def select_column(self, column: Column[S]) -> PatientSeries:
+    def select_column(self, column: Column[S]) -> S:
         """
         Return a PatientSeries containing given column.
 
         Args:
             column: The Column of interest of which you want to retrieve the value.
         """
-        return PatientSeries(self.row.get(column.name))
+        return column.series_type(self.row.get(column.name))
 
 
 class PatientSeries:
@@ -352,16 +352,21 @@ class Predicate:
 @dataclass
 class Column(Generic[S]):
     name: str
+    series_type: type[S]
 
     def is_not_null(self):
         return Predicate(self, "not_equals", None)
 
 
 class IdColumn(Column[IdSeries]):
-    ...
+    def __init__(self, name):
+        return super().__init__(name, IdSeries)
 
 
 class BoolColumn(Column[BoolSeries]):
+    def __init__(self, name):
+        return super().__init__(name, BoolSeries)
+
     def is_true(self) -> Predicate:
         return Predicate(self, "equals", True)
 
@@ -382,6 +387,9 @@ class BoolColumn(Column[BoolSeries]):
 
 
 class DateColumn(Column[DateSeries]):
+    def __init__(self, name):
+        return super().__init__(name, DateSeries)
+
     def __eq__(self, other: str) -> Predicate:  # type: ignore[override]  # deliberately inconsistent with object
         return Predicate(self, "equals", other)
 
@@ -402,6 +410,9 @@ class DateColumn(Column[DateSeries]):
 
 
 class CodeColumn(Column[CodeSeries]):
+    def __init__(self, name):
+        return super().__init__(name, CodeSeries)
+
     def __eq__(self, other: str) -> Predicate:  # type: ignore[override]  # deliberately inconsistent with object
         return Predicate(self, "equals", other)
 
@@ -413,6 +424,9 @@ class CodeColumn(Column[CodeSeries]):
 
 
 class IntColumn(Column[IntSeries]):
+    def __init__(self, name):
+        return super().__init__(name, IntSeries)
+
     def __eq__(self, other: int) -> Predicate:  # type: ignore[override]  # deliberately inconsistent with object
         return Predicate(self, "equals", other)
 
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -46,7 +46,7 @@ for end users.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Generic, TypeVar, Union, overload
+from typing import Generic, TypeVar, overload
 
 from .query_language import (
     BaseTable,
@@ -151,7 +151,7 @@ class EventFrame:
         """
         return SortedEventFrame(self.qm_table, *columns)
 
-    def count_for_patient(self) -> PatientSeries:
+    def count_for_patient(self) -> IntSeries:
         """
         Takes the information from the multiple row per patient EventFrame and counts
         the events per patient.
@@ -160,11 +160,11 @@ class EventFrame:
             None
 
         Returns:
-            PatientSeries: A count of events per patient
+            IntSeries: A count of events per patient
         """
-        return PatientSeries(self.qm_table.count())
+        return IntSeries(self.qm_table.count())
 
-    def exists_for_patient(self) -> PatientSeries:
+    def exists_for_patient(self) -> BoolSeries:
         """
         Takes the information from the multiple row per patient EventFrame and returns a Boolean
         indicating if the Patient has a matching event.
@@ -173,9 +173,9 @@ class EventFrame:
             None
 
         Returns:
-            PatientSeries: A PatientSeries indicating whether each patient has a matching event.
+            BoolSeries: A BoolSeries indicating whether each patient has a matching event.
         """
-        return PatientSeries(self.qm_table.exists())
+        return BoolSeries(self.qm_table.exists())
 
 
 class SortedEventFrame:
@@ -248,46 +248,25 @@ class PatientSeries:
     """
 
     def __init__(self, value: Value | Comparator):
-        """
-        Initialise the PatientSeries
-
-        Args:
-            value: A Value or a Comparator. A Value contains a value such as numeric value,
-                and a Comparator compares the value to a ValueExpression and returns a Boolean.
-        """
         self.value = value
 
     def _is_comparator(self) -> bool:
         return isinstance(self.value, Comparator)
 
-    def __gt__(self, other: ValueExpression) -> PatientSeries:
-        return PatientSeries(value=self.value > other)
-
-    def __ge__(self, other: ValueExpression) -> PatientSeries:
-        return PatientSeries(value=self.value >= other)
-
-    def __lt__(self, other: ValueExpression) -> PatientSeries:
-        return PatientSeries(value=self.value < other)
-
-    def __le__(self, other: ValueExpression) -> PatientSeries:
-        return PatientSeries(value=self.value <= other)
-
-    def __eq__(self, other: ValueExpression) -> PatientSeries:  # type: ignore[override]
-        # All python objects have __eq__ and __ne__ defined, so overriding these method s
+    def __eq__(self, other: PatientSeries | Comparator | str | int) -> BoolSeries:  # type: ignore[override]
+        # All python objects have __eq__ and __ne__ defined, so overriding these methods
         # involves overriding them on a superclass (`object`), which results in
         # a typing error as it violates the violates the Liskov substitution principle
         # https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
         # We are deliberately overloading the operators here, hence the ignore
-        return PatientSeries(value=self.value == other)
+        other_value = other.value if isinstance(other, PatientSeries) else other
+        return BoolSeries(value=self.value == other_value)
 
-    def __ne__(self, other: ValueExpression) -> PatientSeries:  # type: ignore[override]
-        return PatientSeries(value=self.value != other)
+    def __ne__(self, other: PatientSeries | Comparator | str | int) -> BoolSeries:  # type: ignore[override]
+        return BoolSeries(value=self.value != other)
 
-    def __and__(self, other: PatientSeries) -> PatientSeries:
-        return PatientSeries(value=self.value & other.value)
-
-    def __or__(self, other: PatientSeries) -> PatientSeries:
-        return PatientSeries(value=self.value | other.value)
+    def __hash__(self) -> int:
+        return hash(repr(self.value))
 
     def __invert__(self) -> PatientSeries:
         if self._is_comparator():
@@ -301,9 +280,6 @@ class PatientSeries:
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}(value={self.value})"
 
-    def __hash__(self) -> int:
-        return hash(repr(self.value))
-
     def round_to_first_of_month(self):
         return PatientSeries(RoundToFirstOfMonth(self.value))
 
@@ -312,15 +288,45 @@ class PatientSeries:
 
 
 class BoolSeries(PatientSeries):
-    pass
+    def __and__(self, other: BoolSeries) -> BoolSeries:
+        return BoolSeries(value=self.value & other.value)
+
+    def __or__(self, other: BoolSeries) -> BoolSeries:
+        return BoolSeries(value=self.value | other.value)
 
 
 class CodeSeries(PatientSeries):
-    pass
+    def __and__(self, other: CodeSeries) -> BoolSeries:
+        return BoolSeries(value=self.value & other.value)
+
+    def __invert__(self) -> BoolSeries:
+        return BoolSeries(
+            value=(
+                Comparator(lhs=self.value, operator="__ne__", rhs=None, negated=True)
+            )
+        )
 
 
 class DateSeries(PatientSeries):
-    pass
+    def __gt__(self, other: DateSeries | str) -> BoolSeries:
+        other_value = other.value if isinstance(other, DateSeries) else other
+        return BoolSeries(value=self.value > other_value)
+
+    def __ge__(self, other: DateSeries | str) -> BoolSeries:
+        other_value = other.value if isinstance(other, DateSeries) else other
+        return BoolSeries(value=self.value >= other_value)
+
+    def __lt__(self, other: DateSeries | str) -> BoolSeries:
+        other_value = other.value if isinstance(other, DateSeries) else other
+        return BoolSeries(value=self.value < other_value)
+
+    def __le__(self, other: DateSeries | str) -> BoolSeries:
+        other_value = other.value if isinstance(other, DateSeries) else other
+        return BoolSeries(value=self.value <= other_value)
+
+    def __ne__(self, other: DateSeries | str) -> BoolSeries:  # type: ignore[override]
+        other_value = other.value if isinstance(other, DateSeries) else other
+        return BoolSeries(value=self.value != other_value)
 
 
 class IdSeries(PatientSeries):
@@ -328,7 +334,25 @@ class IdSeries(PatientSeries):
 
 
 class IntSeries(PatientSeries):
-    pass
+    def __gt__(self, other: IntSeries | int) -> BoolSeries:
+        other_value = other.value if isinstance(other, IntSeries) else other
+        return BoolSeries(value=self.value > other_value)
+
+    def __ge__(self, other: IntSeries | int) -> BoolSeries:
+        other_value = other.value if isinstance(other, IntSeries) else other
+        return BoolSeries(value=self.value >= other_value)
+
+    def __lt__(self, other: IntSeries | int) -> BoolSeries:
+        other_value = other.value if isinstance(other, IntSeries) else other
+        return BoolSeries(value=self.value < other_value)
+
+    def __le__(self, other: IntSeries | int) -> BoolSeries:
+        other_value = other.value if isinstance(other, IntSeries) else other
+        return BoolSeries(value=self.value <= other_value)
+
+    def __ne__(self, other: IntSeries | int) -> BoolSeries:  # type: ignore[override]
+        other_value = other.value if isinstance(other, IntSeries) else other
+        return BoolSeries(value=self.value != other_value)
 
 
 S = TypeVar("S", bound=PatientSeries)
@@ -579,6 +603,3 @@ def raise_category_errors(errors):
         raise next_category_error
     finally:
         raise_category_errors(errors)
-
-
-ValueExpression = Union[PatientSeries, Comparator, str, int]
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -46,7 +46,7 @@ for end users.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TypeVar, Union, overload
+from typing import Generic, TypeVar, Union, overload
 
 from .query_language import (
     BaseTable,
@@ -138,7 +138,7 @@ class EventFrame:
 
         return EventFrame(predicate.apply_to(self.qm_table))
 
-    def sort_by(self, *columns: Column) -> SortedEventFrame:
+    def sort_by(self, *columns: Column[S]) -> SortedEventFrame:
         """
         Sorts an EventFrame by a specific column, such as a date column, and
         returns a SortedEventFrame with a given sort column.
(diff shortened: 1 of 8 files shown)


tests/test_dsl.py
```diff
@@ -320,3 +327,141 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
     for k in dsl_col_defs:
         assert repr(dsl_col_defs[k]) == repr(qm_col_defs[k])
+
+
+def test_boolseries_and():
+    series1 = BoolSeries(ValueFromRow(source=None, column="test1"))
+    series2 = BoolSeries(ValueFromRow(source=None, column="test2"))
+
+    output = series1 & series2
+    assert isinstance(output, BoolSeries)
```
benbc, line 337: I think we need to verify the structure of the underlying value rather than just the type here. Sorry.

benbc (approved):
Great. Just one comment about the testing.

databuilder/dsl.py
```diff
@@ -302,18 +345,23 @@ def apply_to(self, table: BaseTable) -> BaseTable:
 
 
 @dataclass
-class Column:
+class Column(Generic[S]):
```
rebkwok, line 348: I'm struggling a bit with understanding what this and the `TypeVar` does.  Does this make Column a PatientSeries in some way?

ghickman: I'm going to try and explain this one (@benbc please correct me if I'm wrong!)…

This line makes `Column` a generic Type with the placeholder/typevar `S` to denote the concrete type we pass around.  The TypeVar's `bound` kwarg says that any concrete type used must be a PatientSeries or a subclass of it.

Each Column then declares which Series class it is related to, eg `class IdColumn(Column[IdSeries]):` and sets the appropriate `series_type` instance variable with the relevant Series class in it's `__init__()`.

We've then set up `Column.series_type: type[S]` so that `PatientFrame.select_column()` can return the appropriate Series classed based on the column asked for.  The type system knows how to tie this all together because we're using that same `S` type var in the `select_column()` definition as we are here.

rebkwok: Thanks!

benbc: Yes, I think that's right.

The most important thing is that each column now carries with it information about what kind of series will be created if that column is selected. That information is present both in the runtime object model (the `series_type` field) and in the type system (the type var `S`).

tests/test_dsl.py
ghickman: Updated with asserts on the structure of the underlying comparators, was that what you were thinking @benbc?  I'll add them to the other dunder methods if you think that's enough for `BoolSeries.__and__()`.

benbc: Yes, that's right.

ghickman: @benbc – updated all the series tests with those changes.

databuilder/dsl.py
```diff
@@ -258,34 +258,16 @@ def __init__(self, value: Value | Comparator):
...
-    def __ne__(self, other: ValueExpression) -> PatientSeries:  # type: ignore[override]
-        return PatientSeries(value=self.value != other)
-
-    def __and__(self, other: PatientSeries) -> PatientSeries:
-        return PatientSeries(value=self.value & other.value)
-
-    def __or__(self, other: PatientSeries) -> PatientSeries:
-        return PatientSeries(value=self.value | other.value)
+    def __hash__(self) -> int:
+        return hash(repr(self.value))
```
inglesp, line 270: I know these are alphabetised, but I wonder whether it'd be easier to understand the class if the DSL methods (`__eq__` and `__invert__`) were grouped separately from the other dunder methods (eg `__hash__` and `__repr__`).

tests/test_dsl.py
```diff
@@ -320,3 +327,321 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+    assert output.value.operator is None
+
+    # ValueFromRow & ValueFromRow generates a Comparator for each side of the
+    # __and__ before the BoolSeries __and__ takes over, make sure our inner
+    # Comparators are constructed as expected.
+    assert isinstance(output.value.lhs, Comparator)
+    assert output.value.lhs.lhs.column == "test1"
+
+    assert isinstance(output.value.rhs, Comparator)
+    assert output.value.rhs.lhs.column == "test2"
```
inglesp, line 351: I think I'd prefer tests that look like:

```
def test_boolseries_and():
    series1 = BoolSeries(ValueFromRow(source=None, column="test1"))
    series2 = BoolSeries(ValueFromRow(source=None, column="test2"))

    exp_value = Comparator(
        connector='and_', 
        negated=False, 
        operator=None,
        lhs=Comparator(lhs=series1),
        rhs=Comparator(lhs=series2),
    )

    assert series1 & series2 ==BoolSeries(value=exp_value)
```

A few reasons:

1. Documentation: the behaviour is, by necessity, split up into lots of different functions, and it's helpful to have some tests that show exactly what gets created when eg you combine two series.  I think I'd find it easier to understand tests where the output is spelt out in full, rather than where each different part of the output is checked separately.
2. Maintainability: if we change the way that the output values are constructed, then it'll be fiddly to update each of these small assertions.
3. Coverage + correctness: from reading this, I don't know whether there are any attributes of `output` that are not tested.  There's no coverage-like tool that I know of for checking this, but I think it's pretty important for correctness.

tests/test_dsl.py
```diff
@@ -320,3 +327,321 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+    assert output.value.operator == "__eq__"
+
+    assert isinstance(output.value.rhs, PatientSeries)
+    assert output.value.rhs.value == ""
+
+
+def test_patientseries_is_comparator():
+    assert not PatientSeries(value=Value())._is_comparator()
+
+    assert PatientSeries(value=Comparator())._is_comparator()
```
inglesp, line 645: Why do we need to test this private method?

tests/test_dsl.py
```diff
@@ -320,3 +327,321 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+    series = ~PatientSeries(value=Value())
+    assert isinstance(series.value, Comparator)
+
+
+def test_patientseries_repr():
+    events = tables.clinical_events
+    series = events.sort_by(events.date).first_for_patient().select_column(events.code)
+
+    expected = "CodeSeries(value=ValueFromRow(source=Row(source=Table(name='clinical_events'), sort_columns=('date',), descending=False), column='code'))"
+    assert repr(series) == expected
```
inglesp, line 661: What does this test?

tests/test_dsl.py
benbc: This captures my main feedback on these new tests perfectly.

tests/test_dsl.py
ghickman: We build the repr from the given series of a Column, so here `events.code` is a `CodeColumn`, and should return a `CodeSeries`.  Since we then use a Series' repr when hashing I thought it was useful to test it worked as expected.  Would it be helpful to add this explanation to the test too?

databuilder/dsl.py
ghickman: I tried a variety of ways to order these and eventually just gave up and fell back to alphetising them.  For instance: `__hash__` and `__eq__` are intertwined because of the [way `__hash__` works](https://docs.python.org/3.10/reference/datamodel.html#object.__hash__).

tests/test_dsl.py
ghickman: It wasn't being tested so I thought I'd bump up code coverage!  However, I felt it was useful to have it tested too as it's a core part of the lower level DSL machinery, eg how the base `__invert__` works.

databuilder/dsl.py
inglesp: Fine, not important.

tests/test_dsl.py
inglesp: If it's not being tested then what's it for?  Why not delete it?

tests/test_dsl.py
inglesp: From the perspective of hashing, as long as `PatientSeries.__repr__()` is a pure function (and it'd be batshit if it wasn't, so we don't need to test for that) it doesn't really matter what it returns.

So I think we're just testing this for code coverage, which is fine.

tests/test_dsl.py
inglesp: One more thought on testing: should the DSL tests know anything about the Query Model?

ghickman: @inglesp – updated relevant tests, new version is much nicer, thanks!

---

> should the DSL tests know anything about the Query Model?

Ben and I discussed this when I was starting the tests and came to the outcome that the DSL produces query model objects so the tests should reflect that.

benbc: I think they should. In my mind the Query Model is the core of the system. The DSL exists to create the Query Model. Ultimately I think that all the DSL tests should be assertions of the query model graphs they produce.

tests/test_dsl.py
ghickman: I just traced this down… in the `DateSeries.round_to_first_*` methods we pass in the relevant ValueFromFunction subclasses to `DateSeries`, which are subclasses of Value.  The `_is_comparator` method is then used in `dsl.categorise` to decide if we should remove nulls from the value.

So it's definitely in use, but DateSeries haven't been used in the categorise tests.

tests/test_dsl.py
inglesp: I wasn't clear, sorry.  I think we all agree that the tests should be of the form `assert process(dsl_obj) == qm_obj`.

However, `dsl_obj` is created with knowledge of the Query Model -- for instance `BoolSeries(ValueFromRow(source=None, column="test1"))`.

I'm not sure this is right, because we might end up with tests that create objects that are never actually createable through the DSL.

tests/test_dsl.py
inglesp: Sounds like we're missing a test for `categorise`?

tests/test_dsl.py
benbc: I see, thank you.

So one option is that all the tests just use the facilities available to end users. But then we get much bigger QM graphs than we need for a given test and have to assert against the whole thing (making it over-specified and a bit fragile to change).

Or could do as above and come up with some generic traversal/pattern-matching tooling to pick out just the bit of the graph we actually need to assert against.

Or a smaller change would be to have something like this:

```py
def test_boolseries_and():
    series1 = BoolSeries(a_value())
    series2 = BoolSeries(a_value())

    assert series1 & series2 == BoolSeries(value=Comparator(
        connector='and_', 
        negated=False, 
        operator=None,
        lhs=Comparator(lhs=series1),
        rhs=Comparator(lhs=series2),
    ))
```

Where we hide the value-construction stuff to emphasise that it's not important and make it easier to change if necessary.

tests/test_dsl.py
ghickman: Good point.  Although I think it's important to also keep a test like this around so any errors in `_is_comparator()` are clear!

tests/test_dsl.py
inglesp: Hiding the value construction is a good solution -- it gives readable tests while ensuring that we only use valid constructions.

tests/test_dsl.py
inglesp: I disagree -- I can't see any circumstances under which we learn anything useful from the test failing.

It'll only fail if we change the implementation, and then whoever changes the implementation will have to work out whether it was a useful test or not.

benbc: Yeah, I agree with Peter here.

tests/test_dsl.py
```diff
@@ -6,13 +6,30 @@
...
 
 
+def a_value():
+    """
+    Build a Value instance
+
+    We use these in various tests, but the actual mechanics of creating this is
+    unimportant to tests themselves.
+    """
+    return Value()
```
inglesp, line 30: This is missing my point -- `Value()` is never a valid construction (and so nor is eg `BoolSeries(a_value())`) so we're testing a situation that's never going to be hit in real life.

Perhaps a call tomorrow morning?

ghickman: Sure, is there anything I can look at in the mean time to find what a valid construction should be?

inglesp: Our goals are to ensure that:

* the tests for the DSL only use valid DSL objects (so that we're only testing with realistic values)
* we limit the amount of noise in the tests (so that they're readable)

I originally suggested something like:

```
def test_and():
    series1 = BoolSeries(ValueFromRow(source=None, column="test1"))
    series2 = BoolSeries(ValueFromRow(source=None, column="test2"))

    exp_value = Comparator(
        connector='and_', 
        lhs=Comparator(lhs=series1),
        rhs=Comparator(lhs=series2),
    )

    assert series1 & series2 == BoolSeries(value=exp_value)
```

But this doesn't meet either goal: `ValueFromRow` is not a DSL object, and there's a bit too much detail about the internal structure of `ValueFromRow()`.

Ben's proposal was to have a function that builds some kind of Query Model `Value`, and to use that when building a `BoolSeries` in a test:

```
def a_value():
    ValueFromRow(source=None, column="test1")

def test_and():
    series1 = BoolSeries(a_value())
    series2 = BoolSeries(a_value())

    exp_value = Comparator(
        connector='and_', 
        lhs=Comparator(lhs=series1),
        rhs=Comparator(lhs=series2),
    )

    assert series1 & series2 == BoolSeries(value=exp_value)
```

This makes the test more readable, but `a_value` returns an invalid Query Model object, and so the test is not using a valid DSL object.

The only way to be sure to have a valid DSL object is through using the to DSL.  So we could have one or more functions that build valid DSL objects, by using the DSL:

```
def build_bool_series():
    return clinical_events.exists_for_patient()  # This is about the simplest way to get a valid BoolSeries with the DSL

def test_and():
    series1 = build_bool_series()
    series2 = build_bool_series()

    exp_value = Comparator(
        connector='and_', 
        lhs=Comparator(lhs=series1),
        rhs=Comparator(lhs=series2),
    )

    assert series1 & series2 == BoolSeries(value=exp_value)
```

(`build_bool_series` might be better as a fixture.)

rebkwok: There is a problem with the assertions here, which I discovered accidentally because I'm using this branch temporarily for the date arithmetic ticket.  A `BoolSeries` (or any other Series) has had its `__eq__` method overloaded, so the assertion is just asserting a `Comparator`.  
```
assert series1 & series2 == BoolSeries(value=exp_value)
```
is equivalent to 
```
assert Comparator(...)
```
which is always true.

You could change that line to 
```
assert series1 & series2 == "foo"
```
and the test still passes

ghickman:
@inglesp – updated all the expected outputs for the tests as expected.  The non-BoolSeries ones even ended up being simpler than expected.  Could you do one more review of them?

tests/conftest.py
```diff
@@ -83,3 +84,44 @@ def engine(request, database, spark_database):
...
+    )
+
+
+@pytest.fixture
+def patient_series():
+    return (
+        lambda: clinical_events.sort_by(clinical_events.code)
+        .first_for_patient()
+        .select_column(clinical_events.code)
+    )
```
inglesp, line 127: Is there any reason these all return lambdas?

tests/test_dsl.py
```diff
@@ -320,3 +326,273 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+
+
+def test_boolseries_and(bool_series):
+    series1 = bool_series()
+    series2 = bool_series()
+
+    expected = Comparator(
+        connector="and_",
+        negated=False,
+        operator=None,
```
inglesp, line 338: You can drop the `negated` and `operator` arguments, since they're the default.

tests/test_dsl.py
```diff
@@ -320,3 +326,273 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+        lhs=Comparator(lhs=series1.value, operator="__ne__"),
+        rhs=Comparator(lhs=series2.value, operator="__ne__"),
+    )
+
+    output = series1 & series2
+    assert repr(output) == repr(BoolSeries(value=expected))
```
inglesp, line 376: What does `and`ing a `CodesSeries` mean?

tests/test_dsl.py
```diff
@@ -320,3 +326,273 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+    series = code_series()
+
+    expected = Comparator(
+        connector=None,
+        negated=True,
+        operator="__ne__",
+        lhs=series.value,
+    )
+
+    output = ~series
```
inglesp, line 389: Ditto -- what does negating a `CodeSeries` mean?

databuilder/dsl.py
```diff
@@ -297,21 +279,82 @@ def __invert__(self) -> PatientSeries:
...
+    def __gt__(self, other: ValueExpression) -> BoolSeries:
+        return BoolSeries(value=self.value > other)
+
+    def __ge__(self, other: ValueExpression) -> BoolSeries:
+        return BoolSeries(value=self.value >= other)
 
-    def round_to_first_of_year(self):
-        return PatientSeries(RoundToFirstOfYear(self.value))
+    def __lt__(self, other: ValueExpression) -> BoolSeries:
+        return BoolSeries(value=self.value < other)
```
inglesp, line 313: Should this (and similar) be `BoolSeries(value=self.value < other.value)`?

tests/test_dsl.py
inglesp: As we're now using `repr` for testing equality, I think we can drop this test.

inglesp (changes requested):
Getting there!  But I think where you've got `self.value op other` you want `self.value op other.value`, unless I've misunderstood.

tests/conftest.py
ghickman: The fixtures all need to return a callable so we can get a new Series when we call them.

tests/test_dsl.py
ghickman: Done and done, got all the `connector=None`s too.

tests/conftest.py
benbc: Why not make it a function. I don't think we're gaining anything from this being a fixture.

databuilder/dsl.py
ghickman: No!  `ValueExpression` is defined as `Union[PatientSeries, Comparator, str, int]`, so it's valid to pass in say `"2020-01-01"` (`test_categorise` does this).  Discovered because I made exactly this change and found tests broke… This one caught me out though, so maybe the type of `other` should be more specific for the Series types?

benbc: I don't entirely trust `ValueExpression`. I think that for each method like this in the DSL we should carefully consider what types should be valid and specify them explicitly.

inglesp: Ah, I understand the `"2020-01-01"` case now.

But I think this might be wrong for other cases.

For instance, in this test...

```

def test_intseries_lt(int_series):
    series1 = int_series()
    series2 = int_series()

    expected = Comparator(
        operator="__lt__",
        lhs=series1.value,
        rhs=series2,
    )

    output = series1 < series2
    assert repr(output) == repr(BoolSeries(value=expected))
```

...I'm surprised that `lhs` and `rhs` in the comparator are different kinds of things.  And specifically, `rhs` is a DSL object, and the QM shouldn't know about the DSL.

I think this shows the limitations of this kind of testing.  So I think before this branch lands, we should have at least one integration test showing that a dataset definition that uses these kinds of boolean operations actually work to generate the expected data.

Am aware that this is a massive extra chunk of work!  I'd suggest parking this PR for now, and starting to work on integrating the new DSL into the integration tests.  I'd also suggest discussing approaches in Slack, as @rebkwok or @evansd might have ideas about how to go about doing this.

rebkwok: I actually might have some of those sorts of tests in the date arithmetic branch.  `test_query_engine.py` is a good place to find simple tests that check generated data - they all use the old DSL, so rewriting (some of) them with the new DSL and ensuring they generate the same data would be useful.

Comparators should be expecting either a Value or a Comparator as the lhs, and a Value, Comparator or simple int/str etc as rhs.  I don't think they should ever get a DSL series (as either lhs or rhs).

ghickman: I've tracked this down to the Series comparison methods naively using `other` instead of `other.value`.  This breaks for primitive types so more to fix here.

> Comparators should be expecting either a Value or a Comparator as the lhs

@rebkwok – this is really useful to have spelled out, it hadn't really occurred to be before, thank you!

> So I think before this branch lands, we should have at least one integration test showing that a dataset definition that uses these kinds of boolean operations actually work to generate the expected data.

@inglesp – done in #314, and found some bugs, good call 👍

ghickman:
This comment is left mostly for next-week-me, but please feel free to weigh in if you see a way forward on this.

`self.value op other.value` turns out to be harder than expected.  Most of the comparison methods take a PatientSeries subclass or Python primitive, which of course doesn't have a `.value` attribute.

I've tried checking the type of `other` for the relevant Series class and calling `.value` on that but mypy then complains that I'm trying to do an expression with the wrong types in the return statement (should be a `Comparator | Value` but I'm giving it eg `IntSeries | int`), I _think_ because I'm reusing the `other` value.

Given the comparison methods of `Value` (which is what the expression in the return statements of the Series comparison methods are calling into) all work with Comparators or Values (which aren't typed but that looks correct based on the code) I suspect what I need to do here is wrap the primitives in comparators and then always pass `other.value`.

rebkwok:
> I've tried checking the type of other for the relevant Series class and calling .value on that but mypy then complains that I'm trying to do an expression with the wrong types in the return statement (should be a Comparator | Value but I'm giving it eg IntSeries | int), I think because I'm reusing the other value.

I'm sure I've come across this before, when I was working on the `categorise` stuff.

I did some experimenting, and 
```
    def __eq__(self, other: PatientSeries | Comparator | str | int) -> BoolSeries:  # type: ignore[override]
        if isinstance(other, PatientSeries):
            other = other.value
        return BoolSeries(value=(self.value == other))
```
gives a mypy error ("Incompatible types in assignment (expression has type "Union[Value, Comparator]", variable has type "Union[PatientSeries, Comparator, str, int]")

But 
```
    def __eq__(self, other: PatientSeries | Comparator | str | int) -> BoolSeries:  # type: ignore[override]
        other_value = other.value if isinstance(other, PatientSeries) else other
        return BoolSeries(value=(self.value == other_value))
```
is OK, as far as mypy is concerned.  The variable has to be renamed though.  `other = other.value if isinstance(other, PatientSeries) else other` still raises the error.

ghickman:
@rebkwok – that did it, thank you!  I had tried two separate if blocks which mypy didn't like, I guess because because I wasn't using if/else it couldn't infer that they checked both types.

tests/test_dsl.py
```diff
@@ -343,3 +343,220 @@ def assert_cohorts_equivalent(dsl_cohort, qm_cohort):
...
+    output = series1 == series2
+    assert repr(output) == repr(BoolSeries(value=expected))
+
+
+def test_patientseries_invert():
+    series = ~PatientSeries(value=Comparator())
+    assert isinstance(series.value, Comparator)
+
+    series = ~PatientSeries(value=Value())
+    assert isinstance(series.value, Comparator)
```
inglesp, line 584: Any reason this test doesn't use the same structure as those above?

tests/test_query_engine_dsl.py
```diff
@@ -51,3 +53,123 @@ def test_comparator_order(engine, cohort_with_population):
...
+        patient(4, height=35),
+    ]
+    engine.setup(input_data)
+
+    patients = MockPatients()
+    height = patients.select_column(patients.height)
+
+    lt_25 = (height >= 20) & (height <= 24)
+    gte_25 = (height >= 25) & (height <= 29)
+    in_2020 = lt_25 | gte_25
```
inglesp, line 158: Minor, but these names don't quite make sense.

inglesp (approved):
Great, I think we're basically there.  Thanks for sticking it out!

I have a couple of minor questions/comments, but the only thing that really needs addressing is whether we need `CodeSeries.__and__` and `CodeSeries.__invert__`.  (And if we do, why not `CodeSeries.__or__`?

tests/test_query_engine_dsl.py
ghickman: Copypasta fail, thanks!

ghickman: @inglesp – did you mean `lt_25` and `gte_25` too?

databuilder/dsl.py
```diff
@@ -312,23 +288,71 @@ def round_to_first_of_year(self):
 
 
 class BoolSeries(PatientSeries):
-    pass
+    def __and__(self, other: BoolSeries) -> BoolSeries:
+        return BoolSeries(value=self.value & other.value)
+
+    def __or__(self, other: BoolSeries) -> BoolSeries:
+        return BoolSeries(value=self.value | other.value)
```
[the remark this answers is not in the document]
ghickman: I don't think so, there's a generic `__invert__` inherited from PatientSeries which covers that case, which I think is fine for here?

tests/test_dsl.py
ghickman: nope, just missed that test, good catch!

databuilder/dsl.py
```diff
@@ -312,23 +288,71 @@ def round_to_first_of_year(self):
...
-    pass
+    def __and__(self, other: CodeSeries) -> BoolSeries:
+        return BoolSeries(value=self.value & other.value)
+
+    def __invert__(self) -> BoolSeries:
+        return BoolSeries(
+            value=(
+                Comparator(lhs=self.value, operator="__ne__", rhs=None, negated=True)
+            )
+        )
```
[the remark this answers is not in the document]
ghickman: As in what does the generated comparator mean? I don't know enough of the QueryModel to answer that really, I likely took this from the comparator fork of `PatientSeries.__inver__()`.

tests/test_dsl.py
ghickman: This is used in the categorising code, [example in the tests here](https://github.com/opensafely-core/databuilder/blob/ec0d400c187975f639f39653d83c7b2db164e8f1/tests/test_dsl_categorise.py#L303) to get yes/no categories with a given code.

tests/test_dsl.py
ghickman: This is used in the categorising code, [example in the tests here](https://github.com/opensafely-core/databuilder/blob/ec0d400c187975f639f39653d83c7b2db164e8f1/tests/test_dsl_categorise.py#L331).

databuilder/dsl.py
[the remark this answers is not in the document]
inglesp: I think they're meaningless and potentially confusing, so we should remove them from `CodeSeries`.

databuilder/dsl.py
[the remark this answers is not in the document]
inglesp: Hm -- I think `series.__invert__` only really makes sense if `isinstance(series, BoolSeries)`, so we should only define it here.  (What does inverting a date mean?)

databuilder/dsl.py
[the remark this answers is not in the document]
ghickman: `__and__` and `__invert__` are necessary for categorising to work currently (AIUI).  Probably one for you and @benbc or @rebkwok to discuss and make a call on!

PR #300, opened 2021-12-10
Fix categorise typing

benbc:


databuilder/dsl.py
```diff
@@ -436,9 +447,9 @@ def categorise(
     return PatientSeries(value=ValueFromCategory(value_mapping, default))
 
 
-def _validate_category_mapping(
-    mapping: dict[Expression, PatientSeries], default: Expression | None
-) -> None:
+# Typed as Any because the alternative is another overloaded function here which is unnecessary ceremony for this
+# internal function.
+def _validate_category_mapping(mapping: dict[Any, PatientSeries], default: Any) -> None:
```
ghickman, line 452: Would `object` work in place of `Any` here? Or is that still too specific?

ghickman (approved):

databuilder/dsl.py
benbc: Oh, excellent question. And the answer turns out to be "yes". Thank you.

PR #302, opened 2021-12-13
Remove built-in flake8 ignores for non-built-in variable names

ghickman:
Small tweak to remove some noqas we don't need.


CarolineMorton (approved):
Makes sense to me.

PR #307, opened 2021-12-15
chore: Add some Python caches to `.gitignore`

StevenMaude:


madwort:
do we have a template gitignore anywhere that should be updated? and/or other repos that could use this addition?

StevenMaude:
> do we have a template gitignore anywhere that should be updated? and/or other repos that could use this addition?

This is a good question, and I had a similar thought of how we keep these in sync. I'm not sure if there's a nice, clean way except syncing by hand.

It's this template originally: https://github.com/opensafely-core/repo-template

I can put a PR in there too.

madwort:
Cool, a PR in the template repo sounds good enough for me 👍

StevenMaude:
Actually, this isn't necessary as both caches should actually end up with a `.gitignore` as @ghickman pointed out in Slack.

Copy-paste of my response there:

> 1. The `.mypy_cache` directory I had for `databuilder` didn't have a `.gitignore`. I deleted the directory and re-ran `mypy`, then it appeared. It might have been because I never actually ran `mypy` to completion locally, and just exited early. Not sure.
> 2. Given 1, I just added `.pytest_cache` reflexively after looking at GitHub's Python `.gitignore`, but you're correct in that `.pytest_cache` creates a `.gitignore` too.

PR #308, opened 2021-12-15
ci: Publish a databuilder image

StevenMaude:
For now, this retains the existing cohortextractor-v2 image while publishing a databuilder image.

We also publish both images with a `vX`, `vX.Y` and `vX.Y.Z` for a Git tag of `vX.Y.Z`

The databuilder image won't yet be usable anywhere because job-runner and opensafely-cli (which vendors job-runner) need updating to permit the new image name. See https://github.com/opensafely-core/job-runner/pull/324

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -9,6 +9,7 @@ jobs:
     if: ${{ startsWith(github.ref, 'refs/tags/v') }}
     runs-on: ubuntu-latest
     env:
+      image: ghcr.io/opensafely-core/databuilder
       ce_image: ghcr.io/opensafely-core/cohortextractor-v2
     steps:
       - uses: actions/checkout@v2
@@ -18,14 +19,16 @@ jobs:
         run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
 
       - name: Build image
-        run: docker build . --file Dockerfile --tag ${{ env.ce_image }}:latest
+        run: docker build . --file Dockerfile --tag ${{ env.image }}:latest --tag ${{ env.ce_image }}:latest
 
       - name: Log into GitHub Container Registry
         run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
 
-      - name: Push image to GitHub Container Registry
+      - name: Push images to GitHub Container Registry
         env:
           VERSION: ${{ steps.tags.outputs.tag }}
         run: |
+          docker tag ${{ env.image }} ${{ env.image }}:$VERSION
           docker tag ${{ env.ce_image }} ${{ env.ce_image }}:$VERSION
+          docker push ${{ env.image }}:$VERSION
           docker push ${{ env.ce_image }}:$VERSION
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -14,21 +14,21 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Get current git tag
-        id: tags
-        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
-
+      # This relies on the tag having a version of the form vX.Y.Z
       - name: Build image
-        run: docker build . --file Dockerfile --tag ${{ env.image }}:latest --tag ${{ env.ce_image }}:latest
+        run: |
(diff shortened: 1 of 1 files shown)


StevenMaude:
thought:

* I'm not entirely sure why we tag with `latest` and then tag again with the version, but only use the version tag when pushing. I've followed the existing behaviour for now though. I'm aware that we don't really want to use `latest` at all, so I'm not sure why we tag this way.

StevenMaude:
question:

* The current [cohortextractor-v2 image](https://github.com/opensafely-core/base-docker/pkgs/container/cohortextractor-v2) does have a latest tag. Is this automatically added? I couldn't find a definitive answer.
* Do we want to have a `v0` type tag, as well as the `v0.6.0` we currently have? (I think that's fine in a separate PR though. Given that's it not easy to test this type of workflow change, I'd rather keep the changes minimal.)

bloodearnest:
> thought:
> 
> * I'm not entirely sure why we tag with `latest` and then tag again with the version, but only use the version tag when pushing. I've followed the existing behaviour for now though. I'm aware that we don't really want to use `latest` at all, so I'm not sure why we tag this way.

This is a bug, we should push latest too.

bloodearnest:
> * ave a latest tag. Is this automatically added? I couldn't find a definitive answer.
> * Do we want to have a `v0` type tag, as well as the `v0.6.0` we currently have? (I think that's fine in a se

AIUI it is only added if you specify no tag. It's the default, basically.

I think we should only push specific tags, and not latest, so that no one uses it.

We should probably push `vX.Y.Z`, `vX.Y` and `vX` tags, with the default template using `vX`?

StevenMaude:
> AIUI it is only added if you specify no tag. It's the default, basically.
>
> I think we should only push specific tags, and not latest, so that no one uses it.

My thought is that we might not even control the existence of `latest` in the repository (although `latest` in this case might be the most recently pushed tag, at least). That is apparently how the [Docker Hub](https://stackoverflow.com/q/61540768) works, but I'm not sure about GitHub's registry.

> We should probably push vX.Y.Z, vX.Y and vX tags, with the default template using vX?

:+1:

README.md
```diff
@@ -22,7 +22,7 @@ as described in the [documentation](https://docs.opensafely.org/getting-started/
 If running it directly, it should be run from within the research repository.
 To run the latest version via Docker and access its full help:
 
-    docker run --rm ghcr.io/opensafely-core/cohortextractor-v2 --help
+    docker run --rm ghcr.io/opensafely-core/databuilder --help
```
StevenMaude, line 25: thought: Maybe we shouldn't advocate the latest version and pin to a specific version here, e.g. v0 for now?

StevenMaude: Actually, going to leave this as is for now, because the new image name won't work yet.

.github/workflows/build-and-deploy.yml
```diff
@@ -9,23 +9,26 @@ jobs:
...
         run: docker login https://ghcr.io -u ${{ github.actor }} --password ${{ secrets.DOCKER_RW_TOKEN }}
 
-      - name: Push image to GitHub Container Registry
-        env:
-          VERSION: ${{ steps.tags.outputs.tag }}
+      - name: Push images to GitHub Container Registry
         run: |
-          docker tag ${{ env.image }} ${{ env.image }}:$VERSION
-          docker push ${{ env.image }}:$VERSION
+          docker push --all-tags ${{ env.image }}
```
bloodearnest, line 33: TIL --all-tags, nice!

.github/workflows/build-and-deploy.yml
```diff
@@ -9,23 +9,26 @@ jobs:
...
-        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
-
+      # This relies on the tag having a version of the form vX.Y.Z
       - name: Build image
-        run: docker build . --file Dockerfile --tag ${{ env.image }}:latest
+        run: |
+          PATCH="${GITHUB_REF#refs/*/}"
+          MAJOR="${PATCH%%.*}"
+          MINOR="${PATCH%.*}"
+          docker build . --file Dockerfile \
```
bloodearnest, line 23: Maybe add 

```
echo "MAJOR=$MAJOR"
echo "MINOR=$MINOR"
echo "PATCH=$PATCH"
```
so we get clear debugging output in GHA if something goes wrong here?

bloodearnest (approved):

PR #309, opened 2021-12-16
Increase Coverage… a bit

ghickman:
This tests some low hanging fruit to bump coverage up a little bit futher.

I had to modify the `__eq__` and `__ne__` methods of Comparator to return values, but I think that's correct?


StevenMaude (approved):

StevenMaude:
Looks fine to me.

But, while I likewise assume that returning from `__eq__` and `__ne__` is the right thing to do, I'm not really that familiar with the DSL, so maybe good to confirm with someone who is.

ghickman:
Confirmed by @rebkwok, in Slack, to be the correct fix.

PR #311, opened 2021-12-16
fix: Trigger builds of new releases on merge to main

madwort:
* A successful CI run on main will trigger the build-and-deploy workflow via the `workflow_run` target
* build-and-deploy can also be triggered by manually tagging & pushing via the `create` target
* We only run build-and-deploy if the most recent version tag points to the current commit
* Ideally we would also do some verification of which package versions have already been built & deployed, but despite lots of digging this does not appear to be possible right now

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -3,11 +3,11 @@ name: Build and publish image
 on:
   create:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


StevenMaude:
@madwort: #303 now merged, so this can be worked on.

ghickman (approved):

PR #313, opened 2021-12-17
feat: mirror sqlalchemy logging config for pyhive

ghickman:
This hides pyhives logs when testing to minimise output on failing tests.


evansd (approved):
Excellent, thanks for sorting this!

PR #320, opened 2021-12-17
Prevent equality comparison between column and codelist

### LINKED ISSUE #250: Some Spark tests fail with a SQL error
The following tests (currently marked as xfail) all fail with the same error:
```
tests/test_codelists.py::test_codelist_equals_query[spark]
tests/test_query_engine.py::test_simple_filters[spark-test single equals filter]
tests/test_query_engine.py::test_simple_filters[spark-test multiple equals filter]
tests/test_query_engine.py::test_simple_filters[spark-test not equals filter]
tests/test_query_engine.py::test_simple_filters[spark-test multiple chained filters]
tests/test_query_engine.py::test_in_filter_on_query_values[spark]
tests/test_query_engine.py::test_aggregation[spark-exists-code-expected0]
tests/test_query_engine.py::test_aggregation[spark-count-code-expected1]
tests/test_query_engine.py::test_aggregation[spark-sum-result-expected2]
tests/test_query_engine.py::test_categorise_on_truthiness[spark]
```

This is long convoluted traceback from Spark but the key element appears to be:
```
org.apache.spark.sql.AnalysisException: Correlated scalar subqueries must be aggregated
```

evansd:
This turned out to be the source of the SQL errors in #250. This PR adds a check to prevent doing this and updates all the test cases to use `is_in` instead of `equals`.

Closes #250

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -154,6 +154,14 @@ class BaseTable(QueryNode):
             # convert a between filter into its two components
             return self.filter(*args, on_or_after=value[0], on_or_before=value[1])
 
+        if operator in ("equals", "not_equals") and isinstance(
+            value, (Codelist, Column)
+        ):
+            raise TypeError(
+                f"You can only use '{operator}' to filter a column by a single value.\n"
+                f"To filter using a {value.__class__.__name__}, use 'is_in/not_in'."
+            )
+
         if operator == "is_in" and not isinstance(value, (Codelist, Column)):
             # convert non-codelist in values to tuple
             value = tuple(value)
--- a/tests/acceptance/test_full_long_covid_study.py
+++ b/tests/acceptance/test_full_long_covid_study.py
@@ -118,7 +118,10 @@ class Cohort:
     # Clinical variables
     # Latest recorded BMI
     _bmi_value = (
-        table("clinical_events").filter(code=bmi_code).latest().get("numeric_value")
+        table("clinical_events")
+        .filter("code", is_in=bmi_code)
+        .latest()
+        .get("numeric_value")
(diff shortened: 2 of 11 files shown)


databuilder/query_language.py
```diff
@@ -154,6 +154,14 @@ def filter(self, *args: str, **kwargs: Any) -> BaseTable:  # noqa: A003
...
             return self.filter(*args, on_or_after=value[0], on_or_before=value[1])
 
+        if operator in ("equals", "not_equals") and isinstance(
+            value, (Codelist, Column)
+        ):
+            raise TypeError(
+                f"You can only use '{operator}' to filter a column by a single value.\n"
+                f"To filter using a {value.__class__.__name__}, use 'is_in/not_in'."
+            )
+
```
rebkwok, line 164: Should this validation be done earlier, in the DSL, and raise an error that reports what the user should have written there (in the new-style DSL, that is, when they'd have done something like `events.filter(code==codelist)`?   And/or should the `CodeColumn` in `dsl.py` not define `__eq__` and `__ne__` methods, if we're not allowing those?

rebkwok (approved):
LGTM, just a question about how/where we raise the errors

databuilder/query_language.py
evansd: Oh I think it should definitely be done in the new DSL. This was just a quick fix to catch all the places we were relying on the old behaviour and make sure we don't add any more.

I think the long term goal should be that the DSL does all the user facing validation and can never produce an invalid QM. And then we just have asserts in the QM as a fallback in case we make a mistake in the DSL validation.

Do you think that make sense?

PR #326, opened 2021-12-22
Introduce date arithmetic (subtraction in years) on dateseries

rebkwok:
This introduces the minimal initial date arithmetic, to replace the `age_as_of` function with a more generic date subtraction, returned in years.


inglesp (approved):
Neat.

PR #331, opened 2022-01-04
Convert DateDeltaSeries to days and weeks

rebkwok:
[sc-418]
Calculate differences between dates in days and weeks.

This uses the database-specific `datediff` function (which differs in use between spark and mssql) to calculate the days difference between dates.  It uses the same days difference calculation to calculate the number of full weeks between the two dates.  `datediff` with an interval of "week" (which is only possible in the mssql version of datediff) calculates boundaries crossed rather than full weeks.  

Initially (see the first commit in this PR) I used Julian days to do the date difference calculations in days in a way that could be database agnostic.  This worked, and would mean that we could do the calculations the same way across dbs without using custom db methods, as we do for years and months.  However, it makes for very complex (and I'd assume inefficient) SQL and I think it's better to use the functions available in the dbs backing individual query engines.  If in some future engine we have a db with no `datediff` equivalent, we could reconsider the julian days version.

### DIFF
--- a/databuilder/query_engines/spark.py
+++ b/databuilder/query_engines/spark.py
@@ -59,6 +59,20 @@ class SparkQueryEngine(BaseSQLQueryEngine):
     def get_temp_database(self):
         return self.backend.temporary_database
 
+    def _convert_date_diff_to_days(self, start, end):
+        """
+        Calculate difference in days
+        The sparkSQL datediff function only calculate diff in days, and take the
+        parameters in the order (end, start)
+        """
+        return sqlalchemy.func.datediff(end, start)
+
+    def _convert_date_diff_to_weeks(self, start, end):
+        """
+        Calculate difference in weeks
+        """
+        return sqlalchemy.func.floor(self._convert_date_diff_to_days(start, end) / 7)
+
     def round_to_first_of_month(self, date):
         date = type_coerce(date, sqlalchemy_types.Date())
 
--- a/databuilder/query_engines/spark.py
+++ b/databuilder/query_engines/spark.py
@@ -67,12 +67,6 @@ class SparkQueryEngine(BaseSQLQueryEngine):
         """
         return sqlalchemy.func.datediff(end, start)
 
-    def _convert_date_diff_to_weeks(self, start, end):
-        """
-        Calculate difference in weeks
-        """
-        return sqlalchemy.func.floor(self._convert_date_diff_to_days(start, end) / 7)
-
     def round_to_first_of_month(self, date):
         date = type_coerce(date, sqlalchemy_types.Date())
 
--- a/tests/test_query_engine_dsl.py
+++ b/tests/test_query_engine_dsl.py
@@ -309,7 +309,7 @@ def test_date_arithmetic_conversions(engine, cohort_with_population):
     "current_date,age_data",
     [
         (
-            "2021-09-02",  # end month > 2
+            "2021-09-02",
             {
                 1: dict(dob="2021-09-01", age=1),  # 1 day
                 2: dict(dob="2021-01-15", age=230),
@@ -320,7 +320,7 @@ def test_date_arithmetic_conversions(engine, cohort_with_population):
             },
         ),
         (
-            "2020-01-10",  # end month < 2
+            "2020-01-10",
             {
                 1: dict(dob="2020-01-01", age=9),
                 2: dict(dob="2019-12-20", age=21),  # across year boundary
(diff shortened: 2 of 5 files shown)


rebkwok:
In fact, spark has a datediff function that could be used too, instead of the julian days calculation (it would still needs its own version because it only works for days, and doesn't take an interval parameter), so maybe we don't need this at all.

databuilder/query_engines/spark.py
```diff
@@ -59,6 +59,20 @@ def temp_table_needs_dropping(self, create_table_query):
     def get_temp_database(self):
         return self.backend.temporary_database
 
+    def _convert_date_diff_to_days(self, start, end):
+        """
+        Calculate difference in days
+        The sparkSQL datediff function only calculate diff in days, and take the
+        parameters in the order (end, start)
```
iaindillingham, line 66: 🤦🏻‍♂️

databuilder/query_engines/spark.py
```diff
@@ -59,6 +59,20 @@ def temp_table_needs_dropping(self, create_table_query):
...
+        """
+        return sqlalchemy.func.datediff(end, start)
+
+    def _convert_date_diff_to_weeks(self, start, end):
```
iaindillingham, line 70: This is the same as `BaseSQLQueryEngine._convert_date_diff_to_weeks`, is it not? Do we duplicate it here to ensure `SparkQueryEngine._convert_date_diff_to_days` is called in preference to `BaseSQLQueryEngine._convert_date_diff_to_days`? (i.e. to ensure `self` is an instance of `SparkQueryEngine` rather than `BaseSQLQueryEngine`?)

tests/test_query_engine_dsl.py
```diff
@@ -303,3 +303,91 @@ def test_date_arithmetic_conversions(engine, cohort_with_population):
         {"patient_id": 8, "age_in_years": 9, "age_in_months": 119},
         {"patient_id": 9, "age_in_years": 9, "age_in_months": 118},
     ]
+
+
+@pytest.mark.parametrize(
+    "current_date,age_data",
+    [
```
iaindillingham, line 310: I'd be interested to know more about why you chose these dates and, as a consequence, what cases we're testing. Am I correct in thinking that you chose the end dates to be after, before, and equal to February because February may have a leap day?

tests/test_query_engine_dsl.py
```diff
@@ -303,3 +303,91 @@ def test_date_arithmetic_conversions(engine, cohort_with_population):
...
+    result = engine.extract(data_definition)
+    assert result == [
+        {"patient_id": patient_id, "age_in_days": patient_data["age"]}
+        for patient_id, patient_data in age_data.items()
+    ]
+
+
+def test_date_arithmetic_convert_to_weeks(engine, cohort_with_population):
+    input_data = [
+        # all dobs are subtracted from 2021-03-01, rounded down
```
iaindillingham, line 372: Does the "rounded down" apply to the value of `current_date`, which is `"2021-03-02"`?

iaindillingham (commented):
I'm going to comment rather than approve, because this is the first time I've 👀 Data Builder. That said, there's nothing here that would stop me approving. I've asked a couple of questions to help me understand your approach.

databuilder/query_engines/spark.py
benbc: What Iain said. It would be a shame to have to leave this duplication in.

benbc (approved):

databuilder/query_engines/spark.py
rebkwok: No, there's no reason for the duplication. In fact, I've already taken it out in the the next thing I'm working on locally, but I can add that to this PR

tests/test_query_engine_dsl.py
rebkwok: I did choose dates around Feb because of Feb being fewer days, but the real reason is that the initial iteration of the julian days version of this treated Jan and Feb differently to the rest of the months (it used a formula that involved adding 12 to the month and subtracting 1 from the year if the month is <=2.  Because apparently the year should really start on 1st March, which makes sense really, because then October is month 8, November is month 9...)

THe comments about the end month are probably misleading now that there's nothing special about month 2, I'll take them out

tests/test_query_engine_dsl.py
rebkwok: no, it;s a typo in the comment :)
Should be "subtracted from 2021-03-02, rounded down"

PR #334, opened 2022-01-05
Make it easier to run fast subsets of the tests

benbc:
I thought that `just test-unit` was getting a bit slow, so I had a poke around and discovered that there were several tests that use the database which weren't marked as integration tests. Further poking revealed that I accidentally removed the assertion that all database-using tests were marked as either integration or smoke tests when I was removing the old record/playback mechanism.

Reinstating that assertion proved impossible (for reasons explained in comments), so I've instead introduced auto-marking for any tests that use containers.

I've also codified something that some people have been doing and added an easy way to avoid running the Spark tests because they are so much slower than the other integration tests.

Timings on my laptop (small sample size):

```
Target                      Before    After
test-unit                   ~ 10 sec  < 1 sec
test-smoke                  ~ 10 sec  n/a
test-integration-no-spark   n/a       ~ 45 sec
test-integration            ~ 6 mins  ~ 6 mins  (!?!)
```

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -18,11 +18,11 @@ install dependencies.
 ## Test setup
 
 The test suite includes:
-- smoke tests: tests that run the databuilder end to end in docker
-- integration tests: tests that require a database
-- unit tests that don't require a database
+- integration tests: tests that use Docker (mostly to run a database, but also in some cases to test the data builder's
+  own Docker image)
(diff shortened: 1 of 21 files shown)


inglesp (approved):
Congratulations, you're now the team pytest internals expert!

This all looks good, and it's especially good that we don't have to decide whether a test is an integration test or not.

PR #336, opened 2022-01-06
Add and subtract datedeltas to/from DateSeries

rebkwok:
[sc-418]
This allows adding and subtracting a number of days to/from a date/DateSeries.  It currently ONLY allows days, to avoid/postpone the issue of what it means to add e.g. 1 month to 31st Jan (28th Feb? 1st March?) and the ambiguity of e,g, 2022-01-01 + 3 (as the 3 will always be treated as days).

The number of days can be a simple integer, a DateDeltaSeries that represents a difference between two dates in days (e.g. age as `index_date - date_of_birth`) or an IntSeries created by converting a DateDeltaSeries to days (e.g.  `(index_date - date_of_birth).convert_to_days()`

It also adds methods to add/substract DateDeltaSeries to/from each other, to produce new DateDeltaSeries.

### DIFF
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -55,6 +55,7 @@ from .query_model import (
     Comparator,
     DateAddition,
     DateDifference,
+    DateSubtraction,
     RoundToFirstOfMonth,
     RoundToFirstOfYear,
     Row,
@@ -366,12 +367,23 @@ class DateSeries(PatientSeries):
                 f"Can't {operation} <{delta_value.__class__.__name__}> to a DateSeries."
             )
 
-    def __sub__(self, other: str | DateSeries) -> DateDeltaSeries:
+    def __sub__(
+        self, other: str | DateSeries | DateDeltaSeries | IntSeries | int
+    ) -> DateDeltaSeries | DateSeries:
+        if isinstance(other, (DateDeltaSeries, IntSeries, int)):
+            other_value = self._get_other_datedelta_value(other, "subtract")
+            return DateSeries(DateSubtraction(self.value, other_value))
         return DateDeltaSeries(
             DateDifference(self._get_other_date_value(other), self.value)
         )
 
     def __rsub__(self, other: str | DateSeries) -> DateDeltaSeries:
+        if not isinstance(other, (str, DateSeries)):
+            # consistent with python datetime/timedelta, we cannot subtract a DateSeries from
+            # DateDeltaSeries or integer
+            raise TypeError(
+                f"Can't subtract DateSeries from {other.__class__.__name__}"
+            )
         return DateDeltaSeries(
             DateDifference(self.value, self._get_other_date_value(other))
         )
@@ -400,6 +412,12 @@ class DateDeltaSeries(PatientSeries):
 
     def convert_to_weeks(self):
         return self._convert("weeks")
+
+    def __rsub__(self, other: str) -> DateDeltaSeries:
+        datestring = _validate_datestring(other)
+        # This allows subtraction of a DateDeltaSeries from a date string
+        # e.g. 2020-10-01
+        return DateDeltaSeries(DateSubtraction(datestring, self.convert_to_days()))
 
 
 class IdSeries(PatientSeries):
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -349,28 +349,17 @@ class DateSeries(PatientSeries):
         """Ensure we have either a simple int, or an IntSeries representing a date difference in days"""
         if isinstance(delta_value, DateDeltaSeries):
             return delta_value.convert_to_days()
-        elif isinstance(delta_value, IntSeries):
-            if not isinstance(delta_value.value, DateDifference):
-                raise ValueError(
-                    f"Can't {operation} IntSeries with value <{delta_value.value.__class__.__name__}> to a DateSeries."
-                )
-            if delta_value.value.units != "days":
-                raise ValueError(
-                    f"Can't {operation} {delta_value.value.units} to a DateSeries.  Use this DataDeltaSeries prior to conversion, or convert to days instead."
-                )
-            else:
-                return delta_value
         elif isinstance(delta_value, int):
             return delta_value
         else:
             raise ValueError(
-                f"Can't {operation} <{delta_value.__class__.__name__}> to a DateSeries."
+                f"Can only {operation} integer or DateDeltaSeries (got <{delta_value.__class__.__name__}>)"
             )
 
     def __sub__(
         self, other: str | DateSeries | DateDeltaSeries | IntSeries | int
     ) -> DateDeltaSeries | DateSeries:
-        if isinstance(other, (DateDeltaSeries, IntSeries, int)):
+        if isinstance(other, (DateDeltaSeries, int)):
             other_value = self._get_other_datedelta_value(other, "subtract")
             return DateSeries(DateSubtraction(self.value, other_value))
         return DateDeltaSeries(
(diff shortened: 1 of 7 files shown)


databuilder/dsl.py
```diff
@@ -367,6 +413,12 @@ def convert_to_days(self):
     def convert_to_weeks(self):
         return self._convert("weeks")
 
+    def __rsub__(self, other: str) -> DateDeltaSeries:
+        datestring = _validate_datestring(other)
+        # This allows subtraction of a DateDeltaSeries from a date string
+        # e.g. 2020-10-01
+        return DateDeltaSeries(DateSubtraction(datestring, self.convert_to_days()))
```
rebkwok, line 420: Note this currently only allows you to do things like `"2020-10-01" - <DateDeltaSeries>`.  I'm leaving adding/subtracting int/DateDeltaSeries to/from a DateDeltaSeries for another PR

benbc:
@rebkwok The support for adding and subtracting- `IntSeries` seems to add a lot of complexity. I'm not _convinced_ that it's necessary. Have you seen a use case for it?

rebkwok:
> @rebkwok The support for adding and subtracting- `IntSeries` seems to add a lot of complexity. I'm not _convinced_ that it's necessary. Have you seen a use case for it?

Not specifically; my feeling was that people might calculate a difference in days, and then it'd be logical to want to add/subtract that.  But the simpler thing would be to limit it to an unconverted DateDeltaSeries, which probably makes sense

benbc:
I think you're right that we might discover a need for this -- but it's sufficiently complex that we should only add it if we really need it.

databuilder/dsl.py
rebkwok: The rest of the subtraction is in this PR after all now

databuilder/dsl.py
```diff
@@ -315,43 +319,80 @@ def _validate_datestring(datestring):
...
+        except ValueError:
+            if isinstance(other, str):
+                raise
+            raise TypeError(
+                f"Can't subtract DateSeries from {other.__class__.__name__}"
+            )
+
+    def __add__(self, other: DateDeltaSeries | int) -> DateSeries:
+        other_value = self._get_other_datedelta_value(other, "add")
+        return DateSeries(DateAddition(self.value, other_value))
```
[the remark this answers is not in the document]
inglesp: What error would the user see if they try `10 + date_of_birth`?

I'm weakly in favour of supporting it (and ignoring mypy), in order to be mathematically consistent (since we usually expect `A + B` to be `B + A`) but I suspect it's too much work to be worth doing now.

databuilder/dsl.py
```diff
@@ -367,6 +408,49 @@ def convert_to_days(self):
...
+        return self + other
+
+    def __sub__(self, other: DateDeltaSeries | int) -> DateDeltaSeries:
+        return DateDeltaSeries(
+            DateDeltaSubtraction(self._delta_in_days(self), self._delta_in_days(other))
+        )
+
+    def __rsub__(
+        self, other: str | DateDeltaSeries | int
+    ) -> DateSeries | DateDeltaSeries:
```
inglesp, line 444: I wondered whether `other` could be a `DateSeries`.   But I think I've realised it couldn't be, because that would be handled by `DateSeries.__sub__`.  Could you add a note explaining that?

tests/test_query_engine_dsl.py
```diff
@@ -391,3 +393,274 @@ def test_date_arithmetic_convert_to_weeks(engine, cohort_with_population):
...
+    result = engine.extract(data_definition)
+    assert result == [
+        {
+            "patient_id": 1,
+            "age_in_days_1990": 31,
+            "age_in_days_1991": 396,
+            "age_diff_plus_10": 375,
+            "dob_plus_age_diff_plus_10": date(1991, 8, 20),
+        }
+    ]
```
inglesp, line 666: These tests look sensible, but it's hard as a reader to evaluate whether they're correct, because they're full of magic numbers (and magic dates!).  For instance, where do `396` or `date(1991, 8, 20)` come from?

Something to consider for the future might be to do something like:

```
def test_date_arithmetic_add_and_subtract(engine, cohort_with_population):
    dob = date(1990, 8, 10)
    reference_date = date(1990, 9, 10)

    input_data = [
        patient(1, dob=str(dob),
    ]
    engine.setup(input_data)

    patients = tables.patients
    data_definition = cohort_with_population
    dob_series = patients.select_column(patients.date_of_birth)  # -> DateSeries

    age_in_1990 = str(reference_date) - dob_series  # -> DateDeltaSeries
    data_definition.age_in_days_1990 = age_in_1990.convert_to_days()

    result = engine.extract(data_definition)
    assert result == [
        {
            "patient_id": 1,
            "age_in_days_1990": (reference_date - dob).days,
        }
    ]
```

(I've probably made several mistakes here, but hopefully you get the idea!)

inglesp (approved):
This is a nice piece of work.

For future documentation, it would be helpful to write up notes about what the valid operations on dates and date deltas are while this is still in your mental L1 cache.

databuilder/dsl.py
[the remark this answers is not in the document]
rebkwok: It would be an `unsupported operand between...` message

It works fine if I just add an `__radd__` (it has to allow other to be int or DateSeries, and if it's a DateSeries, return other + self so that it uses the DateSeries `__add__` method).  I had that before and decided to take it out because I couldn't find a way to appease mypy and I wasn't sure about just ignoring it.  I could put that back in with the ignore, or we could just leave it till later

databuilder/dsl.py
```diff
@@ -315,43 +319,80 @@ def _validate_datestring(datestring):
...
 
     def round_to_first_of_month(self) -> DateSeries:
         return DateSeries(RoundToFirstOfMonth(self.value))
 
     def round_to_first_of_year(self) -> DateSeries:
         return DateSeries(RoundToFirstOfYear(self.value))
 
-    def __sub__(self, other: str | DateSeries) -> DateDeltaSeries:
-        return DateDeltaSeries(DateDifference(self._get_other_value(other), self.value))
+    def _get_other_datedelta_value(self, delta_value, operation):
```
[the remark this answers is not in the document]
rebkwok: yes, will update

databuilder/dsl.py
rebkwok: Aaah, this comment has finally made me figure out what mypy was complaining about with the DateSeries __radd__.  I think I can fix that now :)

tests/test_query_engine_dsl.py
rebkwok: good point, it would make sense to use python date subtractions in the tests.  I think I avoided it because of the custom differences we have (e.g. weeks as whole weeks rather than boundaries crossed).  But in the adding/subtracting deltas, we're only using days, so that's not an issue.  I'll take a look at updating the tests.

tests/test_query_engine_dsl.py
```diff
@@ -12,6 +12,10 @@
 pytestmark = pytest.mark.integration
 
 
+def _as_datestring(date_obj):
+    return date_obj.strftime("%Y-%m-%d")
```
inglesp, line 16: Can you not use `str()`?

```
>>> str(date(1990, 8, 10))
'1990-08-10'
```

rebkwok: huh. did not occur to me that would work

rebkwok: @inglesp I've updated that now.  Are you OK with the additional changes I've made since you approved?

inglesp: :shipit:

PR #339, opened 2022-01-11
Unify tables

inglesp:
The main changes are:

* frames are generated from contracts, which means that...
* QM tables know about contracts instead of table names, so that we can support different table names in different backends

To avoid merge conflict catastrophe, there's some further work not included in this PR:

* we need to support different column names in different backends
* we might need to spell out the frame definitions (instead of using `Contract.build_dsl_frame`) to support autocomplete etc
* flesh out new series types (there's likely some refactoring/simplifying here)
* there's probably still some code shuffling and renaming

I've tried to make this reviewable commit-by-commit, but the important files to review are query_model.py and backends/base.py.


tests/acceptance/test_full_long_covid_study.py
```diff
@@ -34,47 +43,47 @@ class Cohort:
 
     # Population
     # Patients registered on 2020-11-01
-    _registrations = table("practice_registrations").date_in_range(index_date)
+    _registrations = Table(WIP_PracticeRegistrations).date_in_range(index_date)
```
benbc, line 46: It would be a kindness to provide something analogous to `tests.lib.frames` for all these `Table()` calls. But maybe you have avoided doing that because all the invocations will be removed or rewritten during the great test upheaval following Dave's work on the query model?

databuilder/query_model.py
```diff
@@ -221,7 +221,7 @@ def aggregate(self, function, column):
 
 @dataclass(frozen=True)
 class Table(BaseTable):
-    name: str
+    contract: None  # Because of circular imports, I can't put the correct type here
```
benbc, line 220: You can mention types by name, as strings (if you have `from __future__ import annotations`), although I haven't checked to see if that helps with circular dependencies.

inglesp:
Superseded by #343.

inglesp:
@benbc thanks for the review here.  Following discussion with Dave, I'd meant to close this sooner.  I think most of your comments are addressed, or don't need addressing, in #343.

PR #343, opened 2022-01-13
Link frames to contracts

inglesp:
This PR contains a bunch of relatively small fiddles and fixes.  The new functionality is in [Add frame definitions](https://github.com/opensafely-core/databuilder/commit/8017cd7936ea5ed302388b87b521e53b996967e5) and [Validate frame definitions](https://github.com/opensafely-core/databuilder/commit/91d527d4b5c359a8947115954eda7d9bf8927761).


databuilder/dsl.py
```diff
@@ -127,6 +128,11 @@ def __init__(self, qm_table: BaseTable):
         """
         self.qm_table = qm_table
 
+    @classmethod
+    def from_contract(cls, contract):
+        qm_table = Table(contract._name)
+        return cls(qm_table)
```
benbc, line 135: The "root" (table) frames don't have any different behaviour from other frames, but they clearly play a special role and that's not clear in the current object model. If we had `EventTable` and `PatientTable` as subclasses then it would clarify that role and they could have constructors that take contracts rather than a special class method.

databuilder/dsl.py
```diff
@@ -235,6 +241,13 @@ def __init__(self, row: Row):
         """
         self.row = row
 
+    @classmethod
+    def from_contract(cls, contract):
+        qm_table = Table(contract._name)
+        # TODO: revisit this!  As things stand, this will generate SQL with an
+        # unnecessary PARTITION OVER, which may carry a performance penalty.
```
benbc, line 250: Since this may be important, could you create a story to cover this rather than leaving it as a TODO?

databuilder/frames.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class WIP_SimplePatientDemographics(PatientFrame):
+    patient_id = IdColumn("patient_id")
+    date_of_birth = DateColumn("date_of_birth")
+
+
+patients = WIP_SimplePatientDemographics.from_contract(
+    contracts.WIP_SimplePatientDemographics
+)
```
benbc, line 127: I think I may have made this point on the restructuring spreadsheet. I don't think "frames" is a helpful name for this module for researchers to understand it. They will consider the objects to be "tables". So we could do with a separate `tables` module that _just_ exports the objects here, not the classes.

databuilder/contracts/base.py
```diff
@@ -84,3 +84,21 @@ def validate_data(cls, backend, table_name):
                 ]
             for constraint in constraints:
                 constraint.validate(backend_table, column_name)
+
+    @classmethod
+    def validate_frame(cls, frame_cls):
+        """Validate that a frame is defined with the correct attributes.
+
+        This uses asserts rather than raising other exceptions, since any invalid frames
+        must be identified and fixed by the development team.
```
benbc, line 93: :+1:

benbc (commented):
General approach is good.

databuilder/frames.py
inglesp: Addressed in f896f1e08c483cbf93da26fc8a97f7a5c37f33b1.

databuilder/dsl.py
inglesp: Addressed in 530a934a9af294cc5b0de618d5368a72a94791ff.

databuilder/contracts/contracts.py
```diff
@@ -6,6 +6,8 @@
 class PatientDemographics(TableContract):
     """Provides demographic information about patients."""
 
+    _name = "patient_demographics"
```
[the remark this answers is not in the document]
inglesp: So that it doesn't clash with a possible field name.

databuilder/contracts/contracts.py
```diff
@@ -138,6 +146,8 @@ class WIP_Hospitalizations(TableContract):
 
 
 class WIP_HospitalizationsWithoutSystem(TableContract):
+    _name = "hospitalizations_"
```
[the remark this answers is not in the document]
inglesp: `hospitalizations_without_system` is a bit of a mouthful, and we can't call it `hospitalizations`, because there's another contract with the same name.

I think I'd thought that since `WIP_HospitalizationsWithoutSystem` isn't fixed, it wasn't worth coming up with a better name now.

Any suggestions?

databuilder/dsl.py
inglesp: See https://app.shortcut.com/ebm-datalab/story/569/decide-how-to-instantiate-patienttable.

databuilder/contracts/contracts.py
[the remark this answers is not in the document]
benbc: `hospitalizations_systemless`? I'd rather have a slightly unsatisfactory name that we hope to improve on (because this is WIP) than introduce what looks like a naming convention.

databuilder/contracts/contracts.py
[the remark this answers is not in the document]
benbc: Aaaah. Yeah, okay. I thought that it was for filtering it out -- I hadn't thought it might actually stomp on something that was needed for a field. I don't love it, but fair enough.

benbc (approved):
Up to you whether you want to pander to my concerns about `hospitalizations_`.

databuilder/contracts/contracts.py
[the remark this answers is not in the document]
inglesp: I've gone with `hospitalizations_without_system` for consistency.

PR #344, opened 2022-01-14
fix: incorrect sense of version tag check

madwort:


.github/workflows/build-and-deploy.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
       # 'Get Previous tag' should only get version tags, but let's verify that
       - name: Fail if not a version tag
```
bloodearnest, line 41: Perhaps:

"Fail if the the tag doesn't start with a v"

madwort: how about "Fail if previous tag is not a version tag (i.e. it does not start with a v)"

bloodearnest (approved):

PR #345, opened 2022-01-14
Test-per-variable approach for study translations

benbc:
Before this PR we had an "acceptance" test with the long covid study translated into the new DSL. However it was written as a single test which had to be `xfail`ed because the DSL is not yet complete. That had two problems: the test was "rotting" as the DSL changed (since it was allowed to fail we didn't notice when we broke it further) and because it was a single test it aborted at the first problem it found so we couldn't easily use it to see how many missing features there were in the DSL before we could support that study.

This PR introduces a new way of writing such acceptance tests, with subtests scoped to a single variable or small related group. So we can now explicitly mark small parts of it as failing, see which bits they are and notice if we break anything that currently work.

(This is the first commit. The others are various bits of tidy-up.)


tests/acceptance/test_long_covid_dsl.py
```diff
@@ -0,0 +1,375 @@
...
+def populated_database(database):
+    setup_data(database)
+    yield database
+
+
+@pytest.fixture
+def cohort(populated_database):
+    class AssertingCohort(Cohort):
+        def assert_results(self, **results):
+            dataset = extract(self, TPPBackend, populated_database)[0]
```
inglesp, line 49: I missed the `[0]` at first.  Suggest splitting over two lines:

```
dataset = extract(self, TPPBackend, populated_database)
row = dataset[0]
```

and then making assertions about `row`.

inglesp (approved):
One tiny suggested change, but only if you can be bothered!

tests/acceptance/test_long_covid_dsl.py
benbc: Good. Thank you.

PR #347, opened 2022-01-18
Track Strict Type Checking Changes

ghickman:
This adds a new justfile target, `check-strict-typing`, and a CI job which calls it.  It runs mypy in strict mode and pipes the output through [`mypy-json-report`](https://pypi.org/project/mypy-json-report/) to generate the `known-mypy-errors.json` file.

Rather than track a "coverage" number a la coverage.py this lets us check to see if the error types per file have changed.

It might be that the code base is too unstable to introduce this currently, but I wanted to see what it would take to get this added as a CI task.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,22 @@ jobs:
       - name: Check formatting, linting and import sorting
         run: ./scripts/do check
 
+  check-strict-typing:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


benbc:
@ghickman Did you decide against this in the end?

ghickman:
@benbc – nope, it's still awating review and fell off my list to chase!

benbc:
@ghickman I think that we need to have a discussion as a team before we put this in. Unless we're all signed up to it there is a risk that it will just become an annoying roadbump because we almost always react to these failures by committing the changes.

PR #348, opened 2022-01-18
ci: Fix tagging of Docker images

StevenMaude:
The versions all currently get set to `main`, possibly because this
workflow is now triggered by a complete `workflow_run` on the `main`
branch.

Instead use the tag, that we've already checked is valid.


StevenMaude:
[Example run](https://github.com/opensafely-core/databuilder/runs/4855607202) where I skipped checking for a tagged commit and didn't push the image. This shows the versions are not just `main`.

madwort (approved):

PR #349, opened 2022-01-18
test: Test SRO measures with DSL

iaindillingham:
This translates the SRO Measures study from cohort-extractor-v2 to the DSL. I've used the long covid study as a pattern to follow, so the first commit moves fixtures from this study to *tests/acceptance/conftest.py*.

I've used `subtest` and `TODO` comments to indicate missing/desired functionality. I've not added much data to `conftest.setup_data`; just enough to make an assertion against a non-`None` result. I've also added a subtest that instantiates each `Measure`. However, as [I don't think this functionality has been implemented][1], I don't think there's anything to make assertions against.

[1]: https://app.shortcut.com/ebm-datalab/story/456/accommodate-measures-with-the-new-dsl

### DIFF
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -1,0 +1,36 @@
+from contextlib import contextmanager
+
+import pytest
+
+from databuilder.backends import TPPBackend
+from databuilder.dsl import Cohort
+from tests.lib.util import extract
+
+
+@pytest.fixture
+def cohort(database):
+    class AssertingCohort(Cohort):
+        def assert_results(self, **results):
+            dataset = extract(self, TPPBackend, database)[0]
+            for variable, expected_value in results.items():
+                assert dataset[variable] == expected_value
+
+    yield AssertingCohort()
+
+
+@pytest.fixture
+def subtest(subtests):
+    @contextmanager
+    def helper(name, missing_feature=""):
+        with subtests.test(msg=name):
+            try:
+                yield
+            except Exception:
+                if missing_feature:
+                    pytest.xfail(f"DSL feature '{missing_feature}' not implemented")
+                raise
+            assert (
+                not missing_feature
+            ), f"Unexpected success, maybe feature '{missing_feature}' has been implemented"
+
+    yield helper
--- a/tests/acceptance/test_sro_measures_dsl.py
+++ b/tests/acceptance/test_sro_measures_dsl.py
@@ -1,0 +1,173 @@
+from databuilder import Measure
+from databuilder.tables import clinical_events, practice_registrations
+from tests.lib.tpp_schema import organisation, patient, registration, snomed_event
+
+from .codelists import (
+    alt_codelist,
+    asthma_codelist,
+    cholesterol_codelist,
+    copd_codelist,
+    hba1c_codelist,
+    medication_review_codelist,
+    qrisk_codelist,
+    rbc_codelist,
+    sodium_codelist,
+    systolic_bp_codelist,
+    tsh_codelist,
+)
+
+
+def test_sro_measures_study(cohort, subtest, database):
+    setup_data(database)
+
+    index_date = "2020-11-01"
+    end_of_index_month = "2020-11-30"
+
+    measures_codelists = {
+        "medication_review": medication_review_codelist,
+        "systolic_bp": systolic_bp_codelist,
+        "qrisk": qrisk_codelist,
+        "cholesterol": cholesterol_codelist,
+        "alt": alt_codelist,
+        "tsh": tsh_codelist,
+        "rbc": rbc_codelist,
+        "hba1c": hba1c_codelist,
+        "sodium": sodium_codelist,
+        "asthma": asthma_codelist,
+        "copd": copd_codelist,
+    }
+
+    with subtest("population"):
+        filtered_practice_registrations = (
+            practice_registrations.filter(
+                practice_registrations.date_start <= index_date
+            )
+            # TODO: Reinstate the following when Predicates support logic operations
+            # .filter(
+            #     (practice_registrations.date_end >= index_date)
+            #     | (practice_registrations.date_end != None)
+            # )
+            .filter(practice_registrations.date_end >= index_date)
+        )
+
+        is_registered = filtered_practice_registrations.exists_for_patient()
+
+        # TODO: The following would be nice
+        # not_dead = patient_demographics.date_of_death == None
+
+        # TODO: Does this work as we expect it to work?
+        # not_dead = (
+        #     patient_demographics.select_column(patient_demographics.date_of_death)
+        #     == None  # noqa: E711
+        # )
+
+        # TODO: This does not work, because population.value is an instance of
+        # Comparator
+        # cohort.set_population(is_registered & not_dead)
+
+        cohort.set_population(is_registered)
+
+    with subtest(
+        "practice_id", missing_feature="add practice ids to practice registrations"
+    ):
+        cohort.practice_id = (
+            filtered_practice_registrations.sort_by(practice_registrations.date_end)
+            .last_for_patient()
+            .select_column(practice_registrations.practice_id)
+        )
+        cohort.assert_results(practice_id=1)
+
+    with subtest("codelists"):
+        for measure_name, target_codelist in measures_codelists.items():
+            filtered_clinical_events = (
+                clinical_events.filter(
+                    clinical_events.code.is_in(target_codelist.codes)
+                )
+                .filter(clinical_events.date >= index_date)
+                .filter(clinical_events.date <= end_of_index_month)
+            )
+
+            cohort.add_variable(
+                measure_name, filtered_clinical_events.exists_for_patient()
+            )
+
+            # TODO: SortedEventFrame.exists_for_patient doesn't exist, so we can't
+            # filter and sort together.
+            sorted_clinical_events = filtered_clinical_events.sort_by(
+                clinical_events.date
+            )
+            measure_event_code = f"{measure_name}_event_code"
+            cohort.add_variable(
+                measure_event_code,
+                sorted_clinical_events.last_for_patient().select_column(
+                    clinical_events.code
+                ),
+            )
+
+        cohort.assert_results(
+            medication_review=True,
+            medication_review_event_code="394720003",
+            systolic_bp=None,
+            systolic_bp_event_code=None,
+            qrisk=None,
+            qrisk_event_code=None,
+            cholesterol=None,
+            cholesterol_event_code=None,
+            alt=None,
+            alt_event_code=None,
+            tsh=None,
+            tsh_event_code=None,
+            rbc=None,
+            rbc_event_code=None,
+            hba1c=None,
+            hba1c_event_code=None,
+            sodium=None,
+            sodium_event_code=None,
+            asthma=True,
+            asthma_event_code="394720003",
+            copd=None,
+            copd_event_code=None,
+        )
+
+    with subtest(
+        "measures", missing_feature="Allow measures to be registered with a cohort"
+    ):
+        # https://app.shortcut.com/ebm-datalab/story/456/accommodate-measures-with-the-new-dsl
+        measures = []
+        for measure_name, target_codelist in measures_codelists.items():
+            measures.extend(
+                [
+                    Measure(
+                        id=measure_name,
+                        cohort=cohort,
+                        numerator=measure_name,
+                        denominator="population",
+                        group_by=["practice", measure_event_code],
+                    ),
+                    Measure(
+                        id=f"{measure_name}_practice_only",
+                        cohort=cohort,
+                        numerator=measure_name,
+                        denominator="population",
+                        group_by=["practice"],
+                    ),
+                ]
+            )
+        # TODO: This function doesn't exist yet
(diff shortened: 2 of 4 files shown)


iaindillingham:
The test check is failing because coverage is less than 100%. We ignore the long covid study when computing coverage; previously, this meant we also ignored the fixtures. I think, then, that I should ignore the SRO Measures study and the fixtures when computing coverage. I'd like to check before editing *pyproject.toml*, though.

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+                not missing_feature
+            ), f"Unexpected success, maybe feature '{missing_feature}' has been implemented"
+
+    yield helper
+
+
+def setup_data(database):
+    database.setup(
+        organisation(organisation_id=1, region="South"),
+        patient(
```
benbc, line 59: It might make the tests clearer if we allow this to vary by study rather than having a single data fixture for all of these tests.

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+            #     (practice_registrations.date_end >= index_date)
+            #     | (practice_registrations.date_end != None)
+            # )
+            .filter(practice_registrations.date_end >= index_date)
+        )
+
+        is_registered = filtered_practice_registrations.exists_for_patient()
+
+        # TODO: The following would be nice
+        # not_dead = patient_demographics.date_of_death == None
```
benbc, line 53: This isn't supported, although we've discussed it.

benbc: (I mean `patient_demographics.date_of_death` as sugar for `atient_demographics.select_column(patient_demographics.date_of_death)`.

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # TODO: Does this work as we expect it to work?
+        # not_dead = (
+        #     patient_demographics.select_column(patient_demographics.date_of_death)
+        #     == None  # noqa: E711
+        # )
```
benbc, line 59: Yes, this is how we'd expect to do this.

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # TODO: This does not work, because population.value is an instance of
+        # Comparator
+        # cohort.set_population(is_registered & not_dead)
```
benbc, line 63: There's something weird going on here. It's a bug, but I'm not sure that it's worth fixing since this will be overhauled soon. @inglesp I think this is probably related to the thing I ran into yesterday.

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+    rbc_codelist,
+    sodium_codelist,
+    systolic_bp_codelist,
+    tsh_codelist,
+)
+
+
+def test_sro_measures_study(cohort, subtest):
+    index_date = "2020-11-01"
+    end_of_index_month = "2020-11-30"
```
benbc, line 21: Can we do this with date arithmetic?

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+                .filter(clinical_events.date >= index_date)
+                .filter(clinical_events.date <= end_of_index_month)
+            )
+
+            cohort.add_variable(
+                measure_name, filtered_clinical_events.exists_for_patient()
+            )
+
+            # TODO: SortedEventFrame.exists_for_patient doesn't exist, so we can't
+            # filter and sort together.
```
benbc, line 92: I don't understand this. Can we discuss? We don't think that this should be necessary.

tests/acceptance/test_sro_measures_dsl.py
```diff
@@ -0,0 +1,157 @@
...
+            qrisk_event_code=None,
+            cholesterol_event_code=None,
+            alt_event_code=None,
+            tsh_event_code=None,
+            rbc_event_code=None,
+            hba1c_event_code=None,
+            sodium_event_code=None,
+            asthma_event_code="394720003",
+            copd_event_code=None,
+        )
```
benbc, line 130: These assertions are actually quite expensive (about 1 sec). Could we do these two together?

benbc: I'd be inclined to pair up the boolean and the code for each codelist.

benbc (commented):
Great. A couple of things to discuss.

Could you add missing things to the lists in the description of the story?

tests/acceptance/conftest.py
iaindillingham: It would, especially as the number of tests increases. We could:

* Move `populated_database` (the fixture) and `setup_data` back to `test_long_covid_dsl`
* Merge `setup_data` into `populated_database`
* Write `test_sro_measures_dsl.populated_database` (i.e. we would have `test_long_covid_dsl.populated_database` and `test_sro_measures_dsl.populated_database`)

What was your rationale for separating `populated_database` and `setup_data`?

Alternatively, we could:

* Keep `populated_database` where it is, but pass it the [`request`][2] object
* Move `setup_data` back to `test_long_covid_dsl`
* Write `test_sro_measures_dsl.setup_data`
* Introspect `request`, for the module-specific `setup_data`

This keeps `populated_database` and `setup_data` separate, but might seem 🪄.

[1]: https://ebmdatalab.slack.com/archives/C02LEF2GC75/p1642700504246100
[2]: https://docs.pytest.org/en/stable/fixture.html#fixtures-can-introspect-the-requesting-test-context

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: I tried several times to understand what was happening, but kept arriving at the definition of `Comparator` and not really understanding what was being compared to what. How is `Comparator` being overhauled?

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: I thought I could do something like this:

```python
# wrangle the clinical events that interest us
events = clinical_events.filter(
    # codelist filter
).filter(
    # index date filter
).filter(
    # end of index month filter
).sort_by(
    # date sort
)

# use them to add variables to our cohort
cohort.add_variable("some_measure", events.exists_for_patient())
cohort.add_variable("some_measure_event_code", events.last_for_patient().select_column(events.code)
```

But `events` is an instance of `SortedEventFrame`, which doesn't have a `exists_for_patient` method 🙁

benbc:
> ignore the SRO Measures study and the fixtures when computing coverage

Yes, I think that's right.

tests/acceptance/conftest.py
benbc: > What was your rationale for separating populated_database and setup_data?

A minor point of taste. Separating the concerns of "define a fixture that populates the database with the data needed for this test" and "define the data needed for this test".

benbc: I don't think it's quite a simple as you suggest (at least in your first option) because the populated database is a dependency (as currently written) of the `cohort` fixture.

Since the `populated_database` fixture is so simple, maybe it's okay to have it duplicated between tests. So you could explicitly pull that into every test.

```py
## conftest.py

@pytest.fixture
def cohort(database):
    # ...
    
## test_long_covid_dsl.py

def setup_data(database):
    # ...

@pytest.fixture
def populated_database(database):
    # ...
    
def test_long_covid_study(cohort, subtest, populated_database):
    # ...
    
## test_sro_measures_dsl.py

def setup_data(database):
    # ...

@pytest.fixture
def populated_database(database):
    # ...
    
def test_sro_measures_study(cohort, subtest, populated_database):
    # ...
```

Maybe?

Maybe that's overkill and this is simpler:

```py
def test_sro_measures_study(cohort, subtest, database):
    setup_data(database)
    
    # ...
```

etc

Since there's only one test in each of these files, we don't really gain anything by pulling the database population out into a fixture, do we?

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+    patient,
+    patient_address,
+    positive_test,
+    registration,
+    snomed_event,
+)
+from tests.lib.util import extract
+
+
+@pytest.fixture(scope="module")
```
benbc, line 21: This module scope is a mistake on my part. There was an older version of this where I had split each variable definition into its own test function and I needed this to avoid re-populating the database every time because that would be too slow. But now that each of these study tests consists of just a single function this is no longer necessary.

tests/acceptance/test_sro_measures_dsl.py
benbc: `Comparator` is part of the old query model. That's going to be replaced.

tests/acceptance/test_sro_measures_dsl.py
benbc: Right. The reason for that is that the _only_ reason to sort an event frame is because you're then going to pick the first or last one. There is never any need to sort and then aggregate with exists or count. So you'd write this instead:

```py
events = clinical_events.filter(
    # codelist filter
).filter(
    # index date filter
).filter(
    # end of index month filter
)

cohort.add_variable("some_measure", events.exists_for_patient())
cohort.add_variable("some_measure_event_code", events.sort_by(events.date).last_for_patient().select_column(events.code)
```

benbc: (Your version isn't semantically incorrect, but there is a redundant sort. We're trying to define the DSL in such a way that they can only ever do operations that make sense.)

tests/acceptance/conftest.py
iaindillingham: I've used the `request` object to introspect the module, as this handles the dependency between `cohort` and `populated_database`. As a consequence, a test that uses the `cohort` fixture will result in a call to the `setup_data` function from within that test's module. Just because we can do it this way, though, doesn't mean we should: how do you feel about the additional 🪄?

Something I'd like to check: regardless of approach, we don't truncate the database between calls. Should we?

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: Would you do this using Data Builder or standard Python? The non-DSL test uses standard Python:

https://github.com/opensafely-core/databuilder/blob/884e7d38ec02e58da9c709b09aeea486b495f607/tests/acceptance/test_sro_measures_study.py#L45-L47

Do we have functionality in Data Builder for this?

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: Thanks for explaining the rationale for not implementing `SortedEventFrame.exists_for_patient`. I'm not sure that "no redundancy" maps to "making sense", though. In this case, it made sense to me to wrangle the clinical events before using them to add variables to the cohort; and as I was going to sort them anyway, the sort certainly didn't feel redundant.

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: I've paired the boolean and the code, as you suggest.

tests/acceptance/conftest.py
iaindillingham: Thanks for the comment. I've removed the scope, so now the fixture is function-scoped (the default).

tests/acceptance/conftest.py
iaindillingham: Ah, but we do truncate the database between calls!

https://github.com/opensafely-core/databuilder/blob/c605cd2a58d2669d4da6010951269c1f058b7e12/tests/lib/databases.py#L97-L98

benbc: I think that's a bit too much magic. Even someone super-familiar with pytest won't have any reason to expect that.

tests/acceptance/test_sro_measures_dsl.py
benbc: I suppose it would be Python, so it doesn't really gain us anything. As you were.

tests/acceptance/conftest.py
iaindillingham: If we took the simpler path, then would you pass `database` as an argument to `AssertingCohort.assert_results`?

iaindillingham: Oh, hold on, we don't have to, do we? We could just make the `cohort` fixture depend on the `database` fixture and remember that the `setup_data` function must be called before `AssertingCohort.assert_results` is called.

benbc: Yes, that's right.

tests/acceptance/test_sro_measures_dsl.py
iaindillingham: Aye aye, 🙂

tests/acceptance/conftest.py
iaindillingham: I've removed the 🪄 and taken the simpler path. Consequently, we no longer need the `populated_database` fixture, in either its original or modified form.

benbc (approved):

PR #350, opened 2022-01-18
Ensure Query Model only contains allowed types

evansd:
This PR adds type checking for the elements of the Query Model which aren't `QueryNode` instances, and fixes some cases where DSL classes previously leaked through.


inglesp (approved):
Hans Brinker salutes you.

PR #352, opened 2022-01-19
Set Required Coverage to 100

ghickman:
This raises the minimum coverage to 100% and fixes each of the remaining failures/warnings in the coverage report.  Each commit explains why a failure/warning was tackled in the way it was to make it to track them in history should we need to.

### DIFF
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -238,7 +238,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         # their elements at runtime. See: https://bugs.python.org/issue46191
         if isinstance(value, (tuple, list)) and any(
             not isinstance(v, SCALAR_TYPES) for v in value
-        ):
+        ):  # pragma: no cover
             assert False, f"Unhandled type {value!r}"
         return value
 
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -116,7 +116,7 @@ class ReifiedQuery(QueryNode):
     columns: tuple[str]
 
     def _get_referenced_nodes(self):
-        return (self.source,)
+        return (self.source,)  # pragma: no cover
 
 
 class MissingString(str):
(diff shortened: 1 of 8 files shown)


databuilder/query_engines/base_sql.py
```diff
@@ -273,6 +273,17 @@ def get_sql_element_no_cache(self, node: QueryNode) -> ClauseElement:
...
+        """
+        Certain places in our Query Model support values which can either be QueryNodes
+        themselves (which require resolving to SQL) or plain static values (booleans,
+        integers, dates, list of dates etc) which we can return unchanged to SQLAlchemy.
+        """
+        if isinstance(value, QueryNode):
+            return self.get_sql_element(value)
+        else:
+            return value
+
```
evansd, line 286: I think a rebase has gone wrong here, and this commit should actually now be a no-op.

databuilder/query_engines/base_sql.py
```diff
@@ -238,9 +238,9 @@ def get_static_value(self, value: Union[Scalar, tuple, list]) -> StaticValue:
         # their elements at runtime. See: https://bugs.python.org/issue46191
         if isinstance(value, (tuple, list)) and any(
             not isinstance(v, SCALAR_TYPES) for v in value
-        ):
+        ):  # pragma: no cover
             assert False, f"Unhandled type {value!r}"
-        return value
+        return value  # pragma: no cover
```
evansd, line 243: Just for my own education, how come this line needs `no cover`? Isn't the return always executed here

evansd (approved):
Thanks for doing this George, I think it's really going to pay off as development continues.

I think there's just one commit that needs removing (ae2ea897b5d90c9b0fde89a00abbf297311e1baa) which I'm pretty sure is just git rebase not following the fact that a function was moved to a different place (in this commit 80de1ce30c6bac51f2251aa8475aa06ae2e080c2).

databuilder/query_engines/base_sql.py
ghickman: I added it without thinking 🤦‍♂️  Fixed and rebased now.

PR #354, opened 2022-01-21
Script Checking for Added Coverage

ghickman:
This adds a workflow to check if our coverage changes if we stop omitting lines with `pragma: no cover` from coverage reporting.  We have a number of these lines in the codebase, 25 at time of writing, and we'd like to know if we've covered those lines with tests we're adding.

I've documented why we're doing this and how to fix a failing test run in `DEVELOPERS.md`, if that's not clear, lets fix that before merging.

### DIFF
--- a/.github/workflows/track-no-cover.yml
+++ b/.github/workflows/track-no-cover.yml
@@ -1,0 +1,37 @@
+---
+name: Check no cover Removal
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "15 9 1 * *"
+
+jobs:
+  check-no-cover:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - uses: extractions/setup-just@v1
+
+      - name: Did coverage change?
+        env:
+          CI: 1
+        run: |
+          set -euo pipefail
+
+          # stop ignoring no cover lines
+          sed -i.bak 's/"pragma: no cover",/# "pragma: no cover",/' pyproject.toml
+
+          just test | tee test-output.txt || true
+
+          # extract missing lines per file fom coverage report in the tests output
+          # python script: get the files table from coverage
+          # awk: break the table down into filename and missing lines number
+          python scripts/parse_coverage_output.py test-output.txt | awk '{ print $1 " " $3 }' - > no-cover.txt
+
+          # compare coverage report to the existing one
+          git diff --exit-code no-cover.txt
--- a/.flake8
+++ b/.flake8
@@ -8,3 +8,4 @@ max-line-length = 88
 per-file-ignores =
     scripts/dbx:INP001
     scripts/dbx-cleanup.py:INP001,F821
+    scripts/parse_coverage_output.py:INP001
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -162,3 +162,21 @@ To do this:
 If using Community Edition, you will need to follow the instructions the
 command outputs to complete the cleanup process, as we cannot fully automate it
 from the cli.
+
+
+### Checking `no cover` Removal
+We have a number of lines omitted from test coverage using `pragma: no cover`.
+We'd like to know when they get test coverage.
+
+To do so we have a scheduled workflow which removes that string from coverage's omitted lines config and runs the tests.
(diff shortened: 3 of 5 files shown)


.github/workflows/track-no-cover.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Check no cover Removal
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "15 9 1 * *"
```
benbc, line 7: Once a month? Why so infrequently?

.github/workflows/track-no-cover.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v2
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "3.9"
+      - uses: extractions/setup-just@v1
+
+      - name: "Stop ignoring pragma: no cover lines"
+        run: |
+          set -eu
+          sed -i.bak 's/"pragma: no cover",/# "pragma: no cover",/' pyproject.toml
```
benbc, line 23: Why is this in a separate step when we have multiple things going on in the other step?

.github/workflows/track-no-cover.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          set -eu
+          set -o pipefail
+
+          just test | tee test-output.txt || true
+
+          # extract missing lines per file fom coverage report in the tests output
+          # python script: get the files table from coverage
+          # awk: break the table down into filename and missing lines number
+          python scripts/parse_coverage_output.py test-output.txt | awk '{ print $1 " " $3 }' - > no-cover.txt
```
benbc, line 37: Seems a bit odd that we're using `awk` to parse the output of a script that we've written expressly for this purpose rather than getting it to output the format we actually want.

.github/workflows/track-no-cover.yml
```diff
@@ -0,0 +1,40 @@
+---
+name: Check no cover Removal
+
+on:
+  workflow_dispatch:
+  schedule:
```
benbc, line 6: In practice, how are we going to notice that this has failed?

benbc (commented):
I'm a bit skeptical about the value of this because of the difficulty in making it likely to be acted on without just running it on every build. But I'm happy for it to go in and see how it goes.

Of course what we _really_ want is a feature in `coveragepy` that does this. I see a closely related feature request (https://github.com/nedbat/coveragepy/issues/711), but the use case there is different. Maybe worth raising a issue for it?

.github/workflows/track-no-cover.yml
madwort, line 37: aside: this seems a little unstable against no-cover lines being moved around in a file, but looking at the contents of `no-cover.txt` a lot of them happen to be at the top of files so it doesn't seem worth worrying about.

madwort: oh - ignore that - `missing lines number` means the count of missing lines, not a list of line numbers that are missing!

ghickman: Yep! I was originally just pulling out the whole coverage table which was far too unstable for this, my hope is count of lines is good enough 🤞

ghickman: Right tool for the job, etc.  It's far easier/quicker to grab tokens 1&3 with awk than do all the breaking down and stripping of leading/trailing spaces in Python.

.github/workflows/track-no-cover.yml
ghickman: Vestige of a previous version that I meant to merge in, I'll pull it into the main step.

ghickman: Fixed and rebased.

.github/workflows/track-no-cover.yml
ghickman: We'll get a failed build, which generates notifications for anyone watching the repo.

ghickman:
> I'm a bit skeptical about the value of this because of the difficulty in making it likely to be acted on without just running it on every build

@benbc – This is my core worry around this route too [related to some of your inline commends too], but I didn't want to push this on every push/PR until we knew how stable it was.  I've tried to make it as stable as I can (eg taking counts per file rather than the full pytest-cov terminal output), but I'm happy to make this a periodic chore until we know how well it works.

benbc (approved):
Let's merge this as it is, but treat it with a healthy dose of skepticism and promise ourselves not to let it become a drag.

PR #360, opened 2022-01-26
fix: correct sense of short-circuit test

madwort:


.github/workflows/build-and-deploy.yml
```diff
@@ -33,9 +33,9 @@ jobs:
         # if we are not on a tag, should look like `v0.6.0-24-gbf6352d`
         run: echo ::set-output name=tag_describe::`git describe --tags`
 
-      - name: Set flag to short circuit workflow if not a tagged commit
+      - name: Verify if we are on a tagged commit, allows us to short-circuit this workflow
```
madwort, line 36:
```suggestion
      - name: Verify whether we are on a tagged commit, allows us to short-circuit this workflow
```

ghickman (approved):

PR #361, opened 2022-01-27
Yet another query model

evansd:
You probably want to look at this in the branch view here:
https://github.com/opensafely-core/databuilder/blob/yet-another-query-model/databuilder/query_model.py

Things to note:

 * I've used the terminology of tables, columns, one-row-per-patient and many-rows-per-patient. These are verbose but I think fairly clear. My feeling is that this stuff is hard enough to get your head round as it is and that pithy neologisms are a hindrance rather than a help here. I am willing to be talked out of this (the mismatch with the DSL is certainly a downside) but I would need to be talked out of it :) If we do need to change the terminology it's a find-and-replace job.

 * The type constraints and validation are (I think and hope) enough to ensure that every constructable query model is meaningful (leaving aside the question of what types are contained _within_ columns, on which more below). There are a few cases where we allow vacuous or pointless operations. It would be easy enough to add validation to reject these if we decided that was a job for the query model. The particular cases I'm thinking of are:
   - aggregating a patient-level column e.g. you could call `exists` on the result of `max` — it would be kind of pointless but not nonsense;
   - sorting by a patient-level column — this wouldn't result in any actual sorting because the value would be constant for each patient, but you could equally well get this behaviour by sorting on a value that you happen to know is constant for each patient, even if it isn't constrained to be such by the query model, so I'm not sure it needs prohibiting. 

 * The big thing we're not validating is, as mentioned, the types contained within columns e.g. there's nothing that stops you comparing a date and a float. This is doable, but needs more thought.

 * I've generated a [translation]( https://gist.github.com/evansd/aeb3fe7084f65956b0ffaa180fc27024) of every query used in the test suite into the new query model. Though obviously these contain more duplication than they would do if you made them by hand, I still found it useful to get a sense of what the structures look like for "real" queries.


databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
 
-    def __post_init__(self):
-        if self.has_categories:
-            raise NotImplementedError("Categorised codelists are currently unsupported")
+    # Containment is a special case: its right-hand side must be something vector-like i.e.
+    # something containing multiple values. To build a column whose values are vectors,
+    # use the `CombineAsSet` aggregation.
+    class In(Column):
+        lhs: Column
+        rhs: Union[Column, tuple[Value], Codelist]
```
benbc, line 255: Do we need `Codelist` as its own type? Could we have a `Code` type instead and then treat a codelist as a set of those?

databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
 
-    def __repr__(self):
-        if len(self.codes) > 5:
-            codes = self.codes[:5] + ("...",)
-        else:
-            codes = self.codes
-        return f"Codelist(system={self.system}, codes={codes})"
+    class NotIn(In):
+        ...
```
benbc, line 258: Do we need `NotIn` rather than `Not(In())`?

databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
+
+    # TODO: Validation of the "inner" types i.e. a column may contain date values and we
+    # want to make it an error if you try to compare these to an int. And there are
+    # several places (e.g. a Filter condition) where we require suppied columns to have
+    # boolean type. We'll need the DSL to get these types from the schema and pass them
+    # in somehow.
+
+    # The one exception to the "common domain" rule is the Join operation which takes
+    # tables from two different domains and produces a new domain
+    if not isinstance(node, Join):
```
benbc, line 306: This exception here is yucky, but I think it's correct. Maybe it would be clearer to include all domain-generating nodes (just tables?) in the exclusion -- I guess the others aren't excluded now because they don't have children.

benbc (commented):
I like this _a lot_.

In particular I like the compromise you've arrived at between typing and dynamic validation that hangs off that (just enough) typing. And the relaxation that you've made to allow certain silly operations. I spent all day yesterday trying to come up with a more strongly typed version and ran into a series of brick walls.

I don't really have any serious objections to this and would be happy to put my name to it.

databuilder/query_model.py
evansd: To be honest, I slapped in `Join` as a bit of a last minute addition just to show how we could handle it so I didn't spend much time looking for an elegant solution here. I'm sure there is a way of doing it more neatly though, as you say. We should probably work out how to represent in with the DSL and handle it in the QE first though!

databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
+    def __repr__(self):
+        # Gives us `self == eval(repr(self))` as for dataclasses
+        return f"{self.__class__.__name__}.{self.name}"
+
+
+@dataclasses.dataclass(frozen=True)
+class Codelist:
+    codes: tuple
+    system: str
+    has_categories: bool = False
```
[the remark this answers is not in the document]
evansd: Yeah, we should. I've been copying that class into all of my various query model attempts and wincing a bit each time.

databuilder/query_model.py
evansd: This is very good suggestion. Done.

databuilder/query_model.py
evansd: We do not. Done.

databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
-    has_categories: bool = False
+    class DateSubtract(BinaryFunction):
+        ...
 
-    def _get_referenced_nodes(self):
-        return ()
+    class DateDifference(Column):
+        start: Union[Column, datetime.date]
+        end: Union[Column, datetime.date]
+        units: str
```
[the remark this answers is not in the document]
evansd: This might just be a case of the function being badly named. This is a function which returns an int and which we use for calculating someone's age on a given date by using it with the `years` unit. I think it also supports `months`, `weeks` and `days`. We don't have date delta as a type in itself at the moment.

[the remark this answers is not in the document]
benbc: Oh okay. I guess that does the job, doesn't it? No point agonizing over it at this point, I think.

databuilder/query_model.py
```diff
@@ -1,452 +1,356 @@
...
+    ...
+
+
+# OPERATIONS
+#
+# Operations indicate the kind of thing they return by subclassing one of the basic
+# types above.
+
+
+class SelectTable(ManyRowsPerPatientTable):
```
inglesp, line 112: I'm unsure how to mentally parse `SelectTable`.

Are we selecting a table?  Or `SELECT`ing from a table?  Or something else?

evansd: We're selecting a table, but I agree it's a bad choice of verb here. Any better suggestions? Off the top of my head ...

`FrameFromTable` — though I quite liked the fact that the public API didn't need to mention either frames or series.

`FromTable` — fits quite well as a SQL-ism, maybe grammatically a bit odd in the context of the other operations

`UseTable` — avoids the confusing presence of "select" but maybe not much clearer

I feel like `SelectColumn` doesn't suffer from the same confusion, but let me know if you disagree.

benbc: I too like fact that frame and series don't appear.

The symmetry between `SelectTable` and `SelectColumn` seems helpful. But maybe the collision with SQL's `Select` is unhelpful.

I also like the fact that all the nodes represent operations and are named for verbs (except the functions which I _suppose_ could be renamed to `ApplyX`).

As well as "use", which Dave suggests, I suppose we could try "pick" or
 "choose".

inglesp: I'm -0 on `SelectTable` and +0 on `UseTable` or `GetTable`.



> I feel like `SelectColumn` doesn't suffer from the same confusion, but let me know if you disagree.

Agree.

PR #368, opened 2022-02-01
test: Test med safety with DSL

iaindillingham:
This PR attempts to recreate [the cohort-extractor-v1 med safety study][1] within Data Builder. As with previous translation studies, most of the missing functionality relates to practice IDs and arithmetic on instances of `BoolSeries`. However, I was also unclear about hospital admissions.

As a small aside, there is some behaviour I can't fathom:

https://github.com/opensafely-core/databuilder/blob/072ebe679da64929f9ddf4670c46d9d0b1f91530/tests/acceptance/test_med_safety_dsl.py#L111-L115

[1]: https://github.com/opensafely/med-safety-PoC/blob/main/analysis/study_definition_ce1.py

### DIFF
--- a/tests/acceptance/test_med_safety_dsl.py
+++ b/tests/acceptance/test_med_safety_dsl.py
@@ -1,0 +1,313 @@
+from databuilder.tables import (
+    clinical_events,
+    hospital_admissions,
+    patient_demographics,
+    practice_registrations,
+)
+from tests.lib.tpp_schema import organisation, patient, registration, snomed_event
+
+from .codelists import (
+    gib_admissions_codelist,
+    oral_nsaid_codelist,
+    placeholder_admissions_codelist,
+    ppi_codelist,
+)
+
+# The following codes are used by APC (Admitted Patient Care), the national dataset for
+# hospital admissions. They are specific to this dataset, so we don't represent them as
+# a codelist. For more information, see:
+# * https://docs.opensafely.org/dataset-apc/#admission_method
+# * https://datadictionary.nhs.uk/attributes/admission_method.html
+EMERGENCY_ADMISSION_CODES = [
+    "21",  # Emergency Admission: Emergency Care Department or dental casualty department of the Health Care Provider
+    "22",  # Emergency Admission: GENERAL PRACTITIONER: after a request for immediate admission has been made direct to a Hospital Provider, i.e. not through a Bed bureau, by a GENERAL PRACTITIONER or deputy
+    "23",  # Emergency Admission: Bed bureau
+    "24",  # Emergency Admission: Consultant Clinic, of this or another Health Care Provider
+    "25",  # Emergency Admission: Admission via Mental Health Crisis Resolution Team
+    "2A",  # Emergency Admission: Emergency Care Department of another provider where the PATIENT  had not been admitted
+    "2B",  # Emergency Admission: Transfer of an admitted PATIENT from another Hospital Provider in an emergency
+    "2D",  # Emergency Admission: Other emergency admission
+    "28",  # Emergency Admission: Other means, examples are:
+    # - admitted from the Emergency Care Department of another provider where they had not been admitted
+    # - transfer of an admitted PATIENT from another Hospital Provider in an emergency
+    # - baby born at home as intended
+]
+
+
+def test_med_safety_study(cohort, subtest, database):
+    setup_data(database)
+
+    index_date = "2019-09-01"
+    index_date_minus_3_months = "2019-06-01"
+
+    with subtest("population"):
+        filtered_practice_registrations = (
+            practice_registrations.filter(
+                practice_registrations.date_start <= index_date
+            )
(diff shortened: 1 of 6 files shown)


tests/acceptance/test_med_safety_dsl.py
```diff
@@ -0,0 +1,289 @@
...
+        # hospital_admissions.primary_diagnosis.
+        cohort.indicator_gib_num_admissions = (
+            hospital_admissions.filter(
+                hospital_admissions.primary_diagnosis.is_in(gib_admissions_codelist)
+            )
+            .filter(
+                # TODO: Query an IntColumn with is_in
+                hospital_admissions.admission_method.is_in(
+                    EMERGENCY_ADMISSIONS_CODELIST
+                )
```
benbc, line 135: Can you explain what this "codelist" is? If the column is an int, I guess it's not a normal codelist.

tests/acceptance/test_med_safety_dsl.py
```diff
@@ -0,0 +1,289 @@
...
+            )
+            .filter(hospital_admissions.episode_is_finished is True)
+            .filter(hospital_admissions.admission_date >= index_date_minus_3_months)
+            .filter(hospital_admissions.admission_date <= index_date)
+            .count_for_patient()
+        )
+        cohort.assert_results(indicator_gib_num_admissions=None)
+
+    with subtest(
+        "indicator_gib_admission_numerator", missing_feature="Missing dependencies"
```
benbc, line 145: I think this is a bit confusing. Maybe define this variable in the same bock as the dependencies? The order doesn't matter too much, does it?

tests/acceptance/test_med_safety_dsl.py
```diff
@@ -0,0 +1,289 @@
...
+    "28",  # Emergency Admission: Other means, examples are:
+    # - admitted from the Emergency Care Department of another provider where they had not been admitted
+    # - transfer of an admitted PATIENT from another Hospital Provider in an emergency
+    # - baby born at home as intended
+]
+
+
+# The v1 study definition uses a placeholder codelist, but for convenience we will use a
+# list.
+PLACEHOLDER_CODELIST = []
```
benbc, line 29: Can you explain this "placeholder"? Why not a proper codelist?

tests/acceptance/test_med_safety_dsl.py
```diff
@@ -0,0 +1,289 @@
...
+    ):
+        # TPPBackend.hospitalizations exists but doesn't expose an admission_method
+        # property. It's also not clear whether hospitalizations.code ==
+        # hospital_admissions.primary_diagnosis.
+        cohort.indicator_gib_num_admissions = (
+            hospital_admissions.filter(
+                hospital_admissions.primary_diagnosis.is_in(gib_admissions_codelist)
+            )
+            .filter(
+                # TODO: Query an IntColumn with is_in
```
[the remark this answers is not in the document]
iaindillingham: This is a missing feature. In each case, I've clarified this by nesting `subtest`. See 5819ac8.

tests/acceptance/test_med_safety_dsl.py
iaindillingham: Thanks for picking me up on this, as I had copied it from the v1 study definition uncritically. There's more information in [this Slack thread][1] about why we don't represent admission method codes as a codelist. See 5819ac8.

[1]: https://ebmdatalab.slack.com/archives/C31D62X5X/p1643797768199749

tests/acceptance/test_med_safety_dsl.py
iaindillingham: The two dependencies are in separate `subtest` blocks, because each has missing features. I will merge this block into the preceding block, so as not to misuse `subtest` blocks. See bdb21f1.

tests/acceptance/test_med_safety_dsl.py
iaindillingham: The v1 study definition references two placeholder codelists:

* `PLACEHOLDER_admissions_codelist`, which is a hard-coded list of codes
* `PLACEHOLDER_drugs_codelist`, which is [an OpenCodelists codelist][1]

I used an empty list as to represent "work in progress", as this seemed to be the intent of the v1 study definition. However, that's confusing 🙂, so I'll use the same hard-coded list of codes for the admissions codelist. I'll also use the same OpenCodelists codelist for the drugs codelist; we already reference this as `oral_nsaid_codelist`, so I'll prefer that name to `PLACEHOLDER_drugs_codelist`. See 41ca62c.

[1]: https://www.opencodelists.org/codelist/pincer/nsaid/23edd06e/

benbc (approved):

tests/acceptance/test_med_safety_dsl.py
iaindillingham: There is already an event from `oral_nsaid_codelist`, which is why the tests failed. I have modified them to reflect the event and provided more information about why we are introducing redundancy. See 04f2442.

PR #370, opened 2022-02-03
Add new query model

evansd:
This PR replaces the old query model with a shiny new one. At present, the new model does only a small fraction of the validation which we intend it to. But extending the validation shouldn't require changes to the API and so can be done in parallel to other development.

Reviewers interested chiefly in the new query model can mostly ignore the translation code and would be best off looking directly at the `query_model` module, rather than the unhelpful diff:
https://github.com/opensafely-core/databuilder/blob/new-query-model/databuilder/query_model.py

Temporarily, the query engine will accept either the old or the new query model. Once all code that produces old query model instances has been removed then we can stop accepting the old query model.

Behind the scenes, the query engine handles the new query model by translating it into the old. But this should be considered an internal implementation detail of the query engine, done simply to allow us to move forward with development without requiring a full query engine refactor first. Eventually this will be refactored away.

We also retain the test module `test_query_engine.py whose test cases are constructed using the old query model and then translated into the new. This provides a useful harness while refactoring the query engine. Eventually we expect this test suite to be replaced with new tests and discarded.


databuilder/query_engines/query_model_convert_to_old.py
```diff
@@ -0,0 +1,224 @@
...
+    while isinstance(node.source, new.Sort):
+        assert isinstance(node.source.sort_by, new.SelectColumn)
+        sort_columns.insert(0, node.source.sort_by.name)
+        node = node.source
+    return old.Row(convert_node(node.source), tuple(sort_columns), descending)
+
+
+@convert_node.register
+def convert_combine_as_set(node: new.AggregateByPatient.CombineAsSet):
+    return convert_node(node.source)
```
benbc, line 153: Is this right? Do we actually support this?

databuilder/query_engines/base_sql.py
```diff
@@ -156,9 +157,18 @@ def get_queries(self) -> tuple[list[Executable], Executable, list[Executable]]:
 
             list_of_setup_queries, query_to_fetch_results, list_of_cleanup_queries
         """
+        # Temporary migration code: we accept definitions using both the old and new
+        # Query Models and interpret the new model by converting it to the old model
+        # first. Very soon (after the relevant bits of code have been deleted) we can
+        # stop accepting the old model. And then we can refactor the Query Eninge to
```
benbc, line 163:
```suggestion
        # stop accepting the old model. And then we can refactor the Query Engine to
```

tests/test_query_engine.py
```diff
@@ -47,7 +63,7 @@ def test_run_generated_sql_get_single_column_default_population(engine):
     class Cohort(OldCohortWithPopulation):
         output_value = table("clinical_events").first_by("patient_id").get("code")
 
-    assert engine.extract(Cohort) == [dict(patient_id=1, output_value="Code1")]
+    assert engine.extract(convert(Cohort)) == [dict(patient_id=1, output_value="Code1")]
```
benbc, line 66: With this conversion, does that mean we can drop the conversion inside the `QueryEngine` once we've done slashing and burning tests elsewhere that use the old v2 DSL? If so, we should add a note to that story (https://app.shortcut.com/ebm-datalab/story/620/drop-support-for-defining-studies-via-the-cohort-extractor-v2-dsl).

benbc (approved):
Great. Minor comments only. I salute your dedication in getting the translation working (both ways!).

databuilder/query_engines/base_sql.py
benbc: My first PR suggestion!

databuilder/query_engines/query_model_convert_to_old.py
evansd: We do support it in the existing query engine, just all in a very ad hoc and under-specified way. But if you try to filter using an IN condition and the thing on the RHS is an EventSeries (or `Column` in the old world) then we do the right thing. So `CombineAsSet` is a no-op in the conversion function — it's just functioning as a way to say: "Yes, I intend to do the thing I am about to do"

databuilder/query_model.py
```diff
@@ -0,0 +1,388 @@
...
+# one which only accepts an int.
+class DefaultSchema:
+    def __init__(self, name):
+        self.name = name
+
+    def __getitem__(self, column_name):  # pragma: no cover
+        return Any
+
+    def __repr__(self):  # pragma: no cover
+        # Gives us `self == eval(repr(self))` as for dataclasses
```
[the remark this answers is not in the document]
evansd: No witchcraft I'm afraid. It just relies on you providing as `name` the thing to which the object is actually bound, which we do on line 73. So the expressions like this have the right repr:
```
SelectTable("events", schema=DEFAULT_SCHEMA)
```

I found it really useful when experimenting to have reprs that could formatted with black, and then pasted into a console for further fiddling.

tests/test_query_engine.py
evansd: Sadly not. In fact, almost the opposite. The thing we need to drop after the slash and burn is having the conversion be _optional_ i.e. we should insist on being supplied with new query model (that behind the scenes we have to convert to old in order to evaluate). But I'll add a note to the story to make that clear.

databuilder/query_model.py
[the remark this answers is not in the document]
benbc: Ha ha. I had forgotten that the variables would be in scope.

databuilder/query_engines/query_model_convert_to_old.py
benbc: Super.

tests/test_query_engine.py
benbc: Oh yes, of course. But we can drop something once that's story's done. Thanks.

databuilder/query_model.py
```diff
@@ -1,452 +1,388 @@
...
+    rhs: Series[T]
 
-    def __ge__(self, other):
-        return self._get_comparator("__ge__", other)
 
-    def __lt__(self, other):
-        return self._get_comparator("__lt__", other)
+# A function is any operation which takes series and values and returns a series. The
+# dimension of the series it returns will be the highest dimension of its inputs i.e. if
+# any of its inputs has many-rows-per-patient then its output will too.  Bellow are all
```
ghickman, line 214:
```suggestion
# any of its inputs has many-rows-per-patient then its output will too.  Below are all
```

databuilder/query_model.py
```diff
@@ -1,452 +1,388 @@
...
+    class Sum(OneRowPerPatientSeries[Numeric]):
+        source: Series[Numeric]
 
-class Value(QueryNode):
-    @staticmethod
-    def _other_as_comparator(other):
-        if isinstance(other, Value):
-            other = boolean_comparator(other)
-        return other
+    # This is unusual aggregation in that while it collapses multiple values per patient
```
ghickman, line 197:
```suggestion
    # This is an unusual aggregation in that while it collapses multiple values per patient
```

ghickman (approved):
Couple of wording suggestions.

Love the really thorough explanations in the comments 👌

databuilder/query_model.py
```diff
@@ -0,0 +1,388 @@
...
+
+# OPERATIONS
+#
+# Operations indicate the kind of thing they return by subclassing one of the basic
+# types above.
+
+
+# `Value` wraps a single static value in a one-row-per-patient series which just has
+# that value for every patient. This simplifies the signature of all other operations
+# which no longer need to care about static values.
```
iaindillingham, line 143: Would it be correct to think that using `Value` is like [broadcasting][1] in numpy? It takes a scalar and stretches it over the number of rows in the one-row-per-patient series.

[1]: https://numpy.org/doc/stable/user/basics.broadcasting.html

benbc: Sort of. Although that's also true of pretty much all the constructs here. And this stretching applies to multi-row-per-patient series as well as the one-row-per-patient ones, depending on the context.

databuilder/query_model.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class PickOneRowPerPatient(OneRowPerPatientFrame):
+    source: SortedFrame
+    position: Position
+
+
+# An aggregation is any operation which accepts many-rows-per-patient series and returns
+# a one-row-per-patient series. Below are all available aggregations (using a class as a
+# namespace).
```
iaindillingham, line 180: Do we intend to add aggregations in the future? `Mean`, `Median`, and `Mode` spring to mind.

evansd: Yes, absolutely. The same with functions too. I should have said "_currently_ available".

PR #371, opened 2022-02-04
Remove new-old DSL

ghickman:
This introduces various no cover pragmas as various parts of the code
base have lost their tests with the removal of the DSL.  They all have a
comment as to why they were removed so we can check they regain coverage
as the Query Language is added.


databuilder/dsl.py
```diff
@@ -46,26 +46,8 @@
...
-    RoundToFirstOfMonth,
-    RoundToFirstOfYear,
-    Row,
-    Table,
-    Value,
-    ValueFromAggregate,
-    ValueFromCategory,
-)
+
+from .query_model_old import BaseTable, Codelist, Comparator, Value, ValueFromAggregate
```
evansd, line 50: I'm sure this is fine for now but, just for my own understanding, why is a tiny bit of the old DSL staying and using the old query model?

evansd (approved):
:fire:

PR #380, opened 2022-02-07
Drop old Query Model Support

ghickman:
This stops us using old Query Model objects in the Query Engines, then deals with the fallout from that by:

* removing all tests which use the old DSL
* removing unused code from the previous point
* marking all remaining untested lines with `no cover`


tests/end_to_end/test_end_to_end_graphnet.py
```diff
@@ -1,60 +1,10 @@
...
-    load_study, database, cohort_extractor_in_process
-):
-    # A dummy data file is ignored if running in a real backend (i.e. DATABASE_URL is set)
-    # This provides an invalid dummy data file, but it is ignored so no errors are raised
-    run_test(
-        load_study,
-        database,
-        cohort_extractor_in_process,
-        "invalid_dummy_data.csv",
-    )
```
benbc, line 60: It seems odd to be removing this since the dummy data path doesn't actually execute the dataset definition. Is it possible to leave it in with a blank/dummy dataset definition?

databuilder/query_engines/base_sql.py
```diff
@@ -165,6 +165,8 @@ def get_queries(self) -> tuple[list[Executable], Executable, list[Executable]]:
         column_definitions = self.column_definitions
         if isinstance(list(column_definitions.values())[0], query_model.Node):
             column_definitions = convert_to_old(column_definitions)
+        else:
```
benbc, line 168: Could you put an explanatory comment on here, please?

benbc (changes requested):
The story suggests refactoring `main.py`, but I think you're right not to have done that because it no longer has tests. I'll create a separate card for that.

tests/end_to_end/test_end_to_end_graphnet.py
ghickman: `test_extracts_data_from_sql_server_ignores_dummy_data_file` is using [this fixture](https://github.com/opensafely-core/databuilder/blob/drop-old-qm-support/tests/fixtures/end_to_end_tests_graphnet/cohort_graphnet.py) which uses the old Query Model's `table`.  That triggers the `assert False` we added to the Query Engine.  I'm not sure can rewrite this currently, should we mark as `xfail` instead of removing it?

benbc: Actually I was wrong about this. I thought that the dummy data codepath wouldn't use the cohort definition but it does. As you were.

ghickman:
@benbc – are you happy for this to be merged now?

benbc (approved):

PR #384, opened 2022-02-09
Add concise syntax for creating tables

iaindillingham:
This means we don't need the query model when creating tables in the query language test.


databuilder/query_language.py
```diff
@@ -32,12 +32,25 @@ def compile(self) -> dict[str, Node]:  # noqa A003
...
 class Series:
-    def __init__(self, qm_node):
-        self.qm_node = qm_node
+    @classmethod
+    def factory(cls, qm_node):
+        # We use a factory method in preference to __init__ because we don't want
+        # self.qm_node to be None.
+        instance = cls()
+        instance.qm_node = qm_node
+        return instance
```
ghickman, line 46: How does this stop `qm_mode` being None?  Couldn't one pass in None here just like in an `__init__`?

iaindillingham: I worded my comment poorly. I wanted to avoid:

```python
class Series:
    def __init__(self, qm_node=None):
        self.qm_node = qm_node
```

Which is potentially confusing, because `Series.qm_node` will never be `None` when `Series` is used as a descriptor.

We could, indeed, pass `None` here -- but we could error, too.

inglesp (approved):
The table definition in `test_query_language.py` is just right, and I think this might be an actually legitimate non-showy-offy use of descriptors -- nice!

I'm not wild about having to call `.factory` everywhere, but if it gets tiresome, we could reinstate `__init__` with an optional `qm_node` parameter.  I think this would be safe, given that (a) we don't expect users to interact with this code and (b) it'll blow up quickly if `qm_node` is `None`.

iaindillingham:
I think I might actually replace `factory` with `__init__` because `Series.qm_node` being `None` will blow up quickly, as you say, but also informatively: the message is, essentially, "The query language isn't bound to the query model".

databuilder/query_language.py
ghickman: > I just haven't yet written the error code

Ah, that's where my confusion came from!

FWIW I don't see what `.factory()` with a check for `qm is None` gives you over an `__init__()` with the same check 🙂

iaindillingham: Yup, agreed. I've replaced `factory` with `__init__` and it looks like it should be easier to reason about, at least to me 🙂 For one thing, `my_series.qm_node` will now always exist, rather than existing in some cases but not in others.

databuilder/query_language.py
```diff
@@ -32,13 +32,21 @@ def compile(self) -> dict[str, Node]:  # noqa A003
...
+        self.qm_node = SelectPatientTable(name=self.__name__)
 
 
 class Series:
-    def __init__(self, qm_node):
+    def __init__(self, qm_node=None):
         self.qm_node = qm_node
 
+    def __set_name__(self, owner, name):
+        self.__name__ = name
```
inglesp, line 44: I missed `__set_name__` on my first readthrough -- what's it for?

inglesp: Oh, I see, it's so we can do `self.__name__` in the descriptor.

iaindillingham: Yes, good question. This is how a descriptor knows the name it has been assigned by the containing class. I expected the name to be passed to `Series.__get__`, but it isn't, as far as I can tell from the documentation and some `pdb`ing. I followed the [Customized names][1] section of the Descriptor HowTo Guide, which uses this to give public and private names.

[1]: https://docs.python.org/3.9/howto/descriptor.html#customized-names

benbc:
I think this falls into the same trap as my earlier solution -- it complicates the fundamental query language objects (specifically `Series`, both in making it a descriptor and the uncertainty about whether it's bound to the query model) in the cause of making it easier to define tables (which aren't really part of that query language).

I've got an alternative approach which keeps the table definition separate, although it's heavily inspired by your use of descriptors, @iaindillingham.

https://github.com/opensafely-core/databuilder/commit/77e33551a2c5b069f7ebd7ffea114616b1c50984

There are several tricks here which are almost entirely orthogonal, so we don't have to accept them all.

1. The use of descriptors to automatically set the name of the columns.
2. The use of descriptors to bind the columns to their table.
3. Automatically setting the name of the table.
4. Use of `__init_subclass__` and arguments to it to remove duplication between the `Table` classes.

I think the last of these is least justified as it uses a fairly obscure language feature to remove a small amount of duplication (the `select()` method).

The thing I'm keenest about in this solution is having a separate class for the descriptor.

iaindillingham:
Thanks @benbc. I like your alternative solution, but am also wary of `__init_subclass__`. Is this the duplication that it's removing?

```python
class Patients(PatientTable):
    select = SelectPatientTable(name="patients")  # Here?
    patient_id = ...
    date_of_birth = ...
````

My other thought is that at runtime `patients.date_of_birth`, for example, is an instance of `DateSeries` rather than `Column`, which could be confusing. However, I'm not sure that it's _that_ confusing.

---

This is my understanding of your comment. I'm going to leave it here, just in case I've misunderstood.

A limitation of this solution is that it complicates `Series`, which is a query language construct:
* Making it a descriptor is undesirable, because it brings tables into the query language (and tables should be separated from the query language)
* Allowing it to be unbound to the query model is undesirable

The alternative solution:
* makes `Column` the descriptor and, as such, separates tables from the query language
* automatically sets `SelectPatientTable.name` (also `SelectTable.name`) so we don't need `Patients.__name__` any more

Both solutions:
* Use a descriptor to set `SelectColumn.name`. However, by reworking the `__get__` and `__set_name__` methods, the alternative solution avoids creating `Column.__name__`.
* Use a descriptor to bind a series/column to its table

**Comment by benbc** [comment]
>  like your alternative solution, but am also wary of __init_subclass__. Is this the duplication that it's removing?

Not quite. We can take out the `Table` superclass altogether at the cost of duplicating this `select()` method between the two table classes.

```py
class PatientTable:
    @classmethod
    def select(cls):
        return SelectPatientTable(cls.__name__.lower())


class EventTable:
    @classmethod
    def select(cls):
        return SelectTable(cls.__name__.lower())
```

benbc:
> have you tried your approach with PyCharm? My concern is that the class attributes of eg Patients will look like Column objects and not Series objects, and so typechecking editors will get confused.

I have now. :-)

Here's a minimal example:

```py
class Series:
    def do_series_thing(self):
        pass

    def do_shared_thing(self):
        pass

class Column:
    def __get__(self, *args):
        return Series()

    def do_column_thing(self):
        pass

    def do_shared_thing(self):
        pass

class Table:
    foo = Column()

t = Table()
t.foo.do_series_thing()
t.foo.do_shared_thing()
t.foo.do_column_thing()
```

PyCharm gets this right -- it highlights the last call as an error, but is happy with the first two.

`mypy` gets it right once I've added type annotations (in particular `__get__` annotated as returning `Series`).

benbc:
That is: the type checkers understand descriptors.

databuilder/query_language.py
iaindillingham: Ah, so, the need for `__set_name__` makes more sense now: the right-hand side of an assignment is executed before the corresponding variable exists, so given:

```python
class Patients(PatientTable):
    patient_id = IdSeries()
    ...
```

`IdSeries` is instantiated before `Patients.patient_id` exists. Prior to the introduction of `__set_name__` in Python 3.6, passing the variable's name to the descriptor necessitated some duplication:

```python
class Patients(PatientTable):
    patient_id = IdSeries("patient_id")
    ...
```

Or a workaround involving an auto-generated name. [_Attribute Descriptors_](https://learning.oreilly.com/library/view/fluent-python/9781491946237/ch20.html) in _Fluent Python_ gives an example.

[This gist](https://gist.github.com/iaindillingham/71aa6d207a0370d236520d4c313d7698) demonstrates the order of execution.

benbc: This all looks correct to me.

iaindillingham:
After much toing and froing on my part, I've used @benbc's approach more-or-less unaltered. Whilst I think it's not straightforward for a newcomer to understand `Table` and `Column`, they make defining tables (e.g. `Patients`) clearer, especially as `mypy` and PyCharm understand the descriptor. (Visual Studio Code doesn't, unfortunately.)

PR #387, opened 2022-02-09
chore: Hazmat Containment

evansd:
This prepares the ground for work on refactoring the query engine to work with the new query model directly. It bundles the old query engine test module, together with a copy of the old query model and the translation code into a `tests/legacy` directory. This shouldn't need to be touched much again until the point where we have sufficient coverage elsewhere and we can just throw it away.

The old query model is moved in the `query_engines` directory as it should now be considered an implementation detail of the query engine and not used elsewhere. (As grep will show, it currently _is_ used in a few other places; possibly these can just be deleted but I didn't want to make that call.) This means we're free to hack about with it as much as we want while refactoring the query engine.

Depends on #386 and so marked as draft until that's merged.


inglesp (approved):
All sensible.

benbc (approved):

PR #389, opened 2022-02-11
fix: Change `Count` and `Exists` to operate on Frames

evansd:
This better matches their actual semantics. Previously we were
arbitrarily selecting the `patient_id` column to pass in here, but that
was a hack that ehrQL was going to hide from the user; we're better off
not requiring it at all.

As discussed in Slack:
https://ebmdatalab.slack.com/archives/C02LEF2GC75/p1644586507472789

### DIFF
--- a/databuilder/query_engines/query_model_convert_to_old.py
+++ b/databuilder/query_engines/query_model_convert_to_old.py
@@ -165,9 +165,13 @@ def convert_combine_as_set(node: new.AggregateByPatient.CombineAsSet):
 
 def convert_aggregation(node):
     function = AGGREGATE_MAP[node.__class__]
-    assert isinstance(node.source, new.SelectColumn)
-    source_table = convert_node(node.source.source)
-    input_column = column_name(node.source.name)
+    if isinstance(node.source, new.Frame):
+        source = new.SelectColumn(node.source, "patient_id")
+    else:
+        source = node.source
+        assert isinstance(source, new.SelectColumn)
+    source_table = convert_node(source.source)
+    input_column = column_name(source.name)
     output_column = f"{input_column}_{function}"
     row = old.RowFromAggregate(source_table, function, input_column, output_column)
     return old.ValueFromAggregate(row, output_column)
--- a/tests/legacy/query_model_convert_to_new.py
+++ b/tests/legacy/query_model_convert_to_new.py
@@ -50,6 +50,9 @@ AGGREGATE_MAP = {
 }
 
 
+AGGREGATIONS_ON_FRAMES = {new.AggregateByPatient.Exists, new.AggregateByPatient.Count}
+
+
 def convert(old_cohort):
     new_cohort = {column: convert_node(node) for column, node in old_cohort.items()}
     convert_node.cache_clear()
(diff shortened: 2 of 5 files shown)


databuilder/query_engines/query_model_convert_to_old.py
```diff
@@ -165,9 +165,12 @@ def convert_combine_as_set(node: new.AggregateByPatient.CombineAsSet):
 
 def convert_aggregation(node):
     function = AGGREGATE_MAP[node.__class__]
-    assert isinstance(node.source, new.SelectColumn)
-    source_table = convert_node(node.source.source)
-    input_column = column_name(node.source.name)
+    if not isinstance(node.source, new.SelectColumn):
```
benbc, line 168: I understand that this isn't a change here, but why is the only `Series` source we accept a `SelectColumn`, not some other `Series` type?

benbc, line 168: I think at least a comment would be in order here to the effect that this is a `Frame`. Or maybe an explicit type test for the `Frame`.

tests/legacy/query_model_convert_to_new.py
```diff
@@ -127,9 +130,10 @@ def convert_value_from_aggregate(node: old.ValueFromAggregate):
     old_aggregate = node.source
     assert isinstance(old_aggregate, old.RowFromAggregate)
     source = convert_node(old_aggregate.source)
-    column = select_column(source, old_aggregate.input_column)
     Aggregation = AGGREGATE_MAP[old_aggregate.function]
-    return Aggregation(column)
+    if Aggregation not in AGGREGATIONS_ON_FRAMES:
+        source = select_column(source, old_aggregate.input_column)
```
benbc, line 135: Perhaps this answers my question above -- the immediate source of an aggregation must be a `SelectColumn` because that's all that the old query model supports. But (I think) the new query model supports an arbitrary source and we'll get weird errors if someone tries it. Perhaps we need an assert with a clear message in `query_model_convert_to_old` for when we hit that case.

tests/test_query_engine.py
```diff
@@ -19,9 +20,10 @@ def test_year_from_date(engine):
     )
 
     patients = SelectPatientTable("patients")
+    registrations = SelectTable("practice_registrations")
 
     class DatasetDefinition:
-        population = AggregateByPatient.Exists(SelectColumn(patients, "patient_id"))
+        population = AggregateByPatient.Exists(registrations)
```
benbc, line 26: Do we not have a test for `Count`?

evansd: We do, but it's in `legacy/test_query_engine.py` and so it's handled by the translation code.

The only reason this test needs updating is that it happens to use `Exists` for the population definition (and that's because we need a definition that's supported by the existing Query Engine).

tests/legacy/query_model_convert_to_new.py
evansd: Yes, the two conversion modules are almost completely uncommented and I didn't make much effort to make them particularly readable either. This was, of course, a calculated decision to encourage us to get rid of them as soon as possible :stuck_out_tongue: 

My approach with the `convert_to_old` module was just to exactly invert the translation done by `convert_to_new` and assert that the constructs are in the specific form that that translation generates. I didn't spend any time thinking about whether the existing query engine could support a wider range of queries.

The state I was aiming for, and which I think we've now achieved with the validation code, is:
 * If you can construct a query with the query model then you've constructed something meaningful.
 * If you get an assertion error from the translation code, or some other error from the query engine, then that means we've hit a limitation of the existing query engine and we either need to do some refactoring work or decide that we're not  yet supporting that construction.
 * If the query executes without error then the results should be correct.

I think all this does make clear though that we need to start the refactoring work sooner rather than later.

benbc: Yeah, fair enough.

PR #393, opened 2022-02-14
Update Docs Generation

ghickman:
This updates the filename for the JSON we produce to be less specific to one part of databuilder, and lets us pass a location to make its use in the docs repo easier.

The docs repo installs databuilder in a subdirectory but we want the JSON that `generate_docs()` produces to be in the top level repo, so the addition of a location flag lets us do that without more justfile config in that repo.

### DIFF
--- a/databuilder/__main__.py
+++ b/databuilder/__main__.py
@@ -58,7 +58,7 @@ def main(args=None):
             url=options.url,
         )
     elif options.which == "generate_docs":
-        generate_docs()
+        generate_docs(location=options.location)
     elif options.which == "print_help":
         parser.print_help()
     else:
@@ -160,6 +160,11 @@ def build_parser():
         help="Generate a JSON representation of the data needed for the public documentation of Backends and Contracts",
     )
     generate_docs.set_defaults(which="generate_docs")
+    generate_docs.add_argument(
+        "--location",
+        type=Path,
+        help="Optional location to write documentation data to.  Uses current working directory otherwise.",
+    )
 
     return parser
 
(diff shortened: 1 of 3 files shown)


databuilder/__main__.py
```diff
@@ -160,6 +160,11 @@ def build_parser():
         help="Generate a JSON representation of the data needed for the public documentation of Backends and Contracts",
     )
     generate_docs.set_defaults(which="generate_docs")
+    generate_docs.add_argument(
```
CarolineMorton, line 163: Nit: Perhaps add location where defaults to if no value for location path passed in.

CarolineMorton (approved):
Great - one minor comment only.

databuilder/__main__.py
ghickman: Not sure I'm parsing that sentence correctly… do you mean add docs to say "this will default to the current directory", setting a default to the flag, or?

The default is currently the current working directory databuilder is invoked from, but maybe I've not surfaced that enough?

CarolineMorton: I think I was not being clear. I mean make it more clear that the documents will end up where the tool is invoked from if no location path is passed in. We might not need it. I just want to avoid people assuming if there is a `docs/` folder already in the location that they are in, that the newly generated (or updated docs) will end up in this folder logically. My read of the code is that this would not be the case.

ghickman: @CarolineMorton – updated and rebased, do you think that's clear enough?

CarolineMorton: Looks good. thanks!

CarolineMorton (approved):

PR #394, opened 2022-02-18
chore: Initial acceptance test

benbc:
This is a temporary measure to help us get the system working end-to-end. The plan is to replace this with tests that pull in dataset definitions from study repos once such things exist for Data Builder.

Also a bunch of little fixes that I made along the way. Commit-by-commit review strongly recommended.


databuilder/query_model.py
```diff
@@ -13,6 +15,8 @@
 
 # The below classes and functions are the public API surface of the query model
 __all__ = [
+    "Node",
+    "Value",
```
benbc, line 19: @evansd I think that `Value` is an oversight, but maybe you deliberately left off `Node` since it shouldn't be needed for defining datasets. I found that it was in use (for example in the query engine) which resulted in warnings from my IDE. I can be talked out of including it here.

databuilder/query_model.py
```diff
@@ -91,10 +95,6 @@ def __repr__(self):
 class Node:
     "Abstract base class for all objects in the Query Model"
 
-    def __init_subclass__(cls, **kwargs):
-        # All nodes in the query model are frozen dataclasses
-        dataclasses.dataclass(cls, frozen=True)
-
```
benbc, line 97: As per my commit message, I'm not entirely happy about this change. I'm open to reverting it as long as we've discussed the tradeoffs.

databuilder/tables.py
```diff
@@ -0,0 +1,55 @@
...
+        self.series = self.series_cls(
+            qm_node=SelectColumn(source=owner.select(), name=name)
+        )
+
+    def __get__(self, instance, owner=None):
+        return self.series
+
+
+class PatientTable(Table, select=SelectPatientTable):
+    pass
```
benbc, line 37: Perhaps `Table`, `Column` and `PatientTable` belong in their own module, since they will be imported by test code that defines its own tables but doesn't need the "official" ones here. I put them in here for now but am open to other suggestions.

tests/lib/tpp_schema.py
```diff
@@ -12,14 +14,19 @@ class Patient(Base):
     DateOfDeath = Column(Date)
 
 
-def patient(patient_id, sex, dob, *entities, date_of_death=None):
-    for entity in entities:
+def patient(patient_id=None, sex=None, dob=None, date_of_death=None, related=None):
+    if not patient_id:
+        patient_id = random.randint(1, 10**6)
```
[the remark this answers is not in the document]
benbc: That's sensible. Thanks.

databuilder/query_model.py
evansd: Yes, I do see the particular value in being able to typecheck dataset definitions so it makes sense to be willing to pay a higher price there — but ideally we wouldn't have to pay it everywhere.

PR #395, opened 2022-02-18
Test query language

iaindillingham:
As part of [623][], @benbc, @inglesp and I have been thinking about how to test that operations on query language (QL) elements return the correct query model (QM) elements. Considerations:

* We like tests that emphasize what is being tested, and de-emphasize the scaffolding.
* We dislike test code that looks like production code.

This PR is one approach. Characteristics:
* The left-hand side of each assertion is the QL, the right-hand side is the QM
* The scaffolding can be used "as is" e.g. `qm_int_series` is a query model series of integers
* Forward and reverse operations are covered

[623]: https://app.shortcut.com/ebm-datalab/story/623/implement-just-enough-of-ehrql-to-run-a-simple-not-necessarily-realistic-study

### DIFF
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -1,5 +1,16 @@
-from databuilder.query_language import Dataset, DateSeries, IdSeries
-from databuilder.query_model import Function, SelectColumn, SelectPatientTable, Value
+from datetime import date
+
+import pytest
+
+from databuilder.query_language import Dataset, DateSeries, IdSeries, IntSeries
+from databuilder.query_model import (
+    Function,
+    SelectColumn,
+    SelectPatientTable,
+    SelectTable,
+    TableSchema,
+    Value,
+)
 from databuilder.tables import Column, PatientTable
 
 
@@ -33,3 +44,50 @@ def test_dataset() -> None:
             rhs=Value(2000),
         ),
     }
+
+
+# The problem: We'd like to test that operations on query language (QL) elements return
+# the correct query model (QM) elements. We like tests that emphasise what is being
+# tested, and de-emphasise the scaffolding. We dislike test code that looks like
+# production code.
+
+# We'd like Series objects with specific "inner" types. How these Series objects are
+# instantiated isn't important.
+qm_table = SelectTable(
+    name="table",
+    schema=TableSchema("schema", int_column=int, date_column=date),
+)
+qm_int_series = SelectColumn(source=qm_table, name="int_column")
+qm_date_series = SelectColumn(source=qm_table, name="date_column")
+
+
+def assert_produces(ql_element, qm_element):
+    assert ql_element.qm_node == qm_element
+
+
+class TestIntSeries:
+    def test_le_value(self):
+        assert_produces(
+            IntSeries(qm_int_series) <= 2000,
+            Function.LE(qm_int_series, Value(2000)),
+        )
+
+    def test_le_value_reverse(self):
+        assert_produces(
+            2000 >= IntSeries(qm_int_series),
+            Function.LE(qm_int_series, Value(2000)),
+        )
+
+    @pytest.mark.xfail(reason="LE comparison with IntSeries not supported")
+    def test_le_intseries(self):
+        assert_produces(
+            IntSeries(qm_int_series) <= IntSeries(qm_int_series),
+            Function.LE(qm_int_series, qm_int_series),
+        )
+
+
+class TestDateSeries:
+    def test_year(self):
+        assert_produces(
+            DateSeries(qm_date_series).year, Function.YearFromDate(qm_date_series)
+        )
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -8,6 +8,8 @@ This contains just enough of ehrQL to be able to extract the following dataset:
 """
 from __future__ import annotations
 
+import dataclasses
+
 from .query_model import Function, Node, Value
 
 
@@ -31,9 +33,9 @@ class Dataset:
         return {name: variable.qm_node for name, variable in self.variables.items()}
 
 
+@dataclasses.dataclass(frozen=True)
 class Series:
-    def __init__(self, qm_node):
-        self.qm_node = qm_node
+    qm_node: Node
 
 
 class IdSeries(Series):


tests/unit/test_query_language.py
```diff
@@ -63,3 +67,45 @@ def test_dataset() -> None:
...
+
+def assert_produces(ql_element, qm_element):
+    assert qm(ql_element) == qm_element
+
+
+class TestIntSeries:
+    def test_le_value(self, select_column):
+        assert_produces(
+            IntSeries(select_column) <= 2000, Function.LE(select_column, Value(2000))
+        )
```
benbc, line 97: You've solved the problem of having to call `.qm_node` in the tests nicely by explicitly constructing the QL object here. :+1:

tests/unit/test_query_language.py
```diff
@@ -63,3 +67,45 @@ def test_dataset() -> None:
...
+@pytest.fixture
+def select_column(select_patient_table):
+    return SelectColumn(select_patient_table, name="column")
+
+
+qm = operator.attrgetter("qm_node")
+
+
+def assert_produces(ql_element, qm_element):
+    assert qm(ql_element) == qm_element
```
benbc, line 90: Once it's hidden away in the assert function, I think it's fine to just call `.qm_node` here -- the function just obscures things a little.

tests/unit/test_query_language.py
benbc, line 97: It would be nice if `select_column` were called something far less specific. I think it could actually be any QM series object, so some name that reflects that would be ideal.

tests/unit/test_query_language.py
```diff
@@ -63,3 +67,45 @@ def test_dataset() -> None:
...
+# tested, and de-emphasise the scaffolding. We dislike test code that looks like
+# production code.
+@pytest.fixture
+def select_patient_table():
+    return SelectPatientTable(name="patient_table")
+
+
+@pytest.fixture
+def select_column(select_patient_table):
+    return SelectColumn(select_patient_table, name="column")
```
benbc, line 83: I'm not sure what benefit we get from making this a fixture rather than a variable.

tests/unit/test_query_language.py
iaindillingham: I think I was just looking for an opportunity to use `operator.attrgetter`. 8853697 removes it.

tests/unit/test_query_language.py
iaindillingham: None 🙂 2bd5724 replaces fixtures with variables.

tests/unit/test_query_language.py
iaindillingham: Thanks for today's discussion about this. 2bd5724 responds to your comment about `select_column` by introducing `qm_select_int` and `qm_select_date` (as variables rather than fixtures). I spent a bit of time thinking about type checking; specifically, why...

```python
Function.LE(
    lhs=SelectColumn(SelectPatientTable(name="some_table"), name="some_column"),
    rhs=Value(2000),
)
```

...didn't raise an error. I then realized this was because `lhs` was `Series[Any]`. So, I've added a `TableSchema` to type the columns.

benbc: Oh, that's great. I'd missed the need for a schema. And adding that schema makes it all the more obvious that putting the types in those variables names is correct.

benbc: I'm not entirely sure about "select" in that name. I think the series-ness is more interesting to someone reading the code than the fact that it's created by a `SelectColumn`. In a real use of `<=`, it could equally be `AggregateByPatient.Count(SelectPatientTable("events"))`, where we don't even to a `SelectColumn`.

tests/unit/test_query_language.py
```diff
@@ -63,3 +73,41 @@ def test_dataset() -> None:
...
+def assert_produces(ql_element, qm_element):
+    assert ql_element.qm_node == qm_element
+
+
+class TestIntSeries:
+    def test_le_value(self):
+        assert_produces(
+            IntSeries(qm_select_int) <= 2000,
+            Function.LE(qm_select_int, Value(2000)),
+        )
```
benbc, line 99: It would be lovely to see a test for the `2000 >= IntSeries(qm_select_int)` case as well to try to prove that we're getting value from these tests beyond just repeating what's in the implementation.

tests/unit/test_query_language.py
iaindillingham: Thanks for your patience -- we discussed this, and then I forgot we discussed this 🤦🏻‍♂️. 9ab9d0a emphasizes the series-ness over the select-ness and also de-emphasizes the patient-ness of the table and schema.

tests/unit/test_query_language.py
iaindillingham: Added in 3628ab8. I was unsure whether to add a second call to `assert_produces` in `test_le_value` or to add a new test. I added a new test. In the former case, I thought an explanatory comment was necessary; this became the name of the test in the second case (`test_le_value_reverse`).

I think we discussed what to call this functionality: [_The Operator Overloading_][1] chapter of _Fluent Python_ uses _forward_ and _reverse_ (from [_Implementing the arithmetic operations_][2] in the `numbers` documentation). However, the [_Data Model_][3] chapter of the Python documentation uses _reflected_.

[1]: https://learning.oreilly.com/library/view/fluent-python/9781491946237/ch13.html
[2]: https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations
[3]: https://docs.python.org/3/reference/datamodel.html

tests/unit/test_query_language.py
```diff
@@ -63,3 +74,50 @@ def test_dataset() -> None:
             rhs=Value(2000),
         ),
     }
+
+
+# The problem: We'd like to test that operations on query language (QL) elements return
+# the correct query model (QM) elements. We like tests that emphasise what is being
+# tested, and de-emphasise the scaffolding. We dislike test code that looks like
+# production code.
```
inglesp, line 82: I don't understand what "we dislike test code that looks like production code" means.

inglesp (approved):
I think this hits the sweet spot.

tests/unit/test_query_language.py
iaindillingham: This was my summary of your concern that chunks of the test code could look like a copy-and-paste from the production code. For example, test code:

```python
class TestIntSeries:
    def test_le_value(self):
        assert_produces(
            IntSeries(qm_int_series) <= 2000,
            Function.LE(qm_int_series, Value(2000)),  # This
        )
```

Production code:

```python
class IntSeries(Series):
    def __le__(self, other):
        return BoolSeries(Function.LE(lhs=self.qm_node, rhs=Value(other)))  # Looks like this
```

In earlier iterations, the resemblance was far more apparent.

PR #397, opened 2022-02-18
Get All Contracts When Generating Docs File

ghickman:
As of #375 we've started organising contracts into a hierarchy.  Since we don't want to make N calls to `inspect.getmembers` for N locations in the hierarchy this switches us to using `__subclasses__`.

### DIFF
--- a/databuilder/docs.py
+++ b/databuilder/docs.py
@@ -37,8 +37,15 @@ def _build_contracts():
         docstring = _reformat_docstring(contract.__doc__)
         dotted_path = f"{contract.__module__}.{contract.__qualname__}"
 
+        # get the contract's hierarchy without the contracts path prefix
+        hierarchy = contract.__module__.removeprefix("databuilder.contracts.")
+
+        # split up on dots and let the docs plugin handle rendering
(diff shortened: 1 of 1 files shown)


databuilder/docs.py
```diff
@@ -31,13 +30,7 @@ def _build_column(name, instance):
...
     """Build a dict representation for each Contract"""
 
-    # get all classes from the contracts module
-    contract_classes = inspect.getmembers(contracts, inspect.isclass)
-
-    # make sure we only use subclasses of TableContract
-    contract_classes = [c for _, c in contract_classes if issubclass(c, TableContract)]
-
-    for contract in contract_classes:
+    for contract in TableContract.__subclasses__():
```
benbc, line 33: I'm sure this is the right approach. But do we need to exclude any contracts created as part of testing? I suppose that when this is run "for real" the test modules won't have been loaded. But we might get some weird behaviour during tests.

benbc (approved):

databuilder/docs.py
ghickman: Something to look out for, definitely 👍

PR #399, opened 2022-02-21
chore: Reinstate dependencies in story dependency graph

benbc:


ghickman (approved):

scripts/story-dependencies
```diff
@@ -43,7 +43,7 @@ def build_graph(stories):
     graph = networkx.DiGraph()
     story_map = {}
 
-    stories = filter(lambda s: not s.archived, stories)
+    stories = list(filter(lambda s: not s.archived, stories))
```
iaindillingham, line 46: I don't wish to aggregate, but why not:

```python
stories = [s for s in stories if not s.archived]
```

I think the answer is simply a preference for `lambda` over list comprehensions, but would like to check my understanding 🙂

benbc: @iaindillingham I think you're right that that would have been better, especially given the necessity for the `list()`. This has auto-merged now or I would change it to your suggestion.

PR #400, opened 2022-02-22
Add Hierarchies to Contracts

ghickman:
This exposes our contract hierarchies in both the contracts reference data and backends' contracts section.  In the docs these will be used for the title of contracts and to link to those contracts from the backends.


CarolineMorton (approved):
Hi 

This looks good to me. Do we have a test case for it yet?

ghickman:
@CarolineMorton – there's some tests for the docs module in `tests/test_docs` which test a few of the keys and some of the content.  The upcoming plugin and docs repo PRs do some more testing.

PR #402, opened 2022-02-22
Tighten domain constraint

evansd:
The practical effect of this PR is to add a single extra condition to the domain constraint validation: it is now no longer possible to combine the results of two divergent filter operations, even if they operate on the same underlying table. This isn't the sort of thing you should ever _want_ to do (an equivalent operation can be expressed using a linear chain of filters) but it's the sort of thing you might accidentally end up doing if you got confused and we'd like to be able to reject such constructs immediately rather than waiting for them to return unexpected results in production.

Enforcing this check doesn't introduce much additional code, and I think the code is _arguably_ a bit less ad-hoc than what went before. But maybe I've just been thinking about it too long.

### DIFF
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -329,7 +329,7 @@ class Categorise(Series[T]):
 
 def has_one_row_per_patient(node):
     "Return whether a Frame or Series has at most one row per patient"
-    return get_domains(node) == {PATIENT_DOMAIN}
+    return get_domain(node) == PATIENT_DOMAIN
 
 
 def get_series_type(series):
@@ -346,15 +346,9 @@ def get_series_type(series):
 def validate_node(node):
     # Check that the types supplied match the types specified
     validate_types(node)
-    # As well as types we need to validate the "domain constraint". Frames and series
-    # which are in one-row-per-patient form can be combined arbitrarily because we can
-    # JOIN using the patient_id and be sure that we're not creating new rows. But for
-    # operations involving many-rows-per-patient inputs we need to ensure that they are
-    # all drawn from the same underlying table. (We call this the "domain" for set
-    # theoretic reasons which the margins of this comment are too small to contain.)
-    # Note that when we add a Join operation that will be the one explicit exception to
-    # this rule.
-    validate_inputs_have_common_domain(node)
+    # As well as types we need to validate the "domain constraint" which specifies how
+    # Frames and Series, potentially drawn from different tables, can be combined
+    validate_input_domains(node)
 
 
 # DOMAIN VALIDATION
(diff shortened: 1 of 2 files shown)


databuilder/query_model.py
```diff
@@ -365,38 +359,83 @@ class DomainMismatchError(Exception):
...
+    # We enforce this by modelling domains as a hierarchy. At the root is the patient
+    # domain, each table has an unique domain descending from this, and each filter
+    # operation creates a new domain descending from the domain of its source. A valid
+    # operation is one where the domains of its inputs are "nested" in the sense that
+    # each domain is an ancestor or descendant of every other.
+    #
+    # We use sets to implement this hierarchy: we start with the patient domain as a set
+    # containing a single unique value, and we create descendent domains by forming the
+    # union of the parent domain with a new unqiue value. The `issubset()` relation then
+    # gives us the "is domain ancestor" semantics we want.
```
inglesp, line 391: This is neat, but would using a list would be clearer?  Instead of `issubset()`, you'd have to write an `isprefix()` function.

inglesp (approved):
Well this turned out to be a lot neater than I was expecting.

databuilder/query_model.py
evansd: I did experiment with this, but it's not just an `isprefix()` function you need to write: you also need to get the sorting working correctly, which means writing a comparison function and then passing that through `cmp_to_key()` to make it a key function you can give to `sorted()`. Also, the comparison function you end up writing has to lie and pretend that all domains are comparable so you need a comment explaining why that's OK in this context. Whereas sets give us exactly the semantics we want out of the box. So I feel like overall it's simpler, though the trick does require some explanation.

databuilder/query_model.py
```diff
@@ -365,38 +359,83 @@ class DomainMismatchError(Exception):
...
 @singledispatch
-def get_domains(node):
-    return set().union(
-        *[get_domains(input_node) for input_node in get_input_nodes(node)]
-    )
+def get_domain(node):
+    return sorted(get_input_domains(node))[-1]
+
+
+# Selecting a table creates a new domain which descends from the patient domain
```
benbc, line 421: Specifically selecting a non-patient table? Maybe "table" (as in `SelectTable`) isn't a specific enough term and we need something that's more clearly in opposition to "patient table".

databuilder/query_model.py
```diff
@@ -365,38 +359,83 @@ class DomainMismatchError(Exception):
...
+    #  * Many-rows-per-patient data drawn from different tables can never be combined
+    #    — at least, not until we add an explicit JOIN operation.
+    #
+    #  * Many-rows-per-patient data drawn from a single underlying table can be combined
+    #    as long as any filters applied to the data can be arranged in a linear
+    #    sequence. That is, it's legal to combine `Table A -> Filter B -> Filter C` and
+    #    `Table A -> Filter B`. But it's not legal to combine `Table A -> Filter D` and
+    #    `Table A -> Filter E`. There's never a need to use such constructions and they
+    #    will almost always be the result of user error. It's better to reject them
+    #    immediately than give the user unexpected results at query time.
```
benbc, line 380: Since this comment is going to be the main way people new to the code have to get their heads round this, I think that an example showing why an illegal construct would be wrong would be helpful.

databuilder/query_model.py
```diff
@@ -365,38 +359,83 @@ class DomainMismatchError(Exception):
...
+    # must be totally ordered by ancestory. We can check this by sorting them (sets
+    # naturally sort by subset-hood in Python) and then confirming that the ordering is
+    # total by testing that each adjacent pair has the required subset relation.
+    domains = sorted(get_input_domains(node))
+    for a, b in zip(domains, domains[1:]):
+        if not a.issubset(b):
+            raise DomainMismatchError(
+                f"Attempt to combine unrelated domains:\n{a} and {b}"
+                f"\nIn node:\n{node}"
+            )
```
benbc, line 403: Maybe slightly cleaner if the detail of this message were part of the exception and we just pass in the data?

tests/test_query_model.py
```diff
@@ -132,6 +132,55 @@ def test_combining_non_patient_level_series_from_different_frames_throws_error()
...
+    events = SelectTable("events", EVENTS_SCHEMA)
+    foo_events = Filter(events, Function.EQ(SelectColumn(events, "code"), Value("foo")))
+    recent_foo_events = Filter(
+        foo_events,
+        # Note that we're filtering `foo_events` using a condition derived from `events`
+        Function.GE(SelectColumn(events, "date"), Value(datetime.date(2022, 1, 1))),
+    )
+    assert recent_foo_events
+
+
```
benbc, line 183: I think it would be helpful to have a test showing that combining non-patient level series with the same domain is okay.

benbc (commented):
One slight reservation: if we add a new operation that produces non-patient frames or series, we need to understand the domain characteristics of that operation and maybe add a specialized method to `get_domain()`. For example `Sort` and `Filter` look like very similar operations on the face of it, but only one needs special handling. The implementation here doesn't prompt us to do that analysis because the default implementation for `get_domain()` accepts anything.

Maybe this is okay and we rely on something like whatever flash of insight caused you to go into this more deeply with `Filter`. Or maybe we should change the signature of the catch-all `get_domain()` so it only accepts operation we know are safe and forces us to provide specific implementations for things like `Sort`.

evansd:
@benbc On your reservation, I wonder if an acceptable compromise might be to have default behaviour defined for any operation which returns `Series` and then explicitly handle the `Sort` operation. That would mean that any new operation on frames would require explicit handling, but we wouldn't be adding extra bureaucracy for every new Function we add (which are likely to be the things we add most frequently).

Given that it turns out to be possible to miss this stuff even when you're explicitly thinking about it, I'm not sure adding an extra checkbox exercise here will really bring a lot of value.

databuilder/query_model.py
evansd: Maybe. I think we are going to want to raise something more structured here (and the same for the `TypeErrors`) so that ehrQL can catch and translate these for its users. But I'd rather leave thinking about this until we do the work in ehrQL and can see what structure would be most helpful.

evansd:
@benbc Thanks for the useful comments. I've addressed most of them, and implemented the compromise I suggested above but feel free to push back on that if you think it's important.

databuilder/query_model.py
benbc: Fine by me.

benbc (approved):
@evansd Yes, this is fine. I was thinking of something like this anyway -- capturing things we know are okay with abstract types rather than explicitly mentioning every concrete type.

PR #403, opened 2022-02-23
Set up databuilder-in-docker Tests

ghickman:
This adds a new tests package, `tests/docker`, for testing parts of databuilder via the docker image we build.  So far it documents that each of the CLI targets can render its help text, and each database is contactable, exercising that the drivers are correctly installed and configured within the image.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -142,11 +142,11 @@ test-unit *ARGS: devenv
     $BIN/python -m pytest -m "not integration" {{ ARGS }}
 
 # run the integration tests only. Optional args are passed to pytest
-test-integration *ARGS: devenv build-databuilder
+test-integration *ARGS: devenv
     $BIN/python -m pytest -m integration {{ ARGS }}
 
 # run the integration tests only, excluding spark tests which are slow. Optional args are passed to pytest
-test-integration-no-spark *ARGS: devenv build-databuilder
+test-integration-no-spark *ARGS: devenv
     $BIN/python -m pytest -m "integration and not spark" {{ ARGS }}
 
 # run the run-databuilder-in-docker tests only
--- a/Justfile
+++ b/Justfile
@@ -149,6 +149,10 @@ test-integration *ARGS: devenv build-databuilder
 test-integration-no-spark *ARGS: devenv build-databuilder
     $BIN/python -m pytest -m "integration and not spark" {{ ARGS }}
 
+# run the run-databuilder-in-docker tests only
+test-docker *ARGS: devenv build-databuilder
+    $BIN/python -m pytest tests/docker  {{ ARGS }}
+
 # run all tests including integration tests. Optional args are passed to pytest
 test-all *ARGS=test_args: devenv build-databuilder
     #!/usr/bin/env bash
--- a/tests/docker/test_cli.py
+++ b/tests/docker/test_cli.py
@@ -1,0 +1,53 @@
+def test_generate_dataset_in_container(run_in_container):
+    output = run_in_container(
+        [
+            "generate_dataset",
+            "--help",
+        ]
+    )
+
+    assert output
+
+
+def test_validate_dataset_definition_in_container(run_in_container):
+    output = run_in_container(
+        [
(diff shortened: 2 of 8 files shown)


Justfile
```diff
@@ -149,6 +149,10 @@ test-integration *ARGS: devenv build-databuilder
 test-integration-no-spark *ARGS: devenv build-databuilder
     $BIN/python -m pytest -m "integration and not spark" {{ ARGS }}
 
+# run the run-databuilder-in-docker tests only
+test-in-docker *ARGS: devenv build-databuilder
```
benbc, line 153: Maybe `test-docker` would be more consistent with our other targets? I don't feel v. strongly.

Justfile
```diff
@@ -149,6 +149,10 @@ test-integration *ARGS: devenv build-databuilder
...
+    $BIN/python -m pytest tests/docker
```
benbc, line 154: Either need to pass on `ARGS` or remove it.

tests/docker/test_cli.py
```diff
@@ -0,0 +1,63 @@
...
+    run_in_container(
+        [
+            "generate_dataset",
+            "--help",
+        ]
+    )
+
+    captured = capsys.readouterr()
+    assert captured.out
+    assert not captured.err
```
benbc, line 11: Are you planning further validation here and in the other tests below?

tests/docker/test_cli.py
```diff
@@ -0,0 +1,63 @@
+def test_generate_dataset_in_container(capsys, run_in_container):
+    run_in_container(
+        [
+            "generate_dataset",
+            "--help",
+        ]
+    )
+
+    captured = capsys.readouterr()
```
evansd, line 9: Using `capsys` (TIL) here is neat, but I feel like it would be better to fix `run_fg` to return its output rather than printing it:
https://github.com/opensafely-core/databuilder/blob/db1a904676d3cdbd943e149df6f268a3a15897fb/tests/lib/docker.py#L48-L49

tests/docker/test_cli.py
ghickman: I wasn't, no.  Do you think we need anything more than an exception isn't thrown?

Justfile
ghickman: My only reasoning for the `in-` was to avoid ambiguity around what exactly the `docker` part refers to since we have databases in docker, build the tool in docker, etc.  I don't feel strongly about it either though.

tests/lib/docker.py
```diff
@@ -47,6 +47,7 @@ def run_fg(self, image, **kwargs):
         try:
             output = self._run(image=image, detach=False, stderr=True, **kwargs)
             print(str(output, "utf-8"))
+            return output
```
benbc, line 50: I think that printing the output as well as returning it is likely to result in noisy tests.

tests/lib/docker.py
```diff
@@ -47,6 +47,7 @@ def run_fg(self, image, **kwargs):
...
         except ContainerError as e:  # pragma: no cover
             print(str(e.stderr, "utf-8"), file=sys.stderr)
             raise
```
benbc, line 53: On a similar note, maybe get rid of the `print()` here while you're passing?

Justfile
benbc: I think consistency is probably more important here. We're going to have five kinds of tests (I think, currently): spec, unit, integration, docker and acceptance, so I think we should have matching targets for the three of them. There's a separate need for developers to understand what each of those kinds _are_, but that will need documentation (which is planned elsewhere).

benbc (commented):
I'm not sure that we learn a lot by enumerating the commands and getting the help for them.

It's definitely good to check that we can pass in a command and have it execute successfully. I'm not 100% convinced that just executing without an error is enough to prove that, though. So we need at least one test that has an observable side-effect or output which we can assert on.

I'd also like to see a test that proves that Data Builder can read from and write to the workspace that we mount.

benbc (approved):

PR #404, opened 2022-02-24
chore: Local acceptance test

benbc:
I've done the minimum here to create a test that exercises `__main__` and reveals the problem with the command line invocation.

We will want to reuse bits of the machinery here both for acceptance tests that pull in dataset definitions from study repos and for the Docker tests. We'll also want to extend this test to exercise the other commands. But I don't want to introduce more abstractions until it's clear what those should be, so this is all inline for now.

Ideally the dataset definition would be even simpler -- with an "everyone" population and a constant variable value, but this is about the simplest that the Data Builder supports right now.


benbc:
Oh, I didn't realize that stuff I eviscerated was still tested. Fix coming.

evansd:
Will await the further fixes, but looks sensible so far.

ghickman (approved):

PR #412, opened 2022-03-01
Remove Mentions of "study"

ghickman:
This switches our use of "study" to "dataset" or "dataset_definition".

### DIFF
--- a/databuilder/dsl.py
+++ b/databuilder/dsl.py
@@ -50,10 +50,6 @@ from .query_engines.query_model_old import Comparator, ValueFromAggregate
 
 
 class Dataset:
-    """
-    Represents the dataset of patients in the defined study.
-    """
-
     def set_population(self, population):
         """
         Sets the population that are included within the Dataset.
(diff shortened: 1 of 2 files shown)


databuilder/measure.py
```diff
@@ -29,23 +29,23 @@ def __init__(
         small_number_suppression=False,
     ):
         """
-        Creates a "measure" using data extracted by the StudyDefinition defined
+        Creates a "measure" using data extracted by the dataset definition defined
```
benbc, line 32: Just "dataset defined"?

databuilder/dsl.py
```diff
@@ -51,7 +51,7 @@
 
 class Dataset:
     """
-    Represents the dataset of patients in the defined study.
+    Represents the dataset of patients in the defined dataset definition.
```
benbc, line 54: I'm not sure that this really means anything any more. TBF it wasn't a great comment before. Maybe just delete it? This module is going away soon anyway.

benbc (approved):

PR #415, opened 2022-03-02
feat: Add basic validation of dummy data

evansd:
This updates the dummy data validation functions to work with the new Query Model. At present, all we validate are that the types are correct.

Note that we validate directly against the CSV here. We could use Pandas to load the CSV into a dataframe and validate against the types Pandas has inferred for us, but then we'd have two problems :)

[sc-779]


databuilder/validate_dummy_data.py
```diff
@@ -1,151 +1,159 @@
...
+    extension = suffixes[-1] if suffixes else ""
+    return extension, gzipped
 
-    validate_column_values(df, column_definitions)
 
+def get_column_specs(column_definitions):
+    # TODO: It may not be universally true that IDs are ints. Internally the EMIS IDs
+    # are SHA512 hashes stored as hex strings which we convert to ints. But we can't
+    # guarantee always to be able to pull this trick.
+    column_specs = {"patient_id": ColumnSpec(int, nullable=False)}
```
benbc, line 59: [Here](https://github.com/opensafely-core/databuilder/blob/0ee050497036e7ed0b242c7b71cde52566881897/databuilder/contracts/types.py#L52) we explicitly allow patient ids to be varchars. Should that affect what you've done here?

databuilder/validate_dummy_data.py
```diff
@@ -1,151 +1,159 @@
...
+        if missing:
+            errors.append(f"Missing columns: {', '.join(missing)}")
+        extra = headers_set - expected_set
+        if extra:
+            errors.append(f"Unexpected columns: {', '.join(extra)}")
+        raise ValidationError("\n".join(errors))
+    elif headers != expected_headers:
+        # We could be more relaxed about things here, but I think it's worth insisting
+        # that columns be in the same order. We've seen analysis code before which is
+        # sensitive to column ordering.
```
benbc, line 110: The variables in a dataset definition go through a lot of contortions to get here as a list of expected headers. Are you sure that the order is stable?

tests/acceptance/test_embedded_study.py
```diff
@@ -13,3 +17,39 @@ def test_generate_dataset(study, database):
...
+
+def test_validate_dummy_data_error_path(tmp_path):
+    dummy_data = "patient_id,year\n1,1971\n2,foo"
+    dataset_def_path = tmp_path / "dataset.py"
+    dummy_data_path = tmp_path / "dummy_data.csv"
+    dataset_path = tmp_path / "dataset.csv"
+    dataset_def_path.write_text(trivial_dataset_definition)
+    dummy_data_path.write_text(dummy_data)
+    with pytest.raises(ValidationError, match="Invalid int"):
+        _generate_dataset(dataset_def_path, dummy_data_path, dataset_path)
```
benbc, line 41: I'm not sure what we gain by testing this error path here. Is there anything this covers that isn't covered by the unit tests?

databuilder/validate_dummy_data.py
evansd: As long as `get_column_definitions()` gives a consistent order then, yes, it is stable. (Since dicts became explicitly insertion-ordered you generally have to go out of your way _not_ to preserve order.)

databuilder/validate_dummy_data.py
```diff
@@ -1,151 +1,159 @@
...
-
-def validate_expected_columns(df, columns_definitions):
-    """Raise DummyDataValidationError if dataframe does not have expected columns."""
-    expected_columns = set(columns_definitions.keys())
-    extra_columns = set(df.columns) - expected_columns
-    if extra_columns:
-        raise DummyDataValidationError(
-            f"Unexpected columns in dummy data: {', '.join(extra_columns)}"
+    for row_n, row in enumerate(reader, start=1):
+        if None in row:
```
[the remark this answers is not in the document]
evansd: Ah yes, a silly omission.

tests/acceptance/test_embedded_study.py
```diff
@@ -13,3 +17,39 @@ def test_generate_dataset(study, database):
...
+
+
+def test_validate_dummy_data_happy_path(tmp_path):
+    dummy_data = "patient_id,year\n1,1971\n2,1992"
+    dataset_def_path = tmp_path / "dataset.py"
+    dummy_data_path = tmp_path / "dummy_data.csv"
+    dataset_path = tmp_path / "dataset.csv"
+    dataset_def_path.write_text(trivial_dataset_definition)
+    dummy_data_path.write_text(dummy_data)
+    _generate_dataset(dataset_def_path, dummy_data_path, dataset_path)
```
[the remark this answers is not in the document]
evansd: I didn't think it would be needed anywhere else, so I didn't think it was worth refactoring the study fixture for it. I don't think the fixture itself is over-engineered: it's going to be re-used over multiple studies so it makes sense to wrap all that logic up.

tests/acceptance/test_embedded_study.py
evansd: I added it because it confirms that the entrypoint does actually call the validation code. If we didn't have this test we could delete those calls and just always copy the dummy data file over and all the tests would still pass, which felt wrong to me.

databuilder/validate_dummy_data.py
benbc: I'm a bit worried that leaving this as-is will leave us with a potential problem that will be first encountered on a researcher's laptop.

databuilder/validate_dummy_data.py
benbc: > As long as get_column_definitions() gives a consistent order

Indeed. At the very least we don't have tests asserting that.

databuilder/validate_dummy_data.py
evansd: I think in that case we should take the varchar ID option out of the contract until we can support it.

tests/acceptance/test_embedded_study.py
benbc: I think that I'd tackle that by refactoring the code so there is a single function which handles dummy data functionality -- both validating and copying the file -- and unit test that. I don't feel strongly about this, though.

databuilder/validate_dummy_data.py
benbc: Yes, I think you're right. I'll create a separate story for that unless you want to do it here.

evansd: If it's just a case of changing that one contract then I may as well do that here (unless there's further complication I've missed). And then I guess another story for working out how we handle it properly and add it back in?

tests/acceptance/test_embedded_study.py
benbc: Maybe? Really up to you.

benbc (approved):

PR #416, opened 2022-03-02
Remove Old DSL Module

ghickman:


CarolineMorton (approved):
This looks good to me. 

I know that the changes in query_utils will remove any backwards compatibility in the DSL but given we don't have this anymore, I guess this doesn't matter anymore!

PR #420, opened 2022-03-03
Reorganise specs

inglesp:
This makes progress towards [781 Unify query engine test harness](https://app.shortcut.com/ebm-datalab/story/781/unify-query-engine-test-harness).

It also unblocks [773 Generate ehrQL reference documentation from the test specifications](https://app.shortcut.com/ebm-datalab/story/773/generate-ehrql-reference-documentation-from-the-test-specifications).


ghickman (commented):
Great use of a fixture!  Just some naming and usage comments 👍

databuilder/query_language.py
```diff
@@ -35,6 +35,9 @@ def __eq__(self, other):
     def __ne__(self, other):
         return BoolSeries(self._make_binary_fn(other, qm.Function.NE))
 
+    def isnull(self):
```
benbc, line 38: It's surprising this isn't `is_null()`. I assume that is a deliberate choice. Can you explain it?

tests/spec/aggregate_frame/test_count_for_patient.py
```diff
@@ -1,44 +1,30 @@
...
+from ..tables import e, p
+
+table_data = {
+    p: """
+          | i1
+        --+----
+        1 | 101
+        2 | 201
+        3 | 301
+        """,
```
benbc, line 10: Do we need `p` here when it's not used directly? I think there is code in `spec_test()` to cope with it missing for population purposes. (Also applies to other tests below.)

benbc: Oh, I've figured this out. You need it to show the result for members of the population missing from the event table. I think that deserves a comment.

tests/spec/combine_series/test_event_series_and_event_series.py
```diff
@@ -1,36 +1,23 @@
...
-    },
-)
 
-
-def test_combine_event_series_and_event_series(in_memory_engine):
-    in_memory_engine.setup(
+def test_event_series_and_event_series(spec_test):
+    spec_test(
+        table_data,
+        (e.i1 + e.i2).sum_for_patient(),
```
benbc, line 18: It seems a bit of a shame that all of these `combine_series` tests are dependent on both integer arithmetic and `sum_for_patient()`. But I imagine you've come to the conclusion that this is the simplest series you can define that "works" for all the cases you want to cover. And I think you're right to favour making them uniform rather then simplifying the ones that don't involve event series.

It would be nice (especially once we get to generating doc from this) to have some overarching text explaining that these specs show the combination behaviour and that the arithmetic details aren't the important thing. We can defer that to the generate doc story; I think we'll either need to have a way of providing module-level text or merge multiple specs into one file.

tests/spec/combine_series/test_value_and_event_series.py
```diff
@@ -1,35 +1,23 @@
...
-    },
-)
 
-
-def test_combine_value_and_event_series(in_memory_engine):
-    in_memory_engine.setup(
+def test_value_and_event_series(spec_test):
+    spec_test(
+        table_data,
+        (1 + e.i1).sum_for_patient(),
```
benbc, line 18: We talked before about covering some of the cases with unit tests rather than specs. This seems like a candidate for that (along with the other inversions).

I see a couple of reasons for not including specs like this:

* When we generate docs from them they will result in what looks like duplication to a human. The repetition may make the semantics harder to grok.

* I fear we're kidding ourselves by trying to cover "all" the combinations. The only way to truly do that would be to use generative testing. (Of course this is an argument for not using unit tests for this either.)

tests/spec/filter/test_drop_with_expr.py
```diff
@@ -1,37 +1,27 @@
...
-    },
-)
 
-
-def test_drop_with_expr(in_memory_engine):
-    in_memory_engine.setup(
+def test_drop_with_expr(spec_test):
+    spec_test(
+        table_data,
+        e.drop((e.i1 + e.i2) < 413).i1.sum_for_patient(),
```
benbc, line 21: This expression seems more complex than is necessary. Could we do without the sum?

tests/spec/filter/test_drop_with_column.py
```diff
@@ -1,43 +1,28 @@
...
-    },
-)
 
-
-def test_drop_with_column(in_memory_engine):
-    in_memory_engine.setup(
+def test_drop_with_column(spec_test):
+    spec_test(
+        table_data,
+        e.drop(e.b1).i1.sum_for_patient(),
```
benbc, line 22: I wonder if we could document the _motivation_ for `drop`/`take` (as opposed to a filter operation) in the specs, showing that `drop(e.b1)` and `take(~e.b1)` don't partition the table.

tests/spec/int_series/test_arithmetic.py
```diff
@@ -1,43 +1,34 @@
...
 
-    results = transpose(in_memory_engine.extract(dataset))
-
-    with subtests.test("negate"):
-        assert results["negate"] == {1: -101, 2: None}
+def test_add(spec_test):
+    spec_test(
+        table_data,
+        p.i1 + p.i2,
+        {1: 212, 2: None},
```
benbc, line 25: I really like the pattern of making the arithmetic explicit in the expected results (`101 + 111` here).

evansd (approved):
:+1: The `spec_test` helper really simplifies these test cases.

Like George, I think that there are some cases where we'd benefit from being slightly more verbose, but I don't see these as blocking issues.

benbc (approved):
Some feedback here isn't strictly on the changes you've made. You can take it as a compliment, if you like, that the specs are now so much clearer as to make such feedback possible. :-) Please defer dealing with points like that if it's easier.

tests/spec/combine_series/test_value_and_event_series.py
inglesp: I was a bit torn here, but I think this is something where we probably do need to carefully exercise the full code path from QL -> QE.  Combining series of different kinds feels a bit risky to me.

> When we generate docs from them they will result in what looks like duplication to a human. The repetition may make the semantics harder to grok.

I agree this is a risk, but I think we can mitigate with good copy.

> I fear we're kidding ourselves by trying to cover "all" the combinations.

We don't have combinatorial explosion here -- we're not testing that you can combine all kinds of series with all kinds of operations, so I think this is reasonable.

databuilder/query_language.py
inglesp: User error, fixed.

tests/lib/contracts.py
```diff
@@ -73,3 +73,21 @@ class Tests(TableContract):
     test_date = Column(
         type=types.Date(),
     )
+
+
+class P(TableContract):
```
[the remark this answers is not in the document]
inglesp: Now called `PatientLevelTable`.

tests/spec/aggregate_frame/test_count_for_patient.py
inglesp: Good point, have added a comment.

tests/spec/combine_series/test_event_series_and_event_series.py
inglesp: Agree with all of this.

tests/spec/int_series/test_arithmetic.py
inglesp: Fixed.

tests/spec/conftest.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+
+from databuilder.query_language import Dataset
+
+
+@pytest.fixture
+def spec_test(in_memory_engine):
+    def run_test(table_data, series, expected_results):
```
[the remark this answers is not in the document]
inglesp: I disagree.  To me, there are two good reasons for using keywords: to differentiate between two parameters of the same type, and so that readers of the code can avoid diving into the implementation of a function to understand what it does.

Neither applies here, and in particular this isn't code that you're ever going to be "just passing through" -- if you're working on this code, you're going to need to understand exactly what `spec_test()` does, in which case you're going to be familiar enough with what the three parameters are that I don't think they need names.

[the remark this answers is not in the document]
inglesp: I've not got time to add keywords now, but won't object if somebody else does.

PR #421, opened 2022-03-04
chore: Various improvements to `main` and its environs

benbc:
This is just part of a larger refactoring story.

[skip-sc]


evansd (approved):
Sadly, I think Peter's probably right and then only future-proof way to trigger an error on compilation to SQL is via a mock.

PR #422, opened 2022-03-04
refactor: Better approach to executable TableSchema reprs

evansd:
Previously we required each schema to have a name which played no other
role than to provide the repr. This alternative approach gives us
executable reprs without such tricks.

### DIFF
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -58,20 +58,22 @@ class Position(Enum):
         return f"{self.__class__.__name__}.{self.name}"
 
 
-class TableSchema:
+class TableSchema(dict):
     "Defines a mapping of column names to types"
 
-    def __init__(self, name, **columns):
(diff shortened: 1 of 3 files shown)


benbc (approved):
Could we get rid of the class and use a [`frozendict`](https://pypi.org/project/frozendict/). Presumably we'd have to give up the repr trick. I don't feel strongly about this.

evansd:
If `frozendict` was a builtin (and why on earth isn't it, given all the other junk that's been added?) I would have done this in a heartbeat. I just have a (possibly irrational) aversion to dragging a third party dependency into our core data structure for something so trivial.

I've no real attachment to the repr thing. It was handy when doing the initial development but probably won't be much use going forward. It's more that we _so nearly_ get it for free with dataclasses that I felt sad about losing it! Yet more irrationality on my part...

PR #423, opened 2022-03-04
chore: Refactoring the entrypoints

benbc:


databuilder/__main__.py
```diff
@@ -66,113 +66,122 @@ def main(args=None):
...
-        "generate_dataset", help="Generate a dataset"
-    )
-    generate_dataset_parser.set_defaults(which="generate_dataset")
-    generate_dataset_parser.add_argument(
+    return parser
+
+
+def add_generate_dataset(subparsers):
+    parser = subparsers.add_parser("generate_dataset", help="Generate a dataset")
+    parser.set_defaults(which="generate_dataset")
```
evansd, line 90: A pattern we use in the CLI tool which avoids the big switch on `which`, is to do:
```py
parser.set_defaults(function=function_to_call)
```

And then the parsing and invocation code can just be:
```py
args = parser.parse_args()
kwargs = vars(args)
function = kwargs.pop("function")
function(**kwargs)
```

That may be distressingly magical for your taste though. I don't have any objection to doing it the more laborious way.

evansd (approved):
This invokes similar feelings to watching someone pressure-wash a particularly grubby patio.

Lovely stuff!

databuilder/__main__.py
benbc: No, not too magic at all. This code is so predictable and verbose that I think that sort of thing is completely justified. I just didn't take the time to work out how to do it. I'll adopt that here as well. Thanks.

benbc: Well...

I can't see a way of doing this that doesn't make me weep (because we also take some inputs as environment variables).

The command line args and env vars aren't really consistent and coherent. So I'm going to merge this as it is and create a story for the backlog to straighten this out at some point.

PR #433, opened 2022-03-24
test: Generative testing for query model

benbc:
Here's an example run with 10k examples and all the stats turned on, to give an idea for the kind of inputs that we're generating.

```
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/benbc/src/opensafely-core/databuilder, configfile: pyproject.toml
plugins: subtests-0.7.0, cov-3.0.0, freezegun-0.4.2, hypothesis-6.39.3, mock-3.7.0
collected 1 item

tests/generative/test_query_model.py .                                   [100%]

6673 unique inputs

1732 unique examples with node count distribution:
  1	11
  2	13
  3	15
  4	33
  5	70
  6	39
  7	19
  8	44
  9	47
 10	35
 11	38
 12	20
 13	14
 14	16
 15	20
 16	13
 17	18
 18	22
 19	20
 20	25
 21	25
 22	21
 23	18
 24	14
 25	24
 26	24
 27	16
 28	38
 29	43
 30	34
 31	40
 32	53
 33	63
 34	86
 35	97
 36	133
 37	115
 38	86
 39	74
 40	93
 41	17
 42	17
 43	8
 44	16
 45	9
 46	4
 47	3
 48	2
 50	2
 52	5
 53	3
 54	3
 55	2
 60	1
 62	1
 63	1
 66	5
 67	1
 72	1
 79	1
 80	1

Node types:
Not                      11644
SelectColumn             9989
Sort                     8097
SelectPatientTable       7744
SelectTable              3501
PickOneRowPerPatient     2180
EQ                       1162
Exists                   1094
IsNull                   892
Filter                   341
Count                    162
Value                    155
Sum                      29
GE                       20
Negate                   16
LT                       11
GT                       9
Or                       8
LE                       7
NE                       6
Subtract                 4
And                      3

4480 unique datasets with size distribution
  0	1
  1	176
  2	264
  3	238
  4	240
  5	200
  6	216
  7	219
  8	215
  9	231
 10	314
 11	246
 12	255
 13	213
 14	231
 15	175
 16	188
 17	165
 18	159
 19	120
 20	126
 21	97
 22	71
 23	39
 24	32
 25	23
 26	11
 27	9
 28	3
 29	2
 31	1
============================ Hypothesis Statistics =============================

tests/generative/test_query_model.py::test_query_model:

  - during reuse phase (0.35 seconds):
    - Typical runtimes: 2-7 ms, ~ 88% in data generation
    - 63 passing examples, 0 failing examples, 0 invalid examples

  - during generate phase (185.48 seconds):
    - Typical runtimes: 2-10 ms, ~ 97% in data generation
    - 9937 passing examples, 0 failing examples, 17067 invalid examples

  - Highest target score: 0  (label='')
  - Stopped because settings.max_examples=10000


======================== 1 passed in 186.27s (0:03:06) =========================
```

### DIFF
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -28,7 +28,9 @@ __all__ = [
     "ValidationError",
     "DomainMismatchError",
     "has_one_row_per_patient",
+    "has_many_rows_per_patient",
     "get_series_type",
+    "get_input_nodes",
 ]
 
 
@@ -335,6 +337,10 @@ def has_one_row_per_patient(node):
     return get_domain(node) == PATIENT_DOMAIN
 
 
+def has_many_rows_per_patient(node):
+    return not has_one_row_per_patient(node)
+
+
 def get_series_type(series):
     "Return the type contained within a Series"
     assert isinstance(series, Series)
@@ -416,18 +422,15 @@ def validate_input_domains(node):
     # union of the parent domain with a new unique value.
 
     if isinstance(node, Filter) or isinstance(node, Sort):
-        source_domain = get_domain(node.source)
-
-        if isinstance(node, Filter):
-            series = node.condition
-        else:
-            series = node.sort_by
+        frame, series = get_input_nodes(node)
+        assert isinstance(frame, Frame)
+        assert isinstance(series, Series)
+        frame_domain = get_domain(frame)
         series_domain = get_domain(series)
-
-        if not series_domain.issubset(source_domain):
+        if not series_domain.issubset(frame_domain):
             raise DomainMismatchError(
                 f"Attempt to combine series with domain:\n{series_domain}"
-                f"\nWith frame with domain:\n{source_domain}"
+                f"\nWith frame with domain:\n{frame_domain}"
                 f"\nIn node:\n{node}"
             )
     else:
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -428,7 +428,7 @@ def get_input_domains(node):
 
 
 # We use an arbitrary unique object to represent the patient domain
-PATIENT_DOMAIN = frozenset([object()])
+PATIENT_DOMAIN = frozenset(["PatientDomain"])
 
 
 @singledispatch
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -378,47 +378,63 @@ def validate_input_domains(node):
     #  * Many-rows-per-patient data drawn from different tables can never be combined
     #    — at least, not until we add an explicit JOIN operation.
     #
-    #  * Many-rows-per-patient data drawn from a single underlying table can be combined
-    #    as long as any filters applied to the data can be arranged in a linear
-    #    sequence. That is, it's legal to combine `Table A -> Filter B -> Filter C` and
-    #    `Table A -> Filter B`. But it's not legal to combine `Table A -> Filter D` and
-    #    `Table A -> Filter E`. There's never a need to use such constructions and they
+    #  * Many-rows-per-patient series data drawn from a single underlying table can be
+    #    combined if they have had the same set of filters applied to them. That is, it's
+    #    legal to combine two series derived from `Table A -> Filter B`, but not a series
+    #    derived from `Table A -> Filter B -> Filter C` with one derived from
+    #    `Table A -> Filter B`. There's never a need to use such constructions and they
     #    will almost always be the result of user error. It's better to reject them
     #    immediately than give the user unexpected results at query time.
     #
-    #    As an example, the below ehrQL would produce an invalid construct:
+    #    As an example, we reject this erhQL construction because the per-patient
+    #    cardinality of the first argument to `+` may be lower than that of the second:
     #
-    #        foo_events = events.filter(code="foo")
-    #        bar_events = events.filter(code="bar")
-    #        bar_dates = bar_events.date
-    #        foobar_events = events.filter(foo_events.date == bar_dates)
+    #        e.take(e.b).i + e.i
     #
-    #    Were we to allow this, it could be executed but it would always return empty
-    #    results because it involves the condition `code = 'foo' AND code = 'bar'`. In
-    #    this case, what the user should have written is:
+    # * Many-rows-per-patient frames can be combined with many-rows-per-patient series
+    #   derived from the same underlying table, as long as any filters applied to each
+    #   one are identical or make the series a "parent" of the frame. That is, it's legal
+    #   to combine `Table A -> Filter B -> Filter C` and a series derived from
+    #   `Table A -> Filter B`. But it's not legal to combine `Table A -> Filter D` and a
+    #   series derived from `Table A -> Filter E`. Combination of this nature takes place
+    #   with the Filter and Sort operations.
     #
-    #        foobar_events = events.filter(foo_events.date.isin(bar_dates))
+    #   That means that filters and sorts can be expressed tersely in ehrQL, like this:
+    #
+    #       events.filter(events.code == "foo").filter(events.date >= start_date)
     #
     # We enforce this by modelling domains as a hierarchy. At the root is the patient
     # domain, each table has an unique domain descending from this, and each filter
-    # operation creates a new domain descending from the domain of its source. A valid
-    # operation is one where the domains of its inputs are "nested" in the sense that
-    # each domain is an ancestor or descendant of every other.
+    # operation creates a new domain descending from the domain of its source.
+    #
+    # A valid series combination operation is one where the non-patient domains are
+    # identical.  A valid filter or sort operation is one where the domains of its series
+    # input's domain is an ancestor of its frame input's domain.
     #
     # We use sets to implement this hierarchy: we start with the patient domain as a set
-    # containing a single unique value, and we create descendent domains by forming the
-    # union of the parent domain with a new unqiue value. The `issubset()` relation then
-    # gives us the "is domain ancestor" semantics we want.
-    #
-    # Another way of expressing the "nested" condition above is that the input domains
-    # must be totally ordered by ancestory. We can check this by sorting them (sets
-    # naturally sort by subset-hood in Python) and then confirming that the ordering is
-    # total by testing that each adjacent pair has the required subset relation.
-    domains = sorted(get_input_domains(node))
-    for a, b in zip(domains, domains[1:]):
-        if not a.issubset(b):
+    # containing a single unique value, and we create descendant domains by forming the
+    # union of the parent domain with a new unique value.
+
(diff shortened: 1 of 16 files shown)


databuilder/query_model.py
```diff
@@ -428,7 +428,7 @@ def get_input_domains(node):
 
 
 # We use an arbitrary unique object to represent the patient domain
-PATIENT_DOMAIN = frozenset([object()])
+PATIENT_DOMAIN = frozenset(["PatientDomain"])
```
benbc, line 453: I'm not dead set on this since it allows in principle for bugs where we accidentally construct another set that just contains `"PatientDomain"` which would be considered equal to this; but that seems pretty unlikely and this was helpful with early debugging when I was still coming to grips with the way it worked.

tests/lib/in_memory/engine.py
```diff
@@ -65,7 +65,7 @@ def visit_Code(self, node):
         assert False
 
     def visit_Value(self, node):
-        return node.value
+        return Column({patient: [node.value] for patient in self.all_patients})
```
benbc, line 65: This seemed too easy as a fix for bare `Value`s not being allowed as variables (and hence, see following commit) for simplifying `use_unrestricted_population()`. But it works...

tests/generative/test_query_model.py
```diff
@@ -0,0 +1,357 @@
...
+    SelectTable, st.sampled_from(sorted(event_tables.keys())), st.just(schema)
+)
+select_patient_table = qm_builds(
+    SelectPatientTable,
+    st.sampled_from(sorted(patient_tables.keys())),
+    st.just(schema),
+)
+select_column = qm_builds(SelectColumn, frame, st.sampled_from(sorted(schema.keys())))
+
+filter = qm_builds(Filter, many_rows_per_patient_frame, series)  # noqa: A001
```
evansd, line 223: I suppose you could use `filter_` here to avoid blatting the builtin, which would fit the convention used elsewhere.

tests/generative/test_query_model.py
```diff
@@ -0,0 +1,357 @@
...
+
+# And here are strategies for generating data. This is complicated by the need for patient ids in patient tables to
+# be unique (see `patient_records()`).
+max_patient_id = 10
+max_num_patient_records = max_patient_id  # <= max_patient_id, hasn't been fine-tuned
+max_num_event_records = max_patient_id  # could be anything, hasn't been fine-tuned
+
+
+def record(table, id_strategy):
+    type_ = (event_tables | patient_tables)[table]
```
evansd, line 266: :ok_hand: PEP584 was long overdue, wasn't it?

tests/generative/test_query_model.py
```diff
@@ -0,0 +1,357 @@
...
+
+Base = sqlalchemy.orm.declarative_base()
+
+
+class P1(Base):
+    __tablename__ = "p1"
+    Id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True, default=next_id)
+    PatientId = sqlalchemy.Column(sqlalchemy.Integer)
+    i = sqlalchemy.Column(sqlalchemy.Integer)
+    b = sqlalchemy.Column(sqlalchemy.Boolean)
```
inglesp, line 40: By not having two columns of the same type in any table, are we missing out on some of the space of possible queries?

tests/generative/test_query_model.py
```diff
@@ -0,0 +1,357 @@
...
+#   "simplicity", using the number of child nodes as a heuristic. It's hard to follow even that, though, because of
+#   the way the concrete types are grouped. And, partly as a consequence, this strategy doesn't shrink very well.
+# * We need to use `st.deferred` in some places to allow (transitively) mutually recursive strategies. We're using it
+#   in _all_ the abstract type strategies in order to allow us to keep the ordering the same as that of the definitions
+#   in `query_model`.
+# * The query model type hierarchy doesn't give us all the information we need, so the approach outlined above will
+#   generate invalid query model graphs (due either to the domains or the types being incorrect). We handle this by
+#   constructing the query model objects using a special-purpose strategy `qm_builds`, which can discard invalid
+#   instances. There is an unexplored optimization avenue here: make the strategies more specific to include type
+#   and/or domain, so that we don't discard so many examples.
```
inglesp, line 120: I can follow how all this works, and I'm reasonably confident that I could add support for a new kind of QM object, but I'm not sure I fully understand why we need two kinds of strategy.  Is that something you're able to explain in this comment?

tests/lib/in_memory/database.py
```diff
@@ -30,10 +30,10 @@ def setup(self, *input_data):
...
         for item in input_data:
             table_name_to_items[item.__tablename__].append(item)
-        self.tables = {
-            table_name: self.build_table(items)
-            for table_name, items in table_name_to_items.items()
-        }
+        self.tables = defaultdict(Table.empty)
+
+        for table_name, items in table_name_to_items.items():
+            self.tables[table_name] = self.build_table(items)
```
inglesp, line 36: This is a reasonable approach, although we no longer raise an error if you try to look up a column on a frame when the column doesn't exist.  I can't see that this matters though.

An alternative would be to pass in a schema.

tests/lib/in_memory/engine.py
inglesp: I should think most things seem too easy after everything else you've put yourself through in this PR...

inglesp (approved):

databuilder/query_model.py
```diff
@@ -378,47 +378,63 @@ def validate_input_domains(node):
...
+        else:
+            series = node.sort_by
+        series_domain = get_domain(series)
+
+        if not series_domain.issubset(source_domain):
+            raise DomainMismatchError(
+                f"Attempt to combine series with domain:\n{series_domain}"
+                f"\nWith frame with domain:\n{source_domain}"
+                f"\nIn node:\n{node}"
+            )
```
evansd, line 438: I think this is fine but fwiw, the approach I took when I sketched this out earlier was to use `@singledispatch` on `validate_input_domains()`  and then have a single function which works with both `Filter` and `Sort`:
```py
@validate_input_domains.register(Filter)
@validate_input_domains.register(Sort)
def validate_inputs_domains_for_frame_and_series_operations(node):
    frame, series = get_input_nodes(node)
    assert isinstance(frame, Frame)
    assert isinstance(series, Series)
    frame_domain = get_domain(frame)
    series_domain = get_domain(series)
    if not series_domain.issubset(frame_domain):
        raise DomainMismatchError(
            f"Attempt to combine series with domain:\n{series_domain}"
            f"\nWith frame with domain:\n{frame_domain}"
            f"\nIn node:\n{node}"
        )
```

I think `@singledispatch` probably isn't helpful here, but I quite like not caring what the argument name of the series is as I feel it makes the intent clearer.

benbc:
@inglesp I meant to say: I took on board your earlier feedback about the inconsistency between strategies that are callables and those that are bare variables. I converted all the strategies to be functions and the whole thing ran massively slowly, so I backed it out.

tests/generative/test_query_model.py
benbc: Yes. I simplified it as much as I could to start with and then didn't add them back. It would be straightforward to add them, but there are lots of ways we could make it richer. I dunno. WDYT?

tests/generative/test_query_model.py
benbc: By two kinds, do you mean the ones for the abstract vs concrete types?

evansd (approved):
This is great. I think I'm at the point where the approach makes sense to me enough that I think I'd be happy enough to extend and build on it; but I'm not sure I _really_ understand how Hypothesis is doing its thing here — and I certainly couldn't have written this code myself.

At some point I'd love to get my head round it, but I don't think there's any hard requirement for me to do so at this stage, right?

tests/lib/in_memory/database.py
benbc: Yeah. I can't (at the moment) see any important errors that we would miss because of that. I think it's probably simpler than having a schema, so let's leave it like this until we find a reason to do otherwise.

databuilder/typing_utils.py
```diff
@@ -83,6 +83,8 @@ def type_matches(spec, target_spec, typevar_context):
 
     # If there's no origin type that means `target_spec` is an ordinary class
     if target_spec_origin is None:
+        if spec == bool and target_spec == int:
```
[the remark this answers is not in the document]
benbc: Good idea.

tests/generative/test_query_model.py
benbc: I've made this change since it was so easy.

tests/generative/test_query_model.py
benbc: Oh yes, thanks.

tests/generative/test_query_model.py
```diff
@@ -0,0 +1,357 @@
...
+bool_values = st.booleans()
+value = qm_builds(Value, st.one_of(int_values, bool_values))
+
+# To simplify data generation, all tables have the same schema.
+schema = TableSchema(i=int, b=bool)
+event_tables = {"e1": E1, "e2": E2}
+patient_tables = {"p1": P1, "p2": P2}
+
+select_table = qm_builds(
+    SelectTable, st.sampled_from(sorted(event_tables.keys())), st.just(schema)
```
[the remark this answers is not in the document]
benbc: Without the explicit sort, I get this error from Hypohthesis:

> Cannot sample from dict_keys(['i1', 'i2', 'b1', 'b2']), not an ordered collection. Hypothesis goes to some length to ensure that the sampled_from strategy has stable results between runs. To replay a saved example, the sampled values must have the same iteration order on every run - ruling out sets, dicts, etc due to hash randomisation.

I guess that's a bug in Hypothesis, which hasn't been updated since dicts became stable.

But I've replaced `sorted()` with `list()` as a gesture of solidarity.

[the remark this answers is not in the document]
evansd: Oh I see! Sorry to have created superfluous work

benbc:
@evansd Could you take a look at this last commit, please? https://github.com/opensafely-core/databuilder/pull/433/commits/07636ea4b862f4202e2075d406f57036332ccb52

Other changes are fairly minor response to reviews rebased into the old commits and don't need rereview.

evansd (approved):
👍 `map_imperatively` makes this much neater, doesn't it?

tests/generative/test_query_model.py
benbc: @inglesp I'm going to merge this to get it off my plate, but please feel free to clarify what you meant here and I'll try to improve the comments later.

PR #437, opened 2022-03-30
Initial work on SQLite Query Engine

evansd:
This passes the first 75% of the spec tests. Sadly, this is the easy 75% as it doesn't involve any filtering, aggregation or combining series, but it's a start.

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -70,6 +70,54 @@ class SQLiteQueryEngine(BaseQueryEngine):
     def get_sql_eq(self, node):
         return operators.eq(self.get_sql(node.lhs), self.get_sql(node.rhs))
 
+    @get_sql.register(Function.NE)
+    def get_sql_ne(self, node):
+        return operators.ne(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.IsNull)
+    def get_sql_is_null(self, node):
+        return operators.is_(self.get_sql(node.source), None)
+
+    @get_sql.register(Function.Not)
+    def get_sql_not(self, node):
+        return sqlalchemy.not_(self.get_sql(node.source))
+
+    @get_sql.register(Function.And)
+    def get_sql_and(self, node):
+        return operators.and_(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.Or)
+    def get_sql_or(self, node):
+        return operators.or_(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.LT)
+    def get_sql_lt(self, node):
+        return operators.lt(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.LE)
+    def get_sql_le(self, node):
+        return operators.le(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.GT)
+    def get_sql_gt(self, node):
+        return operators.gt(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.GE)
+    def get_sql_ge(self, node):
+        return operators.ge(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.Negate)
+    def get_sql_negate(self, node):
+        return operators.neg(self.get_sql(node.source))
+
+    @get_sql.register(Function.Add)
+    def get_sql_add(self, node):
+        return operators.add(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.Subtract)
+    def get_sql_subtract(self, node):
+        return operators.sub(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
     @get_sql.register(SelectColumn)
     def get_sql_select_column(self, node):
         source = self.get_sql(node.source)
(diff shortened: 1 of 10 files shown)


benbc (approved):

databuilder/query_engines/sqlite.py
```diff
@@ -0,0 +1,178 @@
...
+from databuilder.sqlalchemy_utils import get_referenced_tables
+
+from .base import BaseQueryEngine
+
+
+class SQLiteQueryEngine(BaseQueryEngine):
+
+    sqlalchemy_dialect = SQLiteDialect_pysqlite
+
+    def get_query(self, column_definitions):
```
benbc, line 17: Can we switch to "variable" instead of "column" for consistency with terminology elsewhere?

databuilder/query_engines/sqlite.py
```diff
@@ -0,0 +1,178 @@
...
+            table for expr in column_expressions.values() for table in get_tables(expr)
+        }
+        all_tables = population_tables | other_tables
+
+        # Build a column expression for `patient_id`. In the simplest case there's a
+        # single table in the query which is guaranteed to have rows for every patient
+        # and so we can use that. But the flexibility of the query model means that we
+        # need to cope with more complex situations. Because of the outer joins between
+        # tables there may be no one table which has a row for every patient in the
+        # query.
```
benbc, line 38: I have thoughts about population definition which I'm just writing up in a Slack thread.

benbc (commented):
Sorry, a couple of things I forgot in my first review.

databuilder/query_engines/sqlite.py
```diff
@@ -12,12 +15,93 @@ class SQLiteQueryEngine(BaseQueryEngine):
...
+    def get_sql(self, node):
+        assert False, f"Unhandled node: {node}"
+
+    @get_sql.register(Value)
+    def get_sql_value(self, node):
+        return node.value
+
+    @get_sql.register(Function.EQ)
+    def get_sql_eq(self, node):
+        return self.get_sql(node.lhs) == self.get_sql(node.rhs)
```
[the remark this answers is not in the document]
evansd: I think you're quite right. There are definitely unsugared versions of `&` and `|` and I'm sure we can come up with some way of not using operator overloading and making the meaning clear.

databuilder/query_engines/sqlite.py
```diff
@@ -12,12 +15,93 @@ class SQLiteQueryEngine(BaseQueryEngine):
...
+            *[expr.label(name) for name, expr in column_expressions.items()],
+        ]
+
+        return (
+            sqlalchemy.select(columns).select_from(joins).where(population_expression)
         )
 
+    @singledispatchmethod
+    def get_sql(self, node):
+        assert False, f"Unhandled node: {node}"
```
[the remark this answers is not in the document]
evansd: I'd love to have these all as functions too, but I'm not sure it's going to be easily possible because (although this isn't yet the case) it's not just the table mapping for which these methods will need `self`. There are all sorts of places where we need to specialise behaviour per query-engine (e.g. how to create a temporary table) in the middle of walking the graph and I think the most practical way of doing that is to have these as methods on a base class.

> It might be easier to understand this code if we clarify that by having a table mapping object that it can return rather than calling a function on it for each table. This then becomes clearly data.

The reason I liked the idea of using functions here was that I wanted to leave open the possibility of the query engine giving the backend more information about the kind of query it's going to do (e.g these are all the columns I'll need) and letting the backend optimise the kind of query it returns. In my head was something like SQLite's [Virtual Tables](https://www.sqlite.org/vtab.html). I think this stems from worries I had when first designing all this stuff that it would generate awful performance problems and this was my way of mentally waving those away! But we're not doing any that stuff; we've got not immediate plans to; and if we decide we need to do it then we can change the API. So I'd be happy to change it to something declarative if you think that would be an improvement.

> The backend is quite a rich object and we only need a small part of it here.

Yes, I very much agree with this and I've been thinking about change the QueryEngine API for a while, but this didn't feel like the right PR. In my head I had something like this:
```py
query_engine = QueryEngine(database_url, get_table_expression=backend.get_table_expression)
results = query_enginge.fetch_results(dataset_definition)
```
That is, just passing the bit of the backend we care about to the constructor, and not passing the dataset definition itself but supplying that later.

I'm slightly in two minds about passing the database_url to the constructor as well. Should you _have_ to have a database to be able to construct a query engine? Maybe it should be a parameter to `fetch_results()`.

It's worth getting right I think, but also easy enough to change later so maybe doesn't need to much thought right now.

evansd:
@benbc I think this is ready now, if you'd like to take another look over it.

You'll note I've done the simple but inefficient thing with the population table, with the intention of changing this later.

databuilder/query_engines/sqlite.py
```diff
@@ -0,0 +1,159 @@
...
+    def get_results_query(self, variable_expressions, population_expression):
+        # Get all referenced tables
+        tables = set(get_tables(population_expression))
+        for expression in variable_expressions.values():
+            tables.update(get_tables(expression))
+
+        assert len(tables) > 0, "No tables found in query"
+
+        patient_table = self.get_patient_table(tables)
+        other_tables = tables - {patient_table}
```
benbc, line 35: This can't be part of that set, can it?

evansd: Not under the current implementation. But if we changed that to return a vanilla table like `Patients` then it could, and I thought maybe it made the intention clearer.

benbc: Up to you.

PR #449, opened 2022-04-11
test: Include SQLite in generative tests

benbc:


tests/generative/data_setup.py
```diff
@@ -48,10 +47,33 @@ def build_table(name, schema, patient_id_column, ids, registry):
...
-        query_engine_class = "not-needed"
-        patient_join_column = patient_id_column_
+def _make_backend(table_names, schema, patient_id_column):
+    tables = _backend_tables(schema, table_names)
+    class_vars = {
+        "backend_id": "gen-test-backend",
+        "query_engine_class": "not-needed",
+        "patient_join_column": patient_id_column,
+    } | tables
+    return type("Backend", (backends.BaseBackend,), class_vars)
```
inglesp, line 61: :trophy:

tests/generative/variable_strategies.py
```diff
@@ -219,3 +221,112 @@ def composite_filter(draw, wrapped, predicate):
...
+# The only way to detect cases like that would be to replicate SQLAlchemy's interpretation of the entire tree in order
+# to check that the query is non-trivial.
+#
+# We therefore exclude all trivial nodes from the query. That means that there are realistic queries with trivial
+# subtrees that we can't generate with these strategies. We could consider composing known-non-trivial subtrees with
+# possibly-trivial subtrees as a future improvement to these strategies.
+#
+# And advantage to this approach is that we reject huge numbers of trivial queries that Hypothesis was otherwise
+# inclined to generate, so the queries generally are rather more realistic.
+def is_trivial(node):
```
inglesp, line 250: This is a hyperantiautological function, I love it.

tests/generative/variable_strategies.py
```diff
@@ -219,3 +221,112 @@ def composite_filter(draw, wrapped, predicate):
...
+    if type(node) in [
+        Function.Not,
+        Function.IsNull,
+        Function.Negate,
+        Function.RoundToFirstOfMonth,
+        Function.RoundToFirstOfYear,
+        Function.YearFromDate,
+    ]:
```
inglesp, line 258: What happens when we add more unary functions?

tests/unit/test_query_model.py
```diff
@@ -308,6 +309,20 @@ def test_infer_types_where_possible_even_without_schema():
         Function.EQ(event_count, Value("some_string"))
 
 
+def test_combination_of_typed_and_untyped_series():
+    # We're checking here that the type-checking can correctly combine typed series (`IsNull` has type `bool`) and
+    # untyped ones (`SelectColumn` has type `Any` because no schema is provided).
+    assert Function.EQ(
+        lhs=Function.IsNull(Value(None)),
+        rhs=SelectColumn(SelectPatientTable("p0"), "b1"),
+    )
```
inglesp, line 318: What is the `assert` for?

tests/lib/in_memory/database.py
```diff
@@ -209,6 +217,11 @@ def __repr__(self):
             for value in values
         )
 
+    def dict(  # noqa: A003 # builtin shadowing doesn't like a problem for properties
```
inglesp, line 220: Garbled comment?

tests/lib/in_memory/database.py
```diff
@@ -209,6 +217,11 @@ def __repr__(self):
...
+        self,
+    ):
+        return dict(self.patient_to_values)  # pragma: no cover
```
inglesp, line 223: Why is this not covered?

tests/lib/in_memory/test_database.py
```diff
@@ -94,10 +94,12 @@ def test_patient_column_binary_op_patient_column():
         4 | 411
         """
     )
-    assert pc1.binary_op(add, pc2) == Column.parse(
+    assert pc1.binary_op_with_null(add, pc2) == Column.parse(
```
inglesp, line 97: Could we have a test that shows what `.binary_op()` does?

inglesp (approved):
Am happy with the in-memory database changes, and have skimmed the other stuff.

databuilder/query_model.py
```diff
@@ -505,6 +507,23 @@ def get_input_nodes_for_categorise(node):
...
+    nodes = []
+
+    for subnode in get_input_nodes(tree):
+        for node in all_nodes(subnode):
+            nodes.append(node)
+    return [tree] + nodes
+
+
+def count_nodes(tree):  # pragma: no cover
+    return len(all_nodes(tree))
```
evansd, line 521: I'm not sure this is accurate because the `all_nodes()` func it uses doesn't suppress duplicates.

tests/generative/data_setup.py
evansd: My golden rule of Python development is that you're not _really_ having a good time until you're using the three argument form of `type()`.

tests/generative/data_setup.py
```diff
@@ -44,6 +44,10 @@ def _build_orm_class(name, schema, patient_id_column, ids, registry):
     class_ = type(name, (object,), dict(__tablename__=name, metadata=registry.metadata))
     registry.map_imperatively(class_, table)
 
+    # It's helpful to have the classes available as module properties so that we can copy-paste failing test cases
+    # from Hypothesis.
+    globals()[name] = class_
```
evansd, line 49: Even better than using three-argument `type()`!

tests/generative/variable_strategies.py
evansd: It's possible we could determine this (and the other lists of functions) dynamically by introspection over all node types exported by the query model i.e. walk over everything in `query_model.__all__`, is it a QueryNode, if so use `dataclasses.fields()` to check what it accepts.

databuilder/query_engines/sqlite.py
```diff
@@ -84,11 +84,21 @@ def get_sql_not(self, node):
 
     @get_sql.register(Function.And)
     def get_sql_and(self, node):
-        return operators.and_(self.get_sql(node.lhs), self.get_sql(node.rhs))
+        # SQLAlchemy doesn't provide reverse bitwise operations, so `True & Column()` raises a `TypeError`.
+        # See https://github.com/sqlalchemy/sqlalchemy/issues/5846.
+        try:
+            return self.get_sql(node.lhs) & self.get_sql(node.rhs)
```
evansd, line 90: How come the switch back to operator overloading here, rather than using `operators.and_`?

tests/unit/test_query_model.py
evansd: I do this too, just because it feels weird to have a test with no asserts when all I really want to check is that no exception is thrown. I tell myself it's good to check that the function I'm calling has at least returned _something_.

evansd (approved):
Well that was a romp! :tophat: for getting this all to work despite SQLAlchemy's best efforts to stop you.

Peter got there before me with the micro-level comments so I haven't duplicated those.

The more time I've spent working with `exists/count/is_null` the more I've concluded that we really are going to need a fixed universe table provided by the backend. And I think if we had that then some of the complexity here wouldn't be needed. (Although as you say, there are other reasons to want to excluded trivial queries from the generated tests.) But we can discuss that separately.

tests/lib/databases.py
```diff
@@ -273,6 +275,14 @@ def __init__(self, db_name):
             host_from_host=None,
             port_from_host=None,
         )
+        self._engine = None
+
+    def engine(self, dialect=None, **kwargs):
+        # We need to hold a reference to the engine for the lifetime of this database to stop the contents of the
+        # database from being garbage-collected.
```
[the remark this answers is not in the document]
inglesp: Oh, haha.

tests/lib/in_memory/engine.py
```diff
@@ -52,7 +53,19 @@ def tables(self):
 
     @property
     def all_patients(self):
-        return self.database.all_patients
+        tables_mentioned = []
+        for variable in self.column_definitions.values():
+            for node in all_nodes(variable):
+                if type(node) in [SelectPatientTable, SelectTable]:
```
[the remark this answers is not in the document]
benbc: This got reworked to treat the two table types differently.

tests/lib/in_memory/engine.py
```diff
@@ -52,7 +53,19 @@ def tables(self):
 
     @property
     def all_patients(self):
-        return self.database.all_patients
+        tables_mentioned = []
```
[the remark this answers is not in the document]
benbc: Should have been. Now replaced by other code that does use sets.

databuilder/docs/specs.py
```diff
@@ -143,7 +143,7 @@ def convert_output_value(value):
 
 
 class ArgCapturer:
-    def __call__(self, table_data, series, expected_output):
+    def __call__(self, table_data, series, expected_output, population=None):
```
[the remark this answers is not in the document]
benbc: It doesn't change the documentation. I think that's just about okay. It means we're somewhat glossing over the population semantics in the specs. We have a separate task to add specs for the population semantics, which should fill any holes. I'll add a note to that to reconsider this at that point, if you're happy with that.

tests/generative/data_setup.py
```diff
@@ -31,3 +38,12 @@ def build_table(name, schema_, patient_id_column_, ids, registry):
     registry.map_imperatively(class_, table)
 
     return class_
+
+
+def make_backend(patient_id_column_):
```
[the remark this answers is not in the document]
benbc: An artifact of an earlier verison with a name clash. Gone.

tests/generative/data_setup.py
```diff
@@ -24,12 +24,20 @@ def setup(schema):
     )
 
 
-def build_table(name, schema_, patient_id_column_, ids, registry):
+def build_tables(prefix, count, schema, patient_id_column, ids, registry):
+    patient_tables = [
```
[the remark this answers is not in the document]
benbc: Should have been. That stuff's now been somewhat rewritten anyway.

tests/generative/test_query_model.py
```diff
@@ -27,7 +27,9 @@
 
 # To simplify data generation, all tables have the same schema.
 schema = TableSchema(i1=int, i2=int, b1=bool, b2=bool)
-patient_id_column, patient_tables, event_tables, Backend = data_setup.setup(schema)
+patient_id_column, patient_tables, event_tables, Backend = data_setup.setup(
+    schema, 2, 2
```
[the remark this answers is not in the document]
benbc: Good idea. Thanks.

tests/generative/conftest.py
```diff
@@ -28,46 +28,53 @@ def count_nodes(tree):  # pragma: no cover
...
+        type_ for node in get_input_nodes(tree) for type_ in node_types(node)
+    ]
 
 
 def histogram(samples):  # pragma: no cover
     h = defaultdict(int)
     for sample in samples:
         h[sample] = h[sample] + 1
-    return dict(h.items())
+    return sorted(h.items(), key=lambda item: item[0])
```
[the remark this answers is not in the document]
benbc: I suppose that's right. Thanks.

tests/lib/in_memory/database.py
```diff
@@ -58,12 +58,12 @@ class Table:
     name_to_col: dict
 
     @classmethod
-    def empty(cls):
+    def empty(cls):  # pragma: no cover
```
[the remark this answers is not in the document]
benbc: Gone.

tests/generative/variable_strategies.py
benbc: If we add a function that needs to be here then we'll get a test failure when we generate a query that get's unhelpfully optimized away by SQLAlchemy. It may be non-obvious that that's the problem, so a generic approach here would be a _bit_ better, but not worth the effort until we see an actual cost to the current approach, I think.

tests/unit/test_query_model.py
benbc: I do this because Dave does it. (Literally -- I looked around in the codebase for something similar just for uniformity's sake.) I think I'd mildly prefer a comment like `# all's well if we got here`, or something, but not enough to want to change it.

tests/lib/in_memory/database.py
benbc: Indeed. And gone now anyway.

tests/lib/in_memory/database.py
benbc: Can't remember. No longer exists.

tests/lib/in_memory/test_database.py
benbc: This has been reverted.

tests/unit/test_query_model.py
evansd: Haha! And there I was feeling cheered in my idiosyncrasies because "Ben does it too".

tests/generative/test_query_model.py
```diff
@@ -302,19 +300,20 @@ def tune_qm_graph_size(variable):
...
-    engine = QueryEngineFixture("in_memory", InMemoryDatabase(), InMemoryQueryEngine)
-    engine.setup(*instances)
-    engine.extract_qm(
-        {
-            "population": Value(True),
-            "v": variable,
-        }
-    )
+    database = InMemoryDatabase()
+    database.setup(*(instantiate(data)))
```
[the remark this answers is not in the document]
benbc: Oh yes, thanks.

databuilder/query_model.py
benbc: I don't understand this. Where do duplicates arise?

evansd: Anytime there's more than one path to a given node. e.g. a Filter operation will always create two paths to its base SelectTable node: one via the `source` and one via the `condition`.

databuilder/query_engines/sqlite.py
benbc: I think this fix pre-dates your switching to functions and I missed it in the merge. Fixed now.

databuilder/query_model.py
evansd: I mean `count_nodes()` still measures _something_ related to the complexity of the query here, and maybe that's all that matters. But it doesn't literally count the nodes.

benbc: These duplicates are equal but non-identical Python objects? I think I don't consider them duplicates in the query model (although they clearly point to the same conceptual _thing_ in the data).

(In fact this is another whole area for later discussion -- I have been secretly wondering if we would produce more efficient SQL if we de-duplicated these by turning the tree into a graph.)

evansd: No, I think these will be the same Python objects appearing multiple times. (Agree about the de-duplication though.)

evansd: Just to save going back and forth on this:
```
In [1]: from databuilder.query_model import *

In [2]: t = SelectTable("foo")

In [3]: f = Filter(t, Function.EQ(SelectColumn(t, "bar"), Value(10)))

In [4]: [id(i) for i in all_nodes(f)]
Out[4]: 
[140582029858608,
 140582046186800,
 140582035503136,
 140581660203136,
 140582046186800,
 140581660204960]
```

benbc: Oh, this _is_ interesting. Whenever we write QM long-hand they are different objects. The generative tests (I'm pretty sure) are generating different objects. So you're talking about erhQL-generated QM?

This seems like maybe-a-problem if our tests are creating different types of data structures to our production code.

And maybe a bit weird that we have graphs which the engine is "reading" as trees.

evansd: Ohhh, I see where you're coming from now. Well I don't think it's going to be a problem in practice because the frozen dataclasses make equivalent objects behave as pretty much identical unless you start using `is` or `id()`, which I don't think we do. But it's worth bearing in mind.

benbc: I think we're agreed that this is vaguely interesting, but doesn't require changes now, is that right?

evansd: Yes, agreed. That function does the right thing in the context of the generated tests.

evansd (approved):

PR #451, opened 2022-04-15
Support aggregations and "picks" in SQLite Query Engine

evansd:
This gets the rest of the spec tests passing by adding support for `PickOneRowPerPatient` and the various `AggregateByPatient` operations.

I think I'm happy enough that the basic logic here is right, though there's still an open question as to how best to structure the code. The current recursive-everywhere approach was easiest to write in the sense that you don't have to think too hard about how everything hangs together; you just yank the thread and out it comes. But by the same token, it's not that clear how everything _does_ hang together and it would be nice if we could make that more obvious.

There's still a fudge around the "patient universe" question but I'm hoping we can resolve that separately, and soon.

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -21,9 +21,9 @@ class SQLiteQueryEngine(BaseQueryEngine):
             for name, definition in variable_definitions.items()
         }
         population_expression = variable_expressions.pop("population")
-        return self.get_results_query(variable_expressions, population_expression)
+        return self.get_results_query(population_expression, variable_expressions)
 
-    def get_results_query(self, variable_expressions, population_expression):
+    def get_results_query(self, population_expression, variable_expressions):
         # Get all referenced tables
         tables = set(get_tables(population_expression))
         for expression in variable_expressions.values():
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -36,7 +36,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
         return self.get_results_query(population_expression, variable_expressions)
 
     def get_results_query(self, population_expression, variable_expressions):
-        query = self.get_patient_select_query()
+        query = self.get_select_query_for_patient_domain()
         query = query.add_columns(
             *[expr.label(name) for name, expr in variable_expressions.items()]
         )
@@ -44,7 +44,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
         query = prepare_query(query)
         return query
 
-    def get_patient_select_query(self):
+    def get_select_query_for_patient_domain(self):
         """
         Return a SELECT query which is to form the basis of a one-row-per-patient query
         """
@@ -170,7 +170,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
     @get_sql.register(PickOneRowPerPatient)
     def get_sql_pick_one_row_per_patient(self, node):
         domain = get_domain(node.source)
-        query = self.get_select_query_from_domain(domain)
+        query = self.get_select_query_for_domain(domain)
 
         # TODO: Really we only want to select the columns from the base table which
         # we're actually going to use. In the old world we did some pre-processing of
@@ -181,7 +181,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
 
         # Add an extra "row number" column to the query which gives the position of each
         # row within its patient_id partition as implied by the order clauses
-        order_clauses = self.get_order_clauses_from_frame(node.source)
+        order_clauses = self.get_order_clauses(node.source)
         if node.position == Position.LAST:
             order_clauses = [c.desc() for c in order_clauses]
         query = query.add_columns(
@@ -208,7 +208,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
 
     def aggregate_by_patient(self, source_node, aggregation_func):
         domain = get_domain(source_node)
-        query = self.get_select_query_from_domain(domain)
+        query = self.get_select_query_for_domain(domain)
         expression = self.get_sql(source_node)
         query = query.add_columns(aggregation_func(expression).label("value"))
         query = query.group_by(query.selected_columns[0])
@@ -233,7 +233,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
         """
         return query.cte()
 
-    def get_select_query_from_domain(self, domain):
+    def get_select_query_for_domain(self, domain):
         """
         Given a Domain object return the SELECT statement that forms the basis of any
         queries using this domain
@@ -243,11 +243,11 @@ class SQLiteQueryEngine(BaseQueryEngine):
             # a single ManyRowsPerPatientFrame which defines its domain. We fetch this
             # and then use it to generate the corresponding query.
             frame = domain.get_node()
-            return self.get_select_query_from_frame(frame)
+            return self.get_select_query_for_frame(frame)
         else:
-            return self.get_patient_select_query()
+            return self.get_select_query_for_patient_domain()
 
-    def get_select_query_from_frame(self, frame):
+    def get_select_query_for_frame(self, frame):
         """
         Given a ManyRowsPerPatientFrame return the corresponding SELECT query with the
         appropriate filter operations applied
@@ -260,7 +260,7 @@ class SQLiteQueryEngine(BaseQueryEngine):
             query = query.where(sqlalchemy.and_(*where_clauses))
         return query
 
-    def get_order_clauses_from_frame(self, frame):
+    def get_order_clauses(self, frame):
         """
         Given a ManyRowsPerPatientFrame return the order_by clauses created by any Sort
         operations which have been applied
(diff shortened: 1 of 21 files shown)


databuilder/query_engines/sqlite.py
```diff
@@ -151,9 +282,101 @@ def engine(self):
...
+def define_placeholder_tables(query):
+    """
+    Where a backend doesn't provide us with a single "patient universe" table guaranteed
+    to contain all patients, we instead generate a "placeholder" table to use instead.
+
+    This function identifies such placeholder tables, generates an appropriate Common
+    Table Expression which defines what the contents of this table should be, and
+    attaches it to the query.
+    """
+    id_column = query.selected_columns[0]
```
benbc, line 300: I think we need the join-on-first-column-by-convention comment here as well.

databuilder/query_engines/sqlite.py
```diff
@@ -151,9 +282,101 @@ def engine(self):
...
+    placeholder_table = id_column.table
```
benbc, line 301: It's not obvious to me why there is always only one placeholder table in this query that needs replacing.

databuilder/query_engines/sqlite.py
```diff
@@ -151,9 +282,101 @@ def engine(self):
...
+    # Any tables referenced by clauses in the query which aren't yet explicity joined on
+    # will be returned as additional FROM clauses by the `get_final_froms()` method
+    implicit_joins = query.get_final_froms()[1:]
+    for table in implicit_joins:
+        query = query.join(table, table.c[join_column] == join_key, isouter=True)
+    return query
+
+
+# TODO: Hopefully temporary workaround for the fact that we don't know what columns
+# we're going to need in our query so we just have to select all of them
```
benbc, line 356: Can you give a hint as to why this is temporary?

databuilder/query_engines/sqlite.py
```diff
@@ -123,14 +156,112 @@ def get_sql_select_column(self, node):
...
+
+    @get_sql.register(PickOneRowPerPatient)
+    def get_sql_pick_one_row_per_patient(self, node):
+        domain = get_domain(node.source)
+        query = self.get_select_query_for_domain(domain)
+
+        # TODO: Really we only want to select the columns from the base table which
+        # we're actually going to use. In the old world we did some pre-processing of
+        # the graph to make this information available at the point we need it. But for
+        # now we just select everything for the base table.
```
benbc, line 221: This seems like a duplicate of the TODO in the function which contains different information. Maybe merge them?

databuilder/query_engines/sqlite.py
```diff
@@ -151,9 +282,101 @@ def engine(self):
...
+    from patient-level tables in a query because, in effect, there is always an implicit
+    join on `patient_id`. This function makes those implicit joins explicit.
+    """
+    # We use the convention that the column to be joined on is always the first selected
+    # column. This avoids having to hardcode, or pass around, the name of the column.
+    join_key = query.selected_columns[0]
+    join_column = join_key.name
+    # Any tables referenced by clauses in the query which aren't yet explicity joined on
+    # will be returned as additional FROM clauses by the `get_final_froms()` method
+    implicit_joins = query.get_final_froms()[1:]
```
benbc, line 349: Why not the first one?

databuilder/query_model.py
```diff
@@ -414,40 +415,58 @@ def validate_input_domains(node):
...
-def get_input_domains(node):
-    return {get_domain(input_node) for input_node in get_input_nodes(node)}
+@dataclasses.dataclass(frozen=True)
+class Domain:
+    lineage: tuple
+
+    def create_descendent(self, node):
+        return Domain(self.lineage + (node,))
+
+    def is_ancenstor(self, other):
```
benbc, line 449:
```suggestion
    def is_ancestor(self, other):
```

databuilder/query_model.py
```diff
@@ -414,40 +415,58 @@ def validate_input_domains(node):
...
+        return self.lineage == other.lineage[: len(self.lineage)]
 
+    # Defining this operator means Domains work naturally with `sorted()`
+    def __lt__(self, other):
+        return self != other and self.is_ancenstor(other)
```
benbc, line 454:
```suggestion
        return self != other and self.is_ancestor(other)
```

databuilder/query_model.py
```diff
@@ -414,40 +415,58 @@ def validate_input_domains(node):
...
             raise DomainMismatchError(
                 f"Attempt to combine unrelated domains:\n{non_patient_domains}"
                 f"\nIn node:\n{node}"
             )
 
 
-def get_input_domains(node):
-    return {get_domain(input_node) for input_node in get_input_nodes(node)}
+@dataclasses.dataclass(frozen=True)
+class Domain:
```
benbc, line 443: This is much clearer.

databuilder/query_engines/sqlite.py
```diff
@@ -151,9 +282,101 @@ def engine(self):
...
-def get_tables(obj):
-    if isinstance(obj, sqlalchemy.sql.ClauseElement):
-        return get_referenced_tables(obj)
+def prepare_query(query):
+    query = define_placeholder_tables(query)
+    query = apply_patient_joins(query)
+    return query
+
+
+def define_placeholder_tables(query):
```
benbc, line 291: Could this be called "replace_..." or "insert_..." or something to make it clear that it's returning a modified query?

databuilder/query_engines/sqlite.py
```diff
@@ -123,14 +156,112 @@ def get_sql_select_column(self, node):
...
     @cache
     def get_sql_select_table(self, node):
         return self.backend.get_table_expression(node.name)
 
+    # We ignore Filter and Sort operations completely at this point in the code and just
+    # pass the underlying table reference through. It's only later, when building the
+    # SELECT query for a given Frame, that we make use of these. This is in order to
+    # mirror the semantics of SQL whereby columns are selected directly from the
+    # underlying table and filters and sorts are handled separately using WHERE/ORDER BY
+    # clauses.
```
benbc, line 173: I'm going to have to take your and Hypothesis' word for it that this works. I can't reason this through sufficiently to convince myself that it's always okay to lift these up to the outside without leaving any trace.

databuilder/query_engines/sqlite.py
```diff
@@ -123,14 +156,112 @@ def get_sql_select_column(self, node):
...
+        query = query.add_columns(
+            sqlalchemy.func.row_number().over(
+                partition_by=query.selected_columns[0], order_by=order_clauses
+            )
+        )
+
+        query = prepare_query(query)
+
+        # Make the above into a subquery and pull out the relevant columns
+        subquery = query.alias()
```
benbc, line 238: Could this use `reify_query()`?

benbc (approved):
I greatly admire the way you were able to work through this large problem as a sequence of smaller, independent problems and in no way had to repeatedly tear everything up and start again. :grinning:

databuilder/query_engines/sqlite.py
evansd: I've updated the comment, does this make more sense?
```
# The table referenced by `join_key`, and any tables already explicitly joined with
# it, will be returned as the first value from the `get_final_froms()` method. Any
# remaining tables which aren't yet explicitly joined on will be returned as
# additional values.
```

databuilder/query_engines/sqlite.py
evansd: These are created by calling `get_select_query_for_patient_domain()` which we can necessarily only ever do once per SELECT query. The _results_ of that query may then get reified and used in another subsequent query, but at that point the placeholder table has already been dealt with.

databuilder/query_engines/sqlite.py
evansd: Yeah, this is all a bit tricky to reason about. I'd say this move was the key thing that unstuck me in writing this code though. It might be easier to talk about it and see if we can come up with a more convincing narrative that we can introduce here.

databuilder/query_engines/sqlite.py
benbc: Oooh. That suggests that there's an implicit invariant here which hadn't occurred to me -- for every non-trivial [1] branch in the at least one [2] of the children is reified.

[1] that is -- those which have non-`Value` children
[2] maybe both?

databuilder/query_engines/sqlite.py
benbc: Yes. Is this documented behaviour by SQLAlchemy?

databuilder/query_engines/sqlite.py
benbc: I wonder if this is another candidate for (imaginary future) pre-translation transformation.

databuilder/query_engines/sqlite.py
benbc: Yes

databuilder/query_engines/sqlite.py
evansd: Yes, I've wondered exactly that too.

databuilder/query_engines/sqlite.py
evansd: I've further updated the comment. I think this code needs all the context it can get.
```
# The table referenced by `join_key`, and any tables already explicitly joined with
# it, will be returned as the first value from the `get_final_froms()` method
# (because `join_key` is the first column). Any remaining tables which aren't yet
# explicitly joined on will be returned as additional clauses in the list. The best
# explanation of SQLAlchemy's behaviour here is probably this:
# https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#change-4737
```

PR #454, opened 2022-04-22
Run legacy tests against SQLite engine

evansd:
This gets (a slightly modified version of) the legacy tests passing against the SQLite query engine. Although this works I'm leaving it as draft because I'm still very unsure about how best to handle the `CombineAsSet` operation. I nevertheless found it a useful exercise to check we could get something passing.


evansd:
Closing as we no longer interested in the legacy test suite. It was still a useful and reassuring exercise to see that the new query engines could pass it. But there's no value in trying to get this merged.

PR #458, opened 2022-04-29
test: More gen test operators

benbc:


databuilder/query_engines/sqlite.py
```diff
@@ -166,14 +166,19 @@ def get_sql_select_column(self, node):
...
+        return self.table_from_backend(node.name)
+
+    # We have to apply caching here otherwise we generate distinct objects representing
+    # the same table and this confuses SQLAlchemy into generating queries with ambiguous
+    # table references. We don't want to cache on the entire node because that holds a
+    # schema as well as the name, but the schema is irrelevant to the query engine and so
+    # should not be considered in the cache.
+    @cache
+    def table_from_backend(self, name):
+        return self.backend.get_table_expression(name)
```
evansd, line 181: I'm happy to leave this in place for now, but I think we'll need an alternative fix further down the road. Although we don't currently use the schema during execution I think that will change. For instance, we may use it to define appropriate SQLAlchemy column types rather than requiring the backend to specify these as we currently do.

I'd argue that really these are mis-generated queries as instances of the same table should logically have the same schema. But at present we have nothing that enforces that and it's not obvious where the best place to do so would be.

tests/generative/data_setup.py
```diff
@@ -81,3 +96,34 @@ def _backend_column(name, type_):
 
 def _backend_table(name, columns):
     return backends.MappedTable(implements=None, source=name, columns=columns)
+
+
+def _build_query(patient_tables, event_tables):
+    # Note that we do not include a schema for the tables here. This is because we are accessing the patient_id
+    # which isn't included in the schema for simplicity elsewhere. There is no downside to not having the schema
+    # because these queries don't mention any other columns, so we don't miss out on possible type-checking.
```
evansd, line 104: I think this reveals something that's been bothering me about the `~patient_id.is_null()` approach which is that `patient_id` doesn't (and _shouldn't_) exist from ehrQL's point of view. And the only reason it works is because we're not using explicit schemas and so the query model just treats it as a vanilla column with type `Any`.

I think the right solution here to support calling `exists_for_patient()` on patient-level frames and then we can drop these references to `patient_id`.

evansd (approved):
Looks good :+1: 

I've left a couple of comments for things to think about later rather than to be addressed in this PR.

databuilder/query_engines/sqlite.py
benbc: Ooh, yes. I meant to flag this as something that I wasn't really happy with. And this is definitely only viable while we're not using the schema in the engine.

I agree with your other comment below which makes the only driver for this go away. I'll create an issue to cover all this.

benbc: https://github.com/opensafely-core/databuilder/issues/462

PR #464, opened 2022-05-04
Generative test tweaks

benbc:


tests/lib/in_memory/test_database.py
```diff
@@ -192,11 +192,28 @@ def test_column_aggregate_values():
...
+        1 |
+        2 | 201
+        2 |
+        """
+    )
+
+    assert c.aggregate_values(sum, default=None) == Column.parse(
+        """
+        1 |
+        2 |
```
evansd, line 216: Ah! It's excellent that Hypothesis is forcing us to spec this out because actually I don't think this is the behaviour we want for `sum`. I remember discussing this at some point and deciding we needed something with the behaviour of SQlite's [`total()`](https://www.sqlite.org/lang_aggfunc.html#sumunc).

But I think the right thing to do is to merge this as-is, and then make a separate PR which changes the spec and the behaviour of both engines.

evansd (approved):

tests/lib/in_memory/test_database.py
benbc: Thank you. That does sound right. I was conscious as I made that change that it's probably not right for other aggregations either. But as you say we might as well get the current behaviour (whatever it is) on a good footing and improve it later. I see you've created an issue -- thanks.

Issue #438, opened 2022-04-01
Straighten out population semantics


benbc:
We need to do some work to clarify the semantics of population definition (and then record that as spec tests).

I've invented some new terminology that I have found helpful when thinking about this:
* The universe is some set of patients who are up for consideration for inclusion in the dataset.
* The population is some (possiblly improper) subset of the universe for whom the population query returns true. (And the dataset contains one row per member of the population.)

The thing that I think needs pinning down is: how is the universe defined. We have three different implementations today:
* `BaseSQLQueryEngine`: all patients mentioned in the exactly-one table referenced by the population query
* `SQLiteQueryEngine`: all patients mentioned in all tables referenced by the population query or all patients metioned in all tables referenced by any variable if there are no tables in the population query
* `InMemoryQueryEngine`: all patients mentioned in any table in the database

(The difference between the second and third of those has given rise to difference sin query results from generative tests -- the in-memory table returns a row for patient mentioned in a table that's not included in the query.)

My strongly-held opinion is that we should make make a deliberate decision and enforce that with a spec test.

My weakly-held opinion is that all three of those definitions are incorrect and that the correct definition of the universe is: all patients mentioned in all tables referenced by any variable.

I've got two reasons for that:
* (A somewhat feeble opinion.) It gets rid of a special case in the implementation (relative to the current `SQLiteQueryEngine`).
* Consider a study where the desired population is "all patients who aren't dead". By the current `SQLiteQueryEngine` definition that gives you... zero patients in your population (because the only table your population query mentions is ONS Deaths and the population query necessarily returns false for all patients in that table). Of course this is a contrived case, but I think it points to a more general problem with only considering patients in the population query.

PR #468, opened 2022-05-06
Support `count` and `exists` aggregations on patient frames

### LINKED ISSUE #462: Support `exists` for patient tables
Researchers may want to ask "is the patient present in this table". Currently the only _general_ (that is, not schema-specific) way to do that is something like `~ t.patient_id.is_null()`. But we don't really expose the patient ids to researchers, so this will seem like a very odd construct to them.

We think that the right way of fulfilling this need is to support `exists_for_patient()` for all `Frame`s, not just `EventFrame`s. This will require support adding all the way down the stack to the query engines.

Once we've added this support we will be able to replace instances of the `~patient_id.is_null` pattern in tests, for instance in the [population definition](https://github.com/opensafely-core/databuilder/blob/main/tests/generative/data_setup.py#L109-L116) in the generative tests. And once we've done that, we can add schemas to the tables in those population definitions (because they no longer need to access the `patient_id` which isn't part of the schema). And doing _that_ will allow us to revert [this change](https://github.com/opensafely-core/databuilder/commit/e8052e43ab6781fd67c3d6d0bcd7f6ece0ecf4e0).

evansd:
We only really _need_ to support `exists` here as that turns out to be very useful when defining populations ("all the patients who exist in the patients table"). But it's conceptually cleaner to treat all the frame aggregations the same, so `count` comes along for the ride too.

With these operations in place we can change all the population definitions to use `exists_for_patient()` rather than `~p.patient_id.is_null` and remove the workarounds this required.

We can then go one step further and remove the ability to reference `patient_id` from ehrQL altogether. As queries in ehrQL are implicitly run over all patients I would argue that `patient_id` is a column from outside the abstraction, which shouldn't be made available.

Closes #462


databuilder/query_language.py
```diff
@@ -147,7 +147,7 @@ def sort_by(self, series):
         )
 
 
-class SortedEventFrame(Frame):
+class SortedEventFrame(EventFrame):
```
benbc, line 146: I don't think this will quite work as you intend. It enables `e.sort_by().sort_by().first_for_patient()` (which you are quite right that we do need), but `e.sort_by().take().first_for_patient()` because `take()` returns an `EventFrame` which has no `first_for_patient()` method.

It would be _reasonable_, I think, to insist that filtering has to happen before sorting in order to enforce uniformity (I think this was the original intention) -- but there's also an argument for maximizing flexibility.

Maybe you could add specs to cover this behaviour to completely pin it down?

databuilder/query_engines/sqlite.py
```diff
@@ -213,13 +220,37 @@ def aggregate_by_patient(self, source_node, aggregation_func):
...
+        # and these cases require different handling.
+        if has_many_rows_per_patient(source_node):
+            # For many-rows-per-patient frames we use the series aggregation code but
+            # wrapped in some extra logic: these aggregation functions don't take an
+            # argument as they operate over the entire query, so we wrap them in a
+            # lambda which swallows the argument
+            value = self.aggregate_series_by_patient(
+                source_node, lambda _: aggregation_func
+            )
+            # These aggregations are also guaranteed not to be null-valued, even when
```
benbc, line 231: I know you haven't changed this comment, but I think I remember this bringing me to a halt before. The way it's phrased means it's ambiguous whether "these aggregations" refers to `aggregation_func` (it doesn't) or the aggregation that this method is being invoked to implement (it does).

benbc (commented):
Good to remove the `IdSeries`.

tests/spec/conftest.py
```diff
@@ -58,7 +58,7 @@ def run_test(table_data, series, expected_results):
         # variable which is the series under test.
         dataset = Dataset()
         # TODO: See temporary hack above
```
benbc, line 60: Is this still TODO? And is the hack still needed? You closed https://github.com/opensafely-core/databuilder/issues/438 the other day.

databuilder/query_language.py
evansd: Good point. Although this change does make ehrQL exactly reflect the semantics of the QM, so in that sense it does achieve what I intended. But the semantics of the QM are weird in that it allows any mixture of "sorts" and "takes" but require that at least one of the "sorts" immediately precedes the final "pick", and I hadn't thought about that enough.

Assuming we can't live with that behaviour I think we have two options. One is to track sortedness dynamically, as a property that is passed through Frame operations, just as we track domains through Series operations. The difficulty here isn't so much the tracking as it is finding a readable and obvious way of expressing the rules as part of the declaration of the QM classes. At the moment we _mostly_ manage to do this with type-annotations, apart from the domain constraints. (I was happier when there was exactly one domain rule that applied everywhere here, but it's not too bad with just the Sort and Filter exceptions.)

This is going to be tricky, but probably we're going to have to handle this anyway because we're going to need to find some way of preventing aggregations being run on patient-series and that's a structurally parallel problem.

Another option is that we stop requiring sorting at all before a pick operation. Given that we're going to need to impose an [implicit sort order](https://github.com/opensafely-core/databuilder/issues/461) anyway then this is still a well-defined operation. And it would obviously be the technically easiest solution. My worry is that it would be easy for users to assume that `first_by_patient()` came with an implicit date ordering if they're allowed to use it without sorting first. But maybe we should allow it in the QM and then have some extra anti-footgun validation in ehrQL?

Finally, I think we're going to need some way of expanding our auto-documenting spec tests to include specifying what is and isn't valid ehrQL. At the moment, I don't think there's any way of doing this.

tests/spec/conftest.py
evansd: I was going to address that in a separate PR. I think some of the spec examples might need changing as they embed incorrect population semantics.

databuilder/query_engines/sqlite.py
evansd: Good point. I've changed to this:
```
            # We also want to guarantee that our frame level aggregations are never
            # null-valued, even when combined with data involving patient_ids for which
            # they're not defined
```

benbc: Thanks

tests/spec/conftest.py
benbc: I think really this comment is a misplaced raised eyebrow in the direction of you closing #438 when I don't think it's really resolved.

databuilder/query_language.py
benbc: Happy for you to take whatever path you like through this thicket.

benbc (approved):

tests/spec/conftest.py
evansd: I think that is accurate!

PR #475, opened 2022-05-09
Remove workaround for spec populations

### LINKED ISSUE #438: Straighten out population semantics

evansd:
This updates the in-memory database to better cope with references to tables which happen not to have any records. We can then change the population definition to "all patients in either the `p` or `e` tables" and remove the workaround code.

Actually closes #438

### DIFF
--- a/tests/spec/conftest.py
+++ b/tests/spec/conftest.py
@@ -26,7 +26,7 @@ def engine(request):
 
 @pytest.fixture
 def spec_test(request, engine):
-    def run_test(table_data, series, expected_results):
+    def run_test(table_data, series, expected_results, population=None):
         # Create SQLAlchemy model instances for each row of each table in table_data.
         input_data = []
         for table, s in table_data.items():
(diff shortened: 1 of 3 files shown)


benbc (approved):
Great.

PR #480, opened 2022-05-10
Refactor ehrQL implementation

evansd:
The current implementation doesn't distinguish between event series and
patient series. But it needs to, because aggregation operations should
be available on event series but not on patient series.

Distinguishing the two requires a different approach from that currently
taken because whether a  method on a `PatientSeries` should return a
`PatientSeries` or an `EventSeries` depends on the dimension of its
arguments and can't be set statically. So instead we ask the query model
what the dimension of the result is, and use that to determine the
appropriate return type.

Given that we have to do this, then it makes to sense to also use the
query model's type inference code to determine the type of the result
and select the appropriate Series class.

This requires that we pass the query model the TableSchema for any
tables we reference. This replaces the role of the `name_to_series_cls`
mapping.

Depends on #481

### DIFF
--- a/databuilder/contracts/base.py
+++ b/databuilder/contracts/base.py
@@ -83,12 +83,14 @@ class TableContract:
                 )
 
     @classmethod
-    def validate_table(cls, table_cls):
-        """Validate that a table is defined with the correct attributes.
+    def validate_schema(cls, schema):
+        """
+        Validate that a schema is defined with the correct columns
+
         This uses asserts rather than raising other exceptions, since any invalid tables
-        must be identified and fixed by the development team.
-        Note that we could use similar logic to generate a table from a contract.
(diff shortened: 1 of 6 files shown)


inglesp (approved):
This looks sensible and self-contained, and as/when we have to add new series types, it looks like it would be obvious how to do this.

My one concern is that we might have lost support for autocomplete -- is this something you've thought about?

evansd:
> My one concern is that we might have lost support for autocomplete -- is this something you've thought about?

Yes, this is a concern and something I've thought about a bit. But I think I need to do a bit more investigation into what, say, VSCode's python integration gives you in the way of autocomplete to know for certain what the effect will be.

I think we'd already come to the reluctant conclusion that trying to express everything statically use Python's type system isn't really feasible. But I think that's a separate question from helping IntelliSense to make sensible guesses.

PR #486, opened 2022-05-11
Remove patient-level aggregation support and refactor

evansd:
The changes in #480 removed the ability to aggregate patient-level series from ehrQL, but these were still legal constructs in the query model. This PR adds extra validation rules to the query model to disallow these, and then removes a lot of quite gnarly code from SQLiteQueryEngine which turns out only to be needed for this purpose.

As mentioned in the [relevant commit](https://github.com/opensafely-core/databuilder/pull/486/commits/b068a120f4a0cd70f46de9fab88fb8d5cba8b13e), there's scope for improving the way this is implemented in the query model but I don't think it needs tackling now.

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -68,7 +68,11 @@ class SQLiteQueryEngine(BaseQueryEngine):
             # If there's only one table then use the IDs from that
             return sqlalchemy.select(tables[0].c.patient_id)
         else:
-            assert False, "No tables referenced in population expression"
+            # Gracefully handle the degenerate case where the population expression
+            # doesn't reference any tables at all. Our validation rules ensure that such
+            # an expression will never evaluate True so the population will always be
+            # empty. But we can at least return an empty result, rather than blowing up.
+            return sqlalchemy.select(sqlalchemy.literal(None).label("patient_id"))
 
     @singledispatchmethod
     def get_sql(self, node):
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -1,11 +1,9 @@
 import contextlib
-import secrets
 from functools import cache, cached_property, singledispatchmethod
 
 import sqlalchemy
 from sqlalchemy.dialects.sqlite.pysqlite import SQLiteDialect_pysqlite
 from sqlalchemy.sql import operators
-from sqlalchemy.sql.visitors import replacement_traverse
 
 from databuilder.query_model import (
     AggregateByPatient,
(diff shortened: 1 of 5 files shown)


databuilder/query_model.py
```diff
@@ -443,6 +452,10 @@ def validate_input_domains(node):
                 f"Attempt to combine unrelated domains:\n{non_patient_domains}"
                 f"\nIn node:\n{node}"
             )
+        if isinstance(node, AggregatedSeries) and len(non_patient_domains) == 0:
+            raise DomainMismatchError(
+                f"Attempt to aggregate one-row-per-patient series\nIn node:\n{node}"
+            )
```
benbc, line 458: This is sufficiently grimace-inducing to motivate something along the lines that you proposed earlier. But totally okay to defer it.

databuilder/query_engines/sqlite.py
```diff
@@ -68,7 +68,11 @@ def select_patient_id_for_population(self, population_expression):
             # If there's only one table then use the IDs from that
             return sqlalchemy.select(tables[0].c.patient_id)
         else:
-            assert False, "No tables referenced in population expression"
+            # Gracefully handle the degenerate case where the population expression
+            # doesn't reference any tables at all. Our validation rules ensure that such
+            # an expression will never evaluate True so the population will always be
+            # empty. But we can at least return an empty result, rather than blowing up.
+            return sqlalchemy.select(sqlalchemy.literal(None).label("patient_id"))
```
benbc, line 75: What's the reason for handling this? I think it means that we support a population of `Value(False)`, but not one of `Value(True)`, which seems odd.

databuilder/query_engines/sqlite.py
```diff
@@ -327,57 +315,6 @@ def engine(self):
         return engine
 
 
-def prepare_query(query):
-    query = replace_placeholder_table_references(query)
-    query = apply_patient_joins(query)
-    return query
-
-
-def replace_placeholder_table_references(query):
```
benbc, line 336: Oh, glory be.

benbc (approved):

databuilder/query_engines/sqlite.py
evansd: Yes, you're right that we support `Value(False)` but not `Value(True)` and that is a bit odd. It would be reasonable, I think, to reject any population expression which is always False regardless of its inputs. _However_ detecting that class of expression requires a whole load more code similar to, but not the same as, the `evaluate` functions in the current population_validation module. Whereas gracefully handling it like this requires one line.

The alternative is not to handle it and blow up, but I wanted to preempt the smug undertones in Hypothesis's voice — "well, you did _say_ this was a valid population ..."

benbc: Yes, I think you're right that we would need to add the validation. I suppose that's the ideal, most-friendly-possible outcome (on the assumption that any population that always evaluates to `False` must be an error which should be caught as early as possible). But I don't have any reservations about this going in as it is now.

PR #489, opened 2022-05-13
Rudimentary codelist support

evansd:
This adds basic support for querying using codes. We add individual classes for each coding system. This allows us to use the type system to enforce that the right coding system is used.

There's obviously lots more to be done here, but I'd appreciate feedback on the general approach before we go too much further with it.

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -80,7 +80,17 @@ class SQLiteQueryEngine(BaseQueryEngine):
 
     @get_sql.register(Value)
     def get_sql_value(self, node):
-        return sqlalchemy.literal(node.value)
+        if isinstance(node.value, frozenset):
+            value = tuple(self.convert_value(v) for v in node.value)
+        else:
+            value = self.convert_value(node.value)
+        return sqlalchemy.literal(value)
+
+    def convert_value(self, value):
+        if hasattr(value, "_to_primitive_type"):
+            return value._to_primitive_type()
+        else:
+            return value
 
     @get_sql.register(Function.EQ)
     def get_sql_eq(self, node):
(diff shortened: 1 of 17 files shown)


databuilder/query_language.py
```diff
@@ -54,6 +55,11 @@ def is_null(self):
         return _apply(qm.Function.IsNull, self)
 
     def is_in(self, other):
+        # The query model requires a Set type (and an immutable one at that) for
+        # containment queries, but that's a bit user-unfriendly so we accept
+        # lists/tuples etc here and convert them
+        if isinstance(other, Iterable) and not isinstance(other, (str, bytes)):
+            other = frozenset(other)
```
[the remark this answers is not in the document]
benbc: I do think that loosening the types accepted is a good idea.

databuilder/query_engines/sqlite.py
```diff
@@ -80,7 +82,15 @@ def get_sql(self, node):
 
     @get_sql.register(Value)
     def get_sql_value(self, node):
-        return sqlalchemy.literal(node.value)
+        return sqlalchemy.literal(self.convert_value(node.value))
+
+    def convert_value(self, value):
+        if isinstance(value, BaseCode):
+            return value.value
```
benbc, line 89: Another alternative to mentioning `BaseCode` here would be to maintain a registry of mappings that was populated from outside.

benbc: I don't feel strongly which of the three options you take, but we should keep an eye on this and consider refactoring later.

databuilder/query_language.py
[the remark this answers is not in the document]
evansd: You're quite right. I'm not sure what I was thinking here. I have amended.

databuilder/query_engines/sqlite.py
evansd: OK, that sounds sensible. I just wanted to make sure there wasn't some Obviously Correct approach I was missing.

inglesp (approved):

PR #490, opened 2022-05-13
Gph 5765

chrisyeunggraphnet:
We have merged the cureent main branch to this branch today. The major changes are added 2 env variables "GRAPHNET_DB_SCHEMA" and "ENABLE_ISOLATION_LEVEL_AUTOCOMMIT" to allow dynamic schmea and set isolation level to autocommit for sypnase. This 2 env variables will be supplied by the job runner when the graphnet backend is used.


benbc:
We no longer need this as Synapse isn't being used by Graphnet.

PR #495, opened 2022-05-17
Fix some sorting issues

evansd:
This fixes two issues:
 * prevents sorting by things which aren't orderable (like booleans);
 * defines a sort order for NULLs and fixes the in-memory engine to handle them.

It also refactors the SQLite implementation to only select the columns that are actually going to be used in the query. This was done not so much for the performance gains (though we would certainly need to do this before going live anyway) but because it means we no longer rely on the backend to tell us the available columns. And this paves the way for a simpler API, and the ability to run without requiring a backend at all.

Sadly, it doesn't entirely put our sorting issues to bed (see [this comment](https://github.com/opensafely-core/databuilder/issues/461#issuecomment-1129076022)); though I did some now wasted work in the belief that it would.

### DIFF
--- a/databuilder/query_model_transforms.py
+++ b/databuilder/query_model_transforms.py
@@ -1,0 +1,97 @@
+"""
+Apply modifications to the query graph which make it easier to work with or allow us to
+generate more efficient SQL.
+
+This involes adding new kinds of nodes to the query model. However we deliberately
+define these nodes outside of the core query_model module because we're trying to
+maintain separation of three types of concern:
+
+    1. capturing the semantics of the query (which is the query model's job);
+    2. worrying about expressiveness and ease of use (which is ehrQL's job);
+    3. worrying about efficient execution (which is the query engines' job).
+
+The transformations applied here are all about efficient execution, and therefore we
+want to keep them separate from the core query model classes.
+"""
+import copy
+from typing import Any
+
+from databuilder.query_model import PickOneRowPerPatient, SelectColumn, get_input_nodes
+
+
+class PickOneRowPerPatientWithColumns(PickOneRowPerPatient):
+    selected_columns: frozenset[Any]
+
+
+def apply_transforms(variables):
+    # For algorithmic ease we're going to be mutating the query objects, so we make a
+    # copy first to avoid side-effects
+    variables = copy.deepcopy(variables)
+    reverse_index = get_reverse_index(variables.values())
+    add_selected_columns_to_pick_row(reverse_index)
+    return variables
+
+
+def get_reverse_index(nodes):
+    """
+    Return a dict mapping every node in the query to the (possibly empty) set of nodes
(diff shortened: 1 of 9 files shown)


tests/integration/test_query_engines.py
```diff
@@ -80,7 +80,9 @@ def test_multiple_takes_without_chaining(engine):
     }
 
 
-def test_handles_degenerate_poppulaton(engine):
+def test_handles_degenerate_population(engine):
```
inglesp, line 83: Fascinating typo.

tests/spec/sort_and_pick/test_sort_by_column_with_nulls_and_pick.py
```diff
@@ -0,0 +1,38 @@
...
+          |  i1 | i2
+        --+-----+----
+        1 |     | 1
+        1 | 102 | 2
+        1 | 103 | 3
+        2 | 203 | 6
+        2 | 202 | 5
+        2 |     | 4
+        """,
+}
```
inglesp, line 16: Do we need `i2` here?

databuilder/query_model_transforms.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+import copy
+import dataclasses
+from typing import Any
+
+from databuilder.query_model import PickOneRowPerPatient, SelectColumn, get_input_nodes
+
+
+class PickOneRowPerPatientWithColumns(PickOneRowPerPatient):
+    selected_columns: frozenset[Any]
```
inglesp, line 13: What's the rationale for this class living here and not in `query_model.py`?

inglesp (approved):

tests/spec/sort_and_pick/test_sort_by_column_with_nulls_and_pick.py
evansd: No, I suppose we don't do we? I'm not sure now why I thought we did. I think it's worth the effort to make these specs as minimal as possible, so I'll remove.

databuilder/query_model_transforms.py
evansd: Ah good question, and something I should have made more effort to argue for. As I see it, the only job of the query model itself is to capture the semantics of the query. It shouldn't be concerned with ease of expression or readability (which is ehrQL's job) or ease and efficiency of execution (which is the query engine's job). So these kinds of transformation to the query graph are a query engine implementation detail and shouldn't be mixed up in the core query model.

I anticipate us adding quite a few more of these kinds of query-engine-only node types in future to handle things like codelist tables.

Does that help? And do you agree? I'll expand the comments in the code if so.

inglesp: > Does that help? And do you agree?

yes * 2

PR #496, opened 2022-05-17
Begin backend refactor

evansd:
This begins the process of unwinding the slightly muddled relationship between Query Engines and Backends. In particular:
 * database connection details are now passed directly to the Query Engine, rather than via the Backend;
 * the dataset definition is no longer passed to the Query Engine constructor, but rather to the `execute_query()` method


inglesp (approved):
Sensible ratcheting.

PR #497, opened 2022-05-18
Support categorised codelists

evansd:
This adds support for mapping codes to categories using a categorised codelist and, as a side effect, for mapping a series of any arbitrary type from one set of values to another.

Behind the scenes this is implemented using a CASE expression. This involved adding support for arbitrary CASE expressions to the query model, the SQLite query engine, and the in-memory query engine. However we don't yet expose this in ehrQL — just the ability to map from one fixed set of values to another.

It may well be that doing this with CASE expressions is horribly inefficient, at least in some circumstances for some databases, so we may need to switch to using a temporary table and a JOIN. However it should be possible to implement this as a separate optimisation, for which I'll write up a ticket to track.

Please note the final, speculative commit and the question raised in the commit message.

### DIFF
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -55,9 +55,31 @@ class SNOMEDCTCode(BaseCode, system_id="snomedct"):
     "SNOMED CT"
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass()
 class Codelist:
-    codes: frozenset
+    codes: set[BaseCode]
+    category_maps: dict[str, dict[BaseCode, str]]
+
+    def __post_init__(self):
+        # Check that the same codes are used everywhere
+        for category_map in self.category_maps.values():
+            assert self.codes == category_map.keys()
+        # In general, we want categorisations to be accessed as attributes on the
+        # codelist e.g.
+        #
+        #   codelist.group_6_ethnicity
+        #
+        # But in case there are names which aren't valid Python identifiers (or which
+        # clash with other instance attributes) it's always possible to use the
+        # dictionary directly e.g.
+        #
+        #  codelist.category_maps["Group 6 Ethnicity"]
+        #
+        for name, category_map in self.category_maps.items():
+            # Avoid names which could be magic attributes or blat an already existing
+            # attribute
+            if not name.startswith("_") and not hasattr(self, name):
+                setattr(self, name, category_map)
 
 
 def codelist_from_csv(filename, column, system):
@@ -82,8 +104,13 @@ def codelist_from_csv_lines(lines, column, system):
     if column not in reader.fieldnames:
         raise CodelistError(f"No column '{column}' in CSV")
     codes = set()
+    # We treat every other column in the CSV as being a mapping from codes to categories
+    category_maps = {name: {} for name in reader.fieldnames if name != column}
     for row in reader:
         code_str = row[column].strip()
         if code_str:
-            codes.add(code_class(code_str))
-    return Codelist(frozenset(codes))
+            code = code_class(code_str)
+            codes.add(code)
+            for name, category_map in category_maps.items():
+                category_map[code] = row[name].strip()
+    return Codelist(codes, category_maps)
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -55,7 +55,20 @@ class SNOMEDCTCode(BaseCode, system_id="snomedct"):
     "SNOMED CT"
 
 
+@dataclasses.dataclass(frozen=True)
+class Codelist:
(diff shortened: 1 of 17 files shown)


tests/spec/code_series_ops/test_containment.py
```diff
@@ -27,3 +27,27 @@ def test_is_in(spec_test):
...
+            "abc",
+            "ghi",
+        ],
+        column="code",
+        system="snomedct",
+    )
+
+    spec_test(
+        table_data,
+        p.c1.is_in(codelist),
```
inglesp, line 46: Might need to do some work to get this to render sensibly in docs.  An escape hatch might be best here?

databuilder/codes.py
```diff
@@ -55,9 +55,30 @@ class SNOMEDCTCode(BaseCode, system_id="snomedct"):
...
-    codes: frozenset
+    codes: set[BaseCode]
+    category_maps: dict[str, dict[BaseCode, str]]
+
+    def __post_init__(self):
+        # Check that the same codes are used everywhere
+        for category_map in self.category_maps.values():
+            assert self.codes == category_map.keys()
+        # As a convenience we set every "reasonably" named categorisation as an
+        # attribute on the codelist so you can write e..g
```
inglesp, line 68: ...

tests/unit/test_codes.py
```diff
@@ -27,3 +27,21 @@ def test_codelist_from_csv_unknown_system(tmp_path):
     csv_file.touch()
     with pytest.raises(CodelistError, match="not_a_real_system"):
         codelist_from_csv(csv_file, "CodeID", "not_a_real_system")
+
+
+def test_codelist_from_csv_with_categories(tmp_path):
+    csv_file = tmp_path / "codes.csv"
+    csv_file.write_text("CodeID,cat1,__str__\nabc,123,foo\ndef,456,bar,\n ghi ,789\n,")
```
inglesp, line 34: Use a multiline string to make it easier to read?

inglesp (approved):
Great stuff.

tests/lib/in_memory/database.py
```diff
@@ -344,3 +302,58 @@ def parse_value(value):
 def nulls_first_order(key):
     # Usable as a key function to `sorted()` which sorts NULLs first
     return (0 if key is None else 1, key)
+
+
+def apply_function_to_columns(fn, *columns):
```
inglesp, line 307: When we handle a `Case` node, we're not passing in columns.  Could we call this `apply_function`?

inglesp (approved):

tests/lib/in_memory/database.py
```diff
@@ -344,3 +302,58 @@ def parse_value(value):
...
+        default=fn(*[column.default for column in reshaped_columns]),
+    )
+
+
+def reshape_columns(column, *other_columns):
+    reshaped_others = []
+    for other in other_columns:
+        column, other = reshape_column_pair(column, other)
+        reshaped_others.append(other)
+    return [column] + reshaped_others
```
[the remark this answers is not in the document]
evansd: Gah! You're quite right. I don't think this will work. I naively hoped I could convert the existing pairwise code into something more general without having to actually understand what it was doing. I think it's back to the drawing board with this.

tests/integration/test_query_engines.py
```diff
@@ -89,3 +92,37 @@ def test_handles_degenerate_population(engine):
...
+        v=case_operation,
+    )
+
+    results = {r["patient_id"]: r["v"] for r in engine.extract_qm(variables)}
+    assert results == {
+        1: "less_than_two",
+        2: "equals_two",
+        3: "greater_than_two",
+        4: default,
+    }
```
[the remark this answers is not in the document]
evansd: Yes, at this point in history they're not expressible in ehrQL so they have to be in here. But I meant to remove them when I added `case` in the next PR and I forgot. I'll amend the second PR.

tests/lib/in_memory/database.py
evansd: Fine by me!

tests/lib/in_memory/database.py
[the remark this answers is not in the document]
evansd: I _think_ I've got this right now

databuilder/codes.py
evansd: I improved the comment here to better express my intentions.

tests/spec/code_series_ops/test_containment.py
evansd: That sounds sensible. I'm not sure what the means concretely though. Is it something I need to think about now?

PR #498, opened 2022-05-19
Support arbitrary case expressions in ehrQL

evansd:
Happy to bikeshed the syntax a bit more if people would like to.

It strikes me that one advantage of the old syntax, where the values
preceeded the conditions, was that the values then lined up nicely at
the beginning so you could scan them more easily. We could support that
as well/instead of what I've got here easily enough, it's just tricky to
find the right name for the syntactic sugar class because all the
natural ones are Python keywords. Perhaps something like:
```py
case(
  has("abc").when(i1 > 10),
  has("def").when(i1 > 20),
  default="ghi",
)
```

Depends on #497

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -374,3 +374,31 @@ def build_event_table(name, schema, contract=None):
     return EventFrame(
         qm.SelectTable(name, schema=qm.TableSchema(schema)),
     )
+
+
+# CASE EXPRESSION FUNCTIONS
+#
+
+
+# TODO: There's no explicit error handling on using this wrong e.g. not calling `then()`
+# or passing the wrong sort of thing as `condition`. The query model will prevent any
+# invalid queries being created, but we should invest time in making the errors as
+# immediate and as friendly as possible.
+class when:
+    def __init__(self, condition):
+        self._condition = condition
+
+    def then(self, value):
+        new = self.__class__(self._condition)
+        new._value = value
+        return new
+
+
+def case(*when_thens, default=None):
+    cases = _DictArg((case._condition, case._value) for case in when_thens)
+    # If we don't want a default then we shouldn't supply an argument, or else it will
+    # get converted into `Value(None)` which is not what we want
+    if default is None:
+        return _apply(qm.Case, cases)
+    else:
+        return _apply(qm.Case, cases, default)
--- a/tests/integration/test_query_engines.py
+++ b/tests/integration/test_query_engines.py
@@ -1,21 +1,18 @@
-import pytest
 import sqlalchemy
 
 from databuilder.query_model import (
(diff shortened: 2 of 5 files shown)


sebbacon:
On bikeshedding - how about trying it out on some of our researchers?

evansd:
On reflection, I withdraw my offer to bikeshed :)

This is trivial to change; nothing is set in stone until we have researchers using it in anger. Let's just get the basic features in and sweat the syntax later.

databuilder/query_language.py
```diff
@@ -374,3 +374,25 @@ def build_event_table(name, schema, contract=None):
...
+
+
+class when:
+    def __init__(self, condition):
+        self._condition = condition
+
+    def then(self, value):
+        new = self.__class__(self._condition)
+        new._value = value
+        return new
```
inglesp, line 390: What happens if `.then()` isn't called?

databuilder/query_language.py
```diff
@@ -374,3 +374,25 @@ def build_event_table(name, schema, contract=None):
...
+
+
+def case(*when_thens, default=None):
+    cases = _DictArg((case._condition, case._value) for case in when_thens)
+    # If we don't want a default then we shouldn't supply an argument, or else it will
+    # get converted into `Value(None)` which is not what we want
+    default_arg = [default] if default is not None else []
+    return _apply(qm.Case, cases, *default_arg)
```
inglesp, line 398: I'd be tempted to make this really explicit with an `if` statement.

inglesp (approved):
Neat, but needs a little bit of work to make more robust.

databuilder/query_language.py
evansd: It will blow up with an attribute error, and I think that's fine at the moment. We're explicitly not worrying about user friendly error messages at this stage. See e.g. this line whose author may remain nameless:
https://github.com/opensafely-core/databuilder/blob/f89de97622f1a82b232d87bf2e0bb6dc19c12038/databuilder/query_language.py#L28

The query model itself ensures you can't build invalid queries. There's a whole piece of UX work here which will involve catching errors from the query model and translating them into something more "ehrQL appropriate"; and also preventing certain kinds of errors before they make it to the query model.

You're right that I should stick another TODO comment in here, just so it's easy to find places we know need more thought.

databuilder/query_language.py
inglesp: Or even:

```
if default is None:
    return _apply(qm.Case, cases)
else:
    return _apply(qm.Case, cases, default)
```

?

evansd: Ha, oh yes! For some reason I was blind to the possibility of writing the function call twice. That's much simpler in this case.

PR #499, opened 2022-05-19
Support more date operations

evansd:
This adds support for the remaining date operations which I think we'll need for our first study. As with all these PRs, I'm not particularly wedded to the precise method names or syntax here. But it's all pretty easily changed.

Two things in particular would be worth some future discussion. One is the idea of having some sort of `DateDelta` series which you'd create by subtracting one date from another and then access a `.years` attribute on it to get the difference in years.

I think that would be relatively easy to implement, and might be a nice approach; but I was slightly concerned about adding a series-like thing to ehrQL that wasn't _actually_ a series i.e. you can't pass it directly into anything else, you have to use one its units attributes. I also wasn't sure how nicely the syntax would work with method chaining because you'd have to use parens before you can get the attribute e.g.
```py
age_at_event = (events.take(events.code.is_in(codelist).date - patients.date_of_birth).years
```

That's not to say it's definitely a bad idea. I just wasn't immediately excited enough about it to want to implement it now :)

The other thing to think about is whether to add `+`/`-` operator support. I worried a bit about making the units explicit here. Maybe the operators should only accept `timedelta()` instances (with the time component set to zero)?

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -223,6 +223,24 @@ class SQLiteQueryEngine(BaseQueryEngine):
         part_as_str = sqlalchemy.func.strftime(format_str, date)
         return sqlalchemy.cast(part_as_str, sqlalchemy_types.Integer())
 
+    @get_sql.register(Function.DateAddDays)
+    def get_sql_date_add_days(self, node):
+        return self.date_add_days(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    @get_sql.register(Function.DateSubtractDays)
+    def get_sql_date_subtract_days(self, node):
+        return self.date_subtract_days(self.get_sql(node.lhs), self.get_sql(node.rhs))
+
+    def date_add_days(self, date, num_days):
+        num_days_str = sqlalchemy.cast(num_days, sqlalchemy_types.String())
+        modifier = num_days_str.concat(" days")
+        new_date = sqlalchemy.func.date(date, modifier)
+        # Tell SQLAlchemy that the result is a date without doing any CASTing in the SQL
+        return sqlalchemy.type_coerce(new_date, sqlalchemy_types.Date())
+
+    def date_subtract_days(self, date, num_days):
+        return self.date_add_days(date, -num_days)
+
     @get_sql.register(SelectColumn)
     def get_sql_select_column(self, node):
         source = self.get_sql(node.source)
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -168,6 +168,56 @@ class SQLiteQueryEngine(BaseQueryEngine):
     def get_sql_day_from_date(self, node):
         return self.get_date_part(self.get_sql(node.source), "DAY")
 
+    @get_sql.register(Function.DateDifferenceInYears)
+    def get_sql_date_difference_in_years(self, node):
+        return self.date_difference_in_years(
+            self.get_sql(node.lhs), self.get_sql(node.rhs)
+        )
+
+    def date_difference_in_years(self, start, end):
+        year_diff = self.get_date_part(end, "YEAR") - self.get_date_part(start, "YEAR")
+        month_diff = self.get_date_part(end, "MONTH") - self.get_date_part(
+            start, "MONTH"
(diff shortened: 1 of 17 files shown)


databuilder/query_engines/sqlite.py
```diff
@@ -168,6 +168,25 @@ def get_sql_month_from_date(self, node):
...
+        year_d = self.get_date_part(end, "YEAR") - self.get_date_part(start, "YEAR")
+        month_d = self.get_date_part(end, "MONTH") - self.get_date_part(start, "MONTH")
+        day_d = self.get_date_part(end, "DAY") - self.get_date_part(start, "DAY")
+        # The CASE condition below is a simplification of:
+        #
+        #    end_month < start_month OR (end_month == start_month AND end_day < start_day)
+        #
+        # which doesn't involve multiple references to the "get month" function. The
+        # coefficient here is arbitrary as long as it's at least 31 — we use a power of
+        # two for neatness.
```
benbc, line 187: This is quite a thing. Is it a standard trick for comparing dates?

I had to do quite a lot of work to convince myself that it's correct, so I think a more detailed explanation is in order. I'll try to draft something based on my (newly acquired) understanding of what's going on. I'm assuming changes to address my two naming comments above, but these three suggestions are all orthogonal.

I also think that using 32 as the multiplier is misleading because it's hard to shake the belief that it being `31 + 1` is necessary. I'd prefer that we either use 31 or some much larger number. I'll fold an assumption for changing this into my proposal too, but again this is orthogonal.

> The CASE condition below detects the situation where the end day-of-year is earlier than the start day-of-year. In such cases we need to subtract one from the year delta to give the number of whole years that has elapsed between the two dates.
> 
> The most obvious way to determine this duplicates calls to the "get month" function:
>
>     end_month < start_month OR (end_month == start_month AND end_day < start_day)
>
> or equivalently:
> 
>     delta_month < 0 OR (delta_month == 0 AND delta_day < 0)
>
> However this expression has a simplified form that takes advantage of the fact that `delta_day` has a maximum magnitude of `30`:
> 
>     delta_month * -31 > delta_day
>
> To satisfy yourself that these expressions are equivalent, consider this truth table:
>
>      delta_month | delta_day | result
>     =============|===========|========
>          +ve     |    any    | false
>     -------------|-----------|--------
>                  |    +ve    | false
>                  |-----------|--------
>           0      |     0     | false
>                  |-----------|--------
>                  |    -ve    | true
>     -------------|-----------|--------
>          -ve     |    any    | true
>
> By inspection, this truth table is correct for both expressions and therefore they are equivalent.

tests/spec/date_series_ops/test_date_comparisons.py
```diff
@@ -19,7 +17,7 @@
 def test_is_before(spec_test):
     spec_test(
         table_data,
-        p.d1.is_before(date(2000, 1, 1)),
+        p.d1.is_before("2000-01-01"),
```
benbc, line 20: I think we should probably have a separate spec that demonstrates that we can take both types -- for documentation purposes.

benbc (approved):

databuilder/query_engines/sqlite.py
```diff
@@ -168,6 +168,25 @@ def get_sql_month_from_date(self, node):
...
         return self.get_date_part(self.get_sql(node.source), "DAY")
 
+    @get_sql.register(Function.DateDifferenceInYears)
+    def get_sql_date_difference_in_years(self, node):
+        return self.date_difference_in_years(
+            self.get_sql(node.lhs), self.get_sql(node.rhs)
+        )
+
+    def date_difference_in_years(self, start, end):
+        year_d = self.get_date_part(end, "YEAR") - self.get_date_part(start, "YEAR")
```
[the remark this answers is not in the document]
evansd: I used `diff` originally but I was overcome by an irrational desire to get it all on one line! I will amend my ways, and the code.

tests/spec/date_series_ops/test_date_comparisons.py
evansd: You're quite right, we should. I think I hesitated because I couldn't decide whether I ought to duplicate the entire suite with both types or not; so in the end I did nothing. Do you have a view?

PR #501, opened 2022-05-20
Add MSSQL Query Engine

evansd:
This adds a query engine for Microsoft SQL Server which passes the spec tests. Apart from straightforward specialising of date functions the main change this requires was making a distinction between predicates and other boolean-typed expressions (probably best explained by looking at the code).


databuilder/query_engines/base_sql.py
```diff
@@ -83,6 +83,10 @@ def select_patient_id_for_population(self, population_expression):
     def get_sql(self, node):
         assert False, f"Unhandled node: {node}"
 
+    @singledispatchmethod
+    def get_table(self, node):
+        assert False, f"Unhandled node: {node}"
+
```
benbc, line 111: This seems reasonable. Did you consider renaming the other type to `get_column`?

benbc (commented):
I think I was expecting the predicate business to only happen for the MSSQL engine. Since it is compatible with both it's not a problem though.

Have you considered trying the generative tests against MSSQL? May just be too slow, but we could perhaps have a mode to do that which we only run overnight, or soething.

databuilder/query_engines/base_sql.py
evansd: `get_column` wouldn't be _quite_ right in this case. Because `get_table` really does return a reference to a table, but the thing currently called `get_sql` returns an arbitrarily complex expression that can used in places a column reference would be.

`get_sql` is obviously not a great name, but I found it hard to think of the right term if we're going to use `get_expr` to mean "thing which is not a predicate". I think my grasp on the terminology has got a bit shaky. There might be better terms for all the things somewhere.

evansd:
> Have you considered trying the generative tests against MSSQL? May just be too slow, but we could perhaps have a mode to do that which we only run overnight, or soething.

Yes, I did try this but it blew up after not too long after running out of file descriptors! And I didn't have time to investigate right then, and I meant to look into it again this week but haven't had the chance. We should definitely do this though.

Issue #457, opened 2022-04-25
Add missing spec for `sum_for_patient()`


benbc:
It's implicitly specified in other specs (e.g. for drop), but doesn't have its own spec.

PR #503, opened 2022-05-26
Add minimum/maximum series aggregations

evansd:
I realised I had made use of these when spiking the CIPHA study as they turn out to be quite a natural way of expressing certain sorts of query.

We also take the opportunity to add a spec for `sum_for_patient()` as per #457.


benbc (approved):
These need adding to the generative tests' variable strategies. They are there already, I think, but commented out with a TODO.

benbc (approved):

PR #504, opened 2022-05-27
Factor out ORM class auto-generation

evansd:
This factors out the code which automatically generates a matching SQLAlchemy ORM class from a table schema into a reusable method and then uses it to reduce duplication in the spec test setup.

This is particularly useful when constructing large tests with realistic schemas as I am currently doing.

We also change the spec test parser to use schema types when parsing tables. This seems to me more robust and flexible than relying on the first character of the column name. And it means we can add new types to the spec tests without needing to specifically teach the parser about them.

### DIFF
--- a/tests/generative/data_setup.py
+++ b/tests/generative/data_setup.py
@@ -23,8 +23,6 @@ def setup(schema, num_patient_tables, num_event_tables):
     all_patients_query = _build_query(patient_table_names, event_table_names, schema)
 
     return (
-        # We don't currently support different names for this column
-        "patient_id",
         patient_classes,
         event_classes,
         all_patients_query,
--- a/tests/generative/data_setup.py
+++ b/tests/generative/data_setup.py
@@ -1,4 +1,3 @@
-import sqlalchemy
 import sqlalchemy.orm
 
 from databuilder.query_model import (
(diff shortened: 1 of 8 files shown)


tests/generative/data_setup.py
```diff
@@ -7,59 +6,43 @@
...
-        "e", num_event_tables, schema, patient_id_column, registry
+        "e", num_event_tables, schema, base_class
     )
 
     all_patients_query = _build_query(patient_table_names, event_table_names, schema)
 
     return (
-        patient_id_column,
+        # We don't currently support different names for this column
+        "patient_id",
```
benbc, line 27: Don't we? Isn't this the backend patient join column?

benbc (approved):
This is great. Just one query.

tests/generative/data_setup.py
evansd: Ah yes, there's a useful clarification to made here. We support changing this in the SQL schema (for the `patient_join_column` attribute as you say) but in the logical schema it's always called `patient_id`. And where we're generating ORM classes from the logical schema we're implicitly guaranteeing that the two match.

I think it would be better actually to refactor this code so we don't return this value and just have it set as a module level constant. I'll do that now.

benbc (commented):
Fine.

I think I was thrown off by the wording "we don't support" which I took to mean that Data Builder doesn't support it, whereas you actually mean that we don't support it in the tests. This is better anyway.

benbc (approved):

PR #539, opened 2022-06-08
feat: adding narrative to reference docs

Jongmassey:
resolves https://github.com/opensafely/documentation/issues/772

### DIFF
--- a/databuilder/docs/specs.py
+++ b/databuilder/docs/specs.py
@@ -26,14 +26,18 @@ def build_chapter(chapter_id, package_name, module_names):
     """
 
     package = import_module(f"tests.spec.{package_name}")
-    return {
-        "id": chapter_id,
-        "title": package.title,
-        "sections": [
-            build_section(f"{chapter_id}.{ix + 1}", package_name, module_name)
-            for ix, module_name in enumerate(module_names)
-        ],
-    }
+    text = getattr(package, "text", None)
+    return concatenate_optional_text(
+        {
+            "id": chapter_id,
+            "title": package.title,
+            "sections": [
+                build_section(f"{chapter_id}.{ix + 1}", package_name, module_name)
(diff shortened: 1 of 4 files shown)


databuilder/docs/specs.py
```diff
@@ -93,13 +101,26 @@ def build_paragraph(paragraph_id, test_fn):
...
+        },
+        text,
+    )
+
+
+def concatenate_optional_text(dictionary, text):
+    return dict(
+        dictionary,
+        **{"text": text} if text else {},
+    )
```
iaindillingham, line 123: How about...

```suggestion
    return dictionary | {"text": text} if text else dictionary
```

If there is no text, then I think `text` will be an empty string and I think we probably just want to return `dictionary` unchanged, rather than copying it into a new dictionary.

Jongmassey: If there's no text then `text` will be `None` or possible empty string. 

Yes, I agree it's more efficient not to do the dictionary copy unless necessary. I've never seen the pipe operator used to concat dictionaries like that before and in my testing I get `TypeError: unsupported operand type(s) for |: 'dict' and 'dict'`

iaindillingham (approved):
I've not worked on Data Builder for some time, so am hazy on how we're structuring specifications. However, this PR is clear and addresses opensafely/documentation#772.

databuilder/docs/specs.py
iaindillingham: I think you may be running your tests against Python 3.8 rather than 3.9. I've checked out and amended, and the tests pass. I've also verified that I'm running against 3.9.10 (_.python-version_).

Union operators were added to 3.9 by [PEP 584](https://peps.python.org/pep-0584/).

Jongmassey: oh, that's **very** nice

PR #552, opened 2022-06-15
Use temporary tables instead of CTEs for MSSQL

### LINKED ISSUE #515: Replace CTEs with temporary tables in MSSQL query engine
At the moment we use CTEs to "reify" queries:
https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/query_engines/base_sql.py#L421-L428

CTEs are nice and simple to get started with, but they won't work at production scale — at least on the TPP database.

This will involve porting over the code here:
https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/legacy_sqlalchemy_utils.py#L122-L210

And changing the `execute()` method to expect setup and cleanup queries:
https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/query_engines/legacy_base_sql.py#L199-L211

evansd:
Closes #515

### DIFF
--- a/databuilder/query_engines/mssql.py
+++ b/databuilder/query_engines/mssql.py
@@ -1,4 +1,5 @@
 import sqlalchemy
+from sqlalchemy.schema import CreateIndex
 
 from databuilder import sqlalchemy_types
 from databuilder.query_engines.base_sql import BaseSQLQueryEngine
@@ -31,8 +32,17 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
             sqlalchemy.Column(c.name, c.type, key=c.key) for c in query.selected_columns
         ]
         table = GeneratedTable(table_name, sqlalchemy.MetaData(), *columns)
-        # Use the MSSQL `SELECT * INTO ...` construct to create and populate this table
-        table.setup_queries = [SelectStarInto(table, query.alias())]
+        table.setup_queries = [
+            # Use the MSSQL `SELECT * INTO ...` construct to create and populate this
+            # table
+            SelectStarInto(table, query.alias()),
+            # As we always join rows on `patient_id` it makes sense to store them on
+            # disk in `patient_id` order, which is what creating a clustered index does.
+            # (We use `None` as the index name to let SQLAlchemy generate one for us.)
+            CreateIndex(
+                sqlalchemy.Index(None, table.c.patient_id, mssql_clustered=True)
+            ),
+        ]
         # The "#" in `intermediate_table_prefix` ensures this is a session-scoped
         # temporary table so there's no explict cleanup needed
         return table


databuilder/query_engines/mssql.py
```diff
@@ -2,12 +2,17 @@
...
+from databuilder.query_engines.mssql_dialect import MSSQLDialect, SelectStarInto
+from databuilder.sqlalchemy_utils import GeneratedTable
 
 
 class MSSQLQueryEngine(BaseSQLQueryEngine):
     sqlalchemy_dialect = MSSQLDialect
 
+    # The `#` prefix is an MSSQL-ism which automatically makes the tables session-scoped
+    # temporary tables
+    intermediate_table_prefix = "#tmp_"
```
benbc, line 15: It makes me a little uncomfortable that we subclass the _entire_ base engine and then override arbitrary bits of it. So reading the base class there is no way of knowing whether the thing you are looking at is overridden in some subclass. I suppose we might have some sort of "dialect" mechanism where there is a "base" class that contains only logic that doesn't change and delegates to a specializing object that contains all the variation.

I don't feel strongly about this. Just a thought.

benbc (approved):

databuilder/query_engines/mssql.py
evansd: Yeah, this has been niggling at me as well but I sort of pushed the worry aside because I didn't have a clear alternative proposal. But you're quite right that it makes the control flow impossible to follow, and it's only going to get worse as we add further sorts of variation.

I think having some kind of dialect object is probably the way to go. Although that does mean yet another layer of indirection.

An alternative would be, for any behaviour where we want variation, to refuse to implement default behaviour in the base class and to require each subclass to explicitly implement its own behaviour. So for instance `BaseSQLQueryEngine` would just have
```py
def reify_query(self, query):
    raise NotImplementedError()
```

That could get quite verbose and repetitive quickly though.

benbc: Yes, I think it's reasonable to have an inheritance hierarchy with overridden methods to remove duplication. But it would be helpful to have a signal for which methods _might_ be overridden.

databuilder/query_engines/mssql_dialect.py
```diff
@@ -61,3 +63,20 @@ class MSSQLDialect(MSDialect_pymssql):
...
+    def get_children(self):
+        return (self.table, self.selectable)
+
+
+@compiles(SelectStarInto)
+def visit_select_star_into(element, compiler, **kw):
+    return "SELECT * INTO {} FROM {}".format(
+        compiler.process(element.table, asfrom=True, **kw),
+        compiler.process(element.selectable, asfrom=True, **kw),
+    )
```
inglesp, line 82: Does this mean `into_table = sqlalchemy.literal_column(f"* INTO {table.name}")` is going away?

tests/unit/query_engines/test_mssql_dialect.py
```diff
@@ -33,10 +34,34 @@ def test_mssql_date_types():
...
+    table = sqlalchemy.table("foo", sqlalchemy.Column("bar"))
+    query = sqlalchemy.select(table.c.bar).where(table.c.bar > 1)
+    target_table = sqlalchemy.table("test")
+    select_into = SelectStarInto(target_table, query.alias())
+
+    # Check that SelectStarInto supports iteration by confirming that we can get back to
+    # the original column element if we iterate it
+    elements = iterate(select_into)
+    matches = [el for el in elements if el is table.c.bar]
+    assert len(matches) > 0
```
inglesp, line 59: I don't really understand this.  How would it fail?

inglesp (approved):

tests/unit/query_engines/test_mssql_dialect.py
evansd: If you don't implement the `get_children()` method (as I didn't originally) then you won't get any errors from `iterate()`, it will just fail to recurse into the elements children.

I can (and will) make this clearer in the test comments.

PR #567, opened 2022-06-23
Use `Function` class instead of magic `func` namespace

evansd:
This makes it clear that the function names are just arbitrary strings,
rather than being explicitly defined SQLAlchemy objects. It also lets us
supply the return type directly rather than having to faff around with
`type_coerce`.

### DIFF
--- a/databuilder/query_engines/mssql.py
+++ b/databuilder/query_engines/mssql.py
@@ -1,5 +1,6 @@
 import sqlalchemy
 from sqlalchemy.schema import CreateIndex
+from sqlalchemy.sql.functions import Function as SQLFunction
 
 from databuilder import sqlalchemy_types
 from databuilder.query_engines.base_sql import BaseSQLQueryEngine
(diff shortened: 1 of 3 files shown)


databuilder/query_engines/mssql.py
```diff
@@ -15,15 +16,17 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
     intermediate_table_prefix = "#tmp_"
 
     def get_date_part(self, date, part):
-        func = sqlalchemy.func
-        get_part = {"YEAR": func.year, "MONTH": func.month, "DAY": func.day}[part]
-        # Tell SQLAlchemy that the result is an int without doing any CASTing in the SQL
-        return sqlalchemy.type_coerce(get_part(date), sqlalchemy_types.Integer())
+        assert part in {"YEAR", "MONTH", "DAY"}
+        return SQLFunction(part, date, type_=sqlalchemy_types.Integer)
```
iaindillingham, line 20: I enjoyed unpicking `get_part`, but this is much clearer.

iaindillingham (approved):

PR #570, opened 2022-06-24
Fix query model transform code

evansd:
Previously we were using `set()` to collect query model nodes. But query models can contain distinct objects that hash and compare equal. Usually we don't care about this (equal objects can happily substitute for one another) but when start mutating these objects then it matters which specific objects were mutating.

We now collect nodes using their `id()` to ensure that we get all of them.

Who'd have thought that force-mutating frozen objects would have led to problems?

What this obviously shows is the need for a better approach here. But this fix will have to do for now.

### DIFF
--- a/databuilder/query_model_transforms.py
+++ b/databuilder/query_model_transforms.py
@@ -47,13 +47,16 @@ def get_reverse_index(nodes):
     index = {}
     for node in nodes:
         populate_reverse_index(index, node)
-    return index
+    return {node: references.values() for node, references in index.items()}
 
 
 def populate_reverse_index(index, node):
-    index.setdefault(node, set())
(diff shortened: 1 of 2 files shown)


benbc (approved):
"Approve".

<cough> https://en.wikipedia.org/wiki/Zipper_(data_structure) <cough>

evansd:
It says there in black and white that this is the sort of code you approve of, Ben :)

I did have a brief look at those data structures when you first introduced them (despite saying that I wouldn't). I think that's almost certainly the answer. Just not the immediate priority.

benbc:
> Just not the immediate priority.

No, definitely. Just couldn't resist.

PR #576, opened 2022-06-27
Remove backend registration mechanism

### LINKED ISSUE #561: Remove backend registration system
At present, backends define a `backend_id` and register themselves in a global dictionary so that they can be referenced by this ID in environment variables.

However, this requires importing all the backends so they can register themselves. And this causes circular import problems e.g.
```
$ python -c 'from databuilder.query_engines.sqlite import SQLiteQueryEngine'
  File "<string>", line 1, in <module>
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_engines/sqlite.py", line 5, in <module>
    from databuilder.query_engines.base_sql import BaseSQLQueryEngine
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_engines/base_sql.py", line 8, in <module>
    from databuilder.backends.base import DefaultBackend
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/backends/__init__.py", line 3, in <module>
    from .databricks import DatabricksBackend
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/backends/databricks.py", line 2, in <module>
    from ..query_engines.spark import SparkQueryEngine
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_engines/spark.py", line 4, in <module>
    from databuilder.query_engines.base_sql import BaseSQLQueryEngine
ImportError: cannot import name 'BaseSQLQueryEngine' from partially initialized module 'databuilder.query_engines.base_sql' (most likely due to a circular import) (/home/dave/projects/ebmdatalab/databuilder/databuilder/query_engines/base_sql.py)
```

We should get rid of the registration system and allow the environment to specify the backend by its dotted path e.g.
```
OPENSAFELY_BACKEND='databuilder.backends.tpp.TPPBackend'
```

As well as fixing the circular import problem and simplifying the code this naturally supports out-of-tree backends, which I think is desirable.

evansd:
The effect of this PR is that rather than specifying a backend using an
alias like:

    OPENSAFELY_BACKEND='tpp'

The backend is instead specified using its full dotted path:

    OPENSAFELY_BACKEND='databuilder.backends.tpp.TPPBackend'

This fixes a [circular import][1] problem, removes some code, and at the
same time makes it possible to use databuilder with backends definitions
that aren't bundled within databuilder itself.

I don't currently anticipate users needing to supply backend names
themselves on the command line, so the extra verbosity shouldn't be a
problem here. But if it turns out they do need to do so then we can
define a mapping of short aliases to full dotted paths for commonly used
backends.

Closes #561

[1]: https://github.com/opensafely-core/databuilder/issues/561

### DIFF
--- a/databuilder/__main__.py
+++ b/databuilder/__main__.py
@@ -3,7 +3,6 @@ import sys
 from argparse import ArgumentParser, ArgumentTypeError
 from pathlib import Path
 
-from .backends import BACKENDS
 from .main import (
     generate_dataset,
     generate_measures,
@@ -55,7 +54,7 @@ def main(args=None):
         )
     elif options.which == "test-connection":
         test_connection(
-            backend=options.backend,
+            backend_id=options.backend,
             url=options.url,
         )
     elif options.which == "print-help":
@@ -108,7 +107,6 @@ def add_dump_dataset_sql(subparsers):
     parser.add_argument(
         "backend",
         type=str,
-        choices=BACKENDS,  # allow all registered backend subclasses
     )
     parser.add_argument(
         "--dataset-definition",
--- a/databuilder/backends/__init__.py
+++ b/databuilder/backends/__init__.py
@@ -1,7 +1,0 @@
-# All backends need to be imported here so they get registered
-from .base import BACKENDS
-from .databricks import DatabricksBackend
-from .graphnet import GraphnetBackend
-from .tpp import TPPBackend
-
-__all__ = ("BACKENDS", "DatabricksBackend", "TPPBackend", "GraphnetBackend")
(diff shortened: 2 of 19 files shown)


iaindillingham (approved):

evansd:
Thanks @bloodearnest, this is a good point.

I think your suggestion of using `DATABUILDER_BACKEND` instead sounds like the best solution to me. Independent of the backwards compatibility issue it's probably better if the all the databuilder env vars are prefixed `DATABUILDER_`.

I'll make a ticket for it.

PR #578, opened 2022-06-28
Make comparisons between Value node type-sensitive

evansd:
Python treats certain differently typed values as equal (e.g. `1 == True` and `10 == 10.0`) but some of the databases we run against care about the difference between e.g. integers and floats. Because of the caching mechanism used in the BaseSQLQueryEngine, `Value` nodes which compare equal will always produce the same SQL. This means that `Value(10)` can produce SQL corresponding to `Value(10.0)`, if the latter node happens to have been processed and cached first.

The fix here is to define a custom equality method for `Value` which checks the types as well.

### DIFF
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -157,6 +157,16 @@ class AggregatedSeries(OneRowPerPatientSeries):
 class Value(OneRowPerPatientSeries[T]):
     value: T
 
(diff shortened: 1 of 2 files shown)


CarolineMorton (approved):
This looks good to me. I have been following along with your slack thread so understand the context here. I have not run the tests locally.

PR #581, opened 2022-06-29
feat: Add .to_first_of_year and .to_first_of_month

### LINKED ISSUE #545: Add `to_first_of_month` and `to_first_of_year` methods
Exact names up for discussion. But whatever they're called they would need adding to the `DateFunctions` mixin:
https://github.com/opensafely-core/databuilder/blob/2ce31449d6e35b21d45a0e96c746c410be6b7ac9/databuilder/query_language.py#L214-L225

The corresponding operations already defined in the query model, but none of the query engines handle them. Adding the relevant spec tests will drive out exactly what needs doing here.

These are the kind of operations which will need specialising per SQL engine, like `DateAddDays` etc.

rebkwok:
Fixes #545

### DIFF
--- a/databuilder/population_validation.py
+++ b/databuilder/population_validation.py
@@ -223,6 +223,11 @@ def date_subtract_days(date, num_days):
     return date - datetime.timedelta(days=num_days)
 
 
+@register_op(Function.ToFirstOfYear)
+def to_first_of_year(date):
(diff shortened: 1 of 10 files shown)


evansd (approved):
:100:

evansd (approved):
:+1: All looks good

PR #582, opened 2022-06-29
Add a `contains()` method for string series

evansd:
We need this in the short term to query columns which have multiple codes stuffed into a single string. Longer term we have [better ideas][1] for handling this.

There's a wrinkle here in that we specify the `contains()` operator as being case sensitive, but on MSSQL the case sensitivity of the resulting LIKE query is actually determined by the collation of the column being queried. In practice this is unlikely to matter because if the contents of the column are such that case is significant then they shouldn't be stored using a case insensitive collation (and if they are then we have bigger problems). But the fact that database implementation details leak through like this — inevitable though it is — makes me a bit sad.

Note that after merging local MSSQL containers will be need to be rebuilt:

    docker rm --force databuilder-mssql

[1]: https://github.com/opensafely-core/databuilder/issues/574


tests/spec/str_series_ops/test_contains.py
```diff
@@ -61,3 +84,27 @@ def test_contains_value_from_column(spec_test):
             8: None,
         },
     )
+
+
+def test_contains_value_from_column_with_special_characters(spec_test):
```
inglesp, line 89: Probably don't need this in addition to above?

databuilder/query_engines/base_sql.py
```diff
@@ -212,7 +214,15 @@ def get_sql_subtract(self, node):
     def get_sql_string_contains(self, node):
         haystack = self.get_expr(node.lhs)
         needle = self.get_expr(node.rhs)
-        return haystack.contains(needle)
+        escaped_needle = self.string_replace(needle, "/", "//")
+        escaped_needle = self.string_replace(escaped_needle, "%", "/%")
+        escaped_needle = self.string_replace(escaped_needle, "_", "/_")
+        return haystack.contains(escaped_needle, escape="/")
```
inglesp: If we've got a choice of escape characters, why not backlash?

tests/spec/str_series_ops/test_contains.py
inglesp: This is answered in the next commit.

inglesp (approved):

databuilder/query_engines/base_sql.py
evansd: `/` is the SQLAlchemy default. I thought that was odd too and tried using a backslash and then the tests started failing on Spark, and I didn't look much further into it. Worth a comment, do you think?

evansd: No, it's purely internal. I'll add a comment though.

PR #584, opened 2022-06-29
Treat `NOT ...` expressions as predicates

evansd:
In general, failing to correctly classify SQL constructs as predicates
or not causes us to generate SQL which MSSQL can't handle.

I uncovered this particular omission when fiddling around trying to get
the Hypothesis tests working, so Hypothesis gets the credit for this
one.


StevenMaude (approved):

databuilder/sqlalchemy_utils.py
```diff
@@ -27,6 +33,9 @@ def is_predicate(clause):
         return True
     if isinstance(clause, BinaryExpression):
         return clause._is_implicitly_boolean
+    if isinstance(clause, UnaryExpression):
+        # Handles `NOT ...` expressions
+        return clause.operator is operators.inv
```
StevenMaude, line 38: note: I struggled to look up the definition of `operators.inv` and am not familiar with the SQLAlchemy source. Searching the SQLAlchemy docs for it wasn't helpeful.

I believe it's in this [compiler.py](https://github.com/zzzeek/sqlalchemy/blob/67ba6357e4e6dc8619ac63ce1e8f65310b6595bc/lib/sqlalchemy/sql/compiler.py#L271).

evansd: Ah, slightly confusingly I think it actually comes from the stdlib `operator` module (which is why grepping for `def inv` in the SQLAlchemy source won't find anything). But I'm just treating it as an opaque marker for "the thing SQLAlchemy uses when constructing a `NOT` query", which is why I import it from `sqlalchemy.sql.elements`.

evansd: @StevenMaude I've just tweaked this slightly. Is the new version any clearer?

StevenMaude: This is fine; sorry for late reply!

PR #586, opened 2022-06-30
Run generative tests against MSSQL

### LINKED ISSUE #509: Get generative tests running on MSSQL query engine
When I tried this something went wrong and the process ended up exhausting the file descriptor limit. I haven't investigated further.

evansd:
I still don't know exactly what the issue was here, but refactoring things so that we re-use the same `Database` instance across tests seems to have fixed whatever was causing the file descriptor exhaustion problem.

As part of the refactoring we now use the same parametrized `engine` fixture as the other tests so that adding new engine types here automatically includes them in the generative tests.

Closes #509

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -81,29 +81,39 @@ def in_memory_sqlite_database():
     return InMemorySQLiteDatabase()
 
 
-@pytest.fixture(params=["in_memory", "sqlite", "mssql", "spark"])
-def engine(request):
+QUERY_ENGINE_NAMES = ("in_memory", "sqlite", "mssql", "spark")
+
+
+def engine_factory(request, engine_name):
     # We dynamically request fixtures rather than making them arguments in the usual way
     # so that we only start the database containers we actually need for the test run
     fixture = request.getfixturevalue
 
-    name = request.param
-    if name == "in_memory":
+    if engine_name == "in_memory":
         # There are some tests we currently expect to fail against the in-memory engine
         marks = [m.name for m in request.node.iter_markers()]
         if "xfail_in_memory" in marks:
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -82,7 +82,11 @@ def in_memory_sqlite_database():
 
 
 @pytest.fixture(params=["in_memory", "sqlite", "mssql", "spark"])
-def engine(request, in_memory_sqlite_database, mssql_database, spark_database):
+def engine(request):
+    # We dynamically request fixtures rather than making them arguments in the usual way
+    # so that we only start the database containers we actually need for the test run
+    fixture = request.getfixturevalue
```
benbc, line 88: Oh, this is great. Now _you're_ the pytest internals expert. :-P

tests/generative/test_query_model.py
```diff
@@ -45,14 +42,22 @@
 
 
 @pytest.fixture(scope="session")
-def query_engines():
+def query_engines(request):
+    # This is a slight abuse of pytest's internals but it achieves what we want, which
+    # is to get all the values of a parametrized fixture in one go. Usually we want to
+    # run tests against each query engine one at a time, but here we want to run the
+    # same query against multiple query engines together so we can compare the results.
```
benbc, line 49: It does seem a bit odd to use the fixture as a catalogue of engines. Is there a possibility instead of factoring the list of named engines out of the fixture and accessing that instead?

evansd: Oh you're quite right. We can do this without any shenanigans at all — can't we? — with a bit of refactoring. I got carried away with my own terribleness.

benbc (approved):

PR #590, opened 2022-07-01
Allow `take/drop` to accept boolean constants

evansd:
This might look a bit pointless, but it turns out to be convenient to be
able to have a no-op filter condition that can be used as a default to
avoid having to add an extra conditional e.g.
```py
    def some_function(where=True):
        filtered = frame.take(where)
        ...
```

Without this you'd need to something like:
```py
    def some_function(where=None):
        if where is None:
            filtered = frame
        else:
            filtered = frame.take(where)
        ...
```

This involves passing all arguments to ehrQL methods through `_convert`
rather than assuming they are series objects.

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -379,7 +379,7 @@ class EventFrame(Frame):
         return EventFrame(
             qm.Filter(
                 source=self.qm_node,
-                condition=series.qm_node,
+                condition=_convert(series),
             )
         )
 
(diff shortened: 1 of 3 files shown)


rebkwok (approved):
Should `drop()` be tested in the spec tests too?

evansd:
Doh! Yes, I meant to cover `drop` as well. Thanks.

PR #599, opened 2022-07-05
fix: upgrade sqlalchemy and enable caching on dialect subclasses

### LINKED ISSUE #598: Upgrade sqlalchemy
sqlalchemy is currently pinned to version 1.4.27 because upgrading past 1.4.27 results in warnings which cause test errors, e.g.:

SAWarning: Dialect mssql:pymssql will not make use of SQL compilation caching as it does not set the 'supports_statement_cache' attribute to ``True``.  This can have significant performance implications including some performance degradations in comparison to prior SQLAlchemy versions.  Dialect maintainers should seek to set this attribute to True after appropriate development and testing for SQLAlchemy 1.4 caching support.   Alternatively, this attribute may be set to False which will disable this warning.

All our engine dialects inherit from base classes that include `supports_statement_cache = True`, but as of v1.4.28 (not 1.4.5 as it says in the docs), sqlalchemy requires [third party dialects to explicitly declare that](https://docs.sqlalchemy.org/en/14/core/connections.html#caching-for-third-party-dialects) too.

We should be able to fix this by just adding [`supports_statement_cache = True` to each dialect](https://github.com/opensafely-core/databuilder/pull/596#pullrequestreview-1028361329)

rebkwok:
Fixes #598


databuilder/query_engines/spark_dialect.py
```diff
@@ -36,6 +36,9 @@ def visit_create_temporary_view_as(element, compiler, **kw):
 
 
 class SparkDate(sqlalchemy.types.TypeDecorator):
+
+    supports_statement_cache = True
```
evansd: How come this one needs to go the date type and not on the dialect?

evansd (approved):
Looks good.

PR #600, opened 2022-07-05
Add function to import and return all contracts/backends

evansd:
Prior to this we had tests which implicitly relied on various contracts
and backends being imported beforehand. This worked when running the
full test suite but failed when these tests were run individually.

### DIFF
--- a/databuilder/docs/backends.py
+++ b/databuilder/docs/backends.py
@@ -1,11 +1,13 @@
 import operator
 
 from ..backends.base import BaseBackend
+from ..module_utils import get_sibling_subclasses
 from .common import build_hierarchy
 
 
 def build_backends():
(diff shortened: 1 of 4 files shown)


databuilder/contracts/base.py
```diff
@@ -103,3 +105,18 @@ def validate_schema(cls, schema):
             assert (
                 column.type.python_type == schema[col_name]
             ), f"{col_name} doesn't match {column}"
+
+
+def get_all_contracts():
```
rebkwok, line 110: The only difference between this and `get_all_backends` is the class it's looking for subclasses on.  Is it worth making it a shared function?

rebkwok (approved):
LGTM

PR #607, opened 2022-07-07
Databricks updates

rebkwok:
Update the databricks scripts, docs and ensure tests can run (`databricks-sql-connector` was recently upgraded from v0.9.0 to 2.0.2).

It appears that since the `dbx` script was first written, the allowed API calls for the Databricks community edition have been limited.  If using the community edition, you can fetch information about clusters, but can't create or terminate one.  The script and docs are updated accordingly, to note that you need to manually create the cluster in the web console for both community edition and sandbox.

### DIFF
--- a/databuilder/query_engines/spark_dialect.py
+++ b/databuilder/query_engines/spark_dialect.py
@@ -200,14 +200,18 @@ class SparkDialect(HiveHTTPDialect):
         string matching for this and the format of the error messages in the
         specific version of Spark/Hive we are using doesn't match. See:
 
-        https://github.com/dropbox/PyHive/blob/b21c507a24/pyhive/sqlalchemy_hive.py#L275-L297
+        https://github.com/dropbox/PyHive/blob/release/0.6.5/pyhive/sqlalchemy_hive.py#L275-L297
         """
         connection = ConnectionWrapper(connection)
         try:
             return super()._get_table_columns(connection, table_name, schema)
-        except exc.OperationalError as e:
+        except (exc.OperationalError, exc.DatabaseError) as e:
             full_table = table_name if not schema else f"{schema}.{table_name}"
-            if "Table or view not found" in str(e):  # pragma: no cover
+            if "Table or view not found" in str(
+                e
+            ) or f"{full_table} doesn't exist" in str(
+                e
+            ):  # pragma: no cover
                 raise exc.NoSuchTableError(full_table)
             else:
                 raise
--- a/.github/workflows/databricks.yml
+++ b/.github/workflows/databricks.yml
@@ -7,7 +7,7 @@ on:
       test_args:
         description: "Tests to run or other pytest args"
         required: true
-        default: "tests/backend/test_databricks.py"
+        default: "tests/spec -k 'spark'"
 
 jobs:
   test:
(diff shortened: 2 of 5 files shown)


databuilder/query_engines/spark_dialect.py
```diff
@@ -200,14 +200,18 @@ def _get_table_columns(self, connection, table_name, schema):
...
         try:
             return super()._get_table_columns(connection, table_name, schema)
-        except exc.OperationalError as e:
+        except (exc.OperationalError, exc.DatabaseError) as e:
             full_table = table_name if not schema else f"{schema}.{table_name}"
-            if "Table or view not found" in str(e):  # pragma: no cover
+            if "Table or view not found" in str(
+                e
+            ) or f"{full_table} doesn't exist" in str(
+                e
```
rebkwok, line 213: I saw both of these strings, with different underlying errors coming from either hive or dbfs.  The dbfs ones started after I tried running `dbx cleanup`, but I don't know why :shrug:

tests/conftest.py
```diff
@@ -28,6 +29,17 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
...
 
+@pytest.fixture(autouse=True)
+def capture_wrap():  # pragma: no cover
+    """
+    Workaround for issues with pytest's capture logging
+    https://github.com/pytest-dev/pytest/issues/5502
+    """
+    sys.stderr.close = lambda *args: None
+    sys.stdout.close = lambda *args: None
+    yield
```
evansd, line 40: I'm slightly nervous about this because it's such a sweeping bit of monkeypatching rather than something that's just targeting the databricks/thrift problem. Not that I can think of any specific issues it will cause us.

Looking at the relevant bit of `databricks` [code](https://github.com/databricks/databricks-sql-python/blob/6d62baa2c744c5942fb9563f0cc98c9f0080fe44/src/databricks/sql/client.py#L142-L152) it looks like the warning gets triggered because we've got connnections that are still open at the point they get garbage collected. So maybe we can so more careful lifecycle management. Or have something which sticks all open connections in a global somewhere, and then a fixture which automatically closes them before we exit.

I'd be happy to wave this through for now if it unblocks us, but I'd like to replace it with something more targeted (even if equally gnarly) before too long.

evansd (approved):
This looks both great and like it wasn't an awful lot of fun to do. So thanks! As I say, I've got some reservations about the final workaround, but I don't think it's an immediate blocker.

tests/conftest.py
rebkwok: I tried putting a [close_all_sessions](https://docs.sqlalchemy.org/en/13/orm/session_api.html?highlight=close_all_sessions#sqlalchemy.orm.session.close_all_sessions) in either a `pytest_sessionfinish()` or a session fixture, but both still end up with this logging error.

Since we only run the databricks tests manually atm, and the logging error might not cause tests to fail anyway, I could just take the workaround out for now.

evansd: I think `close_all_sessions` is a SQLAlchemy ORM function, rather than a SQLAlchemy Core one — it can be a bit confusing sometimes which features apply to which.

If we're only running them manually then I might be tempted to take it out for now. I might be being overly paranoid, but monkeypatching `sys.stdout` just gives me the heebie jeebies.

It would be worth making an issue describing the problem and the workaround (and a brief note of the things which didn't work) just so the next one of us to hit this has something to go on. And it might be one of those things which Databricks/pytest just fix for us in an update at some point.

rebkwok: I've taken out the workaround, and the CI tests are still passing :shrug:

PR #608, opened 2022-07-07
Rework `dump-dataset-sql` command

evansd:
This turned into a larger change than I originally intended, but I think (hope) it's going to pay off over the long term. The aim here is to make it as ergonomic as possible to inspect the SQL corresponding to a particular dataset definition for debugging purposes.

To that end, I've changed the syntax of `dump-dataset-sql` so that the only required argument is now the dataset file itself. This makes the minimum invocation:

    databuilder dump-dataset-sql path/to/dataset_definition.py

This compiles the dataset definition using the SQLite query engine and the default backend which assumes that the SQL schema exactly matches the contract schema. The resulting SQL is written to stdout.

The `--backend` and `--query-engine` arguments allow specifying either or both of the backend and query engine to use.

The `--output` argument allows writing the output directly to a file, if that's easier than redirecting.

The other important change is that the SQL now contains literal values rather than placeholders (e.g., `foo > 100` rather than `foo > :param_1`. This means it can be executed directly in a db shell, or passed to other tools which will only accept valid SQL.

### DIFF
--- a/databuilder/__main__.py
+++ b/databuilder/__main__.py
@@ -4,11 +4,11 @@ from argparse import ArgumentParser, ArgumentTypeError
 from pathlib import Path
 
 from .main import (
+    dump_dataset_sql,
     generate_dataset,
     generate_measures,
     pass_dummy_data,
     test_connection,
-    validate_dataset,
 )
 
 
@@ -37,7 +37,7 @@ def main(args, environ=None):
                 "error: either --dummy-data-file or DATABASE_URL environment variable is required"
             )
     elif options.which == "dump-dataset-sql":
-        validate_dataset(
+        dump_dataset_sql(
             options.dataset_definition,
             options.output,
             backend_id=options.backend,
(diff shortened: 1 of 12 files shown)


rebkwok (approved):
This all looks good to me.
I think the command was originally called `validate_dataset` (or `validate_cohort` then) because there was a request for users to be able to run an action with `opensafely run` to validate their dataset definition and get useful error messages.  The sql dump was a side effect in case it was useful.  Do we want to get rid of the validate command altogether?

evansd:
> I think the command was originally called validate_dataset (or validate_cohort then) because there was a request for users to be able to run an action with opensafely run to validate their dataset definition and get useful error messages.

Yes, that matches my memory too.

The theory now is that all the validation gets done when importing the dataset definition i.e. it should be impossible to even construct an invalid one. So — again, in theory — there shouldn't be a need for an explicit validation command.

> Do we want to get rid of the validate command altogether?

I think it's already gone. Or more precisely, I think it already got renamed to `dump-dataset-sql` by someone before me (Jon, maybe?). Or is there somewhere else I've missed?

rebkwok:
>I think it's already gone.

yes, I think you're right.  It was just the function name that was still `validate...`

Issue #566, opened 2022-06-23
Command-line configuration errors should give helpful user-facing error messages


StevenMaude:
And not tracebacks.

Examples:

```
$ docker run ghcr.io/opensafely-core/databuilder:v0 test_connection                                                                            
Traceback (most recent call last):                                                                                                                                                                                 
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main                                                                                                                                             
    return _run_code(code, main_globals, None,                                                                                                                                                                     
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code                                                                                                                                                        
    exec(code, run_globals)                                                                                                                                                                                        
  File "/app/databuilder/__main__.py", line 179, in <module>                                                                                                                                                       
    main()                                                                                                                                                                                                         
  File "/app/databuilder/__main__.py", line 59, in main                                                                                                                                                            
    test_connection(                                                                                                                                                                                               
  File "/app/databuilder/main.py", line 68, in test_connection                                                                                                                                                     
    backend = BACKENDS[backend](url, temporary_database=None)                                                                                                                                                      
KeyError: None                                                       
```

```
$ docker run ghcr.io/opensafely-core/databuilder:v0 test_connection --backend graphne
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/app/databuilder/__main__.py", line 179, in <module>
    main()
  File "/app/databuilder/__main__.py", line 59, in main
    test_connection(
  File "/app/databuilder/main.py", line 68, in test_connection
    backend = BACKENDS[backend](url, temporary_database=None)
KeyError: 'graphne'
```

```
$ docker run ghcr.io/opensafely-core/databuilder:v0 test_connection --backend graphnet 
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/app/databuilder/__main__.py", line 179, in <module>
    main()
  File "/app/databuilder/__main__.py", line 59, in main
    test_connection(
  File "/app/databuilder/main.py", line 70, in test_connection
    with query_engine.engine.connect() as connection:
[...]

Issue #533, opened 2022-06-05
Remove unhelpful output from error messages


inglesp:
To generate a dataset, we [import a module containing a dataset definition](https://github.com/opensafely-core/databuilder/blob/main/databuilder/main.py#L86-L96).

If importing that module raises any kind of Python exception, the user sees a stack trace with a lot of unhelpful output.

For instance, if `demo.py` contains the line `asdf`, then we see:

```
$ DATABASE_URL=asdf python -m databuilder generate_dataset --dataset-definition demo.py
2022-06-04 21:34:55 [info     ] Generating dataset for demo.py [databuilder.main] 
Traceback (most recent call last):
  File "/home/inglesp/.pyenv/versions/3.9.5/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/inglesp/.pyenv/versions/3.9.5/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/databuilder/__main__.py", line 179, in <module>
    main()
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/databuilder/__main__.py", line 31, in main
    generate_dataset(
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/databuilder/main.py", line 26, in generate_dataset
    dataset = load_definition(definition_file)
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/databuilder/main.py", line 82, in load_definition
    load_module(definition_file)
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/databuilder/main.py", line 97, in load_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 855, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/demo.py", line 1, in <module>
    asdf
NameError: name 'asdf' is not defined
```

In this case, only these lines are meaningful to the user:

```
2022-06-04 21:34:55 [info     ] Generating dataset for demo.py [databuilder.main] 
Traceback (most recent call last):
  File "/home/inglesp/work/ebmdatalab/opensafely/databuilder/demo.py", line 1, in <module>
    asdf
NameError: name 'asdf' is not defined
```

[...]

PR #609, opened 2022-07-08
Add cleaner error messages

CarolineMorton:
This [ticket](#533) involves returning sensible and useful error messages to the user when they type in a command such as `DATABASE_URL=tpp python -m databuilder generate-dataset --dataset-definition demo_study.py`. This changes aims to pick up code that is incorrect in the dataset-definition python file -in this case, `demo_study.py`. 

We discussed as a team and decided that we wanted a partial traceback that started with the user input and missed out the error being propagated through multiple internal functions with the databuilder architecture. It was not enough however to just take the last 2-3 lines of the traceback however as the error could be within a function defined in the data definition file.

The sorts of error we might see in a dataset-definition file include:

- Variable does not exist
- Incorrect or faulty import
- Function defined in the file does not work

**What I have done:**

- Created a global in the `main.py` to allow us to filter traceback to this point
- Added two functions - 1) checks this global 2) identifies the level at which the traceback needs to start from
- Add try/except to allow us to format the traceback a bit better

**What I have not done:**
- Add friendly_traceback - this is a cool package and from a developer experience would be a fantastic addition but from a researcher user, it adds more to the traceback rather than shortening it into something readable and relevant. I do think adding hints is helpful which is something that this lib does well but that can be done for this section of the codebase by adding information to the parser help. 
- Build Custom Errors - this is something we discussed as a team. There are some errors within the code base in various points e.g. but overall we do not have widespread custom errors and are using Python's standard errors to catch most things. We could revisit this at some point. 
- Address the related ticket in #566 which is bigger and may require more thought.

**Examples:**
`demo_study.py`
```python
sillytest
```
Results in:
```
2022-07-08 13:32:11 [info     ] Generating dataset for demo_study.py [databuilder.main] 
***********
Error Type: NameError
Message: name 'sillytest' is not defined
***********
  File "demo_study.py", line 2, in <module>
    sillytest

NameError: name 'sillytest' is not defined

```

`demo_study.py`
```python
def bad_function:
   pass
```
Results in:
```
2022-07-08 13:32:22 [info     ] Generating dataset for demo_study.py [databuilder.main] 
***********
Error Type: SyntaxError
Message: invalid syntax (demo_study.py, line 4)
***********
  File "demo_study.py", line 4

    def bad_function:

                    ^

SyntaxError: invalid syntax

```

`demo_study.py`
```python
from bad_import import bad_import

bad_import() 
```
`bad_import.py`
```python
def bad_import:
    pass
```
Results in:
```
2022-07-08 13:32:40 [info     ] Generating dataset for demo_study.py [databuilder.main] 
***********
Error Type: SyntaxError
Message: invalid syntax (bad_import.py, line 1)
***********
  File "/Users/carolinemorton/Documents/github/databuilder/bad_import.py", line 1

    def bad_import:

                  ^

SyntaxError: invalid syntax
```


CarolineMorton:
@inglesp Can you review this to see if approach works for you? I will fix tests if it is.

databuilder/main.py
```diff
@@ -14,6 +15,35 @@
...
+
+
+def is_databuilder(tb: traceback) -> bool:
+    """
+    Check if the code in the traceback frame is
+    from the Databuilder Code Base. This aims to remove
+    traceback involving other libaries such as importlib.
+    """
+    db_globals = tb.tb_frame.f_globals
+    return "DATABUILDER_CODE" in db_globals
```
evansd, line 28: This only identifies this specific databuilder module, but won't catch databuilder modules that the user code calls back into. I think it ought to be possible to do the filtering instead using the filename, which I think each frame in the stack trace should have.

databuilder/main.py
```diff
@@ -14,6 +15,35 @@
...
+
+
+def user_code_traceback_level(tb: traceback, error_type: Exception) -> int:
```
evansd, line 31: I'm not convinced this approach to filtering the stack is quite right.

At the point where we hit an exception we've got a kind of stack sandwich:
```mermaid
flowchart TD
  id1["databuilder code (main.py etc)"]
  -->
  id2["user code"]
  -->
  id3["databuilder code (query_langage.py etc)"]
```

And we want to top and tail this to just show the user code.

So I think this involves walking the whole stack and removing everything up to an including the `exec_module` call. And then keeping the rest of the stack up until the first frame from a module inside `databuilder/*`.

So the user doesn't see either our initial code which imported their module _or_ the internals of our query language code which they call into.

This is all a bit theoretical at the moment though. Until we start using this in practice and seeing what the errors actually look like it's hard to know what's going to be most helpful for users.

evansd (changes requested):
Thanks @CarolineMorton. This is a good start but I think there's some more investigation needed into how we're doing to filter the stack trace. I've left a few comments, but imagine some further discussion would be useful so give me (or @inglesp) a shout if you'd like to do that.

And sorry to land you with such a fiddly task!

databuilder/main.py
CarolineMorton: Hi @evansd

Thanks for this. Would you be able to provide an example a dataset definition that would demonstrate this sandwich problem? @inglesp  and I have come up with a number of examples but they do not cause an error in the query_language for example and so do not demonstrate sandwich-stacks. Any ideas?

evansd: I think this is the sort of thing I meant:
```py
from databuilder.query_language import Dataset, build_patient_table

table = build_patient_table("table", {"i": int, "b": bool})

dataset = Dataset()
dataset.v = table.i + table.b
```

If I import that I get a traceback like this:
```
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/__main__.py", line 180, in <module>
    main(sys.argv[1:], environ=os.environ)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/__main__.py", line 40, in main
    dump_dataset_sql(
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/main.py", line 53, in dump_dataset_sql
    dataset_definition = load_definition(definition_file)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/main.py", line 123, in load_definition
    module = load_module(definition_file)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/main.py", line 144, in load_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "error_example.py", line 6, in <module>
    dataset.v = table.i + table.b
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_language.py", line 189, in __add__
    return _apply(qm.Function.Add, self, other)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_language.py", line 335, in _apply
    qm_node = qm_cls(*qm_args)
  File "<string>", line 5, in __init__
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_model.py", line 103, in __post_init__
    validate_node(self)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_model.py", line 378, in validate_node
    validate_types(node)
  File "/home/dave/projects/ebmdatalab/databuilder/databuilder/query_model.py", line 607, in validate_types
    raise TypeValidationError(
databuilder.query_model.TypeValidationError: Add.rhs requires 'databuilder.query_model.Series[int]' but got 'databuilder.query_model.Series[bool]'
In node:
Function.Add(lhs=SelectColumn(source=SelectPatientTable(name='table', schema=TableSchema(i=int, b=bool)), name='i'), rhs=SelectColumn(source=SelectPatientTable(name='table', schema=TableSchema(i=int, b=bool)), name='b'))
```

Leaving aside the fact that the `TypeValidationError` we generate is not very friendly, if we exclude all the `databuilder` and `importlib` frames we get something like:
```
  File "error_example.py", line 6, in <module>
    dataset.v = table.i + table.b
databuilder.query_model.TypeValidationError: Add.rhs requires 'databuilder.query_model.Series[int]' but got 'databuilder.query_model.Series[bool]'
In node:
Function.Add(lhs=SelectColumn(source=SelectPatientTable(name='table', schema=TableSchema(i=int, b=bool)), name='i'), rhs=SelectColumn(source=SelectPatientTable(name='table', schema=TableSchema(i=int, b=bool)), name='b'))
```

Which I think is a big improvement.

CarolineMorton:
Replaced by #691

PR #615, opened 2022-07-11
Support fetching results in batches in MSSQL engine

### LINKED ISSUE #531: Support fetching results in batches in MSSQL engine
To work around the flakiness of the database connection, the MSSQL engine should write the results to a temporary table, and then download the results in batches.

This involves porting [this code from the legacy MSSQL engine](https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/query_engines/legacy_mssql.py#L49-L81) to the [new MSSQL engine](https://github.com/opensafely-core/databuilder/blob/main/databuilder/query_engines/mssql.py).

Jongmassey:
fixes #531


databuilder/query_engines/mssql.py
```diff
@@ -49,3 +52,29 @@ def reify_query(self, query):
...
+        with fetch_results_in_batches(
+            engine=self.engine,
+            queries=setup_queries + [results_query],
+            temp_table_prefix="#",
+            # This value was copied from the previous cohortextractor. I
+            # suspect it has no real scientific basis.
+            batch_size=32000,
+            max_retries=2,
+            sleep=0.5,
+            reconnect_on_error=True,
```
evansd, line 78: Note in passing, I don't think you can enable `reconnect_on_error` here because this is using session-scoped temporary tables.

evansd (commented):
Yes, this is broadly the right direction. But I think we can simplify this a fair bit now that we're not trying to do clever things with resuming downloads across jobs.

I think the only stuff we really need now is the `fetch_table_in_batches` function and the `ReconnectableConnection` class.

The code for checking whether tables exist and are writable can all go  I think. And we already have code in `mssql_dialect` for writing queries to temp tables and creating indexes.

evansd:
Closed in favour of #660

PR #616, opened 2022-07-11
fix: Support multiline series definitions in specs docs

### LINKED ISSUE #605: `case` function calls in spec tests aren't appearing in documentation correctly
case-incomplete-function-call

source: https://github.com/opensafely-core/databuilder/blob/b5b8cdb8c838fa25cb07bc1a262a8b5cae93d6f1/tests/spec/case_expressions/test_case.py#L20-L34

Not quite sure why yet, as I don't know fully end-to-end how all this works. But if `ArgCapturer` is used, then [this assumption might be incorrect](https://github.com/opensafely-core/databuilder/blob/b5b8cdb8c838fa25cb07bc1a262a8b5cae93d6f1/databuilder/docs/specs.py#L166-L174), because `case` takes multiple arguments.

rebkwok:
Fixes #605

### DIFF
--- a/databuilder/docs/specs.py
+++ b/databuilder/docs/specs.py
@@ -1,4 +1,5 @@
 import inspect
+import re
 from importlib import import_module
 
 from tests.spec import toc
@@ -100,9 +101,7 @@ def build_paragraph(paragraph_id, test_fn):
     # Check that the next line is as expected.
     assert source_lines[ix + 1] == "        table_data,"
 
-    # Extract the definition of the series.
-    series_line = source_lines[ix + 2]
-    series = series_line.strip().removesuffix(",")
+    series = get_series_code(source_lines, ix + 2)
 
     # Extract descriptive docstring if any
     text = inspect.getdoc(test_fn)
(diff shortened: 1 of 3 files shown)


databuilder/docs/specs.py
```diff
@@ -119,6 +118,52 @@ def build_paragraph(paragraph_id, test_fn):
...
+    return "\n".join(series_lines).strip().removesuffix(",")
+
+
+def series_is_balanced(series_line_string):
+    """
+    Takes a string representing a series definition and ensures that parentheses are
+    balanced.
+    """
+    stack = []
+    for i in series_line_string:
```
inglesp, line 155: To me, `i` is a surprising variable name -- it suggests an integer value.  Maybe `c` or `char`?

inglesp (approved):

tests/unit/test_docs.py
```diff
@@ -28,3 +31,77 @@ def test_generate_docs():
...
+            # incomplete series definition; this should never happen as it should only exist if
+            # there's a syntax error in a specs tests, which would raise an error earlier than
+            # this code
+            ["p.d1.is_before("],
+            0,
+            "p.d1.is_before(",
+        ),
+    ],
+)
+def test_get_series_code(source_lines, source_index, expected):
```
rebkwok, line 106: I only just noticed that there is a `test_specs.py` which this test belongs in (as this is testing just a bit of `specs.py`, not the docs generation)

evansd (approved):
A nifty solution :+1: 

I wonder if further down the line (depending on how complex these tests get), we'll end up wanting to use the [ast](https://docs.python.org/3/library/ast.html) module to actually parse the test functions for us. But I don't think that's necessary at this point.

rebkwok:
> I wonder if further down the line (depending on how complex these tests get), we'll end up wanting to use the [ast](https://docs.python.org/3/library/ast.html) module to actually parse the test functions for us. But I don't think that's necessary at this point.

I did look at the ast module as an option, but decided not to go there quite yet :)

PR #619, opened 2022-07-12
docs: Add docs for spec tests

rebkwok:


tests/spec/README.md
```diff
@@ -0,0 +1,112 @@
+# Spec tests
+
+These tests specify the expected behaviour of ehrQL, and by extension, the QM.
```
StevenMaude, line 3: suggestion:

```suggestion
These tests specify the expected behaviour of ehrQL, and by extension, the query model.
```

We're not using this abbreviation elsewhere in this document, so I'd remove it to make the reader's life easier.

(I realise it was here already. But if we're improving the whole document in this PR, we might as well do this too.)

tests/spec/README.md
```diff
@@ -0,0 +1,112 @@
...
+    │   ├── aggregate_frame                       <-- chapter
+    │   │   ├── __init__.py                       <-- contains chapter title
+    │   │   ├── test_count_for_patient.py         <-- section
+    │   │   └── test_exists_for_patient.py        <-- section
+```
+
+### Titles
+Chapter titles are specified with a `title` attribute the test directory's `__init__.py`.
+Section titles are specified with a `title` attribute in each test file.
+Paragraph titles are extracted from the test names.
```
StevenMaude, line 31: suggestion: Maybe worth a quick mention that the default paragraph title can be overridden by adding a `title` attribute to the test function?

Iain added this feature in #602.

tests/spec/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+## Test structure
+
+Spec tests follow a standardised structure in order to allow documentation generation.  Each
+spec test uses the `spec_test` fixture, which has the following components:
+
+```
+spec_test(
+    table_data,                          # defines the tables in the test database
```
StevenMaude, line 52: thought: I read this and got momentarily disorientated by not knowing what `table_data` was.

It might be helpful to say "dictionary that defines the tables in the test database": you can't immediately see this is a dictionary here.

tests/spec/README.md
```diff
@@ -0,0 +1,112 @@
...
+        1: (101 + 102),
+        2: 201,
+    },
+)
+```
+
+### Defining table data
+Each test must define a `table_data` dict, with keys `e` (event level data, with multiple rows per patient) and/or `p` (patient level data, with one row per patient)
+
+This sets up the test data.  The first column is always patient ID.  Column names indicate the type of the column in the test database.
```
StevenMaude, line 64: question: Is there a constraint on the patient ID column type? Does it always have to be an integer? If so, worth mentioning.

tests/spec/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+E.g. The following table data dict sets up an event table with an integer column `i1` and a
+boolean column `b1`, with 2 rows for each of 2 patients, and a patient table with a date column
```
StevenMaude, line 67: question: Can you explain what's being counted here, and how, for the reader? It's a little ambiguous to me as written.

To my naive eyes, `b1` in table `e` looks like it has:

* three rows for patient ID `1`
* one non-null row and one null row for patient ID `2`, so two rows total

I can't quite see how `b1` has:

> 2 rows for each of 2 patients

so either it might benefit from explanation, or I'm misreading something here

StevenMaude (approved):
Despite all my comments — they're small points — this is really well explained :+1:

I'll approve and let @rebkwok decide what to do with the suggestions.

tests/spec/README.md
rebkwok: I think I added an additional row for patient 1 and forgot to update the description

rebkwok:
Thanks @StevenMaude.  I've addressed questions/suggestions and also added a section about how to include additional text with a Chapter/Section/Paragraph - no tests appear to use that yet, but it was also added recently, in #539

Issue #564, opened 2022-06-22
We should publish a `v1` tag for the databuilder docker image


evansd:
Initially, this tag would be identical to the `latest` tag. But it gives us the option to break backwards compatibility by freezing `v1` and starting to release to `v2`. The key thing is that all projects should use `databuilder:v1` as the image reference rather than `databuilder:latest` (which should be hard-banned) and that will give us the future flexibility we need.

This is the same approach we used for the [research-action](https://github.com/opensafely-core/research-action) and it served us well there.

Exactly how we apply the same technique using docker tags rather than git tags is still an open question I think.

StevenMaude:
To clarify, we deliberately don't push a `latest` tag for the `databuilder` image to simplify its versioning:

```bash
$ docker pull ghcr.io/opensafely-core/databuilder:latest
Error response from daemon: manifest unknown
```

We made this change when we moved to publishing `databuilder` images instead of just `cohortextractor-v2`. `cohortextractor-v2` does have a now "old" latest tag, but [we'll be rid of that image soon anyway](https://github.com/opensafely-core/job-runner/issues/448).

But, slightly confusingly, [some of the tests in various bits of our code **do** refer to `databuilder:latest`](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Aopensafely-core+%22databuilder%3Alatest%22), so perhaps these should be changed to reflect actual usage.

StevenMaude:
(I'd forgotten that the code search preview wasn't just available for everyone globally, but, yes, that was the search I was going for.)

rebkwok:
> To clarify, we deliberately don't push a latest tag for the databuilder image to simplify its versioning

Does this mean we actually _don't_ want to tag the image as v1?

evansd:
> Does this mean we actually don't want to tag the image as v1?

I might not be quite following here. I thought the idea was that we have a tag called `v1` which, for the moment, behaves exactly like `latest` does on cohortextractor. i.e. it always points to the most recent build of the image. But unlike `latest` this gives us the option further down the road to stop updating the `v1` tag and start publishing a `v2` tag instead —which means we can safely break backwards compatibility.

StevenMaude:
[...]

PR #623, opened 2022-07-13
Add Docker labels to override the defaults

StevenMaude:
The defaults are set in the `base-docker` image.

As mentioned in #564, we want this `databuilder` image to appear in the
`databuilder` repository.

Based on https://github.com/opensafely-core/cohort-extractor/commit/c9b0dfb6f2578e436ee787a6a9b148ce13bc357b

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,15 @@
 # hadolint ignore=DL3007
 FROM ghcr.io/opensafely-core/base-docker:latest
 
+# Some static metadata for this specific image, as defined by:
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+# The org.opensafely.action label is used by the jobrunner to indicate this is
+# an approved action image to run.
+LABEL org.opencontainers.image.title="databuilder" \
+      org.opencontainers.image.description="Data Builder action for opensafely.org" \
+      org.opencontainers.image.source="https://github.com/opensafely-core/databuilder" \
+      org.opensafely.action="databuilder"
+
 # hadolint ignore=DL3008
 RUN \
   apt-get update --fix-missing && \


Dockerfile
```diff
@@ -1,6 +1,15 @@
...
 FROM ghcr.io/opensafely-core/base-docker:latest
 
+# Some static metadata for this specific image, as defined by:
+# https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
+# The org.opensafely.action label is used by the jobrunner to indicate this is
+# an approved action image to run.
+LABEL org.opencontainers.image.title="databuilder" \
+      org.opencontainers.image.description="Data Builder action for opensafely.org" \
+      org.opencontainers.image.source="https://github.com/opensafely-core/databuilder" \
+      org.opensafely.action="databuilder"
```
StevenMaude, line 11: Is this line definitely required? It was included in the `cohortextractor` image configuration that I link to in the commit message, so I've included it.

But I can't quite see where this is checked in job-runner.

There is this [`ALLOWED_IMAGES`](https://github.com/opensafely-core/job-runner/blob/4041d0d7442c8a385c23c9a6d9f0a37e1ceebc0c/jobrunner/config.py#L48-L56) configuration — but I'm not convinced this label is being checked. I could be mistaken.

evansd: Good question. I suspect this might have been aspirational at the time it was written.

StevenMaude: Checked with @bloodearnest and he said we're not using this now, but could in future to:

* ditch the `ALLOWED_IMAGES` list and use this label to determine which images are allowed
* restrict `docker image prune` to these images in `opensafely-cli`

So I'll leave it.

PR #638, opened 2022-07-18
Add utility scripts for working locally with SQLite

evansd:
This adds a pair of new scripts which are desigend to make it easier to
actaully run databuilder locally without needing to set up a database
server and create all the necessary tables.

`create-schema.py` takes a dataset definition and the path to a SQLite
file (which may or may not exist yet) and will create any tables
referenced by the dataset definition.

`generate-dataset-sqlite.py` is a thin wrapper around `generate-dataset`
which runs it against the supplied SQLite database and writes its output
as CSV to stdout.

We still need tooling to make it easy to populate this database, but
this is a step forward.


inglesp (approved):
Nice.

PR #647, opened 2022-07-20
fix: Create documentation PR if docs change

rebkwok:
This PR moves the checking and updating of the automatically generally documentation entirely to this repo (see also https://github.com/opensafely/documentation/pull/881).  

[Workflow run](https://github.com/opensafely-core/databuilder/runs/7433064373?check_suite_focus=true
) with changes needed to public docs since previous release (this was triggered on pushes to this update-docs-workflow branch)

The workflow created [this release](https://github.com/opensafely-core/databuilder/releases/tag/docs-metadata-9ffd278abb5c5915965522d6333dc9942cb3bb9e)
and [this PR](https://github.com/opensafely/documentation/pull/877) in the docs repo, which has triggered all the necessary PR workflows.

[This](https://github.com/opensafely-core/databuilder/runs/7432988181?check_suite_focus=true) is a previous run, testing when the docs hadn't changed since the latest release.

### DIFF
--- a/.github/workflows/update-public-docs.yml
+++ b/.github/workflows/update-public-docs.yml
@@ -63,7 +63,6 @@ jobs:
 
       - name: Upload Release Asset
         if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
-        id: upload_release_asset
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -84,7 +83,7 @@ jobs:
 
       - name: Check if update branch exists
         if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
-        id: check-branch
+        id: check_branch
         run: |
           cd docs
           echo ::set-output name=branch-exists::`git ls-remote --heads origin ${BRANCH}`
@@ -95,7 +94,7 @@ jobs:
           cd docs
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
-          if [ -n "${{ steps.check-branch.outputs.branch-exists }}" ]; then
+          if [ -n "${{ steps.check_branch.outputs.branch-exists }}" ]; then
             echo "Checking out existing branch"
             git fetch origin $BRANCH
             git checkout $BRANCH
@@ -117,7 +116,6 @@ jobs:
       # Open a PR for this branch, if one isn't already open
       - name: Open PR
         if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
-        id: list-pr
         uses: actions/github-script@v6
         env:
           HEAD: "opensafely:${{env.BRANCH}}"
@@ -149,3 +147,5 @@ jobs:
                 base: 'main',
               });
             }
+
+            console.log(results.data.url)
--- a/.github/workflows/update-public-docs.yml
+++ b/.github/workflows/update-public-docs.yml
@@ -8,12 +8,15 @@ on:
       - Build and publish image
     branches:
       - main
-    status:
+    types:
       - completed
 
 jobs:
   update-public-docs:
     runs-on: ubuntu-latest
+
+    env:
+      BRANCH: update-docs-${{ github.sha }}
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/update-public-docs.yml
```diff
@@ -59,31 +62,90 @@ jobs:
...
+          mv new_public_docs.json docs/public_docs.json
+          cd docs
+          git add public_docs.json
+          git commit -m "Update databuilder docs"
+          git push origin $BRANCH
+
+      # Open a PR for this branch, if one isn't already open
+      - name: Open PR
+        if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
+        id: list-pr
```
StevenMaude, line 120: question: Are we using this `id` anywhere?

.github/workflows/update-public-docs.yml
```diff
@@ -59,31 +62,90 @@ jobs:
...
-      - name: Trigger update in docs repo
-        if: ${{ steps.check_docs.outputs.DOCS_CHANGED }}
+      - name: Clone docs repo
+        if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
+        uses: actions/checkout@v3
+        with:
+          repository: opensafely/documentation
+          path: docs
+          ref: 'main'
+          token: ${{ secrets.DOCS_WRITE_TOKEN }}
```
StevenMaude, line 83: comment: I've just merged opensafely/documentation#832 which switches cohort-extractor to now being a submodule. Submodules aren't cloned by default with `actions/checkout@v3`.

I don't believe that will make a difference here, because we're not building the docs, but just mentioning in case.

.github/workflows/update-public-docs.yml
```diff
@@ -59,31 +62,90 @@ jobs:
           prerelease: false
 
       - name: Upload Release Asset
-        if: ${{ steps.check_docs.outputs.DOCS_CHANGED }}
-        id: upload-release-asset
+        if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
+        id: upload_release_asset
```
bloodearnest, line 66: Do we still need the release asset, since we're committing the change?

StevenMaude (approved):
Looks good! Thanks for polishing this process up some more :sparkles: :+1:

.github/workflows/update-public-docs.yml
```diff
@@ -59,31 +62,90 @@ jobs:
...
+          cd docs
+          echo ::set-output name=branch-exists::`git ls-remote --heads origin ${BRANCH}`
+
+      - name: Create or checkout branch
+        if: ${{ steps.check_release_changed.outputs.DOCS_CHANGED }}
+        run: |
+          cd docs
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          if [ -n "${{ steps.check-branch.outputs.branch-exists }}" ]; then
```
bloodearnest, line 98: I think you could get rid of the previous step by embedding the check from it here
```suggestion
          if git ls-remote --heads origin ${BRANCH} >/dev/null 2>&1; then
```

bloodearnest (approved):

.github/workflows/update-public-docs.yml
rebkwok: No, that shouldn't matter

.github/workflows/update-public-docs.yml
rebkwok: No - it's a leftover from a previous version of the workflow

bloodearnest (approved):

.github/workflows/update-public-docs.yml
```diff
@@ -59,31 +62,85 @@ jobs:
...
+          path: docs
+          ref: 'main'
+          token: ${{ secrets.DOCS_WRITE_TOKEN }}
+
+      - name: Create or checkout branch
+        run: |
+          cd docs
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          if [[ $(git ls-remote --heads origin $BRANCH) ]]; then
```
bloodearnest, line 88: nit: am interested in why you went with a subshell and output capturing here?

`if cmd` will branch on whether the `cmd` exits with a 0 or not. `git ls-remote` exits with a 1 if the branch doesn't exist, so the naked command should be all we need?

Not necessarily a problem, but if you want to check the output of `git ls-remote`, I think it's better spelt as `if test -n "$(git ls-remote ..)"`. Bash's [[ is weird and surprising (how it's implemented will horrify you), and I think "test -n" (which would be test --nonzerooutput if test did long options) is more explicit that [['s default.

But feel free to ignore this!

rebkwok: The suggestion you made on the previous PR didn't work (it always ended up evaluating true), and my bash knowledge is only good enough to trial and error find a thing that works

PR #660, opened 2022-08-03
Add rudimentary batch fetching support

evansd:
This adds bare bones support to the MSSQL engine for writing the results of query to a temporary table and then fetching them in batches.

As yet, it makes no attempt to gracefully handle errors and retry; but it should hopefully be enough to attempt some work with "real sized" data.

### DIFF
--- a/tests/unit/test_sqlalchemy_utils.py
+++ b/tests/unit/test_sqlalchemy_utils.py
@@ -5,6 +5,7 @@ from sqlalchemy.dialects.sqlite.pysqlite import SQLiteDialect_pysqlite
 from databuilder.sqlalchemy_utils import (
     GeneratedTable,
     clause_as_str,
+    fetch_table_in_batches,
     get_setup_and_cleanup_queries,
     is_predicate,
 )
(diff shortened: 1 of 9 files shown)


tests/unit/test_sqlalchemy_utils.py
```diff
@@ -133,11 +133,18 @@ def test_clause_as_str():
...
     table = sqlalchemy.Table("foo", sqlalchemy.MetaData(), sqlalchemy.Column("bar"))
     index = sqlalchemy.Index(None, table.c.bar)
     create_index = sqlalchemy.schema.CreateIndex(index)
     dialect = SQLiteDialect_pysqlite()
+
+    # Compiling CreateIndex to a string using literal_binds blows up without a special
+    # workaround. I think this might just be a bug in SQLAlchemy, so if this suddenly
+    # starts working we want to know so we can remove our workaround.
+    with pytest.raises(TypeError, match="unexpected keyword argument 'literal_binds'"):
+        create_index.compile(dialect=dialect, compile_kwargs={"literal_binds": True})
```
inglesp, line 147: It's not clear to me what the workaround is, so could you add a link?

inglesp (approved):

PR #663, opened 2022-08-08
Add class-based schema definitions

evansd:
This PR replaces the old syntax for defining ehrQL schemas e.g.
```py
patients = build_patient_table(
    "patients",
    {
        "sex": str,
        "date_of_birth": datetime.date,
    }
)
```

with a class-based one that looks like this:
```py
@construct
class patients(PatientFrame):
    sex = Series(str)
    date_of_birth = Series(datetime.date)
```

There are some obvious peculiarities here (the decorator and the lowercase class name) which we'll come to later, but the primary motivation for this change is to make it possible to have full [autocomplete](https://github.com/opensafely-core/databuilder/issues/506) for ehrQL when writing dataset definitions in VSCode. As far as I can see, this is only possible if we use classes to define the schema because no other syntax makes the column names available.

This change gives us autocomplete on table and column names, and also on frame methods like `exists_for_patient`, `take` and `drop`. It doesn't yet give us autocomplete on series methods but I have a [proof-of-concept](https://github.com/opensafely-core/databuilder/issues/506#issuecomment-1208411707) which shows that this ought to be possible.

The `@construct` decorator, and the lowercase class name are to avoid having to write something like this:
```py
class Patients(PatientFrame):
    __tablename__ = "patients"
    sex = Series(str)
    date_of_birth = Series(datetime.date)

patients = Patients()
```
Not only would that be more verbose and error prone, but it would leave the `Patients` class in the namespace where it would show as a (misleading) autocomplete option.

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -438,22 +438,6 @@ class SortedEventFrame(BaseFrame):
 #
 
 
-def build_patient_table(name, schema, contract=None):
-    if contract is not None:
-        contract.validate_schema(schema)
-    return PatientFrame(
-        qm.SelectPatientTable(name, schema=qm.TableSchema(schema)),
-    )
-
-
-def build_event_table(name, schema, contract=None):
-    if contract is not None:  # pragma: no cover
-        contract.validate_schema(schema)
-    return EventFrame(
-        qm.SelectTable(name, schema=qm.TableSchema(schema)),
-    )
-
-
 class SchemaError(Exception):
     ...
 
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -367,6 +367,9 @@ class BaseFrame:
         self.qm_node = qm_node
 
     def __getattr__(self, name):
+        return self._select_column(name)
+
+    def _select_column(self, name):
         return _wrap(qm.SelectColumn(source=self.qm_node, name=name))
 
     def exists_for_patient(self):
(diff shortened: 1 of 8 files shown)


databuilder/query_language.py
```diff
@@ -454,6 +454,58 @@ def build_event_table(name, schema, contract=None):
...
+    ...
+
+
+# A class decorator which replaces the class definition with an appropriately configured
+# instance of the class. Obviously this is a _bit_ odd, but I think worth it overall.
+# Using classes to define tables is (as far as I can tell) the only way to get nice
+# autocomplete and type-checking behaviour for column names. But we don't actually want
+# these classes accessible anywhere: users should only be interacting with instances of
+# the classes, and having the classes themselves in the module namespaces only makes
+# autocomplete more confuisng and error prone.
```
iaindillingham, line 451:
```suggestion
# autocomplete more confusing and error prone.
```
🙂

databuilder/tables.py
```diff
@@ -1,15 +1,10 @@
...
-        "date_of_death": datetime.date,
-        "sex": str,
-    },
-    contract=universal.Patients,
-)
+@construct
+class patients(PatientFrame):
+    date_of_birth = Series(datetime.date)
+    date_of_death = Series(datetime.date)
+    sex = Series(str)
```
iaindillingham, line 10: This looks much nicer 🙂

iaindillingham (approved):

PR #667, opened 2022-08-09
Add CSV backed query engine

evansd:
This adds a subclass of the SQLite engine which takes as its DSN (aka `DATABASE_URL`) a path to a directory of CSV files and reads these files into an in-memory database before running the query. This is designed for demonstration/tutorial purposes and not for production use (obvs.).

The contents of each table is expected to be in a CSV file named `<table_name>.csv`.

As the DSN can currently only be supplied as an environment variable, the syntax for invoking this is a little awkward:
```
DATABASE_URL=path/to/my/csvs python -m databuilder generate-dataset \
  --dataset-definition my_definition.py \
  --output /dev/stdout
```

I'm working on a separate PR to address this awkwardness and make the CLI a bit more ergonomic.

As a bonus feature, if you set the environment variable `DATABUILDER_CREATE_MISSING_CSVS=True` then, rather than throwing an error if any expected CSV files are not found, it will create empty CSVs with the appropriate headers.


StevenMaude:
The "awkwardness" as @evansd described it is fine for me for the time being :+1:

***

Two quick usability things which he may already be aware of:

1. As the PR stands, the stdout gets interleaved with stderr, but not a big deal.

```
$ DATABASE_URL=example-data python -m databuilder generate-dataset --dataset-definition dataset_definition_minimal.py  --output /dev/stdout
2022-08-10 14:36:27 [info     ] Generating dataset for dataset_definition_minimal.py [databuilder.main] 
patient_id,date_of_birth
2022-08-10 14:36:27 [info     ] Fetching results               [databuilder.query_engines.base_sql] 
2,2005-10-12
```

2. If a file is missing without the "create blank CSVs" option, you get a traceback. There could be an exception check in `orm_instances_from_csv()`. But also not urgent as the actual error is clear.

The benefits of loading CSV data far outweigh any marginal polish, for working on dataset definitions locally. So I'd be pro-merging sooner rather than later.

***

The code looks OK to my naive eyes. I don't know enough about SQLAlchemy to spot any significant problems there.

StevenMaude (approved):

evansd:
Thanks @StevenMaude. Point 1 I've actually already addressed in #668 because it was annoying me too!

You're right about point 2 as well, but I think it's OK to leave for now.

PR #668, opened 2022-08-10
Trigger errors and log lines before creating output file

evansd:
Because the `results` object is a generator we don't actually execute
any queries until we start consuming it. It gives nicer behaviour if we
start consuming it before creating the output file and writing the
headers so that any errors (or relevant log output) comes first.

### DIFF
--- a/tests/unit/test_itertools_utils.py
+++ b/tests/unit/test_itertools_utils.py
@@ -1,0 +1,34 @@
+import pytest
+
+from databuilder.itertools_utils import eager_iterator
+
+
+def test_eager_iterator():
+    results = eager_iterator(iter([0, 1, 2]))
+    assert list(results) == [0, 1, 2]
+
+
+def test_eager_iterator_works_on_empty_iterators():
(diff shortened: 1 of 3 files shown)


tests/unit/test_itertools_utils.py
```diff
@@ -0,0 +1,34 @@
...
+        yield  # pragma: no cover
+
+    results = generator_with_error()
+    with pytest.raises(ValueError, match="^fail$"):
+        eager_iterator(results)
+
+
+def test_eager_iterator_still_mostly_lazy():
+    i = iter(range(3))
+    assert i.__length_hint__() == 3
```
iaindillingham, line 28: I had no idea this existed!

iaindillingham (approved):
If it helps, then I found this version of the commit much easier to comprehend than the previous version.

evansd:
Thanks Iain. I realised after the first version that I'd end up having to re-implement it when we added other output formats, so I went with this more general approach. Glad it makes more sense into the bargin!

PR #670, opened 2022-08-10
fix: Remove DateSubtractDays query model operation

### LINKED ISSUE #568: Remove `DateSubtractDays` query model operation
Given that we have `DateAddDays` and `Negate` I think this is redundant and I'd like to keep the query model as small as possible.

Note, I'm just talking about the query model here. The existing `subtract_days` method in ehrQL can stay for now, just with the implementation:
```py
return self.add_days(other.__neg__())
```

And then later both `add_days` and `subtract_days` can go when we do the ["date delta"](https://github.com/opensafely-core/databuilder/issues/519) work.

rebkwok:
Fixes #568

### DIFF
--- a/databuilder/population_validation.py
+++ b/databuilder/population_validation.py
@@ -218,11 +218,6 @@ def date_add_days(date, num_days):
     return date + datetime.timedelta(days=num_days)
 
 
-@register_op(Function.DateSubtractDays)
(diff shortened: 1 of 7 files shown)


evansd (approved):
All that lovely red :smile: Thanks!

Issue #646, opened 2022-07-20
It would be friendly of Data Builder to list available backends when you run a command without providing one


inglesp:
For example, currently running `dump-dataset-sql` without specifying a backend gives this:

```
❯ python -m databuilder dump-dataset-sql --dataset-definition tests/acceptance/comparative_booster_study/dataset_definition.py 
usage: databuilder dump-dataset-sql [-h] --dataset-definition DATASET_DEFINITION [--output OUTPUT]
                                    backend
databuilder dump-dataset-sql: error: the following arguments are required: backend
error: Recipe `exec` failed on line 137 with exit code 2
```

Even as someone familiar with the codebase, I'm not sure if I should be giving the class name, the module name or some special registered name to identify the codebase.

inglesp:
Originally added as a draft by @benbc.

evansd:
I think this is obviated by #608.

There's a separate question about making backends and query engines easier to specify by not always requiring the full dotted path (e.g. by defining aliases like `tpp` -> `databuilder.backends.tpp.TPPBackend`).

benbc:
> I think this is obviated by https://github.com/opensafely-core/databuilder/pull/608.

I'm not sure about that. How does the user determine what the valid values of `--backend` and `--query-engine` are?

evansd:
Yes, there is something to be done here — although literally as written the ticket doesn't apply anymore because databuilder no longer throws an error when you don't provide a backend.

But I think it would be good to provide short aliases for the built-in backends/query engines, and enumerate these to the user, while also making clear that any importable dotted path can be used as well.

Issue #580, opened 2022-06-28
Rename `OPENSAFELY_BACKEND` env var to `DATABUILDER_BACKEND`


evansd:
As [noted](https://github.com/opensafely-core/databuilder/pull/576#issuecomment-1167540896) the `OPENSAFELY_BACKEND` env var is, although undocumented, being relied on by [user code](https://github.com/search?l=&q=org%3Aopensafely+OPENSAFELY_BACKEND&type=code) because there's currently no alternative.

It would make the process of moving from cohortextractor to databuilder easier if databuilder used a different environment variable. It's also not totally unreasonable to say that all databuilder environment variables should be prefixed `DATABUILDER_`.

evansd:
Since adding backend "aliases" in
 * #671 
We can actually share the same environment variable between cohortextractor and databuilder because the same value can work with both e.g.
```
OPENSAFELY_BACKEND=tpp
```
It's therefore no longer obligatory to make this change before deploying.

There's still an open question as to whether it would be better to use the `DATABUILDER_` prefix. But we don't have to resolve that immediately.

benbc:
@evansd What's your latest thinking on this?

If we go with `DATABUILDER_BACKEND`, does that mean that we need to change Job Runner to set that as well?

I'm keen to address this soon (if we do it at all) to stabilize the interface between Data Builder and the rest of OpenSAFELY.

evansd:
I think I'd be inclined to stick with `OPENSAFELY_BACKEND`. We would, as you say, need to change Job Runner otherwise and add some extra complexity there, and it's not clear we'd gain anything by it.

PR #671, opened 2022-08-10
Make various CLI improvements

evansd:
This PR is a bit of a grab bag of various things. See the commit sequence for the full details, but the highlights are:
 * Fix package script entrypoint so you can run `databuilder ...` locally instead of `python -m databuilder ...`.
   Note: you'll need to run `pip install --editable .` first.
 * Do away with `--dataset-definition` so you can just write: `generate-dataset path/to/file.py`.
 * Add `--query-engine` and `--backend` arguments to `generate-dataset` so you don't have to use environment variables.
 * Add a `--dsn` argument to `--generate-dataset` instead of having to set `DATABASE_URL`. 
 * Write CSV to stdout by default if no `--output` argument is supplied.
 * Raise helpful errors if you supply the wrong thing to `--backend` or `--query-engine` (see [#566](https://github.com/opensafely-core/databuilder/issues/566)).
 * Add short aliases for the bundled backends and query engines so you don't have to specify the full dottted path every time.
 * List the available backends and query engines in help/error messages (see [#646](https://github.com/opensafely-core/databuilder/issues/646)).

The upshot is that you should hopefully be able to generate a dataset using the local CSV engine with just:

    databuilder generate-dataset path/to/file.py --dsn path/to/csvs

And you could see the SQL that will be run against the TPP backend using:

    databuilder dump-dataset-sql path/to/file.py --backend tpp

### DIFF
--- a/databuilder/__main__.py
+++ b/databuilder/__main__.py
@@ -32,8 +32,8 @@ def main(args, environ=None):
                 definition_file=options.dataset_definition,
                 dataset_file=options.output,
                 db_url=database_url,
-                backend_id=environ.get("OPENSAFELY_BACKEND"),
-                query_engine_id=environ.get("OPENSAFELY_QUERY_ENGINE"),
+                backend_id=options.backend,
+                query_engine_id=options.query_engine,
                 environ=environ,
             )
         elif dummy_data_file:
@@ -100,6 +100,16 @@ def add_generate_dataset(subparsers, environ):
         "--dummy-data-file",
         help="Provide dummy data from a file to be validated and used as the dataset",
         type=Path,
+    )
+    parser.add_argument(
+        "--query-engine",
+        type=str,
+        default=environ.get("OPENSAFELY_QUERY_ENGINE"),
+    )
+    parser.add_argument(
+        "--backend",
+        type=str,
+        default=environ.get("OPENSAFELY_BACKEND"),
     )
 
 
--- a/databuilder/__main__.py
+++ b/databuilder/__main__.py
@@ -24,23 +24,23 @@ def main(args, environ=None):
     options = parser.parse_args(args)
 
     if options.which == "generate-dataset":
-        database_url = environ.get("DATABASE_URL")
-        dummy_data_file = options.dummy_data_file
-
-        if database_url:
+        if options.dsn:
             generate_dataset(
                 definition_file=options.dataset_definition,
                 dataset_file=options.output,
-                db_url=database_url,
+                db_url=options.dsn,
                 backend_id=options.backend,
                 query_engine_id=options.query_engine,
                 environ=environ,
             )
-        elif dummy_data_file:
-            pass_dummy_data(options.dataset_definition, options.output, dummy_data_file)
+        elif options.dummy_data_file:
+            pass_dummy_data(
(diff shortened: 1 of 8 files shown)


databuilder/__main__.py
```diff
@@ -81,21 +100,23 @@ def build_parser(environ):
...
         "--output",
-        help="Path and filename (or pattern) of the file(s) where the dataset will be written",
+        help="Path of the file where the dataset will be written (console by default)",
         type=Path,
     )
+    parser.add_argument(
+        "--dsn",
+        help="Data Source Name: URL of remote database, or path to data on disk",
+        type=str,
+        default=environ.get("DATABASE_URL"),
```
StevenMaude, line 112: question: Should this environment variable name be changed for consistency with the fact it might not be a URL?

StevenMaude: If we change it here, maybe we could be consistent with the other variable names and use an `OPENSAFELY` prefix. That would:

* avoid risk of collision with any other software — should that software similarly use generic names
* makes it obvious what the environment variable is for
* makes it easier to remember, instead of a special case

evansd: I deliberately didn't make that change because it will require coordination with job-runner and there's a [bigger question](https://github.com/opensafely-core/databuilder/issues/580#issuecomment-1211754690) there about whether we should use `OPENSAFELY_` or `DATABUILDER_` as our env var prefix that we ought to settle first.

`DATABASE_URL` is something of a de-facto standard so you could frame this as a risk of collision, or as having things work nicely out of the box :smile:

Maybe the thing to do is close #580 and make a new issue to consider all our env var names together?

StevenMaude: Or maybe name them `DATABUILDER_…` even, as [in @evansd' comment elsewhere](https://github.com/opensafely-core/databuilder/issues/580#issue-1287217034).

StevenMaude: It doesn't need to block this PR in any case, again, as this PR has some useful usability improvements.

StevenMaude:
I think this is actually "fixes #646". Also *possibly* fixes #566; but I can always try deliberately breaking the CLI again.

evansd:
Yeah, I put `Closes #646` in the relevant commit so hopefully that will happen automatically. I wasn't sure if there was more to be covered for #566 so I thought I'd let you decide whether to close it or not.

StevenMaude:
> Yeah, I put `Closes #646` in the relevant commit so hopefully that will happen automatically. I wasn't sure if there was more to be covered for #566 so I thought I'd let you decide whether to close it or not.

Ah, great; I missed that!

For me, it's fine to close #566 too since the specifics have been dealt with here. Better that probably than leaving the issue dangling and never getting pruned. It's probably preferable to create a new, more specific issue if we do encounter any more tracebacks.

StevenMaude:
I'll just give the CLI a quick try, and think this is ready to go then.

StevenMaude:
I've given the new CLI features a go and look like they're working as intended.

StevenMaude (approved):
@evansd: Thanks for these improvements! Along with the recently added option for CSV input, it's made Data Builder much simpler to use :+1:

Issue #665, opened 2022-08-09
Borrow Dockerfile improvements from cohortextractor


evansd:
Simon has done lots of work on the cohortextractor Dockerfile to give it nice caching/layering behaviour, support fast builds and produce compact production images without lots build-time cruft. It also uses the `base-action` rather than `base-docker` image, which I don't think makes much difference right now but is important for maintaining consistency.

We should borrow the approach here for Databuilder.

**cohortextractor `Dockerfile`**:
https://github.com/opensafely-core/cohortextractor/blob/main/Dockerfile

**databuilder `Dockerfile`**:
https://github.com/opensafely-core/databuilder/blob/main/Dockerfile

Lots of this can be copied over directly with just changes in names. The major differences are:
 * We want to use Python 3.9, not the default Python
 * We don't (yet) need to use pyODBC so we don't need the Microsoft apt repos, the unix-odbc dev headers, or the `ACCEPT_EULA` env var.
 * We use `pyproject.toml` rather than `setup.py`

rebkwok:
Progress so far:
I have 2 draft PRs that are similar but slightly different in what they do with the the pyproject.toml.

I think the 2nd one is probably what we want, but both are there in case it's useful to see what I've tried so far:

1) https://github.com/opensafely-core/databuilder/pull/773
This follows the way cohort-extractor's Dockerfile works, including installing from the pyproject.toml and making the executable databuilder available.
BUT - that didn't work with flit, so I had to make changes to the pyproject.toml file, which then means that pip-compile isn't working properly. So this builds the docker image correctly, afaict, and it's pretty quick about it, and changes to app code don't take long to rebuild, but the pyproject.toml is broken for other things. At the moment I don't think we're actually building a databuilder package, but I assume we probably will in future.

2) https://github.com/opensafely-core/databuilder/pull/774
This updates the Dockerfile with similar improvements to those in cohort-extractor, but doesn't pip install using the pyproject.toml. It keeps the entrypoint we had before (python3.9 -m -B databuilder) but moves it out into an entrypoint script so it can be set as the `ACTION_EXEC` env and passed as the entrypoint to the base-action.

Both of these versions work to run a project locally, by:
- building the image with `just build-databuilder-for-os-cli`
- running the [test-age-distribution](https://github.com/opensafely/test-age-distribution) project with os-cli (pointing the action in project.yaml at `databuilder:dev`

[...]

Issue #651, opened 2022-07-28
Latest tag is `vnull` and shouldn't be


StevenMaude:
There was a recent change to how we retrieve the previous tag: 90e0aa7c9ecc333fec84badab9fc63b2eb336c0c.

Somehow this has ended up as `null`. This means we have:

* a `vnull` [tag](https://github.com/opensafely-core/databuilder/releases/tag/vnull)
* a `vnull` [Docker image](https://github.com/opensafely-core/base-docker/pkgs/container/databuilder/29274340?tag=vnull) 

It's possible that there was a temporary outage to GitHub's API, perhaps.

If that was the cause, and there's no way to make this more robust, then we should probably fail at that step if we can't resolve a valid tag.

StevenMaude:
It might be that there's only a release if `public_docs.json` updates, perhaps?

rebkwok:
The issue actually seems to be in the tagging action:
https://github.com/opensafely-core/databuilder/runs/7469012936?check_suite_focus=true#step:3:113

It's setting the `previous_tag` to the last tag, which is a `docs-metadata-xxx` one, and as a result it doesn't parse a previous version, and gets `null` instead.  I had thought that the `tag_prefix: v` [setting](https://github.com/opensafely-core/databuilder/runs/7469012936?check_suite_focus=true#step:3:6) meant that it would parse only tags starting withthe prefix, but apparently it just means it'll label them with it.  

There is an [issue](https://github.com/mathieudutour/github-tag-action/issues/71) in the action repo about confusion with other tags which seems related, although not identical to ours.  (Issue was opened in March 2021 and hasn't had any response from the repo owner though.)

The docs tag and release isn't essential now.  Previously we were using it to upload the public.json file as a release asset which the docs repo workflow would have access to.  However, now we're pulling the docs repo in the databuilder workflow in order to check if the docs have been updated and a new docs PR is needed.  We could manually tag the latest commit that should be tagged with a v0.x.x tag, and get rid of the release part of the docs workflow.

PR #676, opened 2022-08-15
Support `databuilder --version` on CLI

### LINKED ISSUE #624: Support `databuilder --version` on CLI
Unless anyone has better suggestions, this could be done in the same way as we current do for `opensafely-cli`.

1. Have a file at `databuilder/VERSION` whose default contents is `not-from-a-package` or `development` or something like that. (Bonus points if there's something approximating a convention we can follow here.)

2. Read that file at runtime and wire it up to the `--version` argument.

   https://github.com/opensafely-core/opensafely-cli/blob/a9efc8119a979a16fef1dd9ce1e470adad1be2c2/opensafely/__init__.py#L10
   ```py
   ```
   https://github.com/opensafely-core/opensafely-cli/blob/a9efc8119a979a16fef1dd9ce1e470adad1be2c2/opensafely/__init__.py#L21-L23
   ```py
       parser.add_argument(
           "--version", action="version", version=f"opensafely {__version__}"
       )
   ```

3. Have the action which builds the package write the correct version to the file before building.
   https://github.com/opensafely-core/opensafely-cli/blob/a9efc8119a979a16fef1dd9ce1e470adad1be2c2/.github/workflows/build_and_publish.yaml#L44-L46
   ```yaml
    - name: Generate correct value for VERSION file
      run: |
        echo ${{ needs.tag-new-version.outputs.tag }} > opensafely/VERSION
   ```

iaindillingham:
The build backend (`setuptools`/`setuptools_scm`) extracts the version from a tag when building the package, and writes the version to the package's metadata following [`setuptools_scm`'s default versioning scheme][1]. (Because we use tags for versioning both software (`vx.y.z`) and metadata (`docs-metadata-<sha>`) we have to override the call to `git describe`.)

The package is installed when creating the image. However, to ensure the tags are available, git is installed and the .git directory is bind-mounted.[^1] There are better ways of building an image from a Dockerfile; implementing them is the subject of #665.

Closes #624

[1]: https://github.com/pypa/setuptools_scm/#default-versioning-scheme

[^1]: https://github.com/pypa/setuptools_scm/#usage-from-docker


ghickman:
@iaindillingham – I think we have to install the package in the container if we want version numbers here, that's core to how setuptools_scm works.  Given we want to version with git tags (and, I'm assuming, _not_ commit a version file to the repo as part of that workflow) then we have to at the very least build and install.

evansd:
I think we want to install the package in the container in any case. Although it will need the fixes in #671 merged before that will actually work.

There's also ticket #665 open on improving the whole Dockerfile approach based on the one Simon ended up using for cohortextractor.

StevenMaude:
Tagging is also broken at the moment (#651), which may impact this (or at least make it not that useful until fixed).

ghickman:
The Python packaging parts of this look good to me.  Probably best to have someone with more docker knowledge than myself check through the Dockerfile though.

ghickman (approved):

bloodearnest (approved):
It would be good to add the version as a docker label too, so it's inspectable from outside easily. e.g. I would like to add the git ref to the metrics I'm currently working on.

I've been trying to add this as GITREF label, e.g. for the python action image:
https://github.com/opensafely-core/python-docker/blob/main/Makefile#L7
https://github.com/opensafely-core/python-docker/blob/main/Dockerfile#L65

Writing this comment has made me realise that the label is GITREF missing in cohortextractor's docker file.

Not blocking, just a suggestion
I've realised this conventions
For the cohortextractor image:

.dockerignore
```diff
@@ -1,5 +1,3 @@
-.git/
```
evansd, line 1: Don't have time to do a full review right now but, as a quick drive by comment, this doesn't look quite right to me. Doesn't it mean we end up bundling the entire git repo into the Docker image? Which I'm pretty sure we don't want to do.

bloodearnest: It's bind mounted at build time on line 33 of the diff

iaindillingham: Yes, it's bind-mounted: we need to make the repo available, to extract the version from a tag (see the [_Usage from Docker_][1] section of `setuptools_scm`'s documentation).

[1]: https://github.com/pypa/setuptools_scm/#usage-from-docker

evansd: Ah yes. And we never do `COPY .` so it wouldn't get included anyway. Ignore me!

iaindillingham:
You'll find this, buried on line [932][1] of the test failure:

```
  Step 12/13 : RUN --mount=type=bind,target=/app/.git,source=.git python -m pip install --no-cache-dir /app
  ---------------------------- Captured stderr setup -----------------------------
  the --mount option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
```

The error, on line [396][2], relates to the `databuilder_image` fixture:

https://github.com/opensafely-core/databuilder/blob/e6bd8a47389b114002382108e8f6bbb2fc778a0d/tests/conftest.py#L118-L130

I think I need to add `DOCKER_BUILDKIT=1` as the zeroth element in the list that's passed to `subprocess.run`, to [enable BuildKit builds][3]. Is that correct, @bloodearnest? Or should I enable BuildKit builds everywhere?

[1]: https://github.com/opensafely-core/databuilder/runs/8232454546?check_suite_focus=true#step:5:933
[2]: https://github.com/opensafely-core/databuilder/runs/8232454546?check_suite_focus=true#step:5:397
[3]: https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds

iaindillingham:
I've pushed 9e41b5bb and 5a95c133, which build `databuilder` and `databuilder-dev` with BuildKit. I don't know what BuildKit is, but I do know that it allows us to bind-mount the `.git` directory. Could a Docker-maven take a look?

.github/workflows/build-and-deploy.yml
```diff
@@ -98,7 +98,7 @@ jobs:
           echo "MINOR=$MINOR"
           echo "PATCH=$PATCH"
 
-          docker build . --file Dockerfile \
+          DOCKER_BUILDKIT=1 docker build . --file Dockerfile \
```
bloodearnest, line 101: Might be cleaner if you just do

```
export DOCKER_BUILDKIT=1
docker build ...

```

I think that's easier to read, and means if we do other docker builds later.

The way we do this in other projects is have a `just docker-build` command that takes care of any environment variables, and use that in the gh workflow.

iaindillingham: Thanks @bloodearnest. I don't want to extend this to editing the workflow too much, but have made the change you suggested in 7fe47369.

PR #679, opened 2022-08-17
test: Generate patient- and event-level tables differently

inglesp:
This is to support future changes in the in-memory database, which will
need to be able to distinguish between patient- and event-level tables.

### DIFF
--- a/tests/generative/data_setup.py
+++ b/tests/generative/data_setup.py
@@ -14,10 +14,10 @@ def setup(schema, num_patient_tables, num_event_tables):
     base_class = sqlalchemy.orm.declarative_base()
 
     patient_table_names, patient_classes = _build_orm_classes(
-        "p", num_patient_tables, schema, base_class
+        "p", num_patient_tables, schema, base_class, has_one_row_per_patient=True
     )
     event_table_names, event_classes = _build_orm_classes(
-        "e", num_event_tables, schema, base_class
+        "e", num_event_tables, schema, base_class, has_one_row_per_patient=False
     )
 
     all_patients_query = _build_query(patient_table_names, event_table_names, schema)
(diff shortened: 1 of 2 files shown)


inglesp:
@evansd could I have your views on whether this is a good idea?  My changes to the in-memory database require knowing whether a SQLAlchemy table is patient-level or event-level, and my initial hacks involved looking at the name of the table (`if sqla_table.name[0] == "e": ...` etc).  With these changes, we can instead check whether the `patient_id` column is a primary key or not.

A couple of questions:

* Should I be making similar changes to [similar code in the CSV QE](https://github.com/opensafely-core/databuilder/blob/main/databuilder/query_engines/csv.py#L69-L80)?
* There's nothing that I can see in the backends or the contracts that indicate whether a table or contract is patient-level or event-level.  Should there be?  If we're introducing the distinction in the tests, are we setting ourselves up for problems?

evansd:
This seems like a reasonable change to me. It one sense it must be _possible_ to structure the in-memory engine so it doesn't need to know the difference, because the SQL databases don't know the difference and still manage to do the right thing. But I think there's value in any case in setting up the primary keys correctly like this because it stops us inserting invalid test data. And if that ends up making the architecture of the in-memory engine simpler then that's all to the good.

> Should I be making similar changes to [similar code in the CSV QE](https://github.com/opensafely-core/databuilder/blob/main/databuilder/query_engines/csv.py#L69-L80)?

I think it would be sensible to keep them consistent. I went for copy-paste here because I didn't want to think about how best to move this code from `tests.lib` into `databuilder`. But if it's going to get more a lot more complicated then maybe we should do The Proper Thing here.

> There's nothing that I can see in the backends or the contracts that indicate whether a table or contract is patient-level or event-level. Should there be? If we're introducing the distinction in the tests, are we setting ourselves up for problems?

I think this is because the whole backends/contracts design is a bit wonky as it currently stands. I'm working on sorting this out.

inglesp:
> It one sense it must be _possible_ to structure the in-memory engine so it doesn't need to know the difference

It is, but I think the `apply_function` code comes out simpler if we do know the difference.

> maybe we should do The Proper Thing here.

Does it belong in `sqlalchemy_utils.py`, or somewhere else?

evansd:
> I think the `apply_function` code comes out simpler if we do know the difference.

Yeah, I can imagine. I think keeping the in-memory code as comprehensible as possible is an important goal so it's worth taking on a bit of complexity elsewhere.

> Does it belong in `sqlalchemy_utils.py`, or somewhere else?

I think ideally somewhere else. The various `*_utils.py` modules don't currently import (or otherwise expect) any application code and I'd prefer to keep them that way if possible.

Maybe `orm_factory.py` or something like that?

(Incidentally, I'm aware that the `databuilder` package is getting very cluttered and could do with some reorganisation. But I've been putting off doing it because it will be a bit disruptive and never seemed that high priority.)

evansd:
The ideas here were incorporated into #692 (as I understand it)

PR #703, opened 2022-09-08
feat: Add placeholder tutorial tables

StevenMaude:
This PR makes the `import` statements required in the documentation tutorial code work available.

It seems a reasonable time to introduce them now, though they may change or even be separated out from the existing contracts in future[^1], because:

* the addition to the existing code is small and self-contained
* merging should permit easier running of tutorial code either via the released Docker image and OpenSAFELY CLI (pending changes to the examples)

[^1]: We _might_ want to separate in future to insulate the tutorial from changes to existing contracts (which is likely). I'm undecided right now, and this isn't an immediate problem.

### DIFF
--- a/databuilder/tables/examples/tutorial.py
+++ b/databuilder/tables/examples/tutorial.py
@@ -1,0 +1,15 @@
+from databuilder.contracts.universal import patients
+from databuilder.contracts.wip import (
+    clinical_events,
+    hospitalisations,
+    patient_address,
+    prescriptions,
+)
+
+__all__ = [
+    "clinical_events",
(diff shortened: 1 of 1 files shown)


StevenMaude:
Still working on this in conjunction with the tutorial.

Will mark it ready for review once the tutorial is more stable, instead of having to constantly add PRs to update.

For now, I'm fine to run the tutorial checks against this branch.

StevenMaude:
This is now massively simplified from what it was. The tutorial dataset definitions build with this branch.

evansd (approved):

Issue #540, opened 2022-06-08
ehrQL should automatically cast between ints and floats where necessary


evansd:
The query model is strict about types (as it should be) and so will not let you write e.g.
```py
events.take(events.numeric_value > 10)
```
And instead insists on:
```py
events.take(events.numeric_value > 10.0)
```

This is fairly user-unfriendly and it would be better if it just automatically cast the int to a float here.

[This comment](https://github.com/opensafely-core/databuilder/issues/540#issuecomment-1380150744) contains implementation notes.

StevenMaude:
@evansd: The example given here is comparing a table value to a literal value.

Is the intent also to apply casting when working with columns?

(I've encountered this type issue and trying to explain now it in the tutorial. For now, I've put an explanation of the current state, and provided a link to this issue.)

evansd:
Yes, I think if we're going to do this at all we should do it consistently for values drawn from any source. Though my hunch is that this is going to most often used with literal values.

evansd:
We now have a mechanism in place (the `_cast` method, which was added to support date strings) which should allow us to support this fairly easily.


### Implementation

The implementation would look something like:
```py
class FloatFunctions(NumericFunctions):
    @staticmethod
    def _cast(value):
        if isinstance(value, int):
            return float(value)
        if isinstance(value, IntEventSeries, IntPatientSeries):
            return value.as_float()
        return value
```

And then the float [series definitions](https://github.com/opensafely-core/databuilder/blob/bb67580871f025f9df5cdcdbedc7381169788c8b/databuilder/query_language.py#L288-L293) would need changing to:
```py
class FloatEventSeries(FloatFunctions, NumericAggregations, EventSeries):
    _type = float


class FloatPatientSeries(FloatFunctions, PatientSeries):
    _type = float
```
i.e. replacing the generic `NumericFunctions` base class with `FloatFunctions`.


### Testing

As this behaviour concerns just the "ehrQL -> Query Model" translation layer it's not appropriate for the spec tests but should instead be tested in the query language unit tests. We want to test that given operands of certain types we get back a series of the type we expect (see [here](https://github.com/opensafely-core/databuilder/blob/bb67580871f025f9df5cdcdbedc7381169788c8b/tests/unit/test_query_language.py#L347-L385) for an example).


### Limitations

Note that not every form of float-int operation will be covered by this implementation. Everything in these two forms will work **(a)**:

    FloatSeries <operator> IntLiteral
[...]

PR #704, opened 2022-09-08
Support casting between ints and floats

### LINKED ISSUE #658: Support casting between ints and floats
Given that the query model is strict about types (to the point of not even letting you compare a float and an int) we need a way to cast between them.

In ehrQL terms we'd need to add methods called (something like) `as_int`, `as_float` to the `NumericFunctions` mixin:

The query model will need new operations:
```py
class CastToInt(Series[int]):
    source: Series[Numeric]

class CastToFloat(Series[float]):
    source: Series[Numeric]
```

```
databuilder/population_validation.py
tests/lib/in_memory/engine.py
```

It's a little bit clunky having an operation for each type, but I think the complexity involved into trying to make a generic `Cast` operation just isn't worth it for our use case.

There's a separate issue [here](https://github.com/opensafely-core/databuilder/issues/540) about casting automatically where the user's intention is obvious. I think we benefit overall from keeping the query model simple and strict and letting ehrQL handle these kind of UX affordances.

rebkwok:
Fixes #658 

This allows a numeric series to be explicitly cast to int or float.  It doesn't yet do any automatic casting (addressed in #540).

Also adds some additional spec tests for float series.


evansd (commented):
Fundamentally this all looks good. There's one thing you might want to change, depending on how you feel about it, and one small thing that does need changing. The wider question about testing strategy is one for a separate issue I think.

evansd (approved):
Looks great, thanks.

PR #705, opened 2022-09-08
Add public namespaces

### LINKED ISSUE #527: Placeholder: Decide what researchers need to import in order to write a dataset definition
(Was [Make it trivial for researchers to import exactly and only what they need to write dataset definitions](https://app.shortcut.com/ebm-datalab/story/441/make-it-trivial-for-researchers-to-import-exactly-and-only-what-they-need-to-write-dataset-definitions) in Shortcut.)

evansd:
This adds a `databuilder.ehrql` module which provides a "clean" namespace containing just those things users need in order to write dataset definitions.

Similarly, we add imports to `databuilder.tables` to provide the things necessary for defining tables and their constraints. This is less important as it's "provider-facing" rather than user-facing, but still worthwhile I think.

Codelist handling is still a bit of an open question (see #31) but whatever the final API I expect it to be available through `databuilder.ehrql`.

Closes #527

### DIFF
--- a/databuilder/main.py
+++ b/databuilder/main.py
@@ -9,10 +9,9 @@ import structlog
 
 from databuilder.column_specs import get_column_specs
 from databuilder.itertools_utils import eager_iterator
-from databuilder.query_language import Dataset
+from databuilder.query_language import Dataset, compile
 from databuilder.sqlalchemy_utils import clause_as_str, get_setup_and_cleanup_queries
 
-from . import query_language as ql
 from .validate_dummy_data import validate_dummy_data_file, validate_file_types_match
 
 log = structlog.getLogger()
@@ -30,7 +29,7 @@ def generate_dataset(
 
     log.info(f"Generating dataset for {str(definition_file)}")
     dataset_definition = load_definition(definition_file)
-    variable_definitions = ql.compile(dataset_definition)
+    variable_definitions = compile(dataset_definition)
     column_specs = get_column_specs(variable_definitions)
 
     query_engine = get_query_engine(dsn, backend_class, query_engine_class, environ)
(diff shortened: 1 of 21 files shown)


databuilder/main.py
```diff
@@ -146,7 +146,7 @@ def load_definition(definition_file):
         raise AttributeError("A dataset definition must define one 'dataset'")
     assert isinstance(
         dataset, Dataset
-    ), "'dataset' must be an instance of databuilder.query_language.Dataset()"
+    ), "'dataset' must be an instance of databuilder.ehrql.Dataset()"
```
inglesp, line 149: Is there anything we can do to stop imports from the `query_language` module?  Would doing so gain us anything?

tests/unit/test_tables.py
```diff
@@ -0,0 +1,17 @@
+from databuilder import tables
+from databuilder.contracts import constraints
+from databuilder.module_utils import is_proper_subclass
+
+
+def test_all_required_classes_are_exported():
+    # If we add a new Constraint we don't want to forget to make it available via
+    # `databuilder.tables`
```
inglesp, line 8: It's not clear to me why we want to include the constraint classes in `databuilder.tables`.

inglesp (approved):

tests/unit/test_tables.py
evansd: It's all the stuff you need for defining tables. It felt like a cleaner abstraction to me to have that all bundled together (and `databuilder.tables` felt like the most appropriate place). But I don't feel particularly strongly about it.

databuilder/main.py
evansd: I'm not sure there is and I'm not sure it would. As long as the documented modules are clear I don't think there's much to be gained from trying to stop people doing silly things which they have to go out of their way to do.

inglesp: Fine.

tests/unit/test_tables.py
inglesp: Don't feel strongly either.

PR #710, opened 2022-09-09
Remove use of ONS deaths from dataset definition

### LINKED ISSUE #690: Remove use of ONS Deaths from the booster effectiveness study
Graphnet are not able to get hold of the ONS Deaths data. Will has confirmed that this isn't important to the study ("There were about 30 covid related deaths overall in the final study cohort in TPP data. So there will be far fewer for cipha. So it's almost certain too infrequent an outcome to be useful anyway.").

We should remove the use of that table from the dataset definition and update the downstream analytics code so that it doesn't expect to find it.

milanwiedemann:
Removes ONS deaths from `dataset_definition.py`
Fixes #690 

---

@evansd suggested that some references to ONS deaths can stay for now:

- [tests/acceptance/comparative_booster_study/schema.py](https://github.com/opensafely-core/databuilder/blob/300e09e45825748759f39df2ca34fbc84e21c0c9/tests/acceptance/comparative_booster_study/schema.py#L43-L64)
- [tests/acceptance/comparative_booster_study/tpp_schema.py](https://github.com/opensafely-core/databuilder/blob/300e09e45825748759f39df2ca34fbc84e21c0c9/tests/acceptance/comparative_booster_study/tpp_schema.py#L164-L166)
- [tests/acceptance/comparative_booster_study/tpp_backend.py](https://github.com/opensafely-core/databuilder/blob/300e09e45825748759f39df2ca34fbc84e21c0c9/tests/acceptance/comparative_booster_study/tpp_backend.py#L47-L53)

### DIFF
--- a/tests/acceptance/comparative_booster_study/dataset_definition.py
+++ b/tests/acceptance/comparative_booster_study/dataset_definition.py
@@ -8,7 +8,6 @@ from .codelists import combine_codelists
 from .variables_lib import (
     address_as_of,
     age_as_of,
-    cause_of_death_matches,
     create_sequential_variables,
     date_deregistered_from_all_supported_practices,
     emergency_care_diagnosis_matches,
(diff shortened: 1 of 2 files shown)


evansd (approved):
Thanks @milanwiedemann, looks good.

The reason I said to leave the references in the schema, backend etc is that these files aren't _really_ part of the study. I created them to be able to make progress on the study translation without having to commit to an "official" schema. I expect these files to eventually go away and be merged with the official TPP and Graphnet schemas, but until they are I'd rather leave them with the full set of tables.

Issue #709, opened 2022-09-09
"Build and publish image" workflow is broken


iaindillingham:
The "Build and publish image" workflow is broken, so images are neither being built nor published. See, for example, [this run](https://github.com/opensafely-core/databuilder/runs/8271949455?check_suite_focus=true).

I'm unsure how to interpret the output. However, I feel this issue might be related to #702 in that the most recent tag isn't a version (`git tag --sort=refname`).

benbc:
@iaindillingham I'm not sure what you're intending with that sort order. An hour later than you, I get this (with maybe a better sort order?):

```
> git tag --sort=-authordate | head
v0.42.0
v0.41.1
docs-metadata-b61c95630b57a844dcd72d091596e13f9ba5a4b7
docs-metadata-ae390a742a1e9796a33ab8edc6661bae7fa80023
docs-metadata-341e04ebed4f3c9cce4050ad9202adf3ac5a496e
docs-metadata-9380ab6a9bb78def8eb38c110826c280209bbec3
vnull
docs-metadata-74ff1b37ea14aa30074e740853c78b3ea2c55918
docs-metadata-c09d14221e0b1034d51dfc5e2923d41359927c4d
docs-metadata-e21427286a88da9a33babdef51fcf0b603f14b82
```

StevenMaude:
There's something not as expected happening. For example, [this commit](https://github.com/opensafely-core/databuilder/commit/b983c78ea6fca2d751efaee8b316ab80ca2570f5) has been tagged as `v0.45.0`, and isn't on the `main` branch.

StevenMaude:
I have a hunch it's this [action update](https://github.com/opensafely-core/databuilder/pull/689) that might be responsible, in that the timing fits with the tags going wonky.

StevenMaude:
https://github.com/opensafely-core/databuilder/pull/714

StevenMaude:
For reference, because I don't know if you can easily get/link from GitHub's web UI, this is the README diff for github-tag-action from 4.5 to 6.0 (with `.txt` file extension, because GitHub's file upload option requires certain extensions :man_shrugging:)

[github-tag-action.txt](https://github.com/opensafely-core/databuilder/files/9555953/github-tag-action.txt)

There's nothing immediately obvious to my eye there that breaks functionality.

There's also another [tag action](https://github.com/anothrNick/github-tag-action), which may or may not be less glitchy.

remlapmot:
I think you want

[...]

PR #714, opened 2022-09-13
Revert "chore: bump mathieudutour/github-tag-action from 4.5 to 6.0"

StevenMaude:
Reverts opensafely-core/databuilder#689

Following this change, we had issues with tagging (#709).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,7 +62,7 @@ jobs:
           fetch-depth: 0
       - name: Bump version and push tag
         id: tag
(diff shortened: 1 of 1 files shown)


evansd (approved):
Ugggh.

I honestly think this is the kind of circumstance under which it wouldn't be mad to just implement our own action for this.

We only need to support the exact behaviour that we want, and losing time debugging Other People's Code for something as critical and unexciting as our release process can be quite dispiriting.

StevenMaude:
I agree. But, if this fixes things for now, that's still a net improvement.

evansd:
Oh yes, absolutely. Sorry, that was implied but I never actually said it!

PR #715, opened 2022-09-13
Working with os cli

benbc:


databuilder/__main__.py
```diff
@@ -27,6 +29,7 @@
     "databricks": "databuilder.backends.databricks.DatabricksBackend",
     "graphnet": "databuilder.backends.graphnet.GraphnetBackend",
     "tpp": "databuilder.backends.tpp.TPPBackend",
+    "expectations": "databuilder.backends.dummy.DummyBackend",
```
evansd, line 32: Instinctively, I'd take a slightly different approach here. I'll describe it and leave it to you as to whether you think it's an improvement or not.

I'd avoid having `expectations` listed as a valid option in the help text. I feel like this is a temporary fudge to get existing-databuilder working in existing-jobrunner, and not something we want to expose in the public API.

And I don't think we need a new class in `databuilder.backends` to achieve this. Instead we could use a sentinel object:
```py
EXPECTATIONS_BACKEND_PLACEHOLDER = object()
```

And then a guard clause on `backend_from_id`:
```py
def backend_from_id(str_id):
    if str_id == "expectations":
        return EXPECTATIONS_BACKEND_PLACEHOLDER
```

And then the asserts are just:
```py
assert options.backend is not EXPECTATIONS_BACKEND_PLACEHOLDER
```

It's just as horrible a fudge as before, but a bit less leaky in my view.

evansd (approved):
Looks good. I've made some comments on the DumyBackend stuff but nothing critical.

databuilder/__main__.py
benbc: I think this is better than what I have. I started using `None` as a signal and then decided that was _awful_, so I ended up with a whole dummy backend. Didn't occur to me to use a unique object. Thanks.

databuilder/__main__.py
```diff
@@ -43,6 +45,7 @@ def main(args, environ=None):
 
     if options.which == "generate-dataset":
         if options.dsn:
+            assert options.backend != EXPECTATIONS_BACKEND_PLACEHOLDER
```
evansd, line 48: This is fine in this context, but as a general rule I think you're better off using identity checks with sentinel objects (`options.backend is not EXPECTATIONS_BACKEND_PLACEHOLDER`) because of the possibility of the LHS doing funny-business with equality overloading. Plus you can't meaningfully talk about the value of sentinel objects anyway: their identity _is_ their entire role.

benbc: Oh, thank you -- you're quite right. That has not occurred to me. I think the risk here is v low, though, so I'm going to leave it as it's merged. I wouldn't be at all offended if you changed it in passing later (or I might do the same).

evansd: I think the risk is exactly zero here! I certainly don't think it needs changing. I just thought you might be interested in the more general stylistic point.

PR #721, opened 2022-09-15
Add test for `is_not_in`

StevenMaude:


StevenMaude:
I noticed that there's no spec test from the reference not including it. But `is_not_in` works!

This could be intentional. (Maybe the preference to use a negation operator?) But there are both `is_in` and `is_not_in` for codelists, so that seems an inconsistency.

Having both `is_in` and `is_not_in` present for containment *and* codelists perhaps does makes sense.

I could have opened an issue. But just adding a test, then opening it for someone to potentially correct me was as quick.

evansd (approved):
Oh yes, you're quite right. I think this was just an accidental omission from #583. If we've got a test for `is_in` there we should have one for `is_not_in` too.

PR #725, opened 2022-09-16
feat: Smoke test schema

### LINKED ISSUE #701: Remove reference to `databuilder.tables.patients` from test study
Following #680, tables should now be referenced from a module within `databuilder.tables` (e.g. `databuilder.tables.beta.tpp`) rather than from the top level.

However we have left `databuilder.tables.patients` temporarily in place to avoid breaking this test study:
https://github.com/opensafely/test-age-distribution

We should amend the test study, and then remove this table.

This might also be a good time to address the below in order to make these kinds of changes easier in future:

 * https://github.com/opensafely-core/databuilder/issues/681

benbc:
Fixes #701.

### DIFF
--- a/databuilder/tables/beta/smoketest.py
+++ b/databuilder/tables/beta/smoketest.py
@@ -1,0 +1,25 @@
+"""
+This is the minimal schema required to run the OpenSAFELY system smoke test.
+All backends should implement this schema.
+"""
+
+import datetime
+
+from .. import FirstOfMonthConstraint, NotNullConstraint, PatientFrame, Series, table
+
+__all__ = [
+    "patients",
+]
+
+
+@table
+class patients(PatientFrame):
+
+    date_of_birth = Series(
+        datetime.date,
+        description=(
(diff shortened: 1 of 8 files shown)


databuilder/tables/beta/smoketest.py
```diff
@@ -0,0 +1,23 @@
+import datetime
+
+from .. import FirstOfMonthConstraint, NotNullConstraint, PatientFrame, Series, table
+
+# This is the minimal schema required to run the OpenSAFELY system smoke test. All backends
+# should implement this schema.
```
evansd, line 6: I think it you made this the module docstring it would get pulled in by the documentation build.

evansd (approved):
Approved, modulo the faff with `patients`/`patient_demographics` in the Graphnet backend.

databuilder/tables/beta/smoketest.py
benbc: :+1:

Issue #724, opened 2022-09-15
Cannot add `databuilder` requirement as a Git archive with certain commits


StevenMaude:
Not sure what's happening here, will have to investigate.

In the past, I was working from a branch and happily creating requirements based on a GitHub archive zip with a specific hash.

I just tried now, to see if I could do this from `main` in this repository to help Milan run the tutorial.

With `https://github.com/opensafely-core/databuilder/archive/4544d4d8f4c9e743cf7c0396916c521ba0ff8ff2.zip` in the documentation repository's requirements inside `databuilder/requirements.prod.in`, I get:

```
…
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 338, in get_requires_for_build_wheel
        return self._get_build_requires(config_settings, requirements=['wheel'])
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 320, in _get_build_requires
        self.run_setup()
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/build_meta.py", line 335, in run_setup
        exec(code, locals())
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/__init__.py", line 87, in setup
        return distutils.core.setup(**attrs)
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/_distutils/core.py", line 147, in setup
        _setup_distribution = dist = klass(attrs)
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/dist.py", line 476, in __init__
        _Distribution.__init__(
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 280, in __init__
        self.finalize_options()
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools/dist.py", line 900, in finalize_options
        ep(self)
      File "/tmp/pip-build-env-p8bholw6/overlay/lib/python3.9/site-packages/setuptools_scm/integration.py", line 123, in infer_version
        _assign_version(dist, config)
[...]

PR #735, opened 2022-09-21
Support `databuilder --version` on CLI

### LINKED ISSUE #624: Support `databuilder --version` on CLI

### LINKED ISSUE #731: Speed up Docker builds
#624 made Docker builds significantly slower, as Data Builder (and its dependencies) is installed on each build

https://github.com/opensafely-core/databuilder/blob/8fefe6c7578b92c75a17b27df3d561308ced1bea/Dockerfile#L33

to expose a version number. Given that builds are central to testing, this slowdown is intolerable. @evansd reported the issue in [this thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1662743210173519).

We should speed up Docker builds. We should also reconsider `setuptools_scm`.

iaindillingham:
This is the second attempt at adding the version argument. The first attempt (#676, #708), which was based on setuptools-scm, increased the time it took to build the Docker image. Here, we eschew setuptools-scm and simply write the version to a file. Consequently, the time it takes to build the Docker image remains the same.

Closes #624 
Closes #731


pyproject.toml
```diff
@@ -1,11 +1,11 @@
...
-requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
-build-backend = "setuptools.build_meta"
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
 
 [project]
 name = "opensafely-databuilder"
 description = ""
-dynamic = ["version"]
+version = "2+local"
```
ghickman, line 8: This will mean pip reports a different version to what's in the version file, was that intended?

ghickman (approved):

.github/workflows/build-and-deploy.yml
```diff
@@ -98,7 +98,8 @@ jobs:
           echo "MINOR=$MINOR"
           echo "PATCH=$PATCH"
 
-          export DOCKER_BUILDKIT=1
```
StevenMaude, line 101: question: What's the reason for dropping BuildKit?

It was added to make the Dockerfile changes here work, I think. But it's the [more modern Docker way of building images](https://docs.docker.com/build/#overview), in any case. So maybe we still want to retain it?

We use it elsewhere; for example: https://github.com/opensafely-core/base-docker/blob/8d9de9f1c46816eca95b28f15e6aecc128f90b3e/Makefile#L4

StevenMaude:
Because this reverts to flit for building, this might fix #724 whose cause might be a feature of/issue with setuptools_scm.

pyproject.toml
iaindillingham: It wasn't intended. However, I'm unsure how to reconcile dynamic metadata with local version identifiers. Do you know, @ghickman?

.github/workflows/build-and-deploy.yml
iaindillingham: The reason is just as you think: we weren't using it before, and there isn't an immediate need to use it now, so we drop it. We may choose to use it when implementing #665, though.

StevenMaude: :+1: — I think it's fine to leave this PR as is, then semantically re-introduce BuildKit as its own improvement.

pyproject.toml
ghickman: @iaindillingham – you should be able to use your existing creation of a `__version__` attribute on the package to look it up [dynamically with flit](https://flit.pypa.io/en/latest/pyproject_toml.html#new-style-metadata).  Alternatively you could just use the same string for both 😁

PR #736, opened 2022-09-21
Support categorical variables

### LINKED ISSUE #631: Handle categoricals in binary output formats
This was a thing we failed to get right with cohortextractor and I think it's important to get it right in Data Builder.

Often datasets will have columns of string type where the strings can only be one of a fixed set of values. We want to encode these in the binary output as [categoricals](https://arrow.apache.org/docs/python/data.html#dictionary-arrays) rather than strings because the space saving is enormous. (Admittedly, if we're using compression then the savings in disk space may be fairly minimal, but it's the saving in RAM that's really important here.)

When we attempted this in cohortextractor we tried to build the list of categories [on the fly](https://github.com/opensafely-core/cohort-extractor/blob/2b713a5bb7f32e10c049169238edd3bb028578c0/cohortextractor/pandas_utils.py#L120-L137) while processing the output and then convert the column to a categorical [at the end](https://github.com/opensafely-core/cohort-extractor/blob/2b713a5bb7f32e10c049169238edd3bb028578c0/cohortextractor/pandas_utils.py#L80-L88) before writing out the data.

This has some serious disadvantages:
 * It means we need to buffer the whole output in memory before writing out, instead of being able to stream it to disk in chunks.
 * It meant that the order of the categories was not stable and would change depending on the order in which we happened to first encounter each category in the output. This was [particularly problematic](https://github.com/opensafely-core/cohort-extractor/issues/534) for Stata and meant we had to abandon use of categorical variables altogether there.

So I think we should only use categoricals where we know upfront, before retrieving the data, the full set of possible categories.

Fortunately, for the two most common forms of string variable I think we can do this. The two forms are: the output from a categorised codelist; the result of a set of logical conditions in a `case`/`categorised_as` expression.

In the Data Builder query model both of these are implemented using the `Case` operation, so this gives us a fairly simple rule:

 * Where a variable is the result of a `Case` operation, all of whose values are static strings then that variable is a categorical and those strings are its categories.

This does leave a class of string variable that _won't_ be treated as a categorical and that is codes drawn directly from a column of type CTV3, SNOMED-CT etc. The problem here is that, without bundling with databuilder (and keeping up-to-date) the entire set of possible codes for each coding system, we can't determine in advance what the set of required categories will be.

However my sense is that it's relatively rare to attempt to return individual codes like this as opposed to mapping them to categories and returning the categories.

Related databuilder issues:
 * https://github.com/opensafely-core/databuilder/issues/588
 * https://github.com/opensafely-core/databuilder/issues/507

evansd:
This PR adds an internal function to ask, of any variable in a dataset definition: is this a categorical variable and if so what are its categories?

There are two source of categorical variable: those drawn from columns whose definition includes a `CategoricalConstraint`; and those created by a `Case` expression whose outputs are a set of fixed values. Both are handled here. Note that other ways of creating a categorical such as using a categorised codelist to map codes to categories, or using `if_null_then()` to create a new default category, are all compiled down to `Case` expressions and so are handled automatically here.

We use this information to encode such variables as categoricals ([`pyarrow.DictionaryArray`](https://arrow.apache.org/docs/python/data.html#dictionary-arrays)) when writing Arrow files, resulting in significant memory usage and performance benefits over storing them as strings.

We also use this introspection feature when validating user-supplied dummy data to ensure that it uses the same categories as the real thing.

As part of this work we add the basic framework necessary for inferring the range of a numeric variable and selecting the most efficient binary type to represent it — though at present we only use this in one specific case.

Closes #631

---

Although the changes here aren't huge I'm aware that the query model, and the range of different `@singledispatch` functions which do things with it, is growing steadily in complexity and possibly getting a bit out of hand. I think at some point we'll want to step back from all this and come up with a slightly different approach to the query model implementation. _However_, I feel very relaxed about this. I'm confident in the seams we've put in place, and in the fundamental structure of the query model. At some suitable point (which is definitely not now) we can tidy up the implementation to make it more manageable for the long-term. But I don't anticipate that having wider implications on the rest of the codebase.

### DIFF
--- a/databuilder/column_specs.py
+++ b/databuilder/column_specs.py
@@ -19,6 +19,8 @@ class ColumnSpec:
     type: type[T]  # noqa: A003
     nullable: bool = True
     categories: Optional[tuple[T]] = None
+    min_value: Optional[T] = None
+    max_value: Optional[T] = None
 
 
 def get_column_specs(variable_definitions):
@@ -35,11 +37,18 @@ def get_column_specs(variable_definitions):
             continue
         type_ = get_series_type(series)
         categories = get_categories(series)
+        min_value, max_value = get_range(series)
         if hasattr(type_, "_primitive_type"):
             type_ = type_._primitive_type()
             if categories:
                 categories = tuple(c._to_primitive_type() for c in categories)
-        column_specs[name] = ColumnSpec(type_, nullable=True, categories=categories)
+        column_specs[name] = ColumnSpec(
+            type_,
+            nullable=True,
+            categories=categories,
+            min_value=min_value,
+            max_value=max_value,
+        )
     return column_specs
 
 
@@ -88,3 +97,9 @@ def get_categories_for_case(series):
         all_categories.extend(categories)
     # De-duplicate categories while maintaining their original order
     return tuple(dict.fromkeys(all_categories).keys())
+
+
+@singledispatch
+def get_range(series):
+    # For most operations we don't even try to determine the numerical range
+    return None, None
--- a/databuilder/column_specs.py
+++ b/databuilder/column_specs.py
@@ -103,3 +103,12 @@ def get_categories_for_case(series):
 def get_range(series):
     # For most operations we don't even try to determine the numerical range
     return None, None
+
+
+@get_range.register(AggregateByPatient.Count)
+def get_range_for_count(series):
+    # Per-patient row counts can never be negative and we think 65,535 is a reasonable
+    # upper bound, meaning they can be stored in a 16-bit unsigned int, thus reducing
+    # memory pressure. For counts anywhere near this high the user should be classifying
+    # them into buckets rather than retrieving the raw numbers in any case.
+    return 0, 2**16 - 1
--- a/databuilder/column_specs.py
+++ b/databuilder/column_specs.py
@@ -4,7 +4,9 @@ from typing import Optional, TypeVar
 
 from databuilder.query_model import (
     AggregateByPatient,
+    Case,
     SelectColumn,
+    Value,
     get_root_frame,
     get_series_type,
 )
@@ -61,3 +63,28 @@ def get_categories_for_select_column(series):
 def get_categories_for_min_max(series):
     # The min/max aggregations preserve the categories of their inputs
     return get_categories(series.source)
+
+
+@get_categories.register(Value)
+def get_categories_for_value(series):
+    # Static values can be considered categoricals with a single available category
+    return (series.value,)
+
+
+@get_categories.register(Case)
+def get_categories_for_case(series):
+    # The categories for a Case expression are the combined categories of all its output
+    # values, with the proviso that if any output value is non-categorical then the
+    # whole expression is non-categorical also
+    output_values = list(series.cases.values())
+    if series.default is not None:
+        output_values.append(series.default)
+    all_categories = []
+    for output_value in output_values:
+        categories = get_categories(output_value)
+        # Bail if we've got a non-categorical output value
+        if categories is None:
+            return None
+        all_categories.extend(categories)
+    # De-duplicate categories while maintaining their original order
+    return tuple(dict.fromkeys(all_categories).keys())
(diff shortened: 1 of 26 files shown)


databuilder/query_model.py
```diff
@@ -59,23 +59,48 @@ def __repr__(self):
...
-            kwargs.append(f"{name}={prefix}{type_.__name__}")
+        kwargs = [f"{key}={value!r}" for key, value in self.schema.items()]
         return f"{self.__class__.__name__}({', '.join(kwargs)})"
 
+    def get_column_type(self, name):
+        return self.schema[name].type_
+
+    @property
+    def column_names(self):
+        return self.schema.keys()
```
inglesp, line 103: Is it worth returning a list, so that it's indexable?

tests/integration/test_file_formats.py
```diff
@@ -118,3 +119,31 @@ def test_validate_dataset_type_mismatch(tmp_path, extension):
...
+    ]
+    write_dataset(filename, results, column_specs_1)
+
+    column_specs_2 = {
+        "patient_id": ColumnSpec(int),
+        "category": ColumnSpec(str, categories=("x", "y")),
+    }
+
+    errors = {
+        ".arrow": "Unexpected categories in column 'category'",
```
inglesp, line 78: Can we get pyarrow to give something a little more helpful here?

databuilder/file_formats/arrow.py
```diff
@@ -84,6 +84,37 @@ def column_to_pyarrow(column):
     return column_to_pyarrow
 
 
+def smallest_int_type_for_range(minimum, maximum):
+    """
+    Return smallest pyarrow integer type capable of representing all values in the
+    supplied range
+    """
```
inglesp, line 92: Is it worth including the aside from the git commit in this comment?

tests/unit/test_column_specs.py
```diff
@@ -101,3 +101,11 @@ def test_get_categories_for_case_with_mixed_categorical_and_noncategorical_value
...
     i = SelectColumn(SelectTable("t", schema=TableSchema(i=Column(int))), "i")
     assert get_range(i) == (None, None)
+
+
+def test_get_range_for_count():
+    count = AggregateByPatient.Count(events)
+    min_value, max_value = get_range(count)
+    assert min_value == 0
+    num_bits = f"{max_value:b}".count("1")
+    assert num_bits == 16
```
inglesp, line 111: Would this pass if `max_value` was `1010101010101010101010101010101`?

databuilder/column_specs.py
```diff
@@ -103,3 +103,12 @@ def get_categories_for_case(series):
...
     return None, None
+
+
+@get_range.register(AggregateByPatient.Count)
+def get_range_for_count(series):
+    # Per-patient row counts can never be negative and we think 65,535 is a reasonable
+    # upper bound, meaning they can be stored in a 16-bit unsigned int, thus reducing
+    # memory pressure. For counts anywhere near this high the user should be classifying
+    # them into buckets rather than retrieving the raw numbers in any case.
+    return 0, (1 << 16) - 1
```
inglesp, line 114: Not a bit-twiddler myself, I'd find `2 ** 16 - 1` more readable.

tests/unit/test_column_specs.py
inglesp: Also, could we test that a very large value (eg `max_value - 1`) actually gets serialised correctly, and not silently truncated?

inglesp (approved):
All sensible, but share your concerns about steady increase in complexity.  Probably unavoidable though.

tests/integration/test_file_formats.py
evansd: The full error message does include the list of categories it was expecting and the list of categories it found. I just didn't bother matching on it here as it made the test harder to read and it felt like enough to ensure that we were triggering the error we expected to trigger.

evansd:
Thanks @inglesp — good suggestions which I'll incorporate.

On the complexity, I think it is avoidable in one sense in that I think there's some biggish refactoring we could do to the query model which might make it all a bit easier to grasp. But I think the time for that will come later. And I think given the way things are structured the current complexity shouldn't leak too far.

PR #750, opened 2022-09-27
Rename `hospitalization` to `hospitalisation`

### LINKED ISSUE #742: `hospitalizations` has American English spelling
At the risk of bikeshedding, do we want to use `hospitalisations`?

Or even `hospital_admissions` which seems to be the more common NHS term (assuming these two terms have the same accepted meaning)?

(Case: I am writing tutorial information referring to this term. It's a little strange to use my preferred British English spelling in the text referring to the same thing that the code spells with American English. Either you have inconsistency or there's the slight use of American English.)

It's fine if we want to retain this, but thought I'd check.

Example: https://github.com/opensafely-core/databuilder/blob/c661719bc8823b2872c79c0026a9e92415cc2c29/databuilder/contracts/wip.py#L32

StevenMaude:
Fixes #742.

This is the British English spellling and makes the usage consistent throughout the codebase.

We may want to rename the relevant tables and contracts in future, but what they should be called right now is not immediately clear. See the discussion in #742 for more details.

### DIFF
--- a/databuilder/backends/graphnet.py
+++ b/databuilder/backends/graphnet.py
@@ -57,7 +57,7 @@ class GraphnetBackend(BaseBackend):
         ),
     )
 
-    hospitalizations_without_system = MappedTable(
+    hospitalisations_without_system = MappedTable(
         schema=SCHEMA,
         source="Hospitalisations",
         columns=dict(
--- a/databuilder/backends/tpp.py
+++ b/databuilder/backends/tpp.py
@@ -124,7 +124,7 @@ class TPPBackend(BaseBackend):
         """
     )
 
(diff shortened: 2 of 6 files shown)


evansd (approved):

StevenMaude:
@iaindillingham rightly pointed out to me that it's debatable whether `-ise` or `-ize` is actually correct.

I was under the misconception that `ise` was correct for British English and `ize` was correct for American English, [which is not always the case](https://en.wikipedia.org/wiki/Oxford_spelling).

But the NHS tends to favour `hospitalise` which has swung me around on merging this:

s-versus-z-nhs

PR #751, opened 2022-09-27
Add exponential backoff and retry to MSSQL engine

### LINKED ISSUE #532: Add exponential backoff to MSSQL engine
To work around the flakiness of the database connection, the MSSQL engine [catches certain errors from the database, and retries twice, sleeping 0.5s between each attempt](https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/query_engines/mssql_lib.py#L168-L181).  This is good enough to for handling temporary flakiness, but not if the database is unavailable for more than a few seconds.

Instead, we should retry with exponential backoff, with exponentially increasing sleeps between attempts.  In https://github.com/opensafely-core/cohort-extractor/pull/774 we added exponential backoff to cohort-extractor, and we should use the same approach here.  With an initial sleep of 4s, a backoff factor of 4, and making 6 retries, [our attempts are spread over 90 minutes](https://github.com/opensafely-core/cohort-extractor/blob/90f27bedede0f2780ff5ec0bc052a0a6dafe8f2d/cohortextractor/tpp_backend.py#L33-L36).

evansd:
This adds the behaviour we have in cohortextractor whereby certain classes of database error are regarded as transient and are automatically retried after a suitable pause. Furthermore, if we've been able to write our results somewhere persistent then we close the database connection completely before retrying.

The tests feature the distressing presence of `mock.patch`, but given the kind of thing we're trying to test (what happens if this thing throws an error half-way through?) I don't think there's much alternative. And the tests have already exposed that I had forgotten transactions were a thing, so they do have some value.

Closes #532

### DIFF
--- a/databuilder/query_engines/mssql.py
+++ b/databuilder/query_engines/mssql.py
@@ -11,6 +11,7 @@ from databuilder.query_engines.base_sql import BaseSQLQueryEngine
 from databuilder.query_engines.mssql_dialect import MSSQLDialect, SelectStarInto
 from databuilder.sqlalchemy_utils import (
     GeneratedTable,
+    ReconnectableConnection,
     execute_with_retry_factory,
     fetch_table_in_batches,
     get_setup_and_cleanup_queries,
@@ -112,14 +113,29 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
         assert str(results_query) == str(sqlalchemy.select(results_table))
 
         setup_queries, cleanup_queries = get_setup_and_cleanup_queries(results_query)
-        with self.engine.connect() as connection:
+
+        # Because we may be disconnecting and reconnecting to the database part way
+        # through downloading results we need to make sure that the temporary tables we
+        # create, and the commands which delete them, get committed. There's no need for
+        # careful transaction management here: we just want them committed immediately.
+        autocommit_engine = self.engine.execution_options(isolation_level="AUTOCOMMIT")
+
+        with ReconnectableConnection(autocommit_engine) as connection:
             for n, setup_query in enumerate(setup_queries, start=1):
                 log.info(f"Running setup query {n:03} / {len(setup_queries):03}")
                 connection.execute(setup_query)
 
+            # Re-establishing the database connection after an error allows us to
+            # recover from a wider range of failure modes. But we can only do this if
+            # the table we're reading from persists across connections.
+            if results_table.is_persistent:
+                conn_execute = connection.execute_disconnect_on_error
+            else:
+                conn_execute = connection.execute
+
             # Retry 6 times over ~90m
             execute_with_retry = execute_with_retry_factory(
-                connection.execute,
+                conn_execute,
                 max_retries=6,
                 retry_sleep=4.0,
                 backoff_factor=4,
--- a/databuilder/query_engines/mssql.py
+++ b/databuilder/query_engines/mssql.py
@@ -1,6 +1,9 @@
+import datetime
+import secrets
+
 import sqlalchemy
 import structlog
-from sqlalchemy.schema import CreateIndex
+from sqlalchemy.schema import CreateIndex, DropTable
 from sqlalchemy.sql.functions import Function as SQLFunction
 
 from databuilder import sqlalchemy_types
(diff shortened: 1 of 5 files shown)


tests/unit/test_sqlalchemy_utils.py
```diff
@@ -198,3 +202,29 @@ def execute(self, query):
...
+
+@mock.patch("time.sleep")
+def test_execute_with_retry(sleep):
+    execute = mock.Mock(side_effect=[ERROR, ERROR, ERROR, "its OK now"])
+    execute_with_retry = ExecuteWithRetry(
+        execute, max_retries=3, retry_sleep=10, backoff_factor=2
+    )
+    assert execute_with_retry() == "its OK now"
+    assert execute.call_count == 4
+    assert sleep.mock_calls == [mock.call(t) for t in [10, 20, 40]]
```
iaindillingham, line 220: Admirably succinct!

databuilder/query_engines/mssql.py
```diff
@@ -63,11 +66,42 @@ def get_query(self, variable_definitions):
...
+            self.config.get("TEMP_DATABASE_NAME")
+        )
         results_table = temporary_table_from_query(
-            "#results", results_query, index_col="patient_id"
+            table_name, results_query, index_col="patient_id", schema=schema
         )
         return sqlalchemy.select(results_table)
 
+    def get_results_table_name_and_schema(self, temp_database_name):
+        # If we have a temporary database we can write results their which enables
```
iaindillingham, line 79:
```suggestion
        # If we have a temporary database we can write results there which enables
```

databuilder/query_engines/mssql.py
```diff
@@ -63,11 +66,42 @@ def get_query(self, variable_definitions):
...
+            token = secrets.token_hex(6)
+            table_name = f"results_{timestamp:%Y%m%d_%H%M}_{token}"
+            # The `schema` variable below is actually a multi-part identifier of the
+            # form `<database-name>.<schema>`. We don't really care about the schema
+            # here, we just want to use whatever is the default schema for the database.
+            # MSSQL allows you to do this by specifying "." as the schema name. However
+            # I can't find a way of supplying this without SQLAlchemy's quoting
+            # algorithm either mangling it or blowing up. We could work around this by
+            # attaching our own Identifier Preparer to our MSSQL dialect, but that
+            # sounds a lot like hard work. So we're using the default, default schema of
```
iaindillingham, line 93:
```suggestion
            # sounds a lot like hard work. So we're using the default schema of
```

tests/unit/test_sqlalchemy_utils.py
```diff
@@ -3,11 +3,13 @@
 import pytest
 import sqlalchemy
 from sqlalchemy.dialects.sqlite.pysqlite import SQLiteDialect_pysqlite
+from sqlalchemy.engine import Engine
 
 from databuilder.sqlalchemy_utils import (
     ExecuteWithRetry,
     GeneratedTable,
     OperationalError,
+    ReconnectableConnection,
```
iaindillingham, line 12: The comments within the new tests really helped me understand `ReconnectableConnection`.

databuilder/query_engines/mssql.py
```diff
@@ -112,14 +113,29 @@ def get_results(self, variable_definitions):
         assert str(results_query) == str(sqlalchemy.select(results_table))
 
         setup_queries, cleanup_queries = get_setup_and_cleanup_queries(results_query)
-        with self.engine.connect() as connection:
+
+        # Because we may be disconnecting and reconnecting to the database part way
+        # through downloading results we need to make sure that the temporary tables we
+        # create, and the commands which delete them, get committed. There's no need for
+        # careful transaction management here: we just want the committed straight away.
```
iaindillingham, line 120:
```suggestion
        # careful transaction management here: we just want them committed straight away.
```

iaindillingham (approved):
I've made some comments and suggestions; the former are "Why did he do that?" questions, and the latter are fixes for typos. I'm confident I understand what's going on, but don't feel I can offer a substantive critique of the code. Sorry!

databuilder/sqlalchemy_utils.py
```diff
@@ -190,3 +193,42 @@ def fetch_table_in_batches(
             break
         else:
             min_key = row[key_column_index]
+
+
+class ExecuteWithRetry:
```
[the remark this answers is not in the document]
evansd: I think it's mostly because earlier iterations of the design needed a class and then I managed to simplify it but left the class in place because I didn't think it would massively simplify things to do it otherwise. It would be easy enough to make it a closure, `execute_with_retry_factory()` or something. It doesn't seem obviously more readable to me, but I'd be very happy to change it if others would.

databuilder/sqlalchemy_utils.py
```diff
@@ -190,3 +193,42 @@ def fetch_table_in_batches(
...
+    error, using an expontential backoff strategy
+    """
+
+    def __init__(
+        self,
+        execute,
+        max_retries=0,
+        retry_sleep=0,
+        backoff_factor=1,
+        log=lambda *_: None,
```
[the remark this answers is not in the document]
evansd: Yep, that's right. The single underscore variable is a Python convention to mean "I'm syntactically forced to assign a variable here, but I don't intend to ever use it".

tests/unit/test_sqlalchemy_utils.py
```diff
@@ -1,9 +1,13 @@
+from unittest import mock
+
 import pytest
 import sqlalchemy
 from sqlalchemy.dialects.sqlite.pysqlite import SQLiteDialect_pysqlite
 
 from databuilder.sqlalchemy_utils import (
+    ExecuteWithRetry,
     GeneratedTable,
+    OperationalError,
```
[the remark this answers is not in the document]
evansd: I think because I'm treating it semantically as "the kind of error that ExecuteWithRetry looks for, whatever that kind may be" rather than caring what specific sort of error it is. I guess you could argue that either way though.

databuilder/query_engines/mssql.py
evansd: Ah, hilariously, this isn't actually a typo: I really do mean two levels of default. When you create databases in MSSQL they have a default schema set. The default value for this default schema is `dbo`. But it's possible to create databases with a different default schema. I could probably word this a bit more clearly though!

databuilder/sqlalchemy_utils.py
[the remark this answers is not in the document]
evansd: I tried out the closure version and it did look simpler overall, so I'm using that instead. Thanks for the prompt!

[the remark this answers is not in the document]
iaindillingham: It looks simpler to me too 🙂

PR #755, opened 2022-09-28
Replace TPP schema and backend

evansd:
This takes the schema and backend written to test the comparative booster study and makes them the "official" TPP schema and backend (or at least, the official beta version).

We start by tweaking the test schema to match the `patients` contract, fix some bugs and generally bring it more closely inline with structure we use elsewhere.

We then remove the old TPP schema and backend along with its tests. There's some rather ingenious, if horrific, SQL which we lose here. But it was removed in a single clean commit so it should be straightforward to recover it if we decide we want it later.

We then test every table defined in the backend by inserting some data and checking that the table expression returned by the backend gives us what we expect when executed. This exposes some silly bugs which we fix.


databuilder/backends/tpp.py
```diff
@@ -117,11 +117,17 @@ class TPPBackend(BaseBackend):
...
-                carehm.LocationRequiresNursing AS care_home_requires_nursing,
-                carehm.LocationDoesNotRequireNursing AS care_home_does_not_require_nursing
+                CASE
+                    WHEN carehm.LocationRequiresNursing = 'Y' THEN 1
+                    WHEN carehm.LocationRequiresNursing = 'N' THEN 0
+                 END AS care_home_requires_nursing,
+                CASE
+                    WHEN carehm.LocationDoesNotRequireNursing = 'Y' THEN 1
+                    WHEN carehm.LocationDoesNotRequireNursing = 'N' THEN 0
+                 END AS care_home_does_not_require_nursing
```
inglesp, line 130: Yay tests.

Did this creep in from a mistranslation of the cohort-extractor code?  Or is the bug present there too?

inglesp (approved):
This is really good to have, thanks.

databuilder/backends/tpp.py
evansd: It's not present in cohortextractor. It's not even really a mistranslation as there's no direct equivalent in cohortextractor. It was just new code I introduced to make the table "API" nicer, but I was obviously wasn't concentrating very hard at the time.

tests/integration/backends/test_tpp.py
```diff
@@ -0,0 +1,379 @@
...
+    Organisation,
+    Patient,
+    PatientAddress,
+    PotentialCareHomeAddress,
+    RegistrationHistory,
+    SGSS_AllTests_Negative,
+    SGSS_AllTests_Positive,
+    Vaccination,
+    VaccinationReference,
+)
```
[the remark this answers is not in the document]
evansd: Oh yes, this thought occurred to me too but then I forgot about it. I'm sure there's some way of doing this and I think it would be valuable. I'll revisit before merging.

inglesp (approved):
quis scrutatur ipsos scrutatores?

Issue #749, opened 2022-09-27
Updates to Graphnet backend for running smoke test


benbc:

inglesp:
Closed by #768.

PR #768, opened 2022-10-05
Graphnet smoketests

inglesp:
Addresses #749.


databuilder/backends/graphnet.py
```diff
@@ -12,69 +12,14 @@
...
 
-    query_engine_class = MSSQLQueryEngine
+    query_engine_class = SparkQueryEngine
     patient_join_column = "Patient_ID"
-    implements = [databuilder.tables.beta.graphnet, databuilder.tables.beta.smoketest]
+    implements = [databuilder.tables.beta.smoketest]
 
     patients = MappedTable(
         schema=SCHEMA,
         source="Patients",
```
benbc, line 21: The schema spreadsheet has a singular `Patient` here.

benbc (commented):
Just one thing that I think is an error, otherwise this looks correct.

benbc (approved):

databuilder/backends/graphnet.py
inglesp: No, wait, it's got both!

image

@benbc can you ask Graphnet which we should use?

PR #769, opened 2022-10-05
Add scaffolding for supporting dummy data

evansd:
This puts in place all the boring stuff necessary for generating dummy data, without actually generating any dummy data. Specifically:

 * Invoking `generate-dataset` without supplying either a `--dsn` or `--dummy-data-file` no longer throws an error; instead it calls out to a new `generate_dummy_dataset` function.
 * `generate-dataset` accepts a new optional `--dummy-tables` argument which points to a directory of CSV files which be used in place of dynamically generated dummy data.
 * A new `create-dummy-tables` command dumps the dynamically generated dummy data to a directory of CSV files.


inglesp (approved):
Looks sensible!  One small question: is `orm_factory.py` the right name for the module now?  It contains quite a lot of CSV-related code.

evansd:
@inglesp That's a fair point. I think bundling the CSV stuff in there was reasonable, given that it's dealing with serializing and deserializing ORM models, but it doesn't fit the name. I suppose `orm_utils.py` would fit the pattern used elsewhere, semantic cop-out though it is.

I'm sure we've talked before about the need to add some structure to the ever-growing package namespace, but I don't think we'd benefit much from doing it now rather than later.

PR #772, opened 2022-10-06
Promote in-memory engine to not-just-test code

evansd:
We're going to want to use it for dummy data generation.

Putting this into its own, early PR to avoid potential conflicts with others working on the in-memory engine.


Justfile
```diff
@@ -181,7 +181,7 @@ test-spec-no-spark *ARGS: devenv
 # Run the unit tests only. Optional args are passed to pytest.
 test-unit *ARGS: devenv
     $BIN/python -m pytest tests/unit {{ ARGS }}
-    $BIN/python -m doctest tests/lib/in_memory/database.py
+    $BIN/python -m pytest --doctest-modules databuilder
```
StevenMaude, line 184: comment: This seems like a configuration improvement anyway, as well as a workaround for the import issue.

StevenMaude:
Question: (a naive one) why can we get rid of `tests/lib/in_memory/test_full_example.py`?

StevenMaude (approved):

evansd:
> question: (a naive one) why can we get rid of `tests/lib/in_memory/test_full_example.py`?

Because it was entirely commented out!

PR #774, opened 2022-10-06
Update Dockerfile to improve caching/layering

### LINKED ISSUE #665: Borrow Dockerfile improvements from cohortextractor

rebkwok:
Fixes #665 

This is the approach that I've been able to get working for databuilder; see comments in the issue and alternative draft PR #773  for the alternative.

This updates the Dockerfile with similar improvements to those in cohort-extractor, but unlike #773 , it doesn't pip install using the pyproject.toml.  It keeps the entrypoint we had before (`python3.9 -m -B databuilder`) but moves it out into an entrypoint script so it can be set as the `ACTION_EXEC` env and passed as the entrypoint to the base-action.

I didn't notice this being discernable slower to build (and rebuild after code changes) than the version in #773 

Docker tests are now passing, and it also works to run a [local project](https://github.com/opensafely/test-age-distribution) with `opensafely-cli`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,3 +42,4 @@ repos:
     rev: cdefcb096e520a6daa9552b1d4636f5f1e1729cd
     hooks:
     - id: hadolint
+      entry: --entrypoint /bin/hadolint hadolint/hadolint:latest --ignore DL3061 --ignore DL3013 --ignore DL3042
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,9 +37,3 @@ repos:
     - id: check-toml
     - id: check-yaml
     - id: detect-private-key
-
-  - repo: https://github.com/stratasan/hadolint-pre-commit
-    rev: cdefcb096e520a6daa9552b1d4636f5f1e1729cd
-    hooks:
-    - id: hadolint
-      entry: --entrypoint /bin/hadolint hadolint/hadolint:latest --ignore DL3061 --ignore DL3013 --ignore DL3042
--- a/dependencies.txt
+++ b/dependencies.txt
@@ -1,0 +1,11 @@
+# list ubuntu packages needed in production, one per line
+# run time dependencies
+# ensure fully working base python3 installation
+# see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
+python3.9
+python3.9-dev
+python3.9-venv
+python3.9-distutils
+
+# from packages.microsoft.com
+mssql-tools
--- a/dependencies.txt
+++ b/dependencies.txt
@@ -3,7 +3,6 @@
 # ensure fully working base python3 installation
 # see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
 python3.9
-python3.9-dev
 python3.9-venv
 python3.9-distutils
 
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -100,7 +100,8 @@ jobs:
 
           echo "$MAJOR.$MINOR.$PATCH" > databuilder/VERSION
 
-          docker build . --file Dockerfile \
+          DOCKER_BUILDKIT=1 docker build .docker build . --file Dockerfile \
+            --target databuilder
             --tag ${{ env.image }}:$MAJOR \
             --tag ${{ env.image }}:$MINOR \
             --tag ${{ env.image }}:$PATCH
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,63 @@
+# syntax=docker/dockerfile:1.2
+#################################################
+#
+# Initial databuilder layer with just system dependencies installed.
+#
 # hadolint ignore=DL3007
-FROM ghcr.io/opensafely-core/base-docker:latest
+FROM ghcr.io/opensafely-core/base-action:latest as databuilder-dependencies
+
+
+# setup default env vars for all images
+# ACTION_EXEC sets the default executable for the entrypoint in the base-action image
+ENV VIRTUAL_ENV=/opt/venv/ \
+    PYTHONPATH=/app \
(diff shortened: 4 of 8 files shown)


iaindillingham (approved):
Everything here looks very familiar - similar enough to cohort-extractor, without installing Data Builder from _pyproject.toml_ 👍🏻.

When I changed the Dockerfile in #676, @evansd mentioned that I'd made the build noticeably slower. Previously, he could run `pytest -k 'not spark and not mssql and not generative'` in 12s ([thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1663691658505619)). I can run this on my machine in 25s; but my machine isn't Dave's machine 🙂. Perhaps, Dave, you wouldn't mind giving it a try, to give us some reassurance that the build isn't noticeably slower.

rebkwok:
It didn't occur to me to try timing tests (too preoccupied with making them run at all!)
I've run `pytest -k 'not spark and not mssql and not generative'` 5x on my machine in:
1) 19.63s, 2) 16.73s, 3) 19.41s, 4) 18.65s, 5) 16.67s

On `main` (after the initial build), I get: 16.07s, 16.10s, 15.95s, 15.69s, 15.98s - so it does seem to be very slightly slower

evansd:
The key timing here is on
```
time just build-databuilder
```
when a single file has been changed. Currently that's 2.30s for me. (And when there have been zero changes it's 0.93s.)

That's overhead which we add to every single test run which involves the Docker image. I think it would be fine for that to rise by a couple of seconds, but I wouldn't want it to get too high.

rebkwok:
Hmm.  Right, there is definitely a change that needs to be made here so it installs the dev requirements earlier.

rebkwok:
Actually, it doesn't even need the dev requirements, does it?  (At least, the existing Dockerfile doesn't install them, so I assume not).  If I take those out, my `time just build-databuilder` timings with no changes and a single file change are more or less the same.

rebkwok:
Actually faster - about 1.3s for `just build-databuilder` with a single file change

evansd:
Oh great! Well I'm definitely happy with that :smile:

dependencies.txt
```diff
@@ -0,0 +1,11 @@
+# list ubuntu packages needed in production, one per line
+# run time dependencies
+# ensure fully working base python3 installation
+# see: https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e
+python3.9
+python3.9-dev
```
bloodearnest, line 6: We shouldn't need -dev in production, it should only be needed to build packages?

bloodearnest (approved):

bloodearnest:
Left a minor comment

The way to avoid the slowness on local development is to explicitly use a dev image layer that does not copy your files in, and running that dev image with your files mounted in. This is the way cohortextractor, release-hatch and other things work.

The key is a) your dev layer in the Docker that is based on the databuilder-base image (*not* databuilder-prod). This ensures it has no application code at all, only dependencies. Only your prod image has the `COPY databuilder /app/databuilder` step. This means 1) you can't accidently run the prod image when you think you're running dev and 2) it remove the COPY step out of the dependency tree of the dev layer build.

And b) this dev layer must be run with your code mounted in, i.e. `docker run -v .:/app ...`

The way we do this nicely in the other services is via using docker compose, [configuring the mount point](https://github.com/opensafely-core/cohort-extractor/blob/main/docker-compose.yml#L44) in there, then using `docker-compose run test` to wrap it all. This is also nice for starting db services and such.

That way, you can use the local dev image and run it as much as you want, and it should only rebuild when certain files change (requirements*.txt, dependencies*.txt, and currently entrypoint.sh), but never when you're code changes.

bloodearnest:
This image may help with understanding the layering within the dockerfile
```
┌──────────────────┐            ┌──────────────────┐
│   base-python    ├────────────►     builder      │
└────────┬─────────┘            └──────────┬───────┘
         │                                 │
         │                                 │COPY FROM
         │     ┌──────────────────┐        │
         └─────►release-hatch-base◄────────┘
               └────────┬─┬───────┘
                        │ │
                        │ │
┌──────────────────┐    │ │     ┌──────────────────┐
│release-hatch-prod◄────┘ └─────►release-hatch-dev │
└──────────────────┘            └──────────────────┘
```

.pre-commit-config.yaml
```diff
@@ -42,3 +42,4 @@ repos:
     rev: cdefcb096e520a6daa9552b1d4636f5f1e1729cd
     hooks:
     - id: hadolint
+      entry: --entrypoint /bin/hadolint hadolint/hadolint:latest --ignore DL3061 --ignore DL3013 --ignore DL3042
```
bloodearnest, line 45: minor preference to keep the ignores in the Docker file and not here?

rebkwok: I tried that, but it couldn't get it to work - the pre-commit hook kept failing even with the ignores in (including the ones in the same places that they were in cohort-extractor)

rebkwok: Have removed hadolint from the pre-commit hooks altogether and put it in the `just check` command instead (where it seems to respect the ignores in the Dockerfile fine)

PR #776, opened 2022-10-07
Switch CSV engine to use in-memory database

evansd:
The "feature" commit here represents the work I _thought_ this would take. And the other commits represent the rest of the iceberg.


iaindillingham (approved):
As always, thank you for the detailed PR and commit titles and bodies; and the logical sequencing and small size of commits 🙂

PR #784, opened 2022-10-10
Enable sorting for generative tests

### LINKED ISSUE #461: Implicit sort on primary key to ensure stability and consistency of results
When we sort event-dimensioned tables on a column that has identical values, the resulting order is undefined. This can give undetermined values for variables where we pick a column that has different values in two records with identical values for the sort column.

For example, this query

```
e.sort_by(e.i).first_for_patient().b
```

with this data

```
          |  i  | b
        --+-----+---
        1 | 101 | T
        1 | 102 | F 
        2 | 200 | T 
        2 | 200 | F 
```

will always give `True` for patient `1`, but can return either `True` or `False` for patient `2`.

This behaviour has two unsatisfactory consequences:

1. Researchers may see different results when re-running their queries. (This is probably a theoretical problem only because of the scale of real data, but may cause problems when testing with small synthetic datasets.)
2. In our tests we'll see different results for different engines which causes problems for our generative tests.

The proposed solution is that we an an implicit sort by primary key to resolve ties consistently.

This means that we have to insist that all event-dimensioned tables have a primary key. We could potentially allow for a composite "natural" key so that this is less of a constraint on backend implementers.

Note: sorting is disabled in the generative tests to avoid this problem; it should be re-enabled as part fixing this.

benbc:
The driver for this is to allow us to do more thorough testing (fixes #461). Previously we couldn't test sorting with the generative tests because in some cases the sort order differed between databases (and even between runs for a given database).

To enable that testing this PR adds a small and possibly user-invisible feature that makes the results of all sorts stable even in cases where the sort is under-specified by the dataset definition.

There are two things that I'm not sure about in this implementation and would welcome feedback on.

1. Should the new transformation be separate or should it be rolled into the existing one. My weak preference is to separate them because they exist for different reasons, but their implementation ties them together so there is an argument for treating them as a single transformation.

2. The unit tests for the transformations are pretty unreadable (and even hard to write). Should we introduce some kind of QM builder tooling to make tests like this more readable?

### DIFF
--- a/databuilder/query_engines/in_memory.py
+++ b/databuilder/query_engines/in_memory.py
@@ -10,6 +10,7 @@ from databuilder.query_engines.in_memory_database import (
     apply_function,
     handle_null,
 )
+from databuilder.query_model_transforms import apply_transforms
 
 T = True
 F = False
@@ -25,6 +26,8 @@ class InMemoryQueryEngine(BaseQueryEngine):
 
     def get_results(self, variable_definitions):
         self.cache = {}
+
+        variable_definitions = apply_transforms(variable_definitions)
 
         name_to_col = {
             "patient_id": PatientColumn(
@@ -106,7 +109,7 @@ class InMemoryQueryEngine(BaseQueryEngine):
         sort_index = self.visit(node.sort_by).sort_index()
         return source.sort(sort_index)
 
-    def visit_PickOneRowPerPatient(self, node):
+    def visit_PickOneRowPerPatientWithColumns(self, node):
         ix = {
             qm.Position.FIRST: 0,
             qm.Position.LAST: -1,
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -7,6 +7,7 @@ from functools import cache, singledispatch
 from types import GenericAlias
 from typing import Any, Optional, TypeVar
 
+from .codes import BaseCode
 from .typing_utils import get_typespec, get_typevars, type_matches
 
 # The below classes and functions are the public API surface of the query model
@@ -49,7 +50,7 @@ __all__ = [
 # type without specifying what that type has to be
 T = TypeVar("T")
 Numeric = TypeVar("Numeric", int, float)
-Comparable = TypeVar("Comparable", int, float, str, date)
+Comparable = TypeVar("Comparable", int, float, str, date, BaseCode)
 
 
 class Position(Enum):
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -774,3 +774,42 @@ def get_root_frame(frame):
 @get_root_frame.register(SelectPatientTable)
 def get_root_frame_for_table(frame):
     return frame
+
+
+def get_table_and_filters(frame):
+    """
+    Given a ManyRowsPerPatientFrame, destructure it and return the underlying table and
+    any filter operations that have been applied, in application order.
+    """
+    root_frame, filters, _ = get_frame_operations(frame)
+    return filters, root_frame
+
+
+def get_sorts(frame):
+    """
+    Given a ManyRowsPerPatientFrame, destructure it and return any sort operations that
+    have been applied, in application order.
+    """
+    _, _, sorts = get_frame_operations(frame)
+    return sorts
+
+
+def get_frame_operations(frame):
+    """
+    Given a ManyRowsPerPatientFrame, destructure it and return the underlying table and
+    any filter and sort operations that have been applied, in application order.
+    """
+    filters = []
+    sorts = []
+    while True:
+        type_ = type(frame)
+        if type_ is Filter:
+            filters.insert(0, frame)
+            frame = frame.source
+        elif type_ is Sort:
+            sorts.insert(0, frame)
+            frame = frame.source
+        elif type_ is SelectTable:
+            return frame, filters, sorts
+        else:
+            assert False, f"Unexpected type: {frame}"
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -78,6 +78,10 @@ class Column:
 class TableSchema:
     "Defines a mapping of column names to column definitions"
 
+    @classmethod
+    def from_primitives(cls, **kwargs):
+        return cls(**{name: Column(type_) for name, type_ in kwargs.items()})
+
     def __init__(self, **kwargs):
         self.schema = kwargs
 
--- a/databuilder/query_model.py
+++ b/databuilder/query_model.py
@@ -165,11 +165,6 @@ class ManyRowsPerPatientSeries(Series):
     ...
 
 
-# A Frame which has had a Sort operation applied to it
-class SortedFrame(ManyRowsPerPatientFrame):
-    ...
-
-
 # A OneRowPerPatientSeries which is the result of aggregating one or more
 # ManyRowsPerPatientSeries
 class AggregatedSeries(OneRowPerPatientSeries):
@@ -219,13 +214,13 @@ class Filter(ManyRowsPerPatientFrame):
     condition: Series[bool]
 
 
-class Sort(SortedFrame):
+class Sort(ManyRowsPerPatientFrame):
     source: ManyRowsPerPatientFrame
     sort_by: Series[Comparable]
 
 
 class PickOneRowPerPatient(OneRowPerPatientFrame):
-    source: SortedFrame
+    source: Sort
     position: Position
 
 
--- a/databuilder/query_model_transforms.py
+++ b/databuilder/query_model_transforms.py
@@ -14,10 +14,9 @@ The transformations applied here are all about efficient execution, and therefor
 want to keep them separate from the core query model classes.
 """
 import copy
-from abc import ABC
-from collections.abc import Mapping, MutableMapping, MutableSet
 from typing import Any
 
+from databuilder.collections_utils import DefaultIdentityDict, IdentitySet
 from databuilder.query_model import (
     Case,
     Function,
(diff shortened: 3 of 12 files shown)


evansd (approved):
It's very pleasing that this doesn't even touch the query engines.

I've added one, annoying, comment on the boolean handling stuff but I haven't said much beyond that. There were a few comments which I had to go back and delete because you addressed them in later commits.

I wonder if we might end up, further down the road, handling all the sort lifting here rather than in the query engine, which I think might simplify both this code _and_ the query engine. But I don't think it's worth thinking about now.

> Should the new transformation be separate or should it be rolled into the existing one. My weak preference is to separate them because they exist for different reasons, but their implementation ties them together so there is an argument for treating them as a single transformation.

My preferences here are reasonably weak as well. For what it's worth, I tend to swing the other way because I feel like the sort handling code as a whole would be easier to understand if was handled in one go. I think anyone working on any of the sort code would need to understand and be mindful of both parts, and so I don't see that separating them aids comprehension. But I've no real objection to the code as it is; and I'm happy to go with your preferences here given that you've had to write it.

> The unit tests for the transformations are pretty unreadable (and even hard to write). Should we introduce some kind of QM builder tooling to make tests like this more readable?

It's the inside-out ordering that makes them so hard to take in, isn't it? Did you see that [attempt](https://gist.github.com/evansd/924c7c5fe9f9476215df453482a29dd4) I made at a fluent query model? Leaving aside the magical implementation, the basic idea of being able to write `foo.operation(bar)` rather than `Operation(foo, bar)` seems to me to go a long way to increasing readability while keeping very close to "raw" query model. We could implement something like this the tedious, magic-free way if we thought it was worthwhile?

databuilder/query_model.py
```diff
@@ -49,7 +50,7 @@
 # type without specifying what that type has to be
 T = TypeVar("T")
 Numeric = TypeVar("Numeric", int, float)
-Comparable = TypeVar("Comparable", int, float, str, date)
+Comparable = TypeVar("Comparable", int, float, str, date, BaseCode)
```
inglesp, line 53: Commenting here to respond to the commit message:

> This isn't something that researchers will ever want to do, but it
> allows us to use code columns for tie-breaking sorts to ensure sort
> stability.

I think users may want to do this if they want to ensure exactly the same output between runs.  I think we ensure sort stability in parts of cohort-extractor for this reason.

databuilder/query_model_transforms.py
```diff
@@ -31,72 +36,49 @@ def apply_transforms(variables):
...
-    """
-    Replace references to `old_ref` with `new_ref` on `target`
-    """
-    # We're only expecting nodes with a source attribute here
-    assert target.source == old_ref
-    force_setattr(target, "source", new_ref)
-    # We're only expecting nodes which reference a single input
-    assert get_input_nodes(target) == [new_ref]
+
+def get_all_nodes(variables):
```
inglesp, line 144: I'm unkeen on having functions called `all_nodes` and `get_all_nodes` in the same module, without a docstring indicating why they're different and what they're for.

databuilder/query_model_transforms.py
```diff
@@ -31,72 +36,49 @@ def apply_transforms(variables):
...
+    nodes = []
+    for query in variables.values():
+        nodes.extend(all_nodes(query))
+    return nodes
+
+
+def get_dependers(node, nodes):
+    dependers = []
+    for other_node in nodes:
```
inglesp, line 153: Given that this is called from inside a loop over `nodes`, this has quadratic behaviour.  Maybe ok for now?

databuilder/query_model_transforms.py
```diff
@@ -31,72 +36,49 @@ def apply_transforms(variables):
...
-    assert get_input_nodes(target) == [new_ref]
+
+def get_all_nodes(variables):
+    nodes = []
+    for query in variables.values():
+        nodes.extend(all_nodes(query))
+    return nodes
+
+
+def get_dependers(node, nodes):
```
inglesp, line 151: Docstring?

databuilder/query_model_transforms.py
```diff
@@ -68,10 +69,34 @@ def add_selected_columns_to_pick_row(nodes):
...
+    We want to ensure that results are consistent when picking the first or last of a sorted
+    frame, even when the sorting specified in the variable definition isn't sufficient to completely
+    specify the order. To that end we add extra sorts for each column that is ultimately going to
+    be returned from the sorted table.
+
+    This has the additional benefit that the order is the same between different DBMSes, which
+    doesn't help our users but makes testing a bit easier.
+    """
+    for node in nodes:
+        if not isinstance(node, PickOneRowPerPatientWithColumns):
```
inglesp, line 86: Why would this ever be hit?

tests/unit/test_query_model_transforms.py
```diff
@@ -59,3 +59,79 @@ def test_pick_one_row_per_patient_transform():
...
+
+def test_adds_one_selected_column_to_sorts():
+    events = SelectTable(
+        "events",
+        TableSchema(i1=Column(int), i2=Column(int)),
+    )
+    by_i1 = Sort(events, SelectColumn(events, "i1"))
+    variables = dict(
+        v=SelectColumn(
+            PickOneRowPerPatient(source=by_i1, position=Position.FIRST), "i2"
```
inglesp, line 74: Might help with legibility here and elsewhere if you put a trailing comma in this line, and then black would reformat it so that `"i2"` was on a new line.  This would make it clearer (to me) that `"i2"` is the second argument of the `SelectColumn`.

tests/unit/test_query_model_transforms.py
```diff
@@ -59,3 +59,79 @@ def test_pick_one_row_per_patient_transform():
...
+                            name="i3",
+                        ),
+                    }
+                ),
+            ),
+            "i3",
+        ),
+    )
+
+    assert transformed == expected_variables
```
inglesp, line 139: Silly thought: could we write these tests in such a way that we could easily extract the input and expected output, and then as a post-processing step produce ascii diagrams of the two graphs, to make it easier to see what the transformation should do?

benbc:
@evansd I'll merge the transforms. I think you're weak preferences should hold sway over mine in this codebase.

benbc:
@evansd My inclination with the QM writing code is to leave it to a possible future piece of standalone work that the team can prioritize. Are you happy with that?

benbc:
@evansd I forgot something! I put the sortable-codes commit in just to get it working, but it wasn't my intention to force it through without more discussion. Do you want to suggest an alternative?

databuilder/query_model_transforms.py
```diff
@@ -131,6 +134,13 @@ def get_immediate_sorts(node):
     return sorts
 
 
+def make_sortable(col):
+    if get_series_type(col) == bool:
+        # Some databases can't sort booleans (including SQL Server), so we map them to integers
+        return Case(cases={col: Value(1)}, default=Value(0))
```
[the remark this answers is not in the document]
benbc: Thank you. I did think about nulls at some point, but then forgot about it.

tests/spec/sort_and_pick/test_sort_extends_to_all_columns_when_underspecified.py
```diff
@@ -0,0 +1,24 @@
...
+        2 | 101 |  0 |   1
+        """,
+}
+
+
+def test_sorting_extends_to_selected_column(spec_test):
+    spec_test(
+        table_data,
+        e.sort_by(e.i1, e.i2).first_for_patient().i3,
+        {1: 102, 2: 500},
```
inglesp, line 23: I can't see where the new behaviour is actually tested here.  The sort order is entirely specified by the values in `i1` and `i2`, isn't it?

inglesp (approved):
Thanks for doing this.  A few suggestions, but all looks sensible.

This is at the limit of what I can keep in my head a the moment though.  As you've identified, we should think about how we can make the tests more readable.

benbc:
@evansd 

> I wonder if we might end up, further down the road, handling all the sort lifting here rather than in the query engine, which I think might simplify both this code and the query engine.

Yes, I had the same thought.

databuilder/query_engines/in_memory.py
```diff
@@ -100,7 +103,7 @@ def visit_Sort(self, node):
         sort_index = self.visit(node.sort_by).sort_index()
         return source.sort(sort_index)
 
-    def visit_PickOneRowPerPatient(self, node):
+    def visit_PickOneRowPerPatientWithColumns(self, node):
```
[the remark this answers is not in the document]
benbc: No. This latest change is not a transformation, it changes the behaviour by stabilizing under-specified sorts.

databuilder/query_model.py
benbc: My claim is that researchers don't every really want to sort by code for its own sake. They may want to have sort stability, but this isn't the commit that gives that.

databuilder/query_model_transforms.py
benbc: :+1: thanks

databuilder/query_engines/in_memory.py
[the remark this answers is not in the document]
inglesp: Ah yes, of course.

databuilder/query_model.py
inglesp: Makes sense.

databuilder/query_model_transforms.py
benbc: Each transformation inspects every single node.

databuilder/query_model_transforms.py
benbc: It does. I'm v happy to replace with a single pass and a dict if you prefer.

We have a _lot_ of QM graph traversal in the codebase. Could be worth doing some optimization at some point, but I'm not sure this is particularly like to be the worst offender.

databuilder/query_model_transforms.py
benbc: :+1:

tests/unit/test_query_model_transforms.py
benbc: Yeah, I think that's a bit better. I think it's a fools errand to try to make these beautiful without more tooling though.

tests/unit/test_query_model_transforms.py
benbc: That does sound helpful, but I'm skeptical that the effort would pay off at least until we have more transformations.

tests/spec/sort_and_pick/test_sort_extends_to_all_columns_when_underspecified.py
benbc: Yes, you're quite right. This test underwent last-minute revision that has not been kind to it. Thank you.

evansd:
@benbc

> My inclination with the QM writing code is to leave it to a possible future piece of standalone work that the team can prioritize. Are you happy with that?

Yes, I agree.

---

> I forgot something! I put the sortable-codes commit in just to get it working, but it wasn't my intention to force it through without more discussion. Do you want to suggest an alternative?

I think there are two questions here: the important one is whether we want to allow users to sort on code columns, as that becomes part of the public API we need to support — whether we document it or not. There's a secondary question about the implementation (what should import from what) which is something we can revisit later.

I'm coming tentatively to the view that we should allow sorting on **all columns**, including both codes and booleans.

If a dataset definition has an under-specified sort then its results are arbitrary. This PR helpfully ensures that it's _consistently_ arbitrary, which is an important feature; but I think ideally any tie-breaks required should be specified in the dataset definition itself so they can be reviewed for possible statistical implications.

We _could_ require every sort to be fully specified, but that feels quite user-unfriendly. (Maybe we could make it a warning or something though?) But I think we should at least provide the tools necessary to fully specify the sort. And that means being able to sort on all columns.

That's a long-winded way to say: I'm happy with this PR as it is on this point, and I'm thinking we should go further.

evansd (approved):
Remember when we thought this might be a one-line change? :grinning: 

I'm generally happy with this (and of course delighted with the behaviour it adds). The thing which gives me most pause is the change to the query engine to require filters before sorts. It seems to me to not simplify the code all that much. And the `pop_filters`/`pop_sorts` functions will give silently incorrect results if the implicit invariant they rely on is violated.

The mutating-the-immutable approach is also starting to feel not quite as quick a hack as it started out as. But then again, the various IdentityCollections are conceptually very straightforward even if they require some boilerplate. And once we hide them away in a utils module the transform code itself is pretty simple.

databuilder/query_model_transforms.py
```diff
@@ -28,75 +42,184 @@ class PickOneRowPerPatientWithColumns(PickOneRowPerPatient):
...
+def make_sortable(col):
+    if get_series_type(col) == bool:
+        # Some databases can't sort booleans (including SQL Server), so we map them to integers
+        return Case(
+            cases={col: Value(2), Function.Not(col): Value(1)}, default=Value(0)
         )
-        # Update any references to the old node to point to the new one
-        for ref_node in references:
-            update_references(ref_node, node, new_node)
+    return col
```
inglesp, line 184: Do we need a test to demonstrate the new behaviour, that boolean values are sorted as we expect?

inglesp (commented):
I'm afraid I don't have the time to do a thorough review here.

databuilder/query_model_transforms.py
benbc: We do! Added, thanks.

benbc:
@evansd I'm all ready to rip out the change here to canonicalize the sort/filter order in the transforms. But I wonder if I've missed some opportunities to simplify the query engine based on that canonicalization which might tip the balance. I felt as I was writing it that I might be missing something. Do you have a couple of minutes to double check the query engine [as it is in this branch](https://github.com/opensafely-core/databuilder/blob/benbc/generative-test-improvements/databuilder/query_engines/base_sql.py) to see if you can spot any opportunities for further simplification?

databuilder/query_model_transforms.py
```diff
@@ -200,6 +201,27 @@ def get_dependers(node, nodes):
     return dependers
 
 
+def get_frame_operations(frame):
```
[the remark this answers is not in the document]
benbc: I don't think so. It's moved here from `query_model` as a deliberate part of this commit because it's only used here.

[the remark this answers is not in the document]
evansd: Ah sorry, yes. It's moved around a bit and I got confused.

evansd:
@benbc I've taken another look at the code in that branch. And I'm remember spending a while thinking about this when first reviewing this. I don't feel like there are any massive simplifications lying around here (though I'm so used to a certain way of thinking about this code that I may well be missing something obvious).

My reasoning is that, one way or another, we need to walk a chain of operations and the only difference is that in one case the chain co-mingles two types of operation and in the other case the types are cleanly separated. But walking the co-mingled chain just isn't that much more algorithmically difficult (as I see it) than walking the separated one. Particularly if, for the sake of robustness, we decide we need some kind of assertion that the separated chain really _is_ separated.

Again, I might be missing something obvious, but it isn't obvious (yet) to me.

benbc:
@evansd Thank you. I've removed those changes. The commits were pretty much orthogonal to the rest of the PR and rebased out cleanly, so there is no need for you to re-review.

PR #786, opened 2022-10-13
Dummy data: attempt 1

### LINKED ISSUE #764: Add worse-is-better support for generating dummy data
This is mostly a placeholder ticket for gathering thoughts. But as [discussed](https://bennettoxford.slack.com/archives/C03FB777L1M/p1664798224154799) in Slack, we think having some kind of system for generating dummy data (however inadequate) is crucial for getting people to start learning ehrQL.

evansd:
This adds a first pass at a dummy data generator. It has some obvious areas for improvement but it produces non-empty output when run against the comparative booster study (500 rows in 18 seconds) which was our original goal.

While the results are pretty nonsensical (a majority patients have emergency admissions for COVID and many end up spending hundreds of days in critical care) certain things drop out nicely. All the sequential dates (vaccination 1, vaccination 2, admission 1, admission 2 etc) are in the right order and you never end up with a second event without having a first. The columns recording dates for "any COVID vaccine" tie correctly up with the columns for the specific vaccine products. No patients have events after they've died. And the "Age in Aug 2021" agrees sensibly with the "Age at baseline date" column.

Pleasingly, we've managed to get these results with very little domain specific knowledge built in to the generator. It knows that a patient has a date of birth and, potentially, death and that these delineate the range of time during which things can happen to them. And for simplicity, I've made each patient be registered with a single practice from birth. But beyond that it doesn't know anything about the tables other than what's in the contracts.

The current design allows us to define custom generators for specific columns in specific tables, and then it automatically fills in any values which aren't covered by those. You'll see there are two very simple generators: one for `patients` and one for `practice_registrations`.

In terms of magic constants needed to get this particular study to work, there are two. The main one is that we generate dates with a heavy recency bias: an exponential distribution with a mean of one year ago. We also only generate a small handful of rows for each patient in each table (between 0 and 7). These are needed because an inclusion criterion for the study is having your fourth vaccine dose within a specific time frame. My first attempts ended with most patients getting their first doses back in the 80s and being on their 40th dose by the start of the pandemic, so very few made it into the study.

It would be quite simple to apply these constraints just to the `vaccinations` table and have other tables generate more events and over a wider date range. But I wanted to see how little we could get away with at this stage. Eventually I suspect we'll need to teach the generator specifically about the COVID vaccines and their sensible date distributions. 

Closes #764

### DIFF
--- a/databuilder/dummy_data/generator.py
+++ b/databuilder/dummy_data/generator.py
@@ -24,27 +24,12 @@ class DummyDataGenerator:
 
     def __init__(self, variable_definitions):
         self.variable_definitions = variable_definitions
-
-        # TODO: I dislike using today's date as part of the data generation because it
-        # makes the results non-deterministic. However until we're able to infer a
-        # suitable time range by inspecting the query, this will have to do.
-        self.today = date.today()
-        self.rnd = random.Random()
-
-        # Create ORM classes for each of the tables used in the dataset definition
-        self.query_info = QueryInfo.from_variable_definitions(self.variable_definitions)
-        Base = declarative_base()
-        self.orm_classes = {
-            table_info.name: orm_class_from_schema(
-                Base,
-                table_info.name,
-                table_info,
-                table_info.has_one_row_per_patient,
-            )
-            for table_info in self.query_info.tables.values()
-        }
+        self.patient_generator = DummyPatientGenerator(
+            self.variable_definitions, self.random_seed
+        )
 
     def get_data(self):
+        generator = self.patient_generator
         data = []
         found = 0
 
@@ -65,7 +50,7 @@ class DummyDataGenerator:
             # membership) and find those matching the population definition
             patient_batch = {
                 patient_id: list(
-                    self.get_patient_data_for_population_condition(patient_id)
+                    generator.get_patient_data_for_population_condition(patient_id)
                 )
                 for patient_id in range(batch_start, batch_start + self.batch_size)
             }
@@ -76,7 +61,7 @@ class DummyDataGenerator:
                 data.extend(patient_batch[row.patient_id])
                 # Include additional data needed for the dataset but not required just
                 # to determine population membership
-                data.extend(self.get_remaining_patient_data(row.patient_id))
+                data.extend(generator.get_remaining_patient_data(row.patient_id))
                 found += 1
                 if found >= self.population_size:
                     break
@@ -99,10 +84,7 @@ class DummyDataGenerator:
                 # means that we get an empty dataset rather than an error, and can
                 # create empty CSV tables with the right headers.
                 if not data:
-                    data = [
-                        orm_class(patient_id=1)
-                        for orm_class in self.orm_classes.values()
-                    ]
+                    data = generator.get_one_empty_row_for_each_table()
                 return data
 
         # Keep coverage happy: the loop should never complete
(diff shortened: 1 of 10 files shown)


databuilder/dummy_data_generator.py
```diff
@@ -1,20 +1,239 @@
...
+            # There's not much we can about non-categorical strings with no values we
+            # can find in the query, but we shouldn't have many of these
+            return ""
+        elif type_ is date:
+            # Use an exponential distribution to preferentially generate recent events
+            # (mean of one year ago)
+            days_ago = int(self.rnd.expovariate(1 / 365))
+            event_date = self.events_end - timedelta(days=days_ago)
+            # Clip to the available time range
+            return max(event_date, self.events_start)
```
iaindillingham, line 125: I'm really taken with this 🙂 As elsewhere, my main thought is "What if we wanted to use another distribution?" but you've already noted that there are some potential areas for improvement. If we did want to use another distribution, then would you try to accommodate it within this architecture? (e.g. `get_random_value` could be a visitor).

iaindillingham (commented):
I'm going to comment, rather than approve, because this is in draft. What are the next steps? For example, increase the number of rows with non-empty output? Decrease the time taken to generate and query 5,000 patients' worth of data?

inglesp (approved):
A decent first stab, well done.

Out of interest, could we derive 177/5000 from first principles?

databuilder/query_engines/in_memory.py
```diff
@@ -61,8 +65,14 @@ def all_patients(self):
...
     def visit(self, node):
+        try:
+            return self.cache[node]
+        except KeyError:
+            pass
         visitor = getattr(self, f"visit_{type(node).__name__}")
-        return visitor(node)
+        value = visitor(node)
+        self.cache[node] = value
+        return value
```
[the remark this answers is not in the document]
benbc: > Given Ben's fun with frozensets, etc, have we thought through the implications of this?

Shouldn't be a problem if we're not simultaneously busting these objects' contracts by modifying the unmodifiable.

evansd:
Given that the code is significantly different from (although obviously the spiritual successor to) the initial spike I've taken the liberty of resolving all outstanding comments. Feel free to re-open any which you feel still apply.

iaindillingham (approved):
The initial commits were familiar; the latter commits less so. However, I'm reasonably confident that I understand what's going on and certainly don't have any suggestions, constructive or otherwise. I'm going to block out some time next week to consolidate *Dummy Data, Part One*. That shouldn't block merging this PR, though, so 👍🏻

databuilder/dummy_data/query_info.py
```diff
@@ -0,0 +1,128 @@
...
+                )
+                table_info.columns[name] = column_info
+            # Record the ColumnInfo object associated with each SelectColumn node
+            column_info_by_column[column] = column_info
+
+        # Record values used in equality and substring comparisons
+        for node in by_type[Function.EQ] | by_type[Function.StringContains]:
+            # The query model in theory supports "1 == x" style comparisons (i.e.  with
+            # the column on the RHS) but there's no way to generate such constructions
+            # using ehrQL so we only bother handling the "x == 1" orientation here.
```
inglesp, line 113: What if something in ehrQL changes, or if we use another QM front end?

databuilder/dummy_data/generator.py
```diff
@@ -1,20 +1,142 @@
...
-        self.tables = get_table_nodes(*variable_definitions.values())
-        self.orm_classes = orm_classes_from_qm_tables(self.tables)
+
+        # TODO: I dislike using today's date as part of the data generation because it
+        # makes the results non-deterministic. However until we're able to infer a
+        # suitable time range by inspecting the query, this will have to do.
+        self.today = date.today()
+        self.rnd = random.Random()
+
+        # Create ORM classes for each of the tables used in the dataset definition
```
inglesp, line 34: Does this comment belong a line or two below?

databuilder/dummy_data/generator.py
```diff
@@ -1,20 +1,142 @@
...
+        for patient_id in range(1, self.population_size + 1):
+            yield from self.get_patient_data(patient_id)
 
     def get_results(self):
         database = InMemoryDatabase()
         database.setup(self.get_data())
         engine = InMemoryQueryEngine(database)
         return engine.get_results(self.variable_definitions)
+
+    def get_patient_data(self, patient_id):
```
inglesp: Should this be a method on a different class?  I think this class is carrying too much state -- and it's responsible both for generating all dummy data, and for generating data for one patient.

inglesp (approved):
Great, :shipit:

databuilder/dummy_data/query_info.py
evansd: We'll fail to record some values and generate "worse quality" dummy data. Under some circumstances that could mean that we can't generate any matching patients at all. But given all the other ways dummy data generation can go wrong it didn't strike me as hugely high priority. And it's a pain to write tests for because you have to construct the QM by hand.

databuilder/dummy_data/generator.py
evansd: I actually originally had it like that and then I felt like there were too many classes floating around and it didn't help much. But that was before the extra complexity I added with the batching/filtering and lazy generation. So I think you're right now. I'll split them back out.

databuilder/dummy_data/query_info.py
inglesp: Fair enough!

databuilder/dummy_data/generator.py
inglesp: Doesn't have to happen before half term!

PR #828, opened 2022-11-16
Better date arithmetic

### LINKED ISSUE #519: Improve date handling in ehrQL
We want to get rid of `add_days()`, `subtract_days()` and `date_difference_in_years()`, and replace them with a date delta type.

We'd like to support:

```
dataset.some_date = dataset.other_day + days(10)
dataset.some_date = dateset.other_date + days(dataset.some_integer)
dataset.age = (dataset.some_date - patients.date_of_birth).years
```

Note that we plan to make age a reusable variable.

Whoever implements this (ideally not @evansd) should discuss plan of attack with @evansd after some initial experiments.

(edited by @inglesp)

evansd:
This implements the date API roughly as described in #519. Specifically:

1. There are now three "duration" constructors:
   ```py
   from databuilder.ehrql import days, months, years
   ```
   These can be added to, or subtracted from, any date-like thing which means:

   * An ehrQL `DateSeries` e.g. `patients.date_of_birth + years(10)`
   * A `datetime.date` e.g. `date(2020, 1, 1) + months(10)`
   * An ISO date string e.g. `"2020-01-01" - days(10)`

2. An ehrQL `DateSeries` can be subtracted from any date-like thing, and vice versa, to produce a "date difference" object with three attributes: `days`, `months`, `years`.

   Each of these attributes is an ehrQL `IntSeries` e.g.
   ```py
   is_over_60_in_2020 = ("2020-01-01" - patients.date_of_birth).years > 60
   ```

Arithmetic using months and years suffers from ambiguities with multiple defensible resolutions. As you'll see in the comments, I've tried to take a principled approach derived from our original definition of age, which considers the birthday of those born of 29 February to be 1 March in non-leap years. This will need to be explained in the documentation. However, because of these ambiguities, I think the documentation should also push users away from using months and years as units and towards days (and weeks, when we support them). I've made a [ticket](https://github.com/opensafely-core/databuilder/issues/830) to track this.

Related to this, making the month/year behaviour consistent across databases requires some slightly convoluted SQL in places. There may well be further optimisation that can be done here, but this is yet another reason to encourage users away from these units.

The old methods (`difference_in_years`, `add_days`, `subtract_days`) still exist for now but I'd like to remove them as soon as possible. To this end, I plan to submit PRs to all our current ehrQL projects once this PR is merged.

Finally, as mentioned before, we should support `weeks` as a unit. This should be reasonably straightforward once the [multiplication/division](https://github.com/opensafely-core/databuilder/issues/657) ticket is closed. I've made another [ticket](https://github.com/opensafely-core/databuilder/issues/829) to track this.

Closes #519


databuilder/query_language.py
```diff
@@ -319,6 +319,9 @@ def __add__(self, other):
         else:
             return NotImplemented
 
+    def __radd__(self, other):
+        return self.__add__(other)
+
```
benbc, line 330: _Maybe_ worth a comment on `__sub__()` pointing out the `__rsub__()` would be senseless?

tests/spec/date_series_ops/test_date_series_ops.py
```diff
@@ -166,6 +166,29 @@ def test_difference_between_dates_in_years(spec_test):
...
+        (p.d1 - "2000-01-01").days,
+        {
+            1: 0,
+            2: 31,
+            3: -2,
+            5: None,
+        },
+    )
+
+
```
benbc, line 307: Maybe add cases over the end of Feb for leap/non-leap years so the behaviour is super-clear?

databuilder/query_engines/mssql.py
```diff
@@ -49,6 +49,24 @@ def date_add_days(self, date, num_days):
...
+            "DATEFROMPARTS",
+            self.get_date_part(date, "YEAR") + num_years,
+            self.get_date_part(date, "MONTH"),
+            1,
+            type_=sqlalchemy_types.Date,
+        )
+        # Then add on the number of days we're offset from the start of the month which
+        # has the effect of rolling 29 Feb over to 1 Mar as we want
+        return self.date_add_days(start_of_month, self.get_date_part(date, "DAY") - 1)
+
```
benbc, line 87: Rather you than me.

databuilder/query_engines/sqlite.py
```diff
@@ -20,8 +20,14 @@ def get_date_part(self, date, part):
...
-        num_days_str = sqlalchemy.cast(num_days, sqlalchemy_types.String)
-        modifier = num_days_str.concat(" days")
+        return self.date_add("days", date, num_days)
+
+    def date_add_years(self, date, num_years):
+        return self.date_add("years", date, num_years)
+
+    def date_add(self, units, date, value):
+        value_str = sqlalchemy.cast(value, sqlalchemy_types.String)
+        modifier = value_str.concat(f" {units}")
```
benbc, line 33: Good to see that SQLite has you holding its beer, as usual.

benbc (approved):
This has worked out more straight-forwardly than I feared. Well done for finding what I think are reasonable SQL implementations for everything.

Python's operator overloading makes it hard to keep track of which operations are allowed on which types (e.g. `Duration.__add__` doesn't accept series types, but that's fine because it's handled elsewhere). Obviously we have the specs, but even they are too verbose to get an overview easily. Just a comment, really -- I don't have any smart ideas other than some kind of table test. (Oh, I see now that you've kind of added that in your last commit; any chance of extending that to inlcude series, explicitly demonstrate unsupported cases and doing it for `DateDifference`s as well?)

Did you consider making `Duration`s addable to each other?

I think there is another justification for the roll-over behaviour for "short" months. If you ask yourself "have N months/years passed yet" on the last day of the previous month then the answer is clearly "no", whereas if you ask yourself the same question on the first day of the next month then the answer is clearly "yes".

evansd:
> any chance of extending that to inlcude series, explicitly demonstrate unsupported cases and doing it for DateDifferences as well?

That's a good suggestion, I'll do that.

> Did you consider making Durations addable to each other?

I didn't actually, and my first reaction was that this is quite a nice idea. But then I realised that because of All the Terribleness, date addition isn't associative e.g. when adding a year and two days to 29 Feb 2000 you get 3 Mar 2001 if you add the year first and then the days, and you get 2 Mar if you add the days first and then the year. So `years(1) + days(2)` would need to produce some kind of CompositeDuration which keeps track of all its components and applies them in the right order. And then I felt quite tired. We could certainly do it though, if it turns out to be a feature people need.

> I think there is another justification for the roll-over behaviour for "short" months. If you ask yourself "have N months/years passed yet" on the last day of the previous month then the answer is clearly "no", whereas if you ask yourself the same question on the first day of the next month then the answer is clearly "yes".

Yeah, I'm convinced this is the only reasonable approach and that anyone doing it differently is just Wrong On the Internet. I kept thinking of different ways of justifying it and then worrying that there was something question-begging about them. And then I thought maybe it's not that important as long we clearly document our behaviour and apply it consistently. It would be nice to have a succinct knockdown argument for it though.

benbc:
> I'm convinced this is the only reasonable approach

Tell that to all the people with leap day birthdays who celebrate their birthday on Feb 28th.

> date addition isn't associative

Ah-ha-ha-ha-ha. For heaven's sake.

> We could certainly do it though

I think I disagree. I don't think that introducing non-associative addition operations is going to be helping anyone. They can always do it if they really need to by going via ints.

Issue #505, opened 2022-05-27
More helpful error handling in ehrQL


evansd:
This is a grab bag of things that will no doubt need splitting into separate issues once we decide what we're doing here.

---
There are libraries which improve on Python's default traceback and exception display. Given that we control the context in which the dataset definition file is imported, we can ensure one of these libraries is configured at the point where we import the user's code. One thing we could do would be to hide traceback lines involving databuilder code and point directly to the line of user code which triggered it. But something like `friendly` provides many more options for customising what we show the user.
https://github.com/friendly-traceback/friendly-traceback
https://github.com/Qix-/better-exceptions

---
The query model is the ultimate enforcer of validity, but the errors it throws are often not going to be that helpful to the user. We should consider catching these inside the `query_language._wrap` function and translating them into something more friendly. This will probably involve changing the query model errors to be more structured (i.e. include additional attributes beyond a string message) so ehrQL can translate them better.

---
The refactored ehrQL implementation has a distinct Event and Patient Series class for every type of thing (int, date, CTV3Code etc) you can have in a series. This means that a given series object will only have methods corresponding to valid operations for that series. That's a good thing. However it would be very useful to communicate to the user whether they've just typoed a method name (e.g. `values.sum_for_pateint()`) or whether they've used a method which does exist in general, just not for the specific type in question (e.g `some_date_series.sum_for_patient()`).

There's further complexity here because we use attribute access to select columns from frames, so we need to distinguish:
 * Accessing a column which doesn't exist — `frame.bad_column_name`
 * Typoing a method — `frame.sortby()`
 * Calling a method which exists on same frames but not on this one — `not_sorted_frame.first_for_patient()`

inglesp:
Have created #533 to cover part of the first point.

StevenMaude:
Having just deliberately tested things out to break them, I'll add an example of a simple case so there's a real example of the current state:

```
Traceback (most recent call last):
  File "databuilder/ehrql-tutorial-examples/3a_multiple2_dataset_definition.py", line 22, in <module>
    latest_imd_is_at_least_5000 = latest_imd >= 5000.0
databuilder.query_model.TypeValidationError: GE.rhs requires 'databuilder.query_model.Series[int]' but got 'databuilder.query_model.Series[float]'
In node:
[...]

PR #840, opened 2022-11-24
Disallow frames as columns

benbc:
This was a hard-to-debug error I ran into writing a real study with a researcher.

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -45,6 +45,8 @@ class Dataset:
             )
         if name in self.variables:
             raise AttributeError(f"'{name}' is already set and cannot be reassigned")
+        if name == "variables":
+            raise AttributeError("'variables' is not an allowed variable name")
         if name.startswith("__"):
             raise AttributeError(
                 f"Variable names must not start with underscores (you defined a variable '{name}')"
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -31,16 +31,19 @@ class _DictArg(list):
 
 
 class Dataset:
+    def __init__(self):
+        object.__setattr__(self, "variables", {})
+
     def set_population(self, population):
         validate_population_definition(population.qm_node)
-        object.__setattr__(self, "population", population)
+        self.variables["population"] = population
 
     def __setattr__(self, name, value):
         if name == "population":
             raise AttributeError(
                 "Cannot set column 'population'; use set_population() instead"
             )
-        if hasattr(self, name):
+        if name in self.variables:
             raise AttributeError(f"'{name}' is already set and cannot be reassigned")
         if not isinstance(value, BaseSeries):
             raise TypeError(
@@ -50,11 +53,16 @@ class Dataset:
             raise TypeError(
                 f"Invalid column '{name}'. Dataset columns must return one row per patient"
             )
-        super().__setattr__(name, value)
+        self.variables[name] = value
+
+    def __getattr__(self, name):
+        if name in self.variables:
+            return self.variables[name]
+        raise AttributeError(f"Variable '{name}' has not been defined")
 
 
 def compile(dataset):  # noqa A003
-    return {k: v.qm_node for k, v in vars(dataset).items() if isinstance(v, BaseSeries)}
+    return {k: v.qm_node for k, v in dataset.variables.items()}
 
 
 # BASIC SERIES TYPES
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -41,17 +41,17 @@ class Dataset:
     def __setattr__(self, name, value):
         if name == "population":
             raise AttributeError(
-                "Cannot set column 'population'; use set_population() instead"
+                "Cannot set variable 'population'; use set_population() instead"
             )
         if name in self.variables:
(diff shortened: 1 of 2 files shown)


databuilder/query_language.py
```diff
@@ -31,16 +31,19 @@ class _DictArg(list):
 
 
 class Dataset:
+    def __init__(self):
+        object.__setattr__(self, "variables", {})
```
benbc, line 35: This is ugly, but I can't see a better way. Happy to hear suggestions.

databuilder/query_language.py
```diff
@@ -50,11 +53,16 @@ def __setattr__(self, name, value):
             raise TypeError(
                 f"Invalid column '{name}'. Dataset columns must return one row per patient"
             )
-        super().__setattr__(name, value)
+        self.variables[name] = value
```
benbc, line 56: I'm not set on this change. Seems slightly better to be more explicit, but since we're overriding `__setattr__` I don't think that there is a completely clean implementation.

evansd: I think this is a definite improvement, thanks for doing it.

databuilder/query_language.py
```diff
@@ -31,16 +31,19 @@ class _DictArg(list):
...
+    def __init__(self):
+        object.__setattr__(self, "variables", {})
+
     def set_population(self, population):
         validate_population_definition(population.qm_node)
-        object.__setattr__(self, "population", population)
+        self.variables["population"] = population
 
     def __setattr__(self, name, value):
         if name == "population":
```
evansd, line 42: I think this means `variables` is now a reserved (albeit silly) name so we should raise an error if it's used. Which makes me think that `_variables` would be a better name, and then we could reject any names starting with an underscore.

databuilder/query_language.py
```diff
@@ -50,11 +53,16 @@ def __setattr__(self, name, value):
             raise TypeError(
                 f"Invalid column '{name}'. Dataset columns must return one row per patient"
             )
-        super().__setattr__(name, value)
+        self.variables[name] = value
+
+    def __getattr__(self, name):
+        if name in self.variables:
+            return self.variables[name]
+        raise AttributeError(f"Variable '{name}' has not been defined")
```
evansd, line 61: :heart_eyes:

evansd (approved):
I think this is great. As you say, it would be nice not to have to use `object.__setattr__` but I don't see a way round it either.

I've jotted down some thoughts about possible future improvements — but they are very much that, rather than things to change now.

tests/unit/test_query_language.py
```diff
@@ -89,7 +89,7 @@ def test_cannot_reassign_dataset_column():
         dataset.foo = patients.date_of_birth.year + 100
 
 
-def test_cannot_assign_frame_to_column():
```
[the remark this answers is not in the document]
benbc: It did slow me down a little.

tests/unit/test_query_language.py
```diff
@@ -89,6 +89,13 @@ def test_cannot_reassign_dataset_column():
         dataset.foo = patients.date_of_birth.year + 100
 
 
+def test_cannot_assign_frame_to_column():
+    dataset = Dataset()
+    dataset.set_population(patients.exists_for_patient())
```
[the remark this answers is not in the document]
benbc: Indeed. Serves me right for copy-pasting (from another test that also doesn't need it). Fixed.

databuilder/query_language.py
```diff
@@ -42,6 +42,10 @@ def __setattr__(self, name, value):
             )
         if hasattr(self, name):
             raise AttributeError(f"'{name}' is already set and cannot be reassigned")
+        if not isinstance(value, BaseSeries):
+            raise TypeError(
+                f"Invalid column '{name}'. Dataset columns must be values not whole rows"
```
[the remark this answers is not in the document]
benbc: Right. I did consider other error cases briefly but then realized that they are excluded by the next check which calls `value.qm_node`. Obviously the error message will be unhelpful in that case, but at least it'll trigger an error.

I think there is an issue about improving error messages. I'll add a note there.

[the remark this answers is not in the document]
benbc: #505

databuilder/query_language.py
benbc: Oh, that's a good idea -- rules out some other unlikely errors as well.

benbc: Actually, I tell a lie. When I came to implement this I realized that I did think about this and decided it was better not to use an underscore because it's used outside the class (in `compile()`). Your point about it being a reserved word stands though, so I will check for it.

evansd: Ah yes, well actually I've been meaning to move `compile()` to `Dataset._compile()`! I don't think we benefit at all from it being a separate function. It would be much nicer if we could say that a dataset is just an object with a `_compile` method that returns query model.

benbc: I think we separated them originally so that the compile method wasn't available (or not obviously available) to dataset definition writers. Giving it an underscore sort of solves that -- but at the cost of having a different client (`main`) call an underscore method.

PR #845, opened 2022-11-28
Run the gentests on a schedule

benbc:


evansd (approved):
:tada: It's great having all this stuff come together. Thank you!

.github/workflows/generative-tests.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Long-running generative tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "30 9 * * *"
```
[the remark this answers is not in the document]
benbc: Good idea. Done. (Running at midnight.)

.github/workflows/generative-tests.yml
```diff
@@ -0,0 +1,19 @@
+---
+name: Long-running generative tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "0 0 * * *"
```
bloodearnest, line 7: Minor post-merge comment: I think it's a good idea to avoid midnight scheduling with cron, exactly because it's the default.  There will be 10,000s of other `schedule` jobs with `0 0 * *` that are synchronised with this one, and it might take longer to run than other times.

In this particular case, I don't think it's a problem at all. But just thought I'd mention it.

benbc: Thanks, Simon. I've changed this.

PR #852, opened 2022-12-05
Ignore MSSQL's case-nesting limit in gentests

benbc:
Only MSSQL has this limit, which we occasionally hit with deeply-nested structures produced by Hypothesis. We'd like to be able to test that we return the correct results these cases where they are valid.

### DIFF
--- a/tests/generative/recording.py
+++ b/tests/generative/recording.py
@@ -61,10 +61,16 @@ def recorder(request):  # pragma: no cover
         with output_enabled(request):
             show_input_summary(recorder_)
 
+    check_not_too_many_ignored_errors(recorder_)
+
 
 def check_comprehensive(recorder):  # pragma: no cover
     operations_seen = {o for v in recorder.variables for o in node_types(v)}
     variable_strategies.assert_includes_all_operations(operations_seen)
+
+
+def check_not_too_many_ignored_errors(recorder):
+    assert recorder.num_ignored_errors / recorder.num_results <= 0.01
 
 
 def show_input_summary(recorder):  # pragma: no cover


tests/generative/test_query_model.py
```diff
@@ -79,16 +83,30 @@ def run_test(query_engines, data, variable):
         for name, engine in query_engines.items()
     ]
 
-    first_name, first_results = results[0]
-    for other_name, other_results in results[1:]:
+    for first, second in itertools.combinations(results, 2):
```
evansd, line 86: :smile: _Very_ nice

tests/generative/test_query_model.py
```diff
@@ -1,14 +1,18 @@
...
 import hypothesis.strategies as st
 import pytest
+import sqlalchemy.exc
 
 from databuilder.query_model.nodes import Column, TableSchema, count_nodes, node_types
 
 from ..conftest import QUERY_ENGINE_NAMES, engine_factory
 from . import data_setup, data_strategies, variable_strategies
 
+ignore_result = object()
```
evansd, line 14: ronto-nit: I've generally seen sentinels written in all caps as constants.

evansd (approved):
Given how neat your implementation has turned out, I think I like this solution even more than I did at first. But I know you had your reservations and I don't feel hugely strongly that this is _the_ right way of tackling the problem.

tests/generative/test_query_model.py
benbc: Oh, I'm very happy to do that. I started with all caps and changed it because I thought it was my Java heritage showing through.

rebkwok (approved):

PR #853, opened 2022-12-05
QM transform refactoring

benbc:


databuilder/query_model/transforms.py
```diff
@@ -168,6 +174,21 @@ def make_sortable(col):
...
+def collapse_statically_known_non_nulls(node, reverse_index):
+    """
+    Some operations never return null. When such operations are nested inside IsNull, the result
+    is guaranteed to be False; so we can collapse an entire subtree to a single value.
+
+    This allows us to avoid some expressions which would result in SQL that exceeds databases'
+    limits on nested CASE statements.
+    """
+    if isinstance(node.source, Function.IsNull):
+        del reverse_index[node.source]
```
benbc, line 186: Not sure if this is really necessary. It seems like tidying up is a good idea, but we're not doing anything to prune the now-garbage nodes from the list of nodes under consideration, so it's rather cosmetic.

benbc: (It's also not tested.)

databuilder/query_model/transforms.py
```diff
@@ -53,19 +53,28 @@ def apply_transforms(variables):
...
-    variables = copy.deepcopy(variables)  # see comment above
+    for type_, transform in transforms:
+        apply_transform(type_, transform, nodes, reverse_index)
 
+    variables = copy.deepcopy(variables)  # see comment above
     return variables
 
 
-def rewrite_sorts(nodes, reverse_index):
+def apply_transform(type_, transform, nodes, reverse_index):
```
evansd, line 71: This is nice

databuilder/query_model/transforms.py
evansd: It seems reasonable to me too, but I think I'm also prone to bits of purely cosmetic tidying so I'm maybe not the best sounding board!

evansd (approved):
I'd say this is worth merging for the tidying and refactoring alone!

It's unlikely that real queries will end up stacking up `IsNull` in this way, but I think it's reasonably likely that we'll end up seeing `IsNull(AggregateByPatient.Exists)` and I think it would be a tiny change to handle those too. (Not saying to do this now, just that this PR lays the groundwork.)

benbc:
Yes, I was deliberately not trying to cover all the cases here, but leaving clear space for adding them later.

benbc:
I've removed the commit for which this PR was originally created (collapsing stacked `IsNull` operators) because we decided that we don't want it and just left some refactoring that's worth doing anyway.

evansd (approved):

PR #854, opened 2022-12-05
Avoid redundant constructs in gentests

benbc:
These query model constructs are perfectly valid, but they either have constant results (like applying IsNull to an operation that is never null) or can be trivially reduced (like nested Not operations).

Hypothesis seems over-fond of generating big stacks of these redundant constructs and we would prefer to explore more interesting parts of the query space. (And in some rare cases these stacks cause problems with translation to SQL -- they can hit MSSQL's case-nesting limit.)

### DIFF
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -1,3 +1,4 @@
+import hypothesis as hyp
 import hypothesis.errors
 import hypothesis.strategies as st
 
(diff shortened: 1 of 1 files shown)


tests/generative/variable_strategies.py
```diff
@@ -195,12 +204,17 @@ def variable(patient_tables, event_tables, schema, int_values, bool_values):
...
+def qm_builds(type_, *arg_strategies, avoided_inputs=None):
     included_operations.add(type_)
 
     @st.composite
     def strategy(draw, type_, *arg_strategies):
         args = [draw(s) for s in arg_strategies]
+
+        if avoided_inputs:
+            for input_type in avoided_inputs:
+                hyp.assume(not any(isinstance(arg, input_type) for arg in args))
```
benbc, line 216: I'm a bit uncomfortable about this because, I _think_, it gives up on the attempt to build an `IsNull` if this assumption doesn't hold; so it seems like it might result in under-sampling of node types that have avoided inputs.

I've run this locally and did get a reasonable number of `IsNull` operations. I've also got an alternative implementation locally which keeps calling `draw()` until it gets something it likes instead of bailing out of `qm_builds()`. (Currently testing to see if it results in an obviously larger number of `IsNull`s.)

benbc: Over 10k examples I see 33 `IsNull` operations with the version here and 47 with the alternative implementation. Maybe that's a meaningful difference.

benbc:
We prefer to deal with this elsewhere (#852) because we don't want to make the query generation more complex than it already is.

PR #855, opened 2022-12-06
fix: Treat empty MSOA codes as NULL

evansd:
We already handle the magic string `NPC` but we need to also handle the empty string.


StevenMaude (approved):

StevenMaude:
Related to this, for anyone else reading, [NPC means "no postcode"](https://github.com/opensafely-core/databuilder/issues/96#issue-954629807).

Issue #738, opened 2022-09-22
Support is_in with dates


CarolineMorton:
Follow on ticket from #618 and subsequent PR #733. Also relates to #737 

Currently `is_in` is used in the code base to signify if an item in a list. i.e. is a snomed code in a codelist. We might want this for dates. For example, there could be a study where we want to pass all the bank holiday dates for a given year and check what events (perhaps emergency admissions) took place then. 

```
bank_holidays = ["2021-12-25", "2021-12-26", ... etc]

dataset.admitted_to_hospital_bank_holiday = (
    admitted_to_hospital.take(events.date.is_in(bank_holidays))
)
```

evansd:
Just to clarify here: `is_in` works with dates already as that method is defined for series of any type. But it doesn't do the magic convert-strings-to-dates thing which @CarolineMorton implemented in #733 and which, for consistency, it would be nice to support everywhere.

benbc:
I've edited the description to incorporate Dave's comment.

PR #870, opened 2022-12-12
Support is_in with collections of dates-as-strings

Jongmassey:
implements #738

### DIFF
--- a/tests/spec/date_series_ops/test_date_comparisons.py
+++ b/tests/spec/date_series_ops/test_date_comparisons.py
@@ -1,3 +1,5 @@
+from datetime import date
+
 from ..tables import p
 
 title = "Comparisons involving dates"
@@ -128,4 +130,46 @@ def test_is_equal_with_str_date(spec_test):
             3: False,
             4: None,
         },
+    )
+
+
+def test_is_in_dates(spec_test):
+    dates = [date(2010, 1, 1), date(2010, 1, 2)]
+    spec_test(
+        table_data,
+        p.d1.is_in(dates),
+        {
+            1: False,
+            2: False,
+            3: True,
+            4: None,
+        },
+    )
+
+    dates = [date(2010, 1, 2), date(2010, 1, 3)]
+    spec_test(
+        table_data,
+        p.d1.is_in(dates),
+        {
+            1: False,
+            2: False,
+            3: False,
+            4: None,
+        },
+    )
+
+
+def test_is_in_strings(spec_test):
+    d = ["2010-01-01", "2010-01-02"]
+    spec_test(
+        table_data,
+        p.d1.is_in(d),
+        {1: False, 2: False, 3: True, 4: None},
+    )
+
+    d = ["2010-01-02", "2010-01-03"]
+    spec_test(
+        table_data,
+        p.d1.is_in(d),
+        {1: False, 2: False, 3: False, 4: None},
     )


Jongmassey:
tests fail on dumping sql for spark [slack thread](https://bennettoxford.slack.com/archives/C33TWNQ1J/p1670853839628609)

databuilder/query_language.py
```diff
@@ -110,6 +110,9 @@ def is_in(self, other):
         # The query model requires an immutable Set type for containment queries, but
         # that's a bit user-unfriendly so we accept other types here and convert them
         if isinstance(other, (tuple, list, set)):
+            # Parse any ISO date strings in collection
```
evansd, line 113: This will need to be implemented as a new `is_in()` method on the `DateFunctions` class rather than on `BaseSeries` because it applies only to date-typed series. As it stands, this parsing gets applied to series of any type which means that if you try to write:
```py
some_string_series == "2000-01-01"
```
you'll get a type error because it will transform the string into a date.

databuilder/query_language.py
```diff
@@ -110,6 +110,9 @@ def is_in(self, other):
...
+            t = type(other)
```
evansd, line 114: You don't need to worry about preserving the collection type here: everything gets transformed into `frozenset` at the next step anyway.

databuilder/query_language.py
```diff
@@ -274,9 +277,11 @@ class FloatPatientSeries(NumericFunctions, PatientSeries):
 
 def parse_date_if_str(value):
     if isinstance(value, str):
-        return datetime.date.fromisoformat(value)
-    else:
-        return value
+        try:
+            value = datetime.date.fromisoformat(value)
+        except ValueError:
+            pass
```
evansd, line 283: This is another consequence of applying the parsing in `BaseSeries.is_in()`. We want to raise errors for malformed dates but we can only do that if know that the string is supposed to be a date, which means only calling this function in date contexts.

tests/spec/date_series_ops/test_date_comparisons.py
```diff
@@ -129,3 +131,26 @@ def test_is_equal_with_str_date(spec_test):
             4: None,
         },
     )
+
+
+def test_is_in_dates(spec_test):
+    dates = [date(2010, 1, 1), date(2010, 1, 2)]
```
evansd, line 137: I might be tempted to inline these. These tests get formatted as part of the documentation and so it's good to keep syntactic clutter to a minimum. Obviously for a longer list of dates pulling them out into a variable like this would help readability. But as it's only two dates I think it's clearer without the extra line.

evansd, line 137: I wonder if it's worth making the target dates here span one of the example dates to make it clear that the semantics are "exact match to a date in the list" rather than "within the range implied by the list members" e.g. if you used `1990-01-01` and `2010-01-01`.

tests/spec/date_series_ops/test_date_comparisons.py
```diff
@@ -129,3 +131,26 @@ def test_is_equal_with_str_date(spec_test):
...
+        },
+    )
+
+
+def test_is_in_strings(spec_test):
+    d = ["2010-01-01", "2010-01-02"]
+    spec_test(
+        table_data,
+        p.d1.is_in(d),
+        {1: False, 2: False, 3: True, 4: None},
```
evansd, line 155: If you add a trailing comma after `None` then Black will format these one per line as for the other test cases.

databuilder/query_language.py
Jongmassey: but it only passes to parse_date_if_str if it's a collection, and parse_date_if_str returns the input unmodified if it isn't an iso date string format; or am I missing your point here?

evansd: :facepalm: Sorry, I messed up the example. I meant to write:
```py
some_string_series.is_in(["2000-01-01"])
```

Jongmassey: gotcha

databuilder/query_language.py
```diff
@@ -303,6 +303,16 @@ def is_after(self, other):
...
 
+    def is_in(self, other):
+        # The query model requires an immutable Set type for containment queries, but
+        # that's a bit user-unfriendly so we accept other types here and convert them
+        if isinstance(other, (tuple, list, set)):
+            # Parse any ISO date strings in collection
+            t = type(other)
+            other = t(map(parse_date_if_str, other))
+            other = frozenset(other)
+        return _apply(qm.Function.In, self, other)
```
evansd, line 314: Coverage is moaning at you because you've not tested what happens if this condition is false (i.e. you pass something which isn't a tuple, list or set). But I think that exposes an issue here which is that you want to apply the date parsing to _any_ iterable that's passed in, not just those which need converting to `frozenset`. So given that I think you could throw away the conditional all together, which then gives you fewer branches to test. Something like:
```py
# Parse any ISO date strings in collection
other = frozenset(map(parse_date_if_str, other))
return _apply(qm.Function.In, self, other)
```

Jongmassey: I guess on this path it would never get given a Codelist which was what was tripping me up last time, I'll remove the conditional and see what happens.

Jongmassey: `just test-all` seems happier locally with that

evansd (approved):
:+1: Looks good.

This will need a squash merge and, to trigger a new release, you'll need a `fix:` prefix on the title of the squashed commit.

Let me know if you'd like a second pair of eyes on the process. Otherwise I'l leave you to it.

PR #880, opened 2022-12-16
Support multiplication for numeric series

### LINKED ISSUE #864: Support multiplication for numeric series

You should start with a spec test (see [`int_series_ops.test_arithmetic_ops`](https://github.com/opensafely-core/databuilder/blob/main/tests/spec/int_series_ops/test_arithmetic_ops.py)) and [`float_series_ops.test_arithmetic_ops`](https://github.com/opensafely-core/databuilder/blob/main/tests/spec/float_series_ops/test_arithmetic_ops.py)), which will drive out all the necessary implementation.

In ehrQL terms we'd need to add the `__mul__` method to the `NumericFunctions` mixin:

We should also support the reverse methods `__rmul__`.

The query model will need a new `Multiply` operation:
```py
class Multiply(Series[Numeric]):
    lhs: Series[Numeric]
    rhs: Series[Numeric]
```

And support for this will need adding to:
```
tests/lib/in_memory/engine.py
```

After adding that you should get errors from population validation and generative test code which guide you to add support there as well.

Note that the type signature above means that we support multiplying two ints together to get another int (`2 * 3 == 6`), or two floats to get a float (`2.0 * 3.5 = 7.0`), but we don't allow multiplying an int and a float to get a float (`2 * 3.5 = 7.0`).

Instead, the user would have to explicitly cast one of the variables using either the `.to_int()` or `.to_float()` methods before multiplying. This is a bit user-hostile, but we plan to address [implicit casting](https://github.com/opensafely-core/databuilder/issues/540) separately.

milanwiedemann:
Closes #864


evansd (commented):
Looks great! The error is to do with test coverage because the `__rmul__` method never gets called. To cover this you'll need to add a test which does something like:
```py
10 * p.i1
```

iaindillingham:
@milanwiedemann, I've added d3fdfbc to test `NumericFunctions.__rmul__`. We didn't spot this when working locally, because we didn't run `just test-all`. (As well as running all the tests, this command also checks code coverage.)

Before moving on, I'd suggest looking at

https://github.com/opensafely-core/databuilder/blob/7b42371d788f8ca099992786bf640a7d7995ac14/tests/spec/int_series_ops/test_arithmetic_ops.py#L47-L52

and

https://github.com/opensafely-core/databuilder/blob/7b42371d788f8ca099992786bf640a7d7995ac14/tests/spec/float_series_ops/test_arithmetic_ops.py#L55-L60

Notice that the constant is an integer in the first case (`10`) and a float in the second case (`10.0`). What do you think would happen if we made the constant an integer in both cases? Or a float in both cases?

Consider Ben's note on the issue:

> Note that the type signature above means that we support multiplying two ints together to get another int `(2 * 3 == 6)`, or two floats to get a float `(2.0 * 3.5 = 7.0)`, but we don't allow multiplying an int and a float to get a float `(2 * 3.5 = 7.0)`.

I've also added 7b42371, for consistency 🙂

Feel free to squash my commits into your commits, should you accept them, when you're back!

evansd (approved):

Issue #706, opened 2022-09-08
Spec tests should assert that types returned by the database match the query model


evansd:
It's possible for there to be a mismatch between the type that the query model thinks an operation returns and the type actually returned by the corresponding SQL operation in the database. Usually this would be caught by the spec tests, but not always because Python is lose with comparisons between floats, ints and bools.

To ensure that the query model typing is consistent with how the databases really behave, the spec tests should ask the query model what it thinks the type is and then check that the type of the result matches.

This checking can be implemented for all tests by modifying the test harness here:
https://github.com/opensafely-core/databuilder/blob/7e27a67379742759451fe196c1998101241cd73d/tests/spec/conftest.py#L38-L42

Note that this requires a bit of rework of the test harness because currently query model compilation and query execution happens in one step here:
https://github.com/opensafely-core/databuilder/blob/7e27a67379742759451fe196c1998101241cd73d/tests/conftest.py#L62-L64

Once compiled, each variable is a `Series[T]`. The type of the series can be retrieved using [`databuilder.query_model.nodes.get_inner_type()`](https://github.com/opensafely-core/databuilder/blob/7e27a67379742759451fe196c1998101241cd73d/databuilder/query_model/nodes.py#L716). Those types can then be checked against the actual value returned by the database to ensure that they are correct.

To test that the check is working as designed, reintroduce the problem mentioned in [this comment](https://github.com/opensafely-core/databuilder/pull/704#discussion_r966066084) and ensure that the relevant specs fail with a useful error message.

benbc:
This issue originated from a comment on an earlier PR: https://github.com/opensafely-core/databuilder/pull/704#discussion_r966066084.

Jongmassey:
>, reintroduce the problem mentioned in https://github.com/opensafely-core/databuilder/pull/704#discussion_r966066084

Not sure how one might do that!

Jongmassey:
@benbc with your suggested change the test I added in #881 fails, but when I revert your change the test passes. Any hints for writing a test which introduces your suggested scenario within a test and tests that the test I added failed? 

I have tried adding the following to `test_conftest.py`:

```python
@pytest.mark.xfail(strict=True)
def test_run_test_execute_type_mismatch(spec_test):
    class BadCastToInt(Series[Numeric]):
        source: Series[Numeric]
    
    class BadNumericFunctions(ComparableFunctions):
        def as_int(self):
            return _apply(BadCastToInt, self)
            
[...]

PR #881, opened 2022-12-16
Test that types returned by database match the query model

Jongmassey:
Addresses #706

### DIFF
--- a/tests/spec/conftest.py
+++ b/tests/spec/conftest.py
@@ -3,6 +3,8 @@ import datetime
 import pytest
 
 from databuilder.ehrql import Dataset
+from databuilder.query_language import compile
+from databuilder.query_model.nodes import get_series_type
 
 from . import tables
 
@@ -40,6 +42,11 @@ def spec_test(request, engine):
         results_dict = dict(results)
         assert len(results) == len(results_dict), "Duplicate patient IDs found"
         assert results_dict == expected_results
+
+        # assert types are as expected
+        variables = compile(dataset)
+        variable_type = get_series_type(variables["v"])
+        assert results[0][1] is None or isinstance(results[0][1], variable_type)
 
     # Test that we can generate SQL with literal parmeters for debugging purposes
     def run_test_dump_sql(table_data, series, expected_results, population=None):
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -109,3 +109,6 @@ story-dependencies.png
 # Build/test byproducts
 cache/
 .hypothesis
+
+# VS Code configuration
```
Jongmassey, line 113: Sorry if this is a bit of a drive-by. Surely I can't be the only one using vs code?

evansd: You can set this as part of the global gitignore for your system. In general, detritus created by the project should be excluded by the project gitignore, and detritus created by the user's tooling/operating system etc should be excluded by the user's gitignore.

This should be just be as simple as creating a file in `~/.config/git/ignore`, but there are some more details here if that's helpful:
https://git-scm.com/docs/gitignore

Jongmassey: Ideal, shall do thanks!

tests/spec/conftest.py
```diff
@@ -41,6 +43,11 @@ def run_test_execute(table_data, series, expected_results, population=None):
         assert len(results) == len(results_dict), "Duplicate patient IDs found"
         assert results_dict == expected_results
 
+        # assert types are as expected
+        variables = compile(dataset)
+        variable_type = get_series_type(variables["v"])
+        assert results[0][1] is None or isinstance(results[0][1], variable_type)
```
evansd, line 49: This only checks the first result, which is not unreasonable as they should all have the same type unless something _really_ horrible has happened. But, if the first value happens to be `None` then it's not going to do any type-checking at all.

You could make it check the first non-null value, but it's probably simpler to check the type of every value than it is to do that.

evansd (approved):
Happy to approve this as I think in practice it enforces what we need, but have added a note about what happens if the first value is `None`.

tests/spec/conftest.py
Jongmassey: Addressed in latest commit (I hope!)

evansd (approved):
Great, thanks!

PR #886, opened 2022-12-19
`is_between` and `is_on_or_between` operations for dates

### LINKED ISSUE #737: Add `is_between` and `is_on_or_between` operations for dates to ehrQL
We would like researchers to be able to write an expression like this:

```
dataset.covid_date = (
    covid_events.take(covid_events.date.is_between("2021-06-01", "2022-01-01"))
    .sort_by(covid_events.date)
    .last_for_patient()
    .date
)
```

In particular we need to add an `is_between()` method to date-typed `Series`. Like `is_after()` and others this is a date-specific synonym for a standard comparison operation, to be added to [`DateFunctions`](https://github.com/opensafely-core/databuilder/blob/6c73f9d55bdb8afdce9899411decb56e28347159/databuilder/query_language.py#L281).

```
events.date.is_between("2021-06-01", "2022-01-01")
```

is a synonym for

```
(events.date > "2021-06-01") & (events.date < "2022-01-01")
```

We also need a method `is_on_or_between()` with a similar implementation but using `>=` and `<=`.

Both new methods should accept arguments which are instances of `datetime.date`, an ISO-formatted date string, or a date-typed `Series`. (Note that there is existing machinery used in DateFunctions to provide this polymorphism.)

You should start this work by adding a failing test to the specs.

Jongmassey:
fixes  #737

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -316,6 +316,12 @@ class DateFunctions(ComparableFunctions):
     def to_first_of_month(self):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
+    def is_between(self, begin, end):
+        raise NotImplementedError
+
+    def is_on_or_between(self, begin, end):
+        raise NotImplementedError
+
     def __lt__(self, other):
         other = parse_date_if_str(other)
         return _apply(qm.Function.LT, self, other)
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -316,11 +316,11 @@ class DateFunctions(ComparableFunctions):
     def to_first_of_month(self):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
-    def is_between(self, begin, end):
-        return self.__gt__(begin) & self.__lt__(end)
+    def is_between(self, start, end):
+        return self.__gt__(start) & self.__lt__(end)
 
-    def is_on_or_between(self, begin, end):
-        return self.__ge__(begin) & self.__le__(end)
+    def is_on_or_between(self, start, end):
+        return self.__ge__(start) & self.__le__(end)
 
     def __lt__(self, other):
         other = parse_date_if_str(other)
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -317,10 +317,10 @@ class DateFunctions(ComparableFunctions):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
     def is_between(self, start, end):
-        return self.__gt__(start) & self.__lt__(end)
+        return (self > start) & (self < end)
 
     def is_on_or_between(self, start, end):
-        return self.__ge__(start) & self.__le__(end)
+        return (self >= start) & (self <= end)
 
     def __lt__(self, other):
         other = parse_date_if_str(other)
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -320,7 +320,7 @@ class DateFunctions(ComparableFunctions):
         return self.__gt__(begin) & self.__lt__(end)
 
     def is_on_or_between(self, begin, end):
-        return self.__ge__(begin) & self.__lt__(end)
+        return self.__ge__(begin) & self.__le__(end)
 
     def __lt__(self, other):
         other = parse_date_if_str(other)
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -317,10 +317,10 @@ class DateFunctions(ComparableFunctions):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
     def is_between(self, begin, end):
-        raise NotImplementedError
+        return self.__gt__(begin) & self.__lt__(end)
 
     def is_on_or_between(self, begin, end):
-        raise NotImplementedError
+        return self.__ge__(begin) & self.__lt__(end)
 
     def __lt__(self, other):
         other = parse_date_if_str(other)
--- a/tests/spec/date_series_ops/test_date_comparisons.py
+++ b/tests/spec/date_series_ops/test_date_comparisons.py
@@ -11,9 +11,7 @@ table_data = {
         1 | 1990-01-01
         2 | 2000-01-01
         3 | 2010-01-01
-        4 | 2010-01-02
-        5 | 2010-01-03
-        6 |
+        4 |
         """,
 }
 
@@ -26,9 +24,7 @@ def test_is_before(spec_test):
             1: True,
             2: False,
             3: False,
-            4: False,
-            5: False,
-            6: None,
+            4: None,
         },
     )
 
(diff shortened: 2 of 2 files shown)


inglesp (approved):
Right idea, but current implementation of `is_or_or_between` is half-open.  If you'd like to push back, let's continue the discussion on #737.

Also, please can you squash all the commits into one before merging?

databuilder/query_language.py
```diff
@@ -316,6 +316,12 @@ def to_first_of_year(self):
     def to_first_of_month(self):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
+    def is_between(self, begin, end):
+        return self.__gt__(begin) & self.__lt__(end)
```
evansd, line 320: Mixing methods and operators like this feels a bit inconsistent to me. I think it either wants to be:
```py
return (self > begin) & (self < end)
```
or
```py
return self.__gt__(begin).__and__(self.__lt__(end))
```

I think my preference would be for the former as it's a lot more readable, although there's an argument that using operators within the definition of operator methods might be mixing layers a bit.

tests/spec/date_series_ops/test_date_comparisons.py
```diff
@@ -11,7 +11,9 @@
         1 | 1990-01-01
         2 | 2000-01-01
         3 | 2010-01-01
-        4 |
+        4 | 2010-01-02
+        5 | 2010-01-03
+        6 |
```
evansd, line 16: I think it would be better to leave this table as it is and define a new one for the new tests. Otherwise you end up cluttering the other tests with examples that don't really add explanatory value for them.

evansd (commented):
As Peter says, we need some test examples which check the endpoint behaviour. I think it would also be good to have an example where `start` is greater than `end` (which will match nothing) as it's possible that someone might think "between" implies order invariance.

There's also a wider issue here were I think we (as a team) have gone down slightly the wrong path with date testing. This is nothing to do with you as you've just been correctly following established patterns, but I think we need to stop testing every date method with both strings and date objects. I'll think about how best to address this in a separate PR, but for now feel free to just test with one type if that makes your life easier.

databuilder/query_language.py
```diff
@@ -316,6 +316,12 @@ def to_first_of_year(self):
...
+
+    def is_on_or_between(self, begin, end):
+        return self.__ge__(begin) & self.__lt__(end)
```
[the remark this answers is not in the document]
Jongmassey: D'oh, yes sorry that was a brain fart rather than a protest!

databuilder/query_language.py
```diff
@@ -316,6 +316,12 @@ def to_first_of_year(self):
     def to_first_of_month(self):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
+    def is_between(self, begin, end):
```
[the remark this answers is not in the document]
Jongmassey: Agreed

tests/spec/date_series_ops/test_date_comparisons.py
Jongmassey: I tried this and the documentation generation broke - could you advise as the right way to do this? I've seen other examples where there's a table within the test method, but don't wan't to replicate across all four test methods in this case

databuilder/query_language.py
Jongmassey: Good point, done

databuilder/query_language.py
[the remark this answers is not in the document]
Jongmassey: >The spec tests should also be tightened up to catch this 

For some reason "or or between" had (erroneously) been fixed in my head as to being half-open. I've rewritten the tests to reflect the actual spec. Do you think this needs further work?

Jongmassey:
>I think it would also be good to have an example where start is greater than end (which will match nothing) 

Do we want to throw an error in this case?

tests/spec/date_series_ops/test_date_comparisons.py
evansd: Ah, that's annoying. Let me have a fiddle around with it and see what works.

evansd:
> Do we want to throw an error in this case?

In general I'd say that's the right thing to do, but it's slightly tricky in this case because `start` and `end` can be ehrQL series rather than static dates so we can't always tell whether one is greater than the other (indeed, the answer might be different for different patients). So we have to be able to handle them being the wrong way round and specify the behaviour when they are.

Jongmassey:
I think [f58939c](https://github.com/opensafely-core/databuilder/pull/886/commits/f58939c4824072461bf86925f8a968174a30831a) tidies up the tests (notwithstanding our earlier discussion about repetition of date/stringy-date tests) as per your suggestion @evansd  but still I get `AttributeError: 'ArgCapturer' object has no attribute 'table_data'` on `just test-all`

databuilder/query_language.py
```diff
@@ -316,6 +316,14 @@ def to_first_of_year(self):
     def to_first_of_month(self):
         return _apply(qm.Function.ToFirstOfMonth, self)
 
+    def is_between(self, start, end):
+        assert end > start, "end date must be greater than start date"
```
evansd, line 320: Sorry, I wasn't particularly clear in my last comment about error checking. I meant that we _shouldn't_ add the error checking you've added because we can't do it in the general case. The `assert` here won't work if either of `start` or `end` are ehrQL series rather than a static date e.g if someone writes:
```py
events.date.is_between(covid_vaxx_date, covid_vaxx_date + days(14))
```

As an aside, `assert` wouldn't be quite the right thing to use for user-facing errors: `raise ValueError(...)` would probably be the most appropriate standard exception.

In general, assertions are there to catch mistakes made by developers when working on the code and not mistakes made by users of the code. (Python even has an "optimised" mode where it ignores assertions completely.) Where we want to report an error to the user we should raise on exception.

tests/spec/date_series_ops/test_date_comparisons.py
```diff
@@ -183,3 +185,149 @@ def test_is_not_in_strings(spec_test):
...
+            4: True,
+            5: False,
+            6: None,
+        },
+    )
+
+
+def test_is_between_dates_backwards(spec_test):
+    table_data = table_data_for_between_tests
+    with pytest.raises(AssertionError):
```
evansd, line 270: The error you're getting from the docs generator is due to this `pytest.raises` construction, rather than anything to do with the `table_data_for_between_tests` trick.

The spec tests don't currently have a mechanism for giving examples of things you're _not_ allowed to do (though it has been [discussed](https://github.com/opensafely-core/databuilder/issues/521) before).

In this particular case I think the answer is to keep the examples, but change the expected results from `True` to `False`. That at least demonstrates that the arguments are sensitive to order.

databuilder/query_language.py
Jongmassey: Righto, understood. I'll remove that commit

tests/spec/date_series_ops/test_date_comparisons.py
Jongmassey: Sure, that seems good to me

Issue #842, opened 2022-11-24
`Duration`s of the same units should be add-able


evansd:
The below is reasonable code to write and it would be nice if it worked:
```py
covid_to_longcovid_lag = days(84) # 12 weeks
vaccine_effective_lag = days(14) # 2 weeks for vaccine to become effective
vaccine_to_longcovid_lag = covid_to_longcovid_lag + vaccine_effective_lag
```

We can't support adding `Duration`s in general because date addition is not associative (lolcry), but we can certainly support adding ones of the same units.

To avoid endless unit checks, first introduce unit-specific subclasses of `Duration`. This will allow simplification of the existing `Duration.__add__()` method. Then you can add to each of the subclasses' `__add__()` the ability to accept `other` as a instance of the same class.

You should include an example of this construct in the specs and provide comprehensive test coverage in `tests/unit/test_query_language.py`.

benbc:
Note that we already have a [switch](https://github.com/opensafely-core/databuilder/blob/main/databuilder/query_language.py#L316-L324) on this "missing type" for `Duration`/`DateSeries` addition.

Jongmassey:
Have I misunderstood the requirements or is [this](https://github.com/opensafely-core/databuilder/pull/887/commits/c1da9b6e8302dab8bc49fafc8b294982f4ca2a12) not a simple implementation that works without subclasses?

PR #887, opened 2022-12-19
Addition of durations of same unit

Jongmassey:
Implements #842

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -428,6 +428,9 @@ class Duration:
                 return date_utils.date_add_years(other, self.value)
             else:
                 assert False
+        if isinstance(other, Duration):
+            if self.units == other.units:
+                return Duration(units=self.units, value=(self.value + other.value))
         else:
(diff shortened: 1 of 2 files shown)


evansd (commented):
Nice! That turned out quite neatly.

I think eventually we might still end up making separate `Duration` subclasses for each unit because it might help us with adding typing and hence better IDE autocomplete behaviour. But that doesn't change the user-facing API so we can worry about that later, if we do it at all.

One slight wrinkle, and the reason I haven't approved is that since you started this work I've merged a fix to the duration code:
 * #891

Git thinks it can merge your patch in cleanly, but the code it produces is wrong. Are you happy to try rebasing your PR onto `main` and fixing up? Give me a shout if you'd like a second pair of eyes on that.

Jongmassey:
I think I've managed to successfully rebase - could you double-check to make sure it's all good?

evansd (approved):
Lovely stuff!

PR #889, opened 2022-12-20
fix: Cast string literal dates to Date before DATEADD operations

rebkwok:


databuilder/query_engines/mssql.py
```diff
@@ -43,6 +44,10 @@ def get_date_part(self, date, part):
         return SQLFunction(part, date, type_=sqlalchemy_types.Integer)
 
     def date_add_days(self, date, num_days):
+        # DATEADD will turn string literal dates into datetimes, which is not what we
+        # want, so if it's a string literal, explictly cast it to a Date first.
+        if isinstance(date, BindParameter):
+            date = sqlalchemy.cast(date, sqlalchemy_types.Date)
```
evansd, line 50: I wonder if it's worth wrapping this up in a `cast_if_literal()` method as it avoids repeating the explanation.

evansd: Oh, and worth linking to this bit of the SQL Server docs in the comment as well I think:
https://learn.microsoft.com/en-us/sql/t-sql/functions/dateadd-transact-sql?view=sql-server-ver16#return-types

evansd (approved):
Aware you didn't explicitly ask for a review on this, but it all looks good to me!

rebkwok:
> Aware you didn't explicitly ask for a review on this, but it all looks good to me!

I was just waiting for the tests to pass and then got distracted! Thanks :)

PR #892, opened 2022-12-21
Add date arithmentic operations to generative tests

rebkwok:
Fixes #836

PR #893, opened 2022-12-21
fix: Limit date operations to date series in generative tests

rebkwok:
This means that the generative tests need 4-5K examples in order to hit every node rather than 10K (as they did after the date operations were added in #892)

### DIFF
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -53,7 +53,7 @@ from tests.lib.query_model_utils import get_all_operations
 #   instances. There is an unexplored optimization avenue here: make the strategies more specific to include type
 #   and/or domain, so that we don't discard so many examples.
 #
-# **Note that most concrete strategies that take a series as an argument, and that series is an abstract `series`
+# **Note that most concrete strategies take a series as an argument, and that series is an abstract `series`
 #   which can be of any type. However, strategies that require a date-type series as one or more arguments, such
 #   as date arithmetic operations, take a `date_series` strategy, which ensures that they only receive a
 #   series of the correct type. This limits the number of rejected examples, and therefore reduces the number of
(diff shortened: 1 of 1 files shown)


tests/generative/variable_strategies.py
```diff
@@ -51,7 +52,15 @@
 #   constructing the query model objects using a special-purpose strategy `qm_builds`, which can discard invalid
 #   instances. There is an unexplored optimization avenue here: make the strategies more specific to include type
 #   and/or domain, so that we don't discard so many examples.
-from tests.lib.query_model_utils import get_all_operations
+#
+# **Note that most concrete strategies that take a series as an argument, and that series is an abstract `series`
```
evansd, line 56:
```suggestion
# **Note that most concrete strategies take a series as an argument, and that series is an abstract `series`
```

evansd (approved):
Despite having a slightly shaky grasp of Hypothesis I can understand the problem we were having here and I can understand how the fix works, so that's good!

Just thinking out loud here (and not making suggestions for this PR). If we end up having to pull this trick for too many more types I can see it getting a bit out of hand. It feels, naively, like there ought to be some way of generalising this idea. As in, the query model already specifies what types each operation takes. So ideally, we'd have some way of getting Hypothesis to respect that and only generate types it knows the operation will accept. I've _no_ idea where we'd start with that, but I thought I'd at least get the suggestion ruminating in your (and @benbc's) head.

rebkwok:
> If we end up having to pull this trick for too many more types I can see it getting a bit out of hand

Yes, I agree. I think I'd rather wait for Ben's thoughts on how to make it more generic.  I did have a go at using (hypothesis') `filter` to ensure that the series the operations get is of the right type, but that makes it generate the examples and filter ones that don't match, rather than generating examples that already have the right types.

PR #896, opened 2022-12-21
Add cast operations to the generative tests

### LINKED ISSUE #838: Add cast operations to generative tests
See `tests.generative.test_query_model.recorder` for the missing operations.

At the time of writing they are:

```
        Function.CastToFloat,
        Function.CastToInt,
```

(Note that there may be other non-cast operations mentioned there which are not covered by this issue.)

The operations need adding to `tests.generative.variable_strategies`. If you're not familiar with the generative tests then you should get an intro from (probably) Ben before starting this work.

There is a verification step to ensure that we're including all the operations:

```
GENTEST_COMPREHENSIVE=t GENTEST_EXAMPLES=5000 just test-generative
```

Once you've added the operations you'll find that this fails. You will then need to remove the newly added operations from `tests.generative.test_query_model.recorder` and it should pass. If you find you've removed the last of the `known_missing` operations then please remove this exception mechanism altogether.

rebkwok:
Closes #838 

This addes CastToInt and CastToFloat to the generative tests; it also required adding float values to the strategies.

When I added the cast operations, tests started failing every so often when they tried to add or subtract a large number of years from a date.  (e.g. 1900-01-01 - 193 years, which is less than the sqlserver min - 1793-01-01).

We limit int values to 0-10, and the new float values that I've added are limited to 0.0-11.0.  Dates are limited to 1900-01-01 - 2100-12-31.  Previously this was enough to limit the inputs to a DateAddX operation to only values that were valid.

However, there is a Multiply node, which I think returns floats.  Those were previously not being passed to DateAddYears in the tests because it requires an int series.  Now that the tests can cast floats to ints, the result of a Multiply operation can be an input into a  DateAddYears. E.g. if the tests happened to multiply 9 * 10, negate it, cast it to an int and try to add it to a date, it'll try to subtract 900 from the date and fall over.

DateAddX operations take a date series and an int series.  I've solved this for now by filtering out any inputs to a date_add_x strategy that were produced by a CastToInt, because these are ones that may have been the result of a Multiply operation earlier in the tree.

### DIFF
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -64,7 +64,13 @@ from tests.lib.query_model_utils import get_all_operations
 
 
 def variable(
-    patient_tables, event_tables, schema, int_values, bool_values, date_values
+    patient_tables,
+    event_tables,
+    schema,
+    int_values,
+    bool_values,
+    date_values,
+    float_values,
 ):
     frame = st.deferred(
         lambda: st.one_of(
@@ -184,7 +190,9 @@ def variable(
 
     sorted_frame = st.deferred(lambda: st.one_of(sort))
 
-    value = qm_builds(Value, st.one_of(int_values, bool_values, date_values))
+    value = qm_builds(
+        Value, st.one_of(int_values, bool_values, date_values, float_values)
+    )
     date_value = qm_builds(Value, st.one_of(date_values))
 
     select_table = qm_builds(
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -165,6 +165,8 @@ def variable(
             add,
             subtract,
             multiply,
+            cast_to_int,
+            cast_to_float,
             date_add_years,
             date_add_months,
             date_add_days,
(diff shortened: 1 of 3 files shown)


tests/generative/variable_strategies.py
```diff
@@ -252,10 +254,19 @@ def variable(
...
     cast_to_int = qm_builds(Function.CastToInt, series)
     cast_to_float = qm_builds(Function.CastToFloat, series)
-    
-    date_add_years = qm_builds(Function.DateAddYears, date_series, series)
-    date_add_months = qm_builds(Function.DateAddMonths, date_series, series)
-    date_add_days = qm_builds(Function.DateAddDays, date_series, series)
+
+    # We limit the date operations to series that have not been produced by a CastToInt operation on a
+    # float value.  This avoids errors produced by trying to add/subtract large numbers that were
+    # produced by multiplication (which produces floats) and then casting to int
```
inglesp, line 260: This feels to me like the wrong layer to be addressing the issue.  I would imagine we'll find more ways to pass biggish numbers to these date functions, and so we'll have to add further special casing.  And there will probably be other kinds of error that Hypothesis triggers, which we actually consider benign.  So instead of complicating the strategies to avoid triggering those benign errors, I think it might be better to add support for catching and ignoring errors them.

However, this might turn out to be quite a lot of work... so maybe it's OK to have a few special cases!

inglesp (approved):

tests/generative/variable_strategies.py
benbc: There is one way of handling allowed errors in #852. Is that the kind of thing you had in mind, @inglesp?

inglesp: Yes, exactly.

benbc: @rebkwok if you like that approach, I'll try to get that PR merged today.

rebkwok: That looks good, thanks

rebkwok:
I've now removed the commit that tried to filter out bad inputs, as this will be addressed in #919 instead

PR #898, opened 2022-12-23
Refactor test setup machinery

evansd:
This factors out and generalises some of the test setup code from the spec tests so that it can used elsewhere. In particular, it lays the groundwork for supporting end-user tests.

Along the way we trim down the proliferation of ORM utils functions.

### DIFF
--- a/databuilder/backends/base.py
+++ b/databuilder/backends/base.py
@@ -2,7 +2,7 @@ import re
 
 import sqlalchemy
 
-from databuilder.query_language import BaseFrame
+from databuilder.query_language import get_tables_from_namespace
 from databuilder.sqlalchemy_types import type_from_python_type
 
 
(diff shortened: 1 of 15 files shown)


databuilder/backends/base.py
```diff
@@ -41,7 +41,8 @@ def _init_table(cls, table):
 
     @classmethod
     def validate_against_table_namespace(cls, table_namespace):
-        for attr, table in get_tables_from_namespace(table_namespace):
+        for attr, ehrql_table in get_tables_from_namespace(table_namespace):
+            table = ehrql_table.qm_node
```
inglesp, line 45: Elsewhere we've used `ql_xxx` names.  Could we do so here for consistency?

inglesp (approved):
All good refactorings, thanks.

PR #903, opened 2022-12-28
remove date strings from spec tests

### LINKED ISSUE #894: Remove date strings from spec tests
The spec tests have evolved a pattern where every operation involving dates is tested twice, once with ehrQL containing date objects, and once with ehrQL containing date strings.  See [test_date_comparisons.py](https://github.com/opensafely-core/databuilder/blob/5207e81eb102d75f1090b2466b14c2c03c74fe7f/tests/spec/date_series_ops/test_date_comparisons.py) (and possibly elsewhere).

This slows down the tests and adds tedium to adding a new feature, as well as adding overhead for readers of the generated documentation.

Instead, we should only use date objects in tests, and introduce a separate set unit tests that assert that the QM objects produced by the ehrQL with a date string is identical to that produced with a date object.

For instance:

```
assert p.d.is_before("2000-01-01").qm_node == p.d.is_before(date(2000, 1, 1)).qm_node
```

We will need to ensure that it is clear from other documentation that operations on dates can either take date objects or strings.

Jongmassey:
Fixes #894

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -328,6 +328,12 @@ class DateFunctions(ComparableFunctions):
     def is_on_or_between(self, start, end):
         return (self >= start) & (self <= end)
 
+    def if_null_then(self, other):
+        return case(
+            when(self.is_not_null()).then(self),
+            default=parse_date_if_str(other),
+        )
+
     def __lt__(self, other):
         other = parse_date_if_str(other)
         return _apply(qm.Function.LT, self, other)
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -1,4 +1,5 @@
 from datetime import date
+from inspect import signature
 
 import pytest
 
@@ -384,17 +385,17 @@ def test_incompatible_duration_operations(lhs, op, rhs):
 
 
 @pytest.mark.parametrize(
-    "fn_name, n_other",
+    "fn_name",
     [
-        ("is_after", 1),
-        ("is_before", 1),
-        ("is_on_or_after", 1),
-        ("is_on_or_before", 1),
-        ("is_between", 2),
-        ("is_on_or_between", 2),
+        "is_after",
+        "is_before",
+        "is_on_or_after",
+        "is_on_or_before",
+        "is_between",
+        "is_on_or_between",
     ],
 )
-def test_ehrql_date_string_equivalence(fn_name, n_other):
+def test_ehrql_date_string_equivalence(fn_name):
     date_str = "2000-01-01"
     date_date = date(2000, 1, 1)
 
@@ -403,6 +404,7 @@ def test_ehrql_date_string_equivalence(fn_name, n_other):
         d = Series(date)
 
     f = getattr(p.d, fn_name)
-    date_args = [date_date for i in range(n_other)]
-    str_args = [date_str for i in range(n_other)]
+    n_params = len(signature(f).parameters)
+    date_args = [date_date for i in range(n_params)]
+    str_args = [date_str for i in range(n_params)]
     assert f(*date_args).qm_node == f(*str_args).qm_node
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -405,4 +405,9 @@ def test_ehrql_date_string_equivalence(fn_name):
     n_params = len(signature(f).parameters)
     date_args = [date_date for i in range(n_params)]
     str_args = [date_str for i in range(n_params)]
+
+    if "in" in fn_name:
+        date_args = [date_args]
+        str_args = [str_args]
+
     assert f(*date_args).qm_node == f(*str_args).qm_node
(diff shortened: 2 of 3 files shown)


Jongmassey:
This has inadvertently revealed a potential issue with is_[not]_in(): it's possible to pass a non-iterable to it which fails with a potentially unhelpful error: should we handle this more gracefully?

Jongmassey:
the new tests in this PR found that 
`p.d.if_null_then("2001-01-01")` where `d` is a date series gives:

```
./tests/unit/test_query_language.py::test_ehrql_date_string_equivalence[if_null_then] Failed: [undefined]databuilder.query_model.nodes.TypeValidationError: Case.default requires 'typing.Optional[databuilder.query_model.nodes.Series[datetime.date]]' but got 'databuilder.query_model.nodes.Series[str]'
```


I can see `if_null_then` being commonly used with an ISO date string replacement value so will fix along the way

tests/unit/test_query_language.py
```diff
@@ -381,3 +384,36 @@ def test_incompatible_duration_operations(lhs, op, rhs):
...
             assert False
         assert not result
+
+
+@pytest.mark.parametrize(
+    "fn_name",
+    {
+        k
+        for k, v in DateFunctions.__dict__.items()
+        if callable(v) and "add_" not in k and "sub_" not in k
```
inglesp, line 394: I'd like a comment about why we're ignoring the `add_` and `sub_` methods.

tests/unit/test_query_language.py
```diff
@@ -381,3 +384,36 @@ def test_incompatible_duration_operations(lhs, op, rhs):
...
+    }
+    | {
+        k
+        for k, v in BaseSeries.__dict__.items()
+        if callable(v) and "__" not in k and "map" not in k
+    },
+)
+def test_ehrql_date_string_equivalence(fn_name):
+    date_str = "2000-01-01"
+    date_date = date(2000, 1, 1)
```
inglesp, line 404: Micro nit: since these are only used once each, why not inline them?  It's one less thing for the reader to keep in L0 cache.

tests/unit/test_query_language.py
```diff
@@ -381,3 +384,36 @@ def test_incompatible_duration_operations(lhs, op, rhs):
...
+    @table
+    class p(PatientFrame):
+        d = Series(date)
+
+    f = getattr(p.d, fn_name)
+    n_params = len(signature(f).parameters)
+    date_args = [date_date for i in range(n_params)]
+    str_args = [date_str for i in range(n_params)]
+
+    if "in" in fn_name:
```
inglesp, line 415: This is might be quite brittle.  What happens if we add a new method called eg `contains`?

I think it'd be better to be explicit: `if fn_name in ["is_in", "is_not_in"]`.  In addition to being more robust, this makes the reader do less thinking.

tests/unit/test_query_language.py
inglesp: Better might be to have a list of method names that we ignore.  (See my comment below about making the reader do less thinking!)

inglesp (approved):

databuilder/query_language.py
```diff
@@ -328,6 +328,12 @@ def is_between(self, start, end):
     def is_on_or_between(self, start, end):
         return (self >= start) & (self <= end)
 
+    def if_null_then(self, other):
+        return case(
+            when(self.is_not_null()).then(self),
+            default=parse_date_if_str(other),
+        )
```
evansd, line 335: Good catch!

tests/unit/test_query_language.py
```diff
@@ -381,3 +384,45 @@ def test_incompatible_duration_operations(lhs, op, rhs):
         else:
             assert False
         assert not result
+
+
+@pytest.mark.parametrize(
+    "fn_name",
+    (
+        {k for k, v in DateFunctions.__dict__.items() if callable(v)}
```
evansd, line 392: :+1: I'm a fan of using introspection to make sure you can't add things without remembering to test them (the coverage check helps with this, but doesn't solve it completely).

A variant on what you've done here is to define the test parameters manually, but have a separate test that inspects the parameters and makes sure they cover everything e.g.
https://github.com/opensafely-core/databuilder/blob/8c50e843dbc86f62fa73a9a3717778823680611e/tests/unit/file_formats/test_csv.py#L43-L48

Sometimes, generating the test examples themselves by introspection can end up being full of conditional logic and hard to follow, and having manually curated examples and a separate exhaustiveness check can keep the tests simple at the cost of being a bit more laborious to construct.

I think the test here is simple enough to not be a problem, but worth bearing in mind if it starts getting more complicated.

evansd (commented):
Great to see all this red, with the corresponding test speed-up as well.

tests/unit/test_query_language.py
Jongmassey: Cool - I did have some concerns re: the complexity of the derivation of the test parameters but figured it was just about on the right side of readable. I hadn't thought about using another test to check the params of a different test - that's a great idea!

databuilder/query_language.py
Jongmassey: It was only when I rewrote the test to use introspection rather than a static list of functions that I caught this!

PR #908, opened 2022-12-30
Support better reference docs for codelists

Jongmassey:


inglesp:
Code's all good, but we've got a better and simpler solution to the problem.  (See https://github.com/opensafely/documentation/issues/778#issuecomment-1372430027.)

Issue #862, opened 2022-12-07
Speed up the tests


benbc:

inglesp:
Some notes:

The last five successful test CI runs have included test steps taking between 8 and 11 minutes.

On CI, the first ~40-50s of the test step is spent downloading and installing requirements with pip.  Collecting tests and configuring pytest plugins takes a few seconds, as does starting the docker containers.  The remaining time is spent running the tests.

On my machine, a full test run (ie `pytest` with no arguments), without already-running mssql and spark containers, takes ~5 minutes, although I have seen it take longer (with my hunch that the spark tests are particularly high-variance, even accounting for #902).

However, running without spark (`pytest -k 'not spark'`), and without an already-running mssql container, takes ~1 minute.

So one way we can speed up tests with very little effort is to not run tests against spark.  Since spark is not yet used in production, this is something we should consider.

We can speed up the tests by running them in parallel.  However, since the tests share database connections, we'd have to ensure that tests that use the same database are run in sequence.

We can do this with pytest-xdist:

```
# conftest.py

def pytest_collection_modifyitems(session, config, items):
    for item in items:
        group = "other"

        if "engine" in item.fixturenames:
            engine = item.callspec.params["engine"]
            if engine in ["mssql", "spark"]:
                group = engine

        if "mssql_database" in item.fixturenames:
            group = "mssql"
        elif "spark_database" in item.fixturenames:
            group = "spark"

        item._nodeid = f"{item.nodeid}@{group}"
```

This makes the full test suite run about as fast as just running the spark tests.  However, it does not make much of a difference to running tests without spark: both the mssql tests and the non-mssql, non-spark tests take ~30s to run, and so I would have hoped that running them in parallel would also take ~30s.  However, there seems to be a fair bit of overhead, and so they run in ~45s.

More options to explore:

* doing less setup/teardown between tests
* making each test completely independent, allowing more even distribution

inglesp:
> making each test completely independent, allowing more even distribution

I've tried this (see https://github.com/opensafely-core/databuilder/commit/0fb467eca60b4e7cf4866659452e59a15a7be12c) and it's very much not worth it.

PR #911, opened 2023-01-02
Parallelise tests

inglesp:
This PR adds support for running tests in parallel using pytest-xdist, with tests against each slow database (mssql and spark) running in a different process.

The default user experience is unchanged, but you can now add `-n3` (or `-k 'not spark' -n2`) to run tests in parallel.

This leads to a moderate speed-up locally.  For me, non-spark tests can now run as quickly as 33s (down from 47s).  The full test suite can run as quickly as 197s (down from 230s).

I tried to enable it in CI but it didn't work because of a race condition in creating the mssql docker container.  (In any case it is not expected to lead to a significant speed-up, since CI runs also spend ~1m setting up the environment.)

Addresses #862.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,7 @@ add_ignore = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "--tb=native --strict-markers"
+addopts = "--tb=native --strict-markers --dist=loadgroup"
 testpaths = ["tests"]
 markers = [
     "integration: tests that use a container (mostly databases)",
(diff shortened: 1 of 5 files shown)


evansd (approved):
This is much less invasive than I feared it might have to be. I might even go so far as to call it neat. I've certainly got no objection to it going in. I'm not quite sure how much I'll end up using it myself as I've got so used to running smaller subsets of the suite. But maybe I will. And if it saves others time then I'd say it's worth it.

PR #913, opened 2023-01-03
Generalise date series type casting logic

evansd:
We now assume that all methods on all types will need to apply casting, and allow individual types to specify what that casting entails — which by default is a no-op.

This removes the method duplication from `DateFunctions` and gives us the foundations to address #540 when the time comes.

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -91,12 +91,20 @@ class BaseSeries:
             "such as 'a < b < c'.)"
         )
 
+    @staticmethod
+    def _cast(value):
+        # Series have the opportunity to cast arguments to their methods e.g. to convert
+        # ISO date strings to date objects. By default, this is a no-op.
+        return value
+
     # These are the basic operations that apply to any series regardless of type or
     # dimension
     def __eq__(self, other):
+        other = self._cast(other)
         return _apply(qm.Function.EQ, self, other)
 
     def __ne__(self, other):
+        other = self._cast(other)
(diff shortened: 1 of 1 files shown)


databuilder/query_language.py
```diff
@@ -285,6 +305,10 @@ def parse_date_if_str(value):
 
 
 class DateFunctions(ComparableFunctions):
+    @staticmethod
+    def _cast(value):
+        return parse_date_if_str(value)
```
Jongmassey, line 310: could the contents of parse_date_if_str be inlined here?

Jongmassey:
Really nice. Should [the one remaining direct call to `parse_date_if_str`](https://github.com/opensafely-core/databuilder/blob/6f21bdb82d2bfcd4af397d12841d1560480890dd/databuilder/query_language.py#L420) be replaced with a call to `other._cast()` for consistency?

Jongmassey (approved):

databuilder/query_language.py
evansd: I wondered about that too (it would certainly be neater) but there's still one place we need to call it. (And we can't use `other._cast()` here because we can't guarantee that `other` is a `BaseSeries`.) I thought about invoking it as:
```py
other = DateFunctions._cast(other)
```
Which would work correctly, but feels like a hack to me.

Jongmassey: > we can't guarantee that other is a BaseSeries

I hadn't considered that. Good point. I reckon it's good to merge as is

PR #919, opened 2023-01-04
Handle errors in generative tests where test input creates out-of-range dates

### LINKED ISSUE #916: Gentests fail with "year [...] is out of range"
We have occasional generative test failures like this:

```
  File "/home/runner/work/databuilder/databuilder/databuilder/utils/date_utils.py", line 62, in date_add_years
    return datetime.date(date.year + num_years, date.month, date.day)
ValueError: year -3 is out of range
```

[Here](https://github.com/opensafely-core/databuilder/actions/runs/3798074696/jobs/6459515693) is an example.

The problematic query in that case is

```
Function.DateAddYears(
    lhs=Value(value=datetime.date(2000, 1, 1)),
    rhs=Function.DateDifferenceInDays(
        lhs=Value(value=datetime.date(1994, 1, 1)),
        rhs=SelectColumn(
            source=SelectPatientTable(
                name="p1",
                schema=TableSchema(...),
            ),
            name="d1",
        ),
    ),
)
```

The problem here is that `DateDifferenceInDays` is producing a negative integer. That is then used as a number of years to _subtract_ from the `lhs` which results in a date with a negative year (which is invalid).

It's very hard for us to exclude all such constructs, so we should allow these errors to happen and ignore them using the new machinery for that purpose.

There is another closely related error [here](https://github.com/opensafely-core/databuilder/actions/runs/3762908163/jobs/6396010695). This is triggered by `DateAddMonth` rather than `DateAddYears`. We should also make sure that we can handle that case.

In both cases the date construction (in `date_utils.date_add_months` and `date_utils.date_add_years`) has handling for a `ValueError` which assumes a specific cause of the error which is not the case here. That assumption is encoded as an `assert` which is _also_ being triggered in the error cases under discussion here. That handling will need reworking as part of this change.

rebkwok:
On occasion the generative tests create very large integers, e.g. as the result of a multiplication operations, or from a DateDifferenceInDays, which can result in date operations that attempt to return an out-of-range date (with year <=0 or >9999).

Fixes #916

### DIFF
--- a/databuilder/utils/date_utils.py
+++ b/databuilder/utils/date_utils.py
@@ -38,8 +38,9 @@ def date_add_days(date, num_days):
     return date + datetime.timedelta(days=num_days)
 
 
-def valid_year(year):
-    return 0 < year <= 9999
+def assert_valid_year(year):
+    if not 0 < year <= 9999:
+        raise ValueError(f"year {year} is out of range")
 
 
 def date_add_months(date, num_months):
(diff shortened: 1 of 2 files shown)


tests/generative/test_query_model.py
```diff
@@ -112,14 +156,46 @@ def run_test(query_engines, data, variable, recorder):
         ), f"Mismatch between {first_name} and {second_name}"
 
 
+def run_error_test(query_engines, data, variable):
+    instances, variables = setup_test(data, variable)
+    for name, engine in query_engines.items():
+        try:
+            run_with(engine, instances, variables)
+        except Exception as e:  # pragma: no cover
+            assert False, f"{name} engine encountered an error: {e}"
```
benbc, line 165: Sorry -- I don't understand the purpose of this. Doesn't it fail when we get expected errors.

tests/generative/test_query_model.py
```diff
@@ -112,14 +156,46 @@ def run_test(query_engines, data, variable, recorder):
...
+    # The different query engines report errors from out-of-range dates in different ways:
+    # mssql
+    ".+Cannot construct data type date.*",  # DateAddYears, with an invalid calculated year
+    ".+Adding a value to a 'date' column caused an overflow.+",  # DateAddMonths, resulting in an invalid date
+    # sqlite
+    "Couldn't parse date string",
+    # in-memory engine
+    "year -?\\d+ is out of range",  # DateAddYears, with an invalid calculated year
+    "date value out of range",  # DateAddMonths, resulting in an invalid date
+]
```
benbc, line 185: I like the clear documentation of which error is which here.

It's a bit of a shame that the type of the exceptions is disconnected from the error messages. How about catching `Exception` and then testing the type and message against a list of expected pairs?

databuilder/utils/date_utils.py
```diff
@@ -38,6 +38,10 @@ def date_add_days(date, num_days):
     return date + datetime.timedelta(days=num_days)
 
 
+def valid_year(year):
+    return 0 < year <= 9999
```
evansd, line 42: I wonder if it would be cleaner to change this to:
```py
def assert_valid_year(year):
    if not 0 < year <= 9999:
        raise ValueError(f"year {year} is out of range")
```
And then call this before entering the `try` block.

tests/generative/test_query_model.py
rebkwok: The point was to ensure all the errors are caught.  i.e. all the date operations in the `test_handle_date_errors` parameters generate out of range dates in some way. If they raise errors, `run_with` should catch them and return an `IGNORE_RESULT`.  It doesn't assert that, because some of the engines don't raise errors (e.g. sqlite returns None for operations that result in a year > 9999 although it errors for years of <= 0).  I had it catch the exception rather than leaving it to just raise it so that we can report the engine the error occurred with.

rebkwok: It is in fact just that one - sqlite handles dates outside of the upper limit and returns None instead of an error, so I've revised the test to assert that the result is either `IGNORE_RESULT` or `{..., "v": None}` .  That would also have caught the error in the test for days (2000-01-01 + 8000*365 days was a valid calculation because of leap years - updated to + 8000*366 now)

benbc (approved):

databuilder/utils/date_utils.py
```diff
@@ -52,17 +53,17 @@ def date_add_months(date, num_months):
...
-        # 1) An out-of-range year, if new_year results in a value < 1 or > 9999 OR
+        # 1) An out-of-range year, if new_year results in a value < 1 or > 9999 - in this
+        #    case the ValueError is raised; OR
         # 2) A new month which has no corresponding day; in this case we roll forward to the
         #    first of the next month. For a defence of this logic see:
         #    tests/spec/date_series/ops/test_date_series_ops.py::test_add_months
 
         # As no month has more days than December we'll never need to roll forward from
         # December and so we don't need to worry about wrapping round to the next year
+        assert_valid_year(new_year)
```
evansd, line 64: Sorry, I actually meant we could do this outside the `try` block, immediately after we construct `new_year`. That way we don't need to explain and handle two different error conditions inside the exception handler and I think that makes the whole thing easier to understand.

evansd (approved):

databuilder/utils/date_utils.py
rebkwok: Ah, I see - I've done that now

PR #920, opened 2023-01-05
Support division operations for numeric series

### LINKED ISSUE #657: Support division operations for numeric series

This is _mostly_ quite straightforward but there are a few subtleties here.

In ehrQL terms we'd need to add the `__truediv__` and `__floordiv__` methods to the `NumericFunctions` mixin:

We should also support the reverse methods: `__rtruediv__` and `__rfloordiv__`.

The query model will need new `TrueDivide` and `FloorDivide` operations:
```py
class TrueDivide(Series[float]):
    lhs: Series[Numeric]
    rhs: Series[Numeric]

class FloorDivide(Series[int]):
    lhs: Series[Numeric]
    rhs: Series[Numeric]
```

```
tests/lib/in_memory/engine.py
```

After adding that you should get errors from population validation and generative test code which guide you to add support there as well.

There are more complications here than with the multiplication operation. We want to follow Python in having separate operations for float and integer division where the result depends solely on the operation used, rather than the types of the input. This is in contrast to SQL where `1 / 2 = 0` but `1.0 / 2.0 = 0.5`.

Implementing this behaviour in SQL will require some extra work. I imagine that the method which implements this will need to inspect the types of its arguments and, if necessary, do some casting to get the desired behaviour.

Another subtlety is that we want floor division for negative numbers to round towards zero, rather than downwards i.e. `-1 // 2 = 0`. This is in contrast to Python where the result would be `-1`. I actually prefer the Python behaviour here; but matching the default SQL behaviour makes the implementation much simpler, and I doubt many people care about the difference.

Finally, we want division by zero to result in NULL, rather than throwing an error. This is the natural behaviour in both Spark and SQLite but not in MSSQL. Handing this correctly in MSSQL will involve wrapper the divisor in a `NULLIF(divisor, 0)` expression to convert it to NULL if it equals zero.

In all the above though, the key thing is that all the various edge cases are covered by spec tests.

As part of this change, I think most of `tests/spec/int_series_ops` could be moved to `numeric_series_ops` and the cases expanded to cover floats as well (we don't currently test floating point comparisons or arithmetic).

milanwiedemann:
Closes #657

### DIFF
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -31,6 +31,7 @@ from databuilder.query_model.transforms import (
 )
 from databuilder.utils.functools_utils import singledispatchmethod_with_cache
 from databuilder.utils.sqlalchemy_query_utils import (
+    expr_has_type,
     get_setup_and_cleanup_queries,
     is_predicate,
 )
@@ -225,8 +226,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         lhs = self.get_expr(node.lhs)
         rhs = self.get_expr(node.rhs)
         # To ensure TrueDiv behaviour cast one to float if both args are ints
-        if isinstance(lhs.type, sqlalchemy_types.Integer) and isinstance(
-            rhs.type, sqlalchemy_types.Integer
+        if not expr_has_type(lhs, sqlalchemy_types.Float) and not expr_has_type(
+            rhs, sqlalchemy_types.Float
         ):
             lhs = sqlalchemy.cast(lhs, sqlalchemy.Float)
         return self.truedivide(lhs, rhs)
--- a/databuilder/query_engines/base_sql.py
+++ b/databuilder/query_engines/base_sql.py
@@ -220,6 +220,25 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     def get_sql_multiply(self, node):
         return operators.mul(self.get_expr(node.lhs), self.get_expr(node.rhs))
 
+    @get_sql.register(Function.TrueDivide)
+    def get_sql_truedivide(self, node):
+        lhs = self.get_expr(node.lhs)
+        rhs = self.get_expr(node.rhs)
+        # To ensure TrueDiv behaviour cast one to float if both args are ints
+        if isinstance(lhs.type, sqlalchemy_types.Integer) and isinstance(
+            rhs.type, sqlalchemy_types.Integer
+        ):
+            lhs = sqlalchemy.cast(lhs, sqlalchemy.Float)
+        return self.truedivide(lhs, rhs)
+
+    def truedivide(self, lhs, rhs):
+        return lhs / rhs
+
+    @get_sql.register(Function.FloorDivide)
+    def get_sql_floordivide(self, node):
+        float_result = self.get_sql_truedivide(node)
+        return sqlalchemy.cast(SQLFunction("FLOOR", float_result), sqlalchemy.Integer)
+
     @get_sql.register(Function.CastToInt)
     def get_sql_cast_to_int(self, node):
         return sqlalchemy.cast(self.get_expr(node.source), sqlalchemy.Integer)
--- a/databuilder/query_engines/in_memory.py
+++ b/databuilder/query_engines/in_memory.py
@@ -11,7 +11,7 @@ from databuilder.query_engines.in_memory_database import (
 )
 from databuilder.query_model import nodes as qm
 from databuilder.query_model.transforms import apply_transforms
-from databuilder.utils import date_utils
+from databuilder.utils import date_utils, math_utils
 
 T = True
 F = False
@@ -234,6 +234,12 @@ class InMemoryQueryEngine(BaseQueryEngine):
 
     def visit_Multiply(self, node):
         return self.visit_binary_op_with_null(node, operator.mul)
+
+    def visit_TrueDivide(self, node):
+        return self.visit_binary_op_with_null(node, math_utils.truediv)
+
+    def visit_FloorDivide(self, node):
+        return self.visit_binary_op_with_null(node, math_utils.floordiv)
 
     def visit_CastToInt(self, node):
         return self.visit_unary_op_with_null(node, int)
--- a/databuilder/query_model/nodes.py
+++ b/databuilder/query_model/nodes.py
@@ -315,6 +315,14 @@ class Function:
         lhs: Series[Numeric]
         rhs: Series[Numeric]
 
+    class TrueDivide(Series[float]):
+        lhs: Series[Numeric]
+        rhs: Series[Numeric]
+
+    class FloorDivide(Series[int]):
+        lhs: Series[Numeric]
+        rhs: Series[Numeric]
+
     # Casting numeric types
     class CastToInt(Series[int]):
         source: Series[Numeric]
--- a/databuilder/query_model/population_validation.py
+++ b/databuilder/query_model/population_validation.py
@@ -13,7 +13,7 @@ from databuilder.query_model.nodes import (
     get_series_type,
     has_one_row_per_patient,
 )
-from databuilder.utils import date_utils
+from databuilder.utils import date_utils, math_utils
 
 
 def validate_population_definition(population):
@@ -212,6 +212,8 @@ register_op(Function.GE)(operator.ge)
 register_op(Function.Add)(operator.add)
 register_op(Function.Subtract)(operator.sub)
 register_op(Function.Multiply)(operator.mul)
+register_op(Function.TrueDivide)(math_utils.truediv)
+register_op(Function.FloorDivide)(math_utils.floordiv)
 register_op(Function.CastToInt)(int)
 register_op(Function.CastToFloat)(float)
 register_op(Function.StringContains)(operator.contains)
--- a/databuilder/utils/math_utils.py
+++ b/databuilder/utils/math_utils.py
@@ -1,0 +1,18 @@
+def truediv(lhs, rhs):
+    """
+    Implement Python truediv behaviour but return None when dividing by zero.
+    """
+    if rhs == 0:
+        return None
+    else:
+        return lhs / rhs
+
+
+def floordiv(lhs, rhs):
+    """
+    Implement Python floordiv behaviour but return None when dividing by zero.
+    """
+    if rhs == 0:
+        return None
+    else:
+        return int(lhs // rhs)
--- a/tests/spec/float_series_ops/test_arithmetic_ops.py
+++ b/tests/spec/float_series_ops/test_arithmetic_ops.py
@@ -1,6 +1,6 @@
 from ..tables import p
 
-title = "Arithmetic operations"
+title = "Arithmetic operations without division"
 
 table_data = {
     p: """
--- a/tests/spec/float_series_ops/test_division_ops.py
+++ b/tests/spec/float_series_ops/test_division_ops.py
@@ -1,0 +1,127 @@
+from ..tables import p
+
+title = "Arithmetic division operations"
+
+table_data = {
+    p: """
+          |  f1   |   f2
+        --+-------+--------
+        1 | 101.3 |  111.5
+        2 |  -1.3 |  111.5
+        3 | -10.3 |    1.5
+        4 | -10.3 |   -1.5
+        5 |       |   -1.5
+        """,
+}
+
+
+def test_truedivide(spec_test):
+    spec_test(
+        table_data,
+        p.f1 / p.f2,
+        {
+            1: 101.3 / 111.5,
+            2: -1.3 / 111.5,
+            3: -10.3 / 1.5,
+            4: -10.3 / -1.5,
+            5: None,
+        },
+    )
+
+
+def test_truedivide_by_constant(spec_test):
+    spec_test(
+        table_data,
+        p.f1 / 10.0,
+        {
+            1: 101.3 / 10.0,
+            2: -1.3 / 10.0,
+            3: -10.3 / 10.0,
+            4: -10.3 / 10.0,
+            5: None,
+        },
+    )
+
+
+def test_rtruedivide_by_constant(spec_test):
+    spec_test(
+        table_data,
+        10.0 / p.f1,
+        {
+            1: 101.3 / 10.0,
+            2: -1.3 / 10.0,
+            3: -10.3 / 10.0,
+            4: -10.3 / 10.0,
+            5: None,
+        },
+    )
+
+
+def test_truedivide_by_zero(spec_test):
+    spec_test(
+        table_data,
+        p.i1 / 0,
+        {
+            1: None,
+            2: None,
+            3: None,
+            4: None,
+            5: None,
+        },
+    )
+
+
+def test_floordivide(spec_test):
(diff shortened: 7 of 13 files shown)


databuilder/query_model/nodes.py
```diff
@@ -276,6 +276,14 @@ class Multiply(Series[Numeric]):
...
         rhs: Series[Numeric]
 
+    class TrueDivide(Series[Numeric]):
+        lhs: Series[Numeric]
+        rhs: Series[Numeric]
+
+    class FloorDivide(Series[Numeric]):
+        lhs: Series[Numeric]
+        rhs: Series[Numeric]
+
```
evansd, line 286: This signature isn't quite right because, as it's written, it's saying that the operation takes any two numeric types and returns a value of _the same numeric type_ as the inputs. But the behaviour we want from `TrueDivide` is that it always returns a float, even if its arguments are ints. And we want `FloorDivide` to always return an int even if its arguments are floats. So the signature should be something like:
```py
class TrueDivide(Series[float]):
    lhs: Series[Numeric]
    rhs: Series[Numeric]

class FloorDivide(Series[int]):
    lhs: Series[Numeric]
    rhs: Series[Numeric]
```

databuilder/query_engines/in_memory.py
```diff
@@ -230,6 +230,12 @@ def visit_Subtract(self, node):
     def visit_Multiply(self, node):
         return self.visit_binary_op_with_null(node, operator.mul)
 
+    def visit_TrueDivide(self, node):
+        return self.visit_binary_op_with_null(node, operator.div)
+
+    def visit_FloorDivide(self, node):
+        return self.visit_binary_op_with_null(node, operator.floordiv)
```
evansd, line 237: I think we'll end up having to define our own function here, rather than reusing the one from `operator`, because Python's floordiv behaviour is slightly different from the SQL behaviour we want to emulate. But we should just define the desired behaviour in the spec tests and let them drive the necessary changes out.

databuilder/query_engines/base_sql.py
```diff
@@ -221,6 +221,14 @@ def get_sql_subtract(self, node):
     def get_sql_multiply(self, node):
         return operators.mul(self.get_expr(node.lhs), self.get_expr(node.rhs))
 
+    @get_sql.register(Function.TrueDivide)
+    def get_sql_truedivide(self, node):
+        return operators.truediv(self.get_expr(node.lhs), self.get_expr(node.rhs))
```
evansd, line 226: SQL doesn't have distinct truediv/floordiv operators. Instead it says, if both arguments are ints we perform floor division, otherwise we perform true division. So to get the behaviour we want we have to cast types appropriately. Something like this:
```py
lhs = self.get_expr(node.lhs)
rhs = self.get_expr(node.rhs)
# If both arguments are ints, cast one to a float to ensure we get TrueDiv behaviour
if isinstance(lhs.type, sqlalchemy_types.Integer) and isinstance(rhs.type, sqlalchemy_types.Integer):
    lhs = sqlalchemy.cast(lhs, sqlalchemy.Float)
return operators.div(lhs, rhs)
```

databuilder/query_engines/base_sql.py
```diff
@@ -221,6 +221,14 @@ def get_sql_subtract(self, node):
...
+
+    @get_sql.register(Function.FloorDivide)
+    def get_sql_floordivide(self, node):
+        return operators.floordiv(self.get_expr(node.lhs), self.get_expr(node.rhs))
```
evansd, line 230: And the logic here is the inverse of the above:
```py
lhs = self.get_expr(node.lhs)
rhs = self.get_expr(node.rhs)
# Cast any float arguments to int to ensure FloorDiv behaviour
if isinstance(lhs.type, sqlalchemy_types.Float):
    lhs = sqlalchemy.cast(lhs, sqlalchemy.Integer)
if isinstance(rhs.type, sqlalchemy_types.Float):
    rhs = sqlalchemy.cast(rhs, sqlalchemy.Integer)
return operators.div(lhs, rhs)
```

evansd (commented):
Looks like a good start. I've added what I hope are some helpful comments. Do let me know if anything doesn't make sense though.

databuilder/query_engines/in_memory.py
milanwiedemann: > But we should just define the desired behaviour in the spec tests and let them drive the necessary changes out.

Thanks, don't understand where we would need to define our own `floordiv` function and how this relates to the spec tests? What do you mean by "_let them drive the necessary changes out_"?

evansd: Sorry, that was spectacularly opaque on my part! I was referring to this part of the ticket:

> Another subtlety is that we want floor division for negative numbers to round towards zero, rather than downwards i.e. `-1 // 2 = 0`. This is in contrast to Python where the result would be `-1`. I actually prefer the Python behaviour here; but matching the default SQL behaviour makes the implementation much simpler, and I doubt many people care about the difference.

So what I meant was that we'll need to add a case to the spec tests to cover this behaviour e.g.

    -4 // 3 == -1

As soon as you add that test you'll discover that the in-memory engine fails, because Python rounds the result down to -2. And so you'll have to write a custom Python division function with different rounding behaviour to get the tests to pass. (This is what I meant by letting the tests drive the necessary changes out.)

tests/spec/float_series_ops/test_arithmetic_ops.py
```diff
@@ -58,3 +58,35 @@ def test_multiply_with_constant(spec_test):
...
         {1: 10.0 * 111.5, 2: None},
     )
+
+
+def test_truedivide(spec_test):
+    spec_test(
+        table_data,
+        p.f1 / p.f2,
+        {1: 101.3 / 111.5, 2: None},
+    )
```
evansd, line 68: A further thought on testing here. Given that division is quite a lot more complex than any of the other operations and there are many more edges cases to cover I wonder if it's worth breaking out a separate `test_division_ops.py` file with its own `table_data`.

milanwiedemann: that's probably a good idea

databuilder/query_model/population_validation.py
```diff
@@ -211,6 +211,8 @@ def in_(lhs, rhs):
 register_op(Function.Add)(operator.add)
 register_op(Function.Subtract)(operator.sub)
 register_op(Function.Multiply)(operator.mul)
+register_op(Function.TrueDivide)(operator.truediv)
+register_op(Function.FloorDivide)(operator.floordiv)
```
milanwiedemann, line 215: I'm not sure what operator to register here because `operator.floordiv` doesn't exist and we need to define our custom one?

milanwiedemann: or does that happen in `databuilder/query_language.py`? I don't think so, I'm confused


```
    def __floordiv__(self, other):
        other = self._cast(other)
        return _apply(qm.Function.FloorDivide, self, other)
```

evansd: It might be helpful to look at this commit here which introduced the `date_utils` module: https://github.com/opensafely-core/databuilder/commit/e44c327244a22bd63e5ae665b15729ae8b1269cf

Structurally, what we're trying to do here is the same. So I think probably the thing to do is to add a new `maths_utils` module which defines our own `floordiv` function. That could be then be imported and registered in `population_validation.py`:
```py
register_op(Function.FloorDivide)(maths_utils.floordiv)
```

And in `in_memory.py`:
```py
def visit_FloorDivide(self, node):
    return self.visit_binary_op_with_null(node, maths_utils.floordiv)
```

milanwiedemann:
I'll need to look at the failures again tomorrow with a fresh brain but will probably need some guidance how to (1) identify solutions to fix the assertion errors and (2) how to start writing our own `floordiv` function in [utils/math_utils.py](https://github.com/opensafely-core/databuilder/blob/bd8db01c261389cce120120380c7d3ef257db24c/databuilder/utils/math_utils.py)

databuilder/utils/math_utils.py
```diff
@@ -0,0 +1,8 @@
+import operator
+
+
+def floordiv(num_lhs, num_rhs):
+    """
+    NOTE: Match behaviour by SQL that rounds towards zero for negative numbers.
+    """
+    return operator.floordiv(num_lhs, num_rhs)
```
milanwiedemann, line 8: I tried and I think I  succeeded in defining a function that matches the SQL behaviour, but it was long with a couple if else statements ... I used `int()` but any pointers would be great

tests/spec/int_series_ops/test_division_ops.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_floordivide(spec_test):
+    spec_test(
+        table_data,
+        p.i1 // p.i2,
+        {
+            1: 101 // 111,
+            2: 0,
+            3: -1,
```
milanwiedemann, line 50: These two lines test the following:
`-1  // 2 == 0`
`-4 // 3 == -1`

Usually we specify the arithmetic operation here in the spec tests, but if we expect different behaviours between  in_memory and SQL (sqlite, mssql) I'm not sure what the best approach is.

Depending on whether I specify the results that I expect from in_memory (`-1  // 2 == -1` and `-4 // 3 == -2`) or SQL (see above) I get passed test for one or the other.

I thought that defining a `floordiv` function that matches the SQL behaviour (see [here](https://github.com/opensafely-core/databuilder/pull/920/files#r1095765003)) would fix all my problems but so far I couldnt get that to work.

evansd: We definitely don't expect different behaviour between the in-memory and SQL engines. That's part of the point of the spec tests! As for defining a Python function which matches the SQL behaviour, they behave the same when the result is positive and only diverge for negative results. So how about:
```py
def floordiv(lhs, rhs):
    result_is_negative = (lhs < 0) != (rhs < 0)
    absolute_result = abs(lhs) // abs(rhs)
    return -absolute_result if result_is_negative else absolute_result
```

milanwiedemann: thanks! I think this works everywhere but some tests still failing (but I _think_ because of other problems, mainly my not ideal choices for test data and how I specify the actual and expected results)

milanwiedemann: following the tests from date series ops I noticed that I need to specify the actual result that I'm expecting when the engines differ (e.g., [date_series_ops/test_date_series_ops.py](https://github.com/opensafely-core/databuilder/blob/112e6cd91a4cdc1c40a0c080d964e990dcd96028/tests/spec/date_series_ops/test_date_series_ops.py#L197-L209)), instead of rewriting the operation (e.g., [float_series_ops/test_arithmetic_ops.py](https://github.com/opensafely-core/databuilder/blob/112e6cd91a4cdc1c40a0c080d964e990dcd96028/tests/spec/float_series_ops/test_arithmetic_ops.py#L32-L36))

tests/spec/float_series_ops/test_division_ops.py
```diff
@@ -0,0 +1,127 @@
...
+            1: 101.3 / 10.0,
+            2: -1.3 / 10.0,
+            3: -10.3 / 10.0,
+            4: -10.3 / 10.0,
+            5: None,
+        },
+    )
+
+
+def test_rtruedivide_by_constant(spec_test):
```
evansd, line 46: It might be clearer to call this `test_truedivide_constant_by_series`

tests/spec/float_series_ops/test_division_ops.py
```diff
@@ -0,0 +1,127 @@
...
+            1: 101.3 / 10.0,
+            2: -1.3 / 10.0,
+            3: -10.3 / 10.0,
+            4: -10.3 / 10.0,
+            5: None,
+        },
+    )
+
+
+def test_truedivide_by_zero(spec_test):
```
evansd, line 60: I think that, rather than having a separate divide-by-zero test, it might be better to add another row to `table_data` where `f2` is zero and then that case gets covered in the other tests.

tests/spec/float_series_ops/test_division_ops.py
```diff
@@ -0,0 +1,127 @@
...
+            1: 101.3 // 10,
+            2: -1.3 // 10,
+            3: -10.3 // 10,
+            4: -10.3 // 10,
+            5: None,
+        },
+    )
+
+
+def test_rfloordivide_by_constant(spec_test):
```
evansd, line 102: Likewise, maybe call this one `test_floordivide_constant_by_series`

databuilder/query_engines/base_sql.py
```diff
@@ -220,6 +220,25 @@ def get_sql_subtract(self, node):
...
 
+    @get_sql.register(Function.TrueDivide)
+    def get_sql_truedivide(self, node):
+        lhs = self.get_expr(node.lhs)
+        rhs = self.get_expr(node.rhs)
+        # To ensure TrueDiv behaviour cast one to float if both args are ints
+        if isinstance(lhs.type, sqlalchemy_types.Integer) and isinstance(
+            rhs.type, sqlalchemy_types.Integer
+        ):
+            lhs = sqlalchemy.cast(lhs, sqlalchemy.Float)
```
evansd, line 231: There are a few things I've realised here since I first suggested this code. One is that the `isinstance()` check isn't actually enough because it doesn't work with custom types. So you have to use the `epxr_has_type` function from `sqlalchemy_query_utils`.

I've also realised that it's safer to check that values are _not_ floats, rather than check that they _are_ ints. If for some reason the type check doesn't work as expected, it's better to do an unnecessary CAST than to fail to CAST where we need to.

You can see both these at work here (which is where I realised this stuff):
https://github.com/opensafely-core/databuilder/blob/21c0c95d3dbf728e32d87ddfaf741648794df9a6/databuilder/query_engines/mssql.py#L33-L38

evansd (approved):
:tada: It's great that you've persevered with this and I think it all looks good. I've got a few small comments below about the tests, and then I've also realised that some of the advice I initially gave you about how to implement `truedivide` wasn't quite right! But hopefully all fairly small fixes.

tests/spec/float_series_ops/test_division_ops.py
milanwiedemann: ah yes, that's clearer!

tests/spec/float_series_ops/test_division_ops.py
milanwiedemann: yes, that makes sense. I remember getting stuck when I first tried this but found out why and changed it now

databuilder/query_engines/base_sql.py
milanwiedemann: Thanks, I can see how this is safer!

tests/spec/int_series_ops/test_division_ops.py
```diff
@@ -0,0 +1,118 @@
+from ..tables import p
```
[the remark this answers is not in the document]
milanwiedemann: done

milanwiedemann:
thanks @evansd, that all made sense and I implemented all suggestions (see two recent commits). if you're happy I'll squash and merge :)

evansd (approved):
🎉

Issue #839, opened 2022-11-23
Add `Case` to generative tests


benbc:
The operation needs adding to `tests.generative.variable_strategies`. If you're not familiar with the generative tests then you should get an intro from (probably) Ben before starting this work.

There is a verification step to ensure that we're including all the operations:

```
GENTEST_COMPREHENSIVE=t GENTEST_EXAMPLES=5000 just test-generative
```

Once you've added the operation you'll find that this fails. You will then need to remove `Case` from `tests.generative.test_query_model.recorder` and it should pass. If you find you've removed the last of the known_missing operations then please remove this exception mechanism altogether.

rebkwok:
Closed by #925

PR #925, opened 2023-01-06
Add Case to generative tests

rebkwok:


tests/generative/variable_strategies.py
```diff
@@ -273,6 +274,12 @@ def variable(
     to_first_of_year = qm_builds(Function.ToFirstOfYear, date_series)
     to_first_of_month = qm_builds(Function.ToFirstOfMonth, date_series)
 
+    case = qm_builds(
+        Case,
+        st.dictionaries(series, series, min_size=1, max_size=3),
```
rebkwok, line 279: Initially I had `max_size=5`, but with 20K examples it still didn't manage to generate one that used `Case` - presumably because there were too many possible non-homogenous inputs that got thrown away.  With a max of 3, it would work with 10K examples.

databuilder/query_model/nodes.py
```diff
@@ -609,7 +609,12 @@ def validate_types(node):
         # are no ints sitting around to be checked). So the standard, "isinstance()",
         # approach to type checking doesn't work. There may well be a more elegant
         # alternative approach here, but this works for now.
-        typespec = get_typespec(value)
+        try:
+            # sets and mappings require that their items are of homogenous types, and
+            # raise a TypeError during the conversion to type specification.
+            typespec = get_typespec(value)
+        except TypeError as e:
+            raise TypeValidationError(str(e))
```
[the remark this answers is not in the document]
iaindillingham: Might it be useful to [chain the exceptions][1], here?

```suggestion
            raise TypeValidationError(str(e)) from e
```

[1]: https://docs.python.org/3.10/reference/simple_stmts.html#the-raise-statement

iaindillingham (approved):
I'm not yet familiar with hypothesis, but I think I understand what you're doing here and why 🙂 I've made a passing comment about chaining exceptions, which you should feel free to ignore.

tests/generative/recording.py
```diff
@@ -70,7 +70,8 @@ def check_comprehensive(recorder):  # pragma: no cover
 
 
 def check_not_too_many_ignored_errors(recorder):
-    assert recorder.num_ignored_errors / recorder.num_results <= 0.01
+    if recorder.num_results:  # pragma: no cover
```
inglesp, line 73: The comment in the commit would be useful here too.

inglesp (approved):

tests/generative/variable_strategies.py
evansd: Yeah, that makes sense: it's not even that you're trying to get 5-of-a-kind by randomly dealing cards; you need 5 homogeneous-typed values _and_ 5 boolean-typed keys, so the changes get tiny very quickly as you increase N.

This is the sort of thing that the alternative approach Ben is working on will hopefully solve for us.

evansd (approved):

rebkwok:
closes #839

PR #927, opened 2023-01-06
Make sqlite date arithmetic roll over to the 1st of the next month

### LINKED ISSUE #918: Date arithmetic inconsistency between SQLite and in-memory database
This was found by the generative tests [here](https://github.com/opensafely-core/databuilder/actions/runs/3804576792/jobs/6471892745) (we've only seen this once).

```
FAILED tests/generative/test_query_model.py::test_query_model - AssertionError: Mismatch between in_memory and sqlite
assert [{'patient_id': 1, 'v': 1}] == [{'patient_id': 1, 'v': 2}]
  At index 0 diff: {'patient_id': 1, 'v': 1} != {'patient_id': 1, 'v': 2}
  Full diff:
  - [{'patient_id': 1, 'v': 2}]
  ?                         ^
  + [{'patient_id': 1, 'v': 1}]
  ?                         ^
Falsifying example: test_query_model(
    variable=Function.DayFromDate(source=Function.DateAddMonths(lhs=Value(value=datetime.date(2001, 1, 30)), rhs=AggregateByPatient.Count(source=SelectTable(name='e0', schema=TableSchema(i1=Column(int, constraints=()), i2=Column(int, constraints=()), b1=Column(bool, constraints=()), b2=Column(bool, constraints=()), d1=Column(datetime.date, constraints=()), d2=Column(datetime.date, constraints=())))))),
    data=[{'type': data_setup.E0,
      'patient_id': 1,
      'i1': 0,
      'i2': 0,
      'b1': False,
      'b2': False,
      'd1': datetime.date(2000, 1, 1),
      'd2': datetime.date(2000, 1, 1)}],
    query_engines={'in_memory': <tests.conftest.QueryEngineFixture at 0x7f17fe6db400>,
     'sqlite': <tests.conftest.QueryEngineFixture at 0x7f17fe6db1c0>,
     'mssql': <tests.conftest.QueryEngineFixture at 0x7f17fe6f03d0>},
    recorder=<bound method ObservedInputs.record of <tests.generative.recording.ObservedInputs object at 0x7f17ee37c3d0>>,
)

You can reproduce this example by temporarily adding @reproduce_failure('6.56.3', b'AXicE2FiY2BkkJViAAJGBlQAAAfmAFY=') as a decorator on your test case
```

Looks as though adding one month to 30th Jan 2001 gives 1st Mar for one database and 2 Mar for the other.

rebkwok:
SQLite performs a normalisation when a date arithmetic operation results in a date with an invalid number of days (e.g. 30 Feb 2000); namely it rolls over to the next month by incrementing the month and subtracting the number of days in the month.  For February dates, this results in dates between 1-3 March, depending on how far off the calculated date is (i.e. 31 Feb in a non-leap year will become 03 Mar). We now apply a correction so that we always roll over to the 1st of the next month.

Fixes #918

### DIFF
--- a/databuilder/query_engines/sqlite.py
+++ b/databuilder/query_engines/sqlite.py
@@ -23,7 +23,47 @@ class SQLiteQueryEngine(BaseSQLQueryEngine):
         return self.date_add("days", date, num_days)
 
     def date_add_months(self, date, num_months):
-        return self.date_add("months", date, num_months)
+        new_date = self.date_add("months", date, num_months)
+        # In cases of day-of-month overflow, SQLite *usually* rolls over to the first of
+        # the next month as want it to.
+        # It does this by performing a normalisation when a date arithmetic operation results
+        # in a date with an invalid number of days (e.g. 30 Feb 2000); namely it
+        # rolls over to the next month by incrementing the month and subtracting
+        # the number of days in the month. For all months except February, the invalid day is only
+        # ever at most 1 day off (i.e. the 31st, for a month with only 30 days), and so this
+        # normalisation results in a rollover to the first of the next month. However for February,
+        # the invalid day can be 1 to 3 days off, which means date rollovers can result in the 1st,
+        # 2nd or 3rd March.
+        #
+        # The SQLite docs (https://sqlite.org/lang_datefunc.html) state:
+        # "Note that "±NNN months" works by rendering the original date into the YYYY-MM-DD
+        # format, adding the ±NNN to the MM month value, then normalizing the result. Thus,
+        # for example, the date 2001-03-31 modified by '+1 month' initially yields 2001-04-31,
+        # but April only has 30 days so the date is normalized to 2001-05-01"
+        #
+        # i.e. 2001-03-31 +1 month results in 2001-04-31, but as the number of days is invalid,
+        # SQLite rolls over to the next month by incrementing the month by 1, and subtracting
+        # the number of days in April (30) from the days, resulting in 2001-05-01
+        #
+        # In the case of February, a calculation that results in a date with an invalid number of
+        # days follows the same normalisation method:
+        # 2000-02-30: increment month by 1, subtract 29 (leap year) days -> 2000-03-01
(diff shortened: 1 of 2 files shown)


tests/spec/date_series_ops/test_date_series_ops.py
```diff
@@ -132,19 +132,34 @@ def test_add_months(spec_test):
...
+        8  | 2004-03-29 | -1
+        9  | 2003-03-30 | -1
+        10 | 2004-03-30 | -1
+        11 | 2003-03-31 | -1
+        12 | 2004-03-31 | -1
+        13 | 2003-01-29 | -1
+        14 | 2000-10-31 |  11
+        15 | 2000-10-31 | -11
+        16 | 2000-10-31 |  4
+        17 | 1999-11-30 |  3
```
rebkwok, line 162: These test cases may be overkill, especially given that it ends up in the docs, but I wanted to assure myself it worked.  I could remove some if it's going to make the docs confusing.

evansd: Yeah, I can see why you'd add all these while working on it but I think we need to be trying to keep a lid on the spec tests otherwise it's very easy for them to get out of control. I think what we should be aiming for is the minimal set that clearly excludes SQLite's native behaviour.

databuilder/query_engines/sqlite.py
```diff
@@ -23,7 +23,47 @@ def date_add_days(self, date, num_days):
         return self.date_add("days", date, num_days)
 
     def date_add_months(self, date, num_months):
-        return self.date_add("months", date, num_months)
+        new_date = self.date_add("months", date, num_months)
+        # In cases of day-of-month overflow, SQLite *usually* rolls over to the first of
```
evansd, line 27: For some reason, it's always our date-related code that ends up needing these comment essays :) But in any case, I think this kind of thing is super valuable and thanks for taking the time to write it.

evansd (approved):
:tada: 

I probably shouldn't feel quite so happy that other people are now sharing my date arithmetic pain, but I'd be lying if I said I wasn't.

PR #929, opened 2023-01-09
fix: Allow the in-memory engine to combine differently sorted event series

### LINKED ISSUE #917: Combining event-level series where one has been sorted is not supported by the in-memory database
This is an error that has cropped up in the generative tests [once](https://github.com/opensafely-core/databuilder/actions/runs/3784903709/jobs/6434557272).

The relevant QM construct is:

```
DateDifferenceInYears(
    lhs=SelectColumn(SelectTable("e0"), "d1"),
    rhs=SelectColumn(
        source=Sort(
            source=SelectTable("e0"),
            sort_by=SelectColumn(SelectTable("e0"), "i2"),
        ),
        name="d1",
    ),
)
```

This triggers an error in the in-memory database:

```
File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory.py", line 257, in visit_DateDifferenceInYears
    return self.visit_binary_op_with_null(node, date_utils.date_difference_in_years)
  File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory.py", line 153, in visit_binary_op_with_null
    return self.visit_binary_op(node, handle_null(op))
  File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory.py", line 150, in visit_binary_op
    return apply_function(op, lhs, rhs)
  File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory_database.py", line 441, in apply_function
    {
  File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory_database.py", line 442, in <dictcomp>
    p: apply_function_to_rows_and_values(fn, [col[p] for col in columns])
  File "/home/runner/work/databuilder/databuilder/databuilder/query_engines/in_memory_database.py", line [461](https://github.com/opensafely-core/databuilder/actions/runs/3784903709/jobs/6434557272#step:5:462), in apply_function_to_rows_and_values
    assert len(set(keys_list)) == 1
```

I guess that this is because we're trying to combine two event-level series that derive from the same table, but where one has been sorted and the other hasn't.

You can reproduce this problem by following the instructions in Hypothesis' error message in the build output (it may be necessary to check out the exact version of the code where the failure happened):

```
You can reproduce this example by temporarily adding @reproduce_failure('6.56.3', b'AXickxZlYGJgYGAGYiYGRiDJCEYwgMQEAwAKbQA8') as a decorator on your test case
```

rebkwok:
Fixes #917 

We have a check in the in-memory database when applying a function to Rows (i.e. combining event-series data) which asserts that the keys for each `Rows` instance that's being combined are the same.  If one of the event series is the result of a Sort, the keys may be the same, but in a different order.  That doesn't matter for combining the series, as long as the keys themselves still represent the same rows (which they do), so the assertion on keys now ignores their order.

### DIFF
--- a/tests/spec/combine_series/test_event_series_and_event_series.py
+++ b/tests/spec/combine_series/test_event_series_and_event_series.py
@@ -4,12 +4,12 @@ title = "Combining two event series"
 
 table_data = {
     e: """
-          |  i1 |  i2
-        --+-----+-----
-        1 | 101 | 111
-        1 | 102 | 112
-        2 | 201 | 211
-        2 | 202 | 212
+          |  i1 |  i2 | s1
+        --+-----+-----+---
+        1 | 101 | 111 | b
+        1 | 102 | 112 | a
+        2 | 201 | 211 | b
+        2 | 202 | 212 | a
     """,
 }
 
@@ -21,5 +21,19 @@ def test_event_series_and_event_series(spec_test):
         {
             1: (101 + 111) + (102 + 112),
             2: (201 + 211) + (202 + 212),
+        },
+    )
+
+
+def test_event_series_and_sorted_event_series(spec_test):
+    """
+    The sort order of the underlying event series does not affect their combination.
+    """
+    spec_test(
+        table_data,
+        (e.i1 + e.sort_by(e.s1).i2).minimum_for_patient(),
+        {
+            1: (101 + 111),
+            2: (201 + 211),
         },
     )
--- a/databuilder/query_engines/in_memory_database.py
+++ b/databuilder/query_engines/in_memory_database.py
@@ -462,7 +462,29 @@ def apply_function_to_rows_and_values(fn, args):
     # test data, and not a check that the QM has provided two frames with the same
     # domain.
     keys_list = [tuple(a) for a in args if isinstance(a, Rows)]
-    assert len(set(keys_list)) == 1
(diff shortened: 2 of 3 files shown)


tests/spec/combine_series/test_event_series_and_event_series.py
```diff
@@ -23,3 +23,17 @@ def test_event_series_and_event_series(spec_test):
...
+    The sort ordering of the underlying event series does not affect their combination.
+    """
+    spec_test(
+        table_data,
+        (e.i1 + e.sort_by(e.s1).i2).minimum_for_patient(),
+        {
+            1: (101 + 111),
+            2: (201 + 211),
+        },
+    )
```
rebkwok, line 39: This test reassured me that the sorting doesn't affect the outcome of the series combination (and also that the other engines can already deal with it) - but I'm not sure if it makes sense to be included in the docs

evansd: I had a go at what it would look like to move this test out of the specs tests and into the query engine integration tests and there ends up being quite a lot of boilerplate. I agree that it doesn't make a lot of sense in the docs, but I'd say we just leave it in here for now and we can decide if have to move it out later. It does feel like a useful property to test.

evansd (approved):

PR #930, opened 2023-01-10
Groundwork for table-from-file feature

evansd:
This PR lays the groundwork for implementing the [table-from-file](https://github.com/opensafely-core/databuilder/issues/793) feature, which means the necessary query model changes, the minimal ehrQL needed to write a test, and a minimal failing spec test (currently marked as `xfail`).

To help break the task down, this PR introduces a `@table_from_rows` decorator which takes its data as a list of row tuples. I don't expect this to stick around: it should be replaced by a `@table_from_file` decorator which takes a path to a file. But it enables us to worry about loading data from files as a separate step once we have the other core bits of machinery in place.

The next step is to teach the various query engines how to handle the `InlinePatientTable` node. I suggest starting with the `InMemoryQueryEngine` before tackling `BaseSQLQueryEngine`. There's a bit more detail in the "Query Engines" section of the notes [here](https://github.com/opensafely-core/databuilder/issues/793#issuecomment-1373957956).

### DIFF
--- a/tests/unit/query_model/test_nodes.py
+++ b/tests/unit/query_model/test_nodes.py
@@ -13,6 +13,8 @@ from databuilder.query_model.nodes import (
     Filter,
     Frame,
     Function,
+    InlinePatientTable,
+    IterWrapper,
     PickOneRowPerPatient,
     Position,
     SelectColumn,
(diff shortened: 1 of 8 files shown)


tests/unit/query_model/test_nodes.py
```diff
@@ -116,6 +118,29 @@ def test_query_reprs_round_trip(queries):
...
+    inline_table = InlinePatientTable(
+        rows=IterWrapper(
+            [
+                (1, 100),
+                (2, 200),
+            ]
+        ),
+        schema=TableSchema(i=Column(int)),
+    )
+    assert hash(inline_table)
```
inglesp, line 134: Could you add a comment about what exactly this line is testing?

inglesp (approved):

PR #931, opened 2023-01-11
Rework external acceptance tests

### LINKED ISSUE #681: Rework acceptance tests which use external repos

At the moment we have some acceptance tests which [pull in code](https://github.com/opensafely-core/databuilder/blob/f69754da48ddedc547fd2a421e905798a9446cce/tests/acceptance/test_age_distribution_study.py) from external repos and attempts to run it. The idea being that we should know if we've accidentally broken APIs that were actually being used.

However while our APIs are in flux, and while there are no actual studies using them, these tests make certain kinds of changes logistically painful. See related Slack thread here:
https://bennettoxford.slack.com/archives/C03FB777L1M/p1656596804056129

And even when we _do_ have stable APIs this behaviour is problematic because it makes CI non-deterministic (your PR can fail but not because of any changes within the PR itself).

As mentioned in the above thread, I think the solution is as follows:
 * acceptance test targets are vendored in to the databuilder repo, rather than being pulled in at runtime;
 * we have a script which does this vendoring and makes it easy to update the vendored version from the upstream;
 * we have a scheduled action that regularly checks out databuilder, runs the update script, and then runs the acceptance tests against the updated studies.

This final step is what guards against the vendored versions going stale, but in a way which keeps our CI runs hermetic and deterministic.

If you then need to make a breaking change to Data Builder, the process is:

1. Make the breaking change in a a PR.
2. In the same PR, make the corresponding change to the vendored dataset definition.
3. Get the PR merged.
4. Raise a PR against the original study with the same dataset definition changes.
5. Get that PR merged.
6. Run the update script to get the vendored dataset definition back in sync.

Note that there is currently only one such acceptance test, but we'll add others after this issue has been done.

evansd:
This reworks tests which dynamically pull in code from external repos to instead use vendored copies of that external code, as described in #681. This restores determinism to our test suite and also gives us a more straightforward workflow for making breaking changes where we deliberately decide to do so.

As part of this change we move the `comparative_booster_study` test to its [own repo](https://github.com/opensafely/comparative-booster-ehrql-poc) and then re-add it using the new mechanism.

Longer term we might replace this study with one under more active development, but for now it's useful to have at least one reasonably large and complex test study.

Closes #681


tests/acceptance/test_external_studies.py
```diff
@@ -0,0 +1,88 @@
...
+    # Test that we can compile the dataset definition to a valid query model graph. I
+    # think this is sufficient for these tests which are intended to ensure we don't
+    # accidentally break the API. If we're unable to execute a valid query, that's a
+    # separate class of problem for which we need separate tests.
+    dataset = load_dataset_definition(dataset_def_path)
+    variable_definitions = compile(dataset)
+    assert variable_definitions
+
+
+# SCRIPT TO UPDATE EXTERNAL STUDIES
```
inglesp, line 54: Feels like this should be in a separate file.

.github/workflows/update-external-studies.yml
```diff
@@ -0,0 +1,40 @@
...
+          branch: bot/update-external-studies
+          base: main
+          commit-message: "chore: Update `external_studies` test code"
+          title: "Update `external_studies` test code"
+          body: ""
+
+      # The PR will still require manual approval, this just reduces it to a one-click process
+      - name: Enable automerge
+        if: steps.create_pr.outputs.pull-request-operation == 'created'
+        uses: peter-evans/enable-pull-request-automerge@60812ab1c2c6c6a8932b4d6e059becafaf386256
```
inglesp, line 37: Nice touch

inglesp (approved):
Nice, and much less work than I'd thought.

> There's still an open question as to what studies we should include here. Might [cis-pop-validation-ehrql](https://github.com/opensafely/cis-pop-validation-ehrql) be a candidate?

Could we discuss in Slack?

tests/acceptance/test_external_studies.py
evansd: Yeah, fair enough. I was trying to keep it self-contained but it does mix up the concerns a bit too much.

PR #940, opened 2023-01-16
Add explanatory comment for frames in gentests

benbc:


evansd (approved):
:ok: Oh _of course_. You have erased my previous objection with this comment.

PR #942, opened 2023-01-16
Add IsNull to generative tests

rebkwok:


tests/generative/variable_strategies.py
```diff
@@ -112,6 +113,11 @@ def add(draw, type_, frame):
     def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
 
+    @st.composite
+    def is_null(draw, _type, frame):
+        type_ = draw(any_type())
+        return Function.IsNull(draw(series(type_, frame)))
```
rebkwok, line 119: I think this is what IsNull needs - a series of any type, but it can be a one or many rows per patient series, so we can just pass on `frame` without reducing it with `one_row_per_patient_frame_or`?

benbc: Yes, I think this is fine.

tests/generative/variable_strategies.py
```diff
@@ -70,6 +70,7 @@ class DomainConstraint:
             value: ({int, float, bool, datetime.date}, DomainConstraint.PATIENT),
             select_column: ({int, float, bool, datetime.date}, DomainConstraint.ANY),
             eq: ({bool}, DomainConstraint.ANY),
+            is_null: ({bool}, DomainConstraint.ANY),
```
benbc, line 73: I think maybe I'd argue that `IsNull` is "simpler" than `Eq`, but I don't feel strongly.

benbc (approved):
Yes, this is exactly what I'm expecting.

PR #946, opened 2023-01-18
fix: Ensure Arrow dictionaries are Pandas-compatibile

### LINKED ISSUE #945: Pandas cannot read Arrow dictionaries with unsigned index type
We encode categorical variables using PyArrow [`DictionaryArray`](https://arrow.apache.org/docs/python/generated/pyarrow.DictionaryArray.html). This means the values get encoded as an array of integer indices into a dictionary. When selecting the type for this integer array we choose the smallest type capable of representing all the necessary values:
https://github.com/opensafely-core/databuilder/blob/49f1477c4854e88dd6ffcba7cec26de434a5906d/databuilder/file_formats/arrow.py#L67-L73

However, PyArrow is unable to convert a `DictionaryArray` to a pandas [`Categorical`](https://pandas.pydata.org/docs/reference/api/pandas.Categorical.html) if its index type is an unsigned integer, giving an error like:
```
ArrowTypeError: Converting unsigned dictionary indices to pandas not yet supported, index type: uint8
```

This is marked as a "TODO" in PyArrow test code [here](https://github.com/apache/arrow/blob/4e439f6a597180c5fc8ff1552c860cecd33736c5/python/pyarrow/tests/test_pandas.py#L3758-L3762):
```py
        if index_type[0] == 'u':
            # TODO: unsigned dictionary indices to pandas
            with pytest.raises(TypeError):
                d1.to_pandas()
            continue
```
But I can't find a corresponding ticket. The closest one is this closed one which mentions other languages but not Python:
 * https://github.com/apache/arrow/issues/25354

In any case, the fix here is hopefully straightforward: we should always use signed integer types for dictionary indices.

In theory, this means that where categories number between 127 and 255 we'll be forced to use a two-byte `int16` where we could have got away with a one byte `uint8`. In practice I expect the number of categories to be much smaller than this anyway and for this change to have no memory usage impact.

evansd:
This means ensuring that we always used a signed integer type for a `DictionaryArray` index.

There's a theoretical loss of storage efficiency here (a categorical with 127 < N <= 255 categories will get encoded as `int16` rather than `uint8`) but we expect it to be rare to have that many categories, and being readable by Pandas is a key feature.

We don't currently have a mechanism for testing this within Data Builder so I tested manually with the following `test_dataset.py` definition:
```py
from databuilder.ehrql import Dataset
from databuilder.tables.beta import tpp

dataset = Dataset()
dataset.sex = tpp.patients.sex
dataset.set_population(tpp.patients.exists_for_patient())
```

Without the fix, generating a dataset file and loading it with Pandas fails:
```
databuilder generate-dataset test_dataset.py --output test.arrow
python -c 'import pandas; pandas.read_feather("test.arrow")'
```

```
...
pyarrow.lib.ArrowTypeError: Converting unsigned dictionary indices to pandas not yet supported, index type: uint8
```

After applying the fix the file is read successfully.

Closes #945

### DIFF
--- a/databuilder/file_formats/arrow.py
+++ b/databuilder/file_formats/arrow.py
@@ -65,7 +65,11 @@ def get_field_and_convertor(name, spec):
         type_ = PYARROW_TYPE_MAP[spec.type]()
 
     if spec.categories is not None:
-        index_type = smallest_int_type_for_range(0, len(spec.categories) - 1)
+        # Although pyarrow.dictionary indices can obviously never be negative we use
+        # `-1` as the minimum below so we always get a signed type; this is because
+        # Pandas can't read dictionaries with unsigned index types. See:
+        # https://github.com/opensafely-core/databuilder/issues/945
+        index_type = smallest_int_type_for_range(-1, len(spec.categories) - 1)
         value_type = type_
         type_ = pyarrow.dictionary(index_type, value_type, ordered=True)
         column_to_pyarrow = make_column_to_pyarrow_with_categories(
--- a/tests/integration/file_formats/test_arrow.py
+++ b/tests/integration/file_formats/test_arrow.py
@@ -27,6 +27,6 @@ def test_write_dataset_arrow(tmp_path):
     assert output_columns == list(column_specs.keys())
     assert output_rows == results
     assert categories == ["M", "F", "I"]
-    assert index_type == pyarrow.uint8()
+    assert index_type == pyarrow.int8()
     assert table.column("patient_id").type == pyarrow.int64()
     assert table.column("year_of_birth").type == pyarrow.uint16()


databuilder/file_formats/arrow.py
```diff
@@ -65,7 +65,11 @@ def get_field_and_convertor(name, spec):
         type_ = PYARROW_TYPE_MAP[spec.type]()
 
     if spec.categories is not None:
-        index_type = smallest_int_type_for_range(0, len(spec.categories) - 1)
+        # Although dictionary indices can obviously never be negative we use `-1` as the
```
iaindillingham, line 68: I might have misunderstood. I thought `dict` indices could be any hashable type, which includes negative integers:

```python
>>> d = {-1: "minus one"}
>>> d[-1]
'minus one'
```

iaindillingham (approved):
I tested this by building a `dev` image from the associated branch, and calling this from [opensafely/winter-pressures](https://github.com/opensafely/winter-pressures). The current image (v0.66.0) fails; the image built from the associated branch passes 🚀

databuilder/file_formats/arrow.py
evansd: Ah yeah, two different sorts of dictionary here: Python ones and PyArrow ones. The PyArrow dictionary index is just the offset into the dictionary array and so is necessarily positive. I can make the comment clearer.

PR #947, opened 2023-01-18
Add Comparison and Boolean operations to generative tests

### LINKED ISSUE #943: Add comparison operations to new generative tests
Add tests for:
```
    Function.LE,
    Function.NE,
    Function.GT,
    Function.LT,
    Function.GE,
    Function.And,
    Function.Or,
    Function.Not,
```

rebkwok:
Closes #943 

Adds tests for:
```
    Function.LE
    Function.NE
    Function.GT
    Function.LT
    Function.GE
    Function.And
    Function.Or
    Function.Not
```

### DIFF
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -157,10 +157,7 @@ def variable(patient_tables, event_tables, schema, value_strategies):
 
     @st.composite
     def add(draw, type_, frame):
-        return Function.Add(
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame)))),
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame)))),
-        )
+        return draw(binary_operation(type_, frame, Function.Add))
 
     def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -71,6 +71,14 @@ def variable(patient_tables, event_tables, schema, value_strategies):
             select_column: ({int, float, bool, datetime.date}, DomainConstraint.ANY),
             is_null: ({bool}, DomainConstraint.ANY),
             eq: ({bool}, DomainConstraint.ANY),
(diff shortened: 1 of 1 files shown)


tests/generative/variable_strategies.py
```diff
@@ -96,31 +104,78 @@ def select_column(type_, frame):
         column_names = [n for n, t in schema.column_types if t == type_]
         return st.builds(SelectColumn, st.just(frame), st.sampled_from(column_names))
 
+    @st.composite
+    def is_null(draw, _type, frame):
+        type_ = draw(any_type())
+        return Function.IsNull(
+            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
```
rebkwok, line 111: This one already existed and is (mostly) just reordered so the functions are in the same order they appear in `series_constraints`.  It also now passes `draw(one_row_per_patient_frame_or(frame))` to `series` instead of just `frame` - because I'd misunderstood what `one_row_per_patient_frame_or` was doing before and thought it was always generating one-row-per-patient

tests/generative/variable_strategies.py
```diff
@@ -114,13 +166,19 @@ def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
 
     @st.composite
-    def is_null(draw, _type, frame):
-        type_ = draw(any_type())
-        return Function.IsNull(draw(series(type_, frame)))
+    def operation(draw, type_, frame, operator_func):
```
evansd, line 166: I might be tempted to call this `binary_operation` or `binary_op` or something like that just because I tend to use "operation" to refer to _any_ query model node whereas this is specifically for the two-place ones.

evansd (approved):
Looks great. Factoring out the `operation` method is a good move. I've included a bit of bikeshedding over the name which you should feel free to ignore.

tests/generative/variable_strategies.py
rebkwok: Thanks - I couldn't work out what to call it :)

PR #948, opened 2023-01-18
Add Exists operation to generative tests

rebkwok:


tests/generative/variable_strategies.py
```diff
@@ -80,6 +80,7 @@ class DomainConstraint:
             le: ({bool}, DomainConstraint.ANY),
             ge: ({bool}, DomainConstraint.ANY),
             add: ({int, float}, DomainConstraint.ANY),
+            exists: ({bool}, DomainConstraint.PATIENT),
```
inglesp, line 83: Am no expert, but I'd have thought that `exists` (and maybe also `count`?) was simpler than most of the types above.

inglesp (approved):
By analogy with `count`, this looks right!

tests/generative/variable_strategies.py
evansd: Yeah, I think my instincts agree with Peter here that both `Exists` and `Count` produce a series directly from a frame which is generally going to be simpler than operations which combine different series to produce a new one.

I don't know if we've got a good workflow for deciding this kind of thing though. Do we just have to try with both and decide which set of outputs looks "better"?

evansd (approved):

tests/generative/variable_strategies.py
rebkwok: Happy to move them - I'm struggling to make decisions on which is simpler.  Ben put `count` after `add`, so I'd left `exists` there too, since it's similar.  Should we consider anything that takes only one series argument to be simpler than anything that takes two or more?  i.e. should `not_` and `negate` come before the other comparison and arithmetic operations?

evansd: I've no real idea either to be honest :) But now you say it, the single place operations _do_ seem simpler, don't they?

PR #959, opened 2023-01-20
Add Date operations to the generative tests

### LINKED ISSUE #950: Add date operations to new generative tests
Needs to add:
```
    Function.YearFromDate
    Function.MonthFromDate
    Function.DayFromDate
    Function.ToFirstOfYear
    Function.ToFirstOfMonth
    Function.DateAddYears
    Function.DateAddMonths
    Function.DateAddDays
    Function.DateDifferenceInYears
    Function.DateDifferenceInMonths
    Function.DateDifferenceInDays
```

rebkwok:
Adds:
```
    Function.YearFromDate
    Function.MonthFromDate
    Function.DayFromDate
    Function.ToFirstOfYear
    Function.ToFirstOfMonth
    Function.DateAddYears
    Function.DateAddMonths
    Function.DateAddDays
    Function.DateDifferenceInYears
    Function.DateDifferenceInMonths
    Function.DateDifferenceInDays
```

Also adds a new ignored error for sqlite, which was caused in one of my local runs by a many-nested query of date operations, and fixes the ignored error counting in the test recordings.

Also had to add the suppression of the too-many-filters health check.  It's still passing the comprehensive check with only ~300 examples, and producing around 275 combinations and 190 unique queries.

Fixes #950

### DIFF
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -135,7 +135,7 @@ def run_test(query_engines, data, variable, recorder):
     ]
 
     recorder.record_results(
-        len(results), len([r for r in results if r is IGNORE_RESULT])
+        len(results), len([r for (_, r) in results if r is IGNORE_RESULT])
     )
 
     for first, second in itertools.combinations(results, 2):
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -186,6 +186,9 @@ IGNORED_ERRORS = [
         sqlalchemy.exc.OperationalError,
         re.compile(".+Case expressions may only be nested to level 10.+"),
     ),
+    # SQLite raises a parser stack overflow error if the variable strategy generates queries
+    # that result in many nested queries
+    (sqlalchemy.exc.OperationalError, re.compile(".+parser stack overflow")),
     # OUT-OF-RANGE DATES
     # The variable strategy will sometimes result in date operations that construct
     # invalid dates (e.g. a large positive or negative integer in a DateAddYears operation
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -160,13 +160,11 @@ def variable(patient_tables, event_tables, schema, value_strategies):
         type_ = draw(any_type())
         return draw(binary_operation(type_, frame, Function.NE))
 
-    @st.composite
-    def and_(draw, type_, frame):
-        return draw(binary_operation(type_, frame, Function.And))
+    def and_(type_, frame):
+        return binary_operation(type_, frame, Function.And)
 
-    @st.composite
-    def or_(draw, type_, frame):
-        return draw(binary_operation(type_, frame, Function.Or))
+    def or_(type_, frame):
+        return binary_operation(type_, frame, Function.Or)
 
     @st.composite
     def lt(draw, _type, frame):
@@ -188,59 +186,37 @@ def variable(patient_tables, event_tables, schema, value_strategies):
         type_ = draw(comparable_type())
         return draw(binary_operation(type_, frame, Function.GE))
 
-    @st.composite
-    def add(draw, type_, frame):
-        return draw(binary_operation(type_, frame, Function.Add))
+    def add(type_, frame):
+        return binary_operation(type_, frame, Function.Add)
 
-    @st.composite
-    def subtract(draw, type_, frame):
-        return draw(binary_operation(type_, frame, Function.Subtract))
+    def subtract(type_, frame):
+        return binary_operation(type_, frame, Function.Subtract)
 
-    @st.composite
-    def multiply(draw, type_, frame):
-        return draw(binary_operation(type_, frame, Function.Multiply))
+    def multiply(type_, frame):
+        return binary_operation(type_, frame, Function.Multiply)
 
-    @st.composite
-    def date_add_years(draw, type_, frame):
-        return draw(
-            binary_operation_with_types(type_, int, frame, Function.DateAddYears)
-        )
+    def date_add_years(type_, frame):
+        return binary_operation_with_types(type_, int, frame, Function.DateAddYears)
 
-    @st.composite
-    def date_add_months(draw, type_, frame):
-        return draw(
-            binary_operation_with_types(type_, int, frame, Function.DateAddMonths)
-        )
+    def date_add_months(type_, frame):
+        return binary_operation_with_types(type_, int, frame, Function.DateAddMonths)
 
-    @st.composite
-    def date_add_days(draw, type_, frame):
-        return draw(
-            binary_operation_with_types(type_, int, frame, Function.DateAddDays)
-        )
+    def date_add_days(type_, frame):
+        return binary_operation_with_types(type_, int, frame, Function.DateAddDays)
 
-    @st.composite
-    def date_difference_in_years(draw, type_, frame):
-        return draw(
-            binary_operation(datetime.date, frame, Function.DateDifferenceInYears)
-        )
+    def date_difference_in_years(type_, frame):
+        return binary_operation(datetime.date, frame, Function.DateDifferenceInYears)
 
-    @st.composite
-    def date_difference_in_months(draw, type_, frame):
-        return draw(
-            binary_operation(datetime.date, frame, Function.DateDifferenceInMonths)
-        )
+    def date_difference_in_months(type_, frame):
+        return binary_operation(datetime.date, frame, Function.DateDifferenceInMonths)
 
-    @st.composite
-    def date_difference_in_days(draw, type_, frame):
-        return draw(
-            binary_operation(datetime.date, frame, Function.DateDifferenceInDays)
-        )
+    def date_difference_in_days(type_, frame):
+        return binary_operation(datetime.date, frame, Function.DateDifferenceInDays)
 
-    @st.composite
-    def binary_operation(draw, type_, frame, operator_func):
+    def binary_operation(type_, frame, operator_func):
         # A strategy for operations that take lhs and rhs arguments of the
         # same type
-        return draw(binary_operation_with_types(type_, type_, frame, operator_func))
+        return binary_operation_with_types(type_, type_, frame, operator_func)
 
     @st.composite
     def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -90,6 +90,9 @@ def variable(patient_tables, event_tables, schema, value_strategies):
             add: ({int, float}, DomainConstraint.ANY),
             subtract: ({int, float}, DomainConstraint.ANY),
             multiply: ({int, float}, DomainConstraint.ANY),
+            date_add_years: ({datetime.date}, DomainConstraint.ANY),
+            date_add_months: ({datetime.date}, DomainConstraint.ANY),
+            date_add_days: ({datetime.date}, DomainConstraint.ANY),
         }
         series_types = series_constraints.keys()
 
@@ -219,11 +222,35 @@ def variable(patient_tables, event_tables, schema, value_strategies):
         return draw(binary_operation(type_, frame, Function.Multiply))
 
     @st.composite
+    def date_add_years(draw, type_, frame):
+        return draw(
+            binary_operation_with_types(type_, int, frame, Function.DateAddYears)
+        )
+
+    @st.composite
+    def date_add_months(draw, type_, frame):
+        return draw(
+            binary_operation_with_types(type_, int, frame, Function.DateAddMonths)
+        )
+
+    @st.composite
+    def date_add_days(draw, type_, frame):
+        return draw(
+            binary_operation_with_types(type_, int, frame, Function.DateAddDays)
+        )
+
+    @st.composite
     def binary_operation(draw, type_, frame, operator_func):
         # A strategy for operations that take lhs and rhs arguments of the
         # same type
-        lhs = draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        rhs = draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
+        return draw(binary_operation_with_types(type_, type_, frame, operator_func))
+
+    @st.composite
+    def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
+        # A strategy for operations that take lhs and rhs arguments with specified lhs
+        # and rhs types (which may be different)
+        lhs = draw(series(lhs_type, draw(one_row_per_patient_frame_or(frame))))
+        rhs = draw(series(rhs_type, draw(one_row_per_patient_frame_or(frame))))
         return operator_func(lhs, rhs)
 
     def any_type():
@@ -311,9 +338,6 @@ known_missing_operations = {
     AggregateByPatient.Sum,
     Function.CastToFloat,
     Function.CastToInt,
-    Function.DateAddYears,
-    Function.DateAddMonths,
-    Function.DateAddDays,
     Function.DateDifferenceInYears,
     Function.DateDifferenceInMonths,
     Function.DateDifferenceInDays,
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -128,51 +128,28 @@ def variable(patient_tables, event_tables, schema, value_strategies):
     @st.composite
     def is_null(draw, _type, frame):
         type_ = draw(any_type())
-        return Function.IsNull(
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        )
+        return Function.IsNull(draw(series(type_, frame)))
 
-    @st.composite
-    def not_(draw, type_, frame):
-        return Function.Not(
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def not_(type_, frame):
+        return st.builds(Function.Not, series(type_, frame))
 
-    @st.composite
-    def year_from_date(draw, _type, frame):
-        return Function.YearFromDate(
-            draw(series(datetime.date, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def year_from_date(_type, frame):
+        return st.builds(Function.YearFromDate, series(datetime.date, frame))
 
-    @st.composite
-    def month_from_date(draw, _type, frame):
-        return Function.MonthFromDate(
-            draw(series(datetime.date, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def month_from_date(_type, frame):
+        return st.builds(Function.MonthFromDate, series(datetime.date, frame))
 
-    @st.composite
-    def day_from_date(draw, _type, frame):
-        return Function.DayFromDate(
-            draw(series(datetime.date, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def day_from_date(_type, frame):
+        return st.builds(Function.DayFromDate, series(datetime.date, frame))
 
-    @st.composite
-    def to_first_of_year(draw, _type, frame):
-        return Function.ToFirstOfYear(
-            draw(series(datetime.date, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def to_first_of_year(_type, frame):
+        return st.builds(Function.ToFirstOfYear, series(datetime.date, frame))
 
-    @st.composite
-    def to_first_of_month(draw, _type, frame):
-        return Function.ToFirstOfMonth(
-            draw(series(datetime.date, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def to_first_of_month(_type, frame):
+        return st.builds(Function.ToFirstOfMonth, series(datetime.date, frame))
 
-    @st.composite
-    def negate(draw, type_, frame):
-        return Function.Negate(
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        )
+    def negate(type_, frame):
+        return st.builds(Function.Negate, series(type_, frame))
 
     @st.composite
     def eq(draw, _type, frame):
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -247,20 +247,27 @@ def variable(patient_tables, event_tables, schema, value_strategies):
         # A strategy for operations that take lhs and rhs arguments with specified lhs
         # and rhs types (which may be different)
 
-        # we can combine this frame with a series drawn from:
-        # - a new one-row-per-patient-frame
-        # - this frame itself
-        # If it's already a one-row-per-patient-frame, it uses the database, so can
-        # be combined with a value or a series
-        # If it's not a one-row-per-patient-frame, avoid combining with a value so we
-        # don't end up with operations that don't hit the database
+        # A binary operation has 2 inputs, which are
+        # 1) A series drawn from the specified frame
+        # 2) one of:
+        #    a) A series drawn from the specified frame
+        #    b) A series drawn from a one-row-per-patient-frame
+        #    c) A Value
+
+        # Define other_frame and other_series (#2 above)
+        # pick either a one-row-per-patient-frame or this frame
         other_frame = draw(st.one_of(one_row_per_patient_frame(), st.just(frame)))
         if is_one_row_per_patient_frame(other_frame):
+            # if other_frame is patient-level, the input will either be a value or a
+            # series drawn from other_frame
             other_strategy = draw(st.sampled_from([value, series]))
         else:
+            # if other_frame is event-level, the input will be a series drawn from the
+            # other_frame
             other_strategy = series
 
-        # Pick the order of the lhs and rhs from the two frames and associated strategies
+        # Pick the order of the lhs and rhs inputs built from the two frames and
+        # associated strategies
         lhs_frame, lhs_strategy = draw(
             st.sampled_from([(frame, series), (other_frame, other_strategy)])
         )
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -73,6 +73,11 @@ def variable(patient_tables, event_tables, schema, value_strategies):
             count: ({int}, DomainConstraint.PATIENT),
             is_null: ({bool}, DomainConstraint.ANY),
             not_: ({bool}, DomainConstraint.ANY),
+            year_from_date: ({int}, DomainConstraint.ANY),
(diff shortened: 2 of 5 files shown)


tests/generative/variable_strategies.py
```diff
@@ -85,6 +90,12 @@ class DomainConstraint:
             add: ({int, float}, DomainConstraint.ANY),
             subtract: ({int, float}, DomainConstraint.ANY),
             multiply: ({int, float}, DomainConstraint.ANY),
+            date_add_years: ({datetime.date}, DomainConstraint.ANY),
+            date_add_months: ({datetime.date}, DomainConstraint.ANY),
+            date_add_days: ({datetime.date}, DomainConstraint.ANY),
+            date_difference_in_years: ({int}, DomainConstraint.ANY),
+            date_difference_in_months: ({int}, DomainConstraint.ANY),
+            date_difference_in_days: ({int}, DomainConstraint.ANY),
         }
```
rebkwok, line 99: I am not at all confident that these are in the right simplest-first order.  The year-from and to-first-of ones only take a single series input, but feel like date-related operations are probably more complex than comparisons, but negate can return more than one type which might make it more complex.  The date_add and date_diff ones take 2 series inputs and are dates so I've put them at the end, but :shrug:

tests/generative/test_query_model.py
```diff
@@ -135,7 +136,7 @@ def run_test(query_engines, data, variable, recorder):
     ]
 
     recorder.record_results(
-        len(results), len([r for r in results if r is IGNORE_RESULT])
+        len(results), len([r for (_, r) in results if r is IGNORE_RESULT])
```
benbc, line 139: Oooh. It would be interesting to know how much we can see ignored over a longish run with this fixed.

tests/generative/test_query_model.py
```diff
@@ -65,6 +65,7 @@
 settings = dict(
     max_examples=(int(os.environ.get("GENTEST_EXAMPLES", 100))),
     deadline=None,
+    suppress_health_check=[hyp.HealthCheck.filter_too_much],
```
benbc, line 68: Do you understand why this is needed? It's just the filtering to ensure that we actually hit the database that we're doing too much of?

tests/generative/test_query_model.py
```diff
@@ -63,7 +63,7 @@
     value_strategies,
 )
 settings = dict(
-    max_examples=(int(os.environ.get("GENTEST_EXAMPLES", 100))),
+    max_examples=(int(os.environ.get("GENTEST_EXAMPLES", 200))),
```
benbc, line 66: Do I gather that you found it was sometimes hitting as many as two errors in 100 examples, but you haven't ever seen three or more over 200? This seems like it's right on the border, so it's bound to be a bit flaky isn't it? Maybe 1% is too tight a constraint.

benbc (commented):
Just looking at this for interest. I haven't done a thorough review.

tests/generative/test_query_model.py
rebkwok: I think it must be - that's the only filtering we have.  I'd guessed that it was just because there are more possible operations now so it generates more that don't hit the database, but I may have misunderstood.  It's still managing to pass the comprehensive test with only about 300 examples

tests/generative/test_query_model.py
rebkwok: Locally it was doing OK with 200, but it's failed again on that check.  I didn't want to make the overall test runs slower by upping the examples too much; maybe it's better to have make the constraint dependent on the number of examples, so it's e.g. 5% if GENTEST_EXAMPLES is < 500

tests/generative/test_query_model.py
benbc: > It's still managing to pass the comprehensive test with only about 300 examples

Oh, that is cool. What an improvement!

benbc: > I'd guessed that it was just because there are more possible operations now so it generates more that don't hit the database

Yes, it kind of must be that, mustn't it, but it seem surprising (to me) because you haven't introduced any new non-database terminal nodes. Have you looked at sample of the output queries? I think it's worth doing that from time to time just to make sure you're not missing something. For example if we weren't able to generate `date`-typed `SelectColumn` nodes for some reason, then all the `date`-typed terminals would be `Value` nodes -- and that might result in more filtering.

tests/generative/test_query_model.py
benbc: Or turn off that check for the build and only have it on for the long-running overnight tests.

tests/generative/test_query_model.py
evansd: I wonder if what's happening here is that we're over-using `one_row_per_patient_frame_or(frame))`. Each time we do that we give Hypothesis the chance to "drop down" from an event frame to patient frame; and the more operations we stack up that use that function, the greater the chance that it will eventually choose to do so. Once it's got a patient frame it's highly incentivised to choose a `Value` (the simplest of all possible operations).

I think it might be worth only using `one_row_per_patient_from_or` for binary operations where it's somewhat interesting to combine event and patient frames in a single operation. For unary operations I think it's pointless: if we'd wanted a patient frame we could have just chosen a patient frame in the first place.

evansd: Expanding this thought further: once Hypothesis has chosen the domain/frame for a series I don't think we should be giving it the option to change its mind. This means that for unary operations we should just pass the frame straight through. For binary operations, we can always use a patient frame _in addition to_, but not instead of, the original frame without changing the domain. So we want something like:
```py
@st.composite
def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
    other_frame = draw(one_row_per_patient_frame_or(frame))
    # I don't know off the top of my head how to do this in Hypothesis, but I
    # assume it's possible
    lhs_frame, rhs_frame = made_up_funcs.shuffle([frame, other_frame])
    lhs = draw(series(lhs_type, lhs_frame))
    rhs = draw(series(rhs_type, rhs_frame))
    return operator_func(lhs, rhs)
```

rebkwok: I think this makes sense to me (the not using `one_row_per_patient_frame_or` everywhere, at least).  

> For binary operations, we can always use a patient frame in addition to, but not instead of, the original frame without changing the domain

Can you explain this a bit more?  I'm not sure I fully understand what domains are.

rebkwok: I changed all `one_row_per_patient_frame_or` in unary operators to just pass on the original frame, and changed the function above like this (just using random to shuffle the two frames):
```
    @st.composite
    def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
        other_frame = draw(one_row_per_patient_frame_or(frame))
        frames = [frame, other_frame]
        random.shuffle(frames)
        lhs_frame, rhs_frame = frames
        lhs = draw(series(lhs_type, lhs_frame))
        rhs = draw(series(rhs_type, rhs_frame))
        return operator_func(lhs, rhs)
```
And it still complains that it's filtering out too much with 100 or 200 examples, but with 300 it's OK. 
**However** I've realised that the same is true before I made these changes - i.e. with `one_row_per_patient_frame_or` used everywhere, it complains about the filtering with 200 examples, but not with 300.

 If the binary operator also just uses the frame it's passed for both lhs and rhs frames, it doesn't complain with even 100 examples (which is to be expected, I guess).

evansd: Yeah, this is pretty much the hardest bit to understand of the query model — and the bit which took us the longest to work out.

You can think of a series' domain as the answer to the question "which other series can I compose this with?". For a many-rows-per-patient series, its domain is the frame its values were drawn from. For all one-row-per-patient series (which include static values) it's a sort of shared base, or root, domain we call the "patient domain".

The basic rule is that you can't combine different many-rows-per-patient domains, but you _can_ combine a many-rows-per-patient domain with a one-row-per-patient domain.
 This is what's behind the `one_row_per_patient_or(frame)` strategy: it allows us to test mixing a many-rows-per-patient frame with a one-row-per-patient one.

But if we call it twice we might end up throwing away the original frame, and I think that's part of the problem here. So my suggestion was to always keep the original frame and then decide if we're going to combine it with a series drawn from the same frame or whether to mix it with a one-row-per-patient one.

Does that make any sense?

benbc: I was conscious when I wrote this that the implementation would "demote" the domain to the patient domain some proportion of the time. I didn't like it all that much, but since it was valid I left it be.

But you're right, Dave, that

1) it's not really consistent with contract of the concrete strategies (because the demotion is observable from outside, rather than just being an implementation detail); but it would be (consistent and just an implementation detail) if only _one_ of the inputs to a binary operator was chosen to be demoted
2) it could be causing this problem

So I agree that the right thing to do is drop the demotion for unary operators and never demote both of the inputs for binary operators.

evansd: I'm not sure how this is going to behave with `random.shuffle` as it's not under Hypothesis' control and is non-deterministic. I think you might be able to do it with:
```py
swap = draw(st.booleans())
lhs_frame, rhs_frame = (frame, other frame) if not swap else (other_frame, frame)
```

benbc: I guess that Becky's latest findings support the general thrust here:

> If the binary operator also just uses the frame it's passed for both lhs and rhs frames, it doesn't complain with even 100 examples

But it's a bit sad that the approach of only demoting one and only sometimes doesn't fix the problem.

I suppose we could try biasing the distribution further (trivially but clunkily by putting more than one instance of `frame` into the shuffled list).

benbc: A couple of points on the new implementation as it stands.

Can we inline `one_row_per_patient_frame_or()`, if that's the only place it's being used now? It will make it much clearer what's going on, I think. I think that will also allow a clearer implementation where we chose between three options for the inputs: `frame, patient_frame`, `patient_frame, frame` or `frame, frame`.

You shouldn't use `random` because it clashes with Hypothesis' reproducibility needs. But Hypothesis has a replacement for it that you can use if you need. Here I suspect that you don't and you can compose what you need out of existing strategies instead, like the use of `one_of()` in `one_row_per_patient_frame_or()`.

evansd: Ooh, further thought here: the root problem is that Hypothesis is too keen to pick `Values` (because they're simple) but by doing so it ends up building expressions which don't touch the database at all.

What if we said that the _only_ place it should generate `Values` is in combination with series which aren't `Values`? I don't think that's overlay restrictive: operations that only involve static values are pointless; we can resolve them statically before compiling to SQL if we want to.

Changing this would involve taking `value` out of the `series_constraints` dict and then amending the logic in `binary_operation_with_types` to something that chooses whether we're combining the original frame with either itself, or a new one-row-per-patient frame, or a value; and then draws series accordingly.

I think if we did that we might even be able to do away with the "uses database" check because we'd never generate examples which failed it.

evansd: Yes, you're right about `Case` and `In`. I think any operation involving more than one argument needs some generalisation of the logic for binary operations.

Just experimenting with what Hypothesis lets us do, I think something like this might work (though it's not particularly elegant):
```py
@st.composite
def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
    other_frame = draw(st.one_of(one_row_per_patient_frame(), st.just(frame)))
    if is_one_row_per_patient_frame(other_frame):
        other_strategy = draw(st.sampled_from([value, series]))
    else:
        other_strategy = series
    (lhs_frame, lhs_strategy), (rhs_frame, rhs_strategy) = st.permutations(
        [(frame, series), (other_frame, other_strategy)]
    )
    lhs = draw(lhs_strategy(lhs_type, lhs_frame))
    rhs = draw(rhs_strategy(rhs_type, rhs_frame))
    return operator_func(lhs, rhs)
```

rebkwok: I've updated this PR now with a version of @evansd's suggestion above (had to use an alternative to `st.permutations` because you can't unpack a a strategy), and it seems to be doing the right thing - no need for the filter check suppression, and it's still passing the comprehensive tests with 300 examples.

From the last run (300 examples):
```
247 unique input combinations

173 unique queries

with a total of 1564 nodes

with this node count distribution
  2	18
  3	7
  4	20
  5	22
  6	19
  7	26
  8	5
  9	10
 10	4
 11	5
 12	5
 13	2
 14	3
 15	5
 16	1
 18	5
 19	3
 20	1
 22	2
 23	1
 25	1
 26	1
 27	1
 30	2
 44	1
 45	2
 71	1

and these node types
Value                    387
SelectPatientTable       257
SelectColumn             246
Multiply                 68
Subtract                 65
Add                      57
ToFirstOfMonth           56
DateAddMonths            46
Negate                   46
Or                       45
DateAddYears             35
LE                       34
DateAddDays              31
ToFirstOfYear            31
DateDifferenceInMonths   27
MonthFromDate            16
DateDifferenceInDays     15
DateDifferenceInYears    15
YearFromDate             14
Count                    13
GT                       10
LT                       9
And                      7
DayFromDate              6
SelectTable              5
GE                       4
IsNull                   4
Not                      4
EQ                       3
Exists                   3
NE                       3
Filter                   2

166 unique datasets

with this size distribution
  0	1
  1	41
  2	7
  3	5
  4	7
  5	5
  6	6
  7	5
  8	2
  9	9
 10	12
 11	3
 12	5
 13	8
 14	13
 15	13
 16	3
 17	5
 19	2
 20	2
 21	5
 22	6
 26	1

7 errors ignored
```

tests/generative/variable_strategies.py
```diff
@@ -128,51 +128,35 @@ def count(_type, _frame):
...
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        )
+        return Function.IsNull(draw(series(type_, frame)))
 
     @st.composite
     def not_(draw, type_, frame):
-        return Function.Not(
-            draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        )
+        return Function.Not(draw(series(type_, frame)))
```
benbc, line 135: When you've only got one `draw()` left (here and elsewhere) you don't need a `@st.composite`.

```
def not_(type_, frame):
    return st.builds(Function.Not, series(type_, frame))
```

tests/generative/test_query_model.py
benbc: I don't think the logic is quite right (or it is right but it's unclear so I'm misunderstanding it).

I think that the logic should be that of the two inputs to a binary operator:
* one input should always be an arbitrary series on the specified frame
* the other input should be _either_ an arbitrary series on the specified frame _or_ an arbitrary series on a patient-domain frame _or_ a `Value`

(With the caveat that if the frame is already patient-level then that second case collapses to just two possibilities.)

And we remove `Value` from the list of arbitrary series.

rebkwok: > I think that we may need to allow for values Case as well, and In in case that is given special treatment.

I thought this meant that we *shouldn't* remove `value` from the `series_constraints`.  Did you mean that Case/In (when we get to them) need to be treated differently to other binary operators?

evansd: > it seems to be doing the right thing

Well that's encouraging.


> had to use an alternative to `st.permutations` because you can't unpack a strategy

Ah yeah, I find it so hard to keep it straight in my head whether I'm dealing with a strategy or the result of a strategy. I think you might be able to do with with `draw(st.permutations(...))` but what you've got does the trick as well.


> I don't think the logic is quite right (or it is right but it's unclear so I'm misunderstanding it).

I'm 80% confident it's right, and 100% confident it's unclear :smile: 

I think it's equivalent to what you say, but instead of picking directly from the trio `(original-frame, patient-frame, value)` it first picks from `(original-frame, patient-frame)` and then splits the second branch into `(patient-frame, value)`.

I've just experimented with some alternatives but it's quite tricky to express this perspicaciously in Hypothesis.

I don't think this comment's quite right though:
```
        # we can combine this frame with a series drawn from:
        # - a new one-row-per-patient-frame
        # - this frame itself
        # If it's already a one-row-per-patient-frame, it uses the database, so can
        # be combined with a value or a series
        # If it's not a one-row-per-patient-frame, avoid combining with a value so we
        # don't end up with operations that don't hit the database
```

rebkwok: > I think that the logic should be that of the two inputs to a binary operator:
> 
>  - one input should always be an arbitrary series on the specified frame
>  - the other input should be either an arbitrary series on the specified frame or an arbitrary series on a patient-domain frame or a Value

I think this is what's happening.  I'll see if I can make it clearer.

evansd: > Did you mean that Case/In (when we get to them) need to be treated differently to other binary operators?

Well I don't think of `Case` as a binary operator. I guess it does take two arguments, but one of those is a dictionary with arbitrary many series in it. So we're going to have to generalise the logic for binary operators somehow. I think this means that we'll need to pick at least one "condition" series to use the original frame, and then the rest can use the original frame, or a one-row-per-patient frame, or a value.

`In` is a bit different because we don't currently have a way of specifying anything _other_ than a `Value` on the RHS.

But I propose worrying about both of these when we come to them. Either way this means removing `value` from `series_constraints`.

rebkwok: On second thoughts, I do think the logic is what we want.  I've tried to improve the comment.

benbc: > I assume you mean that if it's already patient-level then the only 2 options are a series on the specified frame or a value. Why could it not also be a series on another patient-level frame?

@rebkwok You're quite right about this.

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
     def multiply(draw, type_, frame):
         return draw(binary_operation(type_, frame, Function.Multiply))
 
+    @st.composite
+    def date_add_years(draw, type_, frame):
+        return draw(
+            binary_operation_with_types(type_, int, frame, Function.DateAddYears)
+        )
```
benbc, line 207: No need for `@st.composite` here or in any of the functions below that do `return draw(...)`. If there is only one `draw()`, it can usually be eliminated. In this case you can write it as:

```
    def date_add_years(draw, type_, frame):
        return binary_operation_with_types(type_, int, frame, Function.DateAddYears)
```

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
...
+    @st.composite
+    def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
+        # A strategy for operations that take lhs and rhs arguments with specified lhs
+        # and rhs types (which may be different)
+
+        # A binary operation has 2 inputs, which are
+        # 1) A series drawn from the specified frame
+        # 2) one of:
+        #    a) A series drawn from the specified frame
+        #    b) A series drawn from a one-row-per-patient-frame
```
benbc, line 254: Maybe "any one-row-per-patient frame" to make it clear that it can be different in the case where such a thing is passed in?

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
...
+        # pick either a one-row-per-patient-frame or this frame
+        other_frame = draw(st.one_of(one_row_per_patient_frame(), st.just(frame)))
+        if is_one_row_per_patient_frame(other_frame):
+            # if other_frame is patient-level, the input will either be a value or a
+            # series drawn from other_frame
+            other_strategy = draw(st.sampled_from([value, series]))
+        else:
+            # if other_frame is event-level, the input will be a series drawn from the
+            # other_frame
+            other_strategy = series
```
benbc, line 267: Now I think that the comment is correct and this logic may be wrong :-). I don't think that the series strategy should vary based on whether the frame is one-row-per-patient or not, should it? There's no reason not to combine a one-row-per-patient frame with a value.

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
...
+    def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
+        # A strategy for operations that take lhs and rhs arguments with specified lhs
+        # and rhs types (which may be different)
+
+        # A binary operation has 2 inputs, which are
+        # 1) A series drawn from the specified frame
+        # 2) one of:
+        #    a) A series drawn from the specified frame
+        #    b) A series drawn from a one-row-per-patient-frame
+        #    c) A Value
```
benbc, line 255: Great. This comment is v clear now.

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
...
+
+        # Define other_frame and other_series (#2 above)
+        # pick either a one-row-per-patient-frame or this frame
```
benbc, line 258: This comment refers to `other_series`, which I think is correct and clearer than `other_strategy` which is used in the code.

tests/generative/variable_strategies.py
```diff
@@ -183,12 +200,84 @@ def subtract(draw, type_, frame):
...
+
+        # Pick the order of the lhs and rhs inputs built from the two frames and
+        # associated strategies
+        lhs_frame, lhs_strategy = draw(
+            st.sampled_from([(frame, series), (other_frame, other_strategy)])
+        )
+        if lhs_frame == frame:
+            rhs_frame, rhs_strategy = other_frame, other_strategy
+        else:
+            rhs_frame, rhs_strategy = frame, series
```
benbc, line 277: It _might_ be clearer here to draw from two four-tuples rather than having the switching logic after the draw.

benbc (commented):
Sorry for being picky. I do think this is getting better and better.

tests/generative/variable_strategies.py
rebkwok: But other_series could be a value, which is confusing.

I've just rewritten this function to pick the input type (series or value) first, which I think makes it clearer, and still does the right thing:

```
    @st.composite
    def binary_operation_with_types(draw, lhs_type, rhs_type, frame, operator_func):
        # A strategy for operations that take lhs and rhs arguments with specified lhs
        # and rhs types (which may be different)

        # A binary operation has 2 inputs, which are
        # 1) A series drawn from the specified frame
        # 2) one of:
        #    a) A series drawn from the specified frame
        #    b) A series drawn from a one-row-per-patient-frame
        #    c) A Value

        # first pick a input strategy, either a value or a series
        other_input = draw(st.sampled_from([value, series]))
        if other_input == value:
            # A value strategy ignores its frame input, so we can give it anything here
            other_frame = None
        else:
            # The input is a series, so now pick a frame for it to be drawn from
            # The other frame will either be a new one-row-per-patient-frame or this frame
            other_frame = draw(st.one_of(one_row_per_patient_frame(), st.just(frame)))

        # Pick the order of the lhs and rhs inputs built from the two frames and
        # associated strategies
        lhs_frame, lhs_input, rhs_frame, rhs_input = draw(
            st.sampled_from(
                [
                    (frame, series, other_frame, other_input),
                    (other_frame, other_input, frame, series)
                ]
            )
        )
        lhs = draw(lhs_input(lhs_type, lhs_frame))
        rhs = draw(rhs_input(rhs_type, rhs_frame))

        return operator_func(lhs, rhs)
```

tests/generative/variable_strategies.py
evansd: You can also do it like this:
```py
(lhs_frame, lhs_strategy), (rhs_frame, rhs_strategy) = draw(
    st.permutations([(frame, series), (other_frame, other_strategy)])
)
```
(I forgot the `draw` when I originally suggested `permutations` which is why it didn't work.)

tests/generative/variable_strategies.py
benbc: I think that's clear and correct and I'd be happy for you to merge it as-is.

But of course I have some thoughts...

* A `Value` _is_ a `Series`, it's just not one that can be created by the `series` strategy. This is a point where our generation code does not align directly with the type structure of the query model because the query model can express stuff which is both unrealistic and unhelpful. We have similar non-generality in the `Filter`-construction strategy.
* While it's correct that `Value` will ignore its input frame, there are plenty of other places where we knowingly pass ignored values to series strategies (for example type to `eq()` and frame to `count()`). I'm not sure that avoiding doing that here is worth the extra branching that it requires (but I'm quite happy to be wrong).

rebkwok: So it can be reduced to this?
```
        # first pick an "other" input series (i.e. #2 above), either a value series 
        # or a series drawn from a frame
        other_series = draw(st.sampled_from([value, series]))
        # Now pick a frame for the series to be drawn from
        # The other frame will either be a new one-row-per-patient-frame or this frame
        # (Note if the other_series is a value, the frame will be ignored)
        other_frame = draw(st.one_of(one_row_per_patient_frame(), st.just(frame)))
```
Also, yes, of course, I was thinking of value and series as two different things.

tests/generative/variable_strategies.py
```diff
@@ -171,24 +186,72 @@ def ge(draw, _type, frame):
...
+    def date_difference_in_days(type_, frame):
+        return binary_operation(datetime.date, frame, Function.DateDifferenceInDays)
 
     @st.composite
     def binary_operation(draw, type_, frame, operator_func):
         # A strategy for operations that take lhs and rhs arguments of the
         # same type
-        lhs = draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
-        rhs = draw(series(type_, draw(one_row_per_patient_frame_or(frame))))
+        return draw(binary_operation_with_types(type_, type_, frame, operator_func))
```
benbc, line 220: Still one excess `draw()` here. (Sorry!)

tests/generative/variable_strategies.py
```diff
@@ -171,24 +186,72 @@ def ge(draw, _type, frame):
...
+        # Pick the order of the lhs and rhs inputs built from the two frames and
+        # associated strategies
+        lhs_frame, lhs_input, rhs_frame, rhs_input = draw(
+            st.sampled_from(
+                [
+                    (frame, series, other_frame, other_series),
+                    (other_frame, other_series, frame, series),
+                ]
+            )
+        )
```
benbc, line 251: Or see Dave's alternative suggestion.

benbc (approved):

tests/generative/variable_strategies.py
rebkwok: Unless it's more efficient, I prefer this one for clarity

benbc: :+1:

rebkwok:
@benbc @evansd Failing coverage reminded me that we no longer need the `uses_the_database` filter.  Just noting here as I've removed that after Ben's :heavy_check_mark:

Issue #967, opened 2023-01-24
Generated documentation Markdown needs to be manually updated


StevenMaude:
Following the documentation move to this repository in #938.

In the documentation repository, there were automatic pull requests to update this file. At present, developers would have to rebuild the file themselves to get the CI check to pass. 

But perhaps that's acceptable following a change to the docs, for now.

Another improvement would be to ditch this file entirely.

StevenMaude:
Note: we're not currently enforcing the documentation checks yet, so developers don't need to have updated the `public_docs.json` to get their PRs to pass. In future, we could make this a required check to help keep the `public_docs.json` current.

evansd:
> Another improvement would be to ditch this file entirely.

This feels like the solution to me. I think we only went down this route to try to produce a static artefact that the docs site could consume. Now that it's all in the repo I'd hope we could do without it. I'm sure that will require some re-wiring, but hopefully it will all end up simpler.

StevenMaude:
As of #985, this will be instead that `just generate-docs` need to be run manually when changing relevant code, but this is still a simpler overall process.

That might be acceptable for the time being.

Potentially this could be a pre-commit hook, perhaps?

Issue #966, opened 2023-01-24
Handle updating of the Docker image in the tutorial `project.yaml`


StevenMaude:
As part of the documentation move (#808), we moved the current ehrQL tutorial dataset definition examples to this repository.

Previously, the version was pinned by the Docker image version specified in `ehrql-tutorial-examples/project.yaml` and the appropriate `requirements` file.

On moving to this repository, we now effectively unpin Data Builder as a requirement, and use the current repository version when performing tasks like validating the dataset definitions work. This is likely to go out of sync with a pinned Docker version.

One example of a problem would be the following conditions are met:

1. Some new feature is in the repository version of Data Builder; for example in a PR branch.
2. That same feature is used in the dataset definitions.
3. The most recent Docker image is based on the last release on `main` and will not include the feature.

The "check documentation" CI will then always fail. 

One workaround for that could be to build the same Docker image as part of CI, and substituting that version into `project.yaml` to ensure that we use the CI-built image. This gets the branch CI passing. But the image version specified would still need a resync following a release.

(Even though this is a problem, the benefit of this new way is that the documentation will then reflect the latest version of Data Builder. For example, the existing examples used `subtract_days()` which has been removed from Data Builder.)

StevenMaude:
On the updating side, maybe this could be a workflow that finds and replaces the relevant versions, and opens a PR, triggered by a new tag push?

StevenMaude:
We should reintroduce the check that was removed in 5403c43fc5f15117126fff02743289cb82b49107.

PR #968, opened 2023-01-24
Make the docs CI pass for now

StevenMaude:
This PR implements a couple of fixes that get the docs CI passing for now.

This is still being run manually for the time being.

It also flags some other issues following the docs migration to this repository:
* #966
* #967


StevenMaude:
The documentation check, which is not yet mandatory, is passing as of ebf85c8:

documentation-check-ebf85c8

Because of the reasons given in #966 and #967, it may fail again soon-ish, but this PR includes a couple of required fixes anyway.

rebkwok (approved):
If the documentation check workflow is now passing, I'd suggest making it trigger on PR/push to make sure the public_docs.json is kept up to date. Or if it's going to be flakey on the tutorial examples, maybe pull the public docs to run alongside the main tests for now.  

I don't think it's a burden to expect developers who make changes that change the specs/contracts/backends to have to update and commit the public_docs.json file though.

PR #971, opened 2023-01-25
Dates of death are not rounded to first of month

evansd:
This removes an incorrectly applied constraint from the `date_of_death` column and tidies up some redundant and misleading column descriptions.

### DIFF
--- a/databuilder/contracts/universal.py
+++ b/databuilder/contracts/universal.py
@@ -10,15 +10,12 @@ class patients(PatientFrame):
 
     date_of_birth = Series(
         datetime.date,
-        description=(
-            "Patient's year and month of birth, provided in format YYYY-MM-01. "
-            "The day will always be the first of the month."
-        ),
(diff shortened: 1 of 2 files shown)


rebkwok (approved):
Why don't we need the column descriptions any more? Are they coming from somewhere else?

Also, I suspect this PR was opened before the new "check docs" workflow that looks for changes in the public_docs.json and will fail on merge, so you might want to preempt it :)

evansd:
Ah thanks @rebkwok, I'll rebuild now.

On the descriptions, any constraints applied (like "this is rounded to the first of the month", or "this must be one of these specific values") get rendered into the docs and I felt like the descriptions were mostly just repeating those. And then all that was left was things like "date_of_birth -> Patient's date of birth" which felt a bit pointless.

But thinking about it more, maybe it's worth having those in by way of example, and to avoid having blank spaces in the docs. So I think I've talked myself into re-adding them :)

Issue #972, opened 2023-01-25
Extract extra CIS data for ONS study


rebkwok:
External researchers at ONS have begun [a cohort-extractor study on mental health](https://github.com/opensafely/MH_pandemic/).  They would like to access new data in the `ONS_CIS_New` table.  This table is now included in the [TPP schema report](https://reports.opensafely.org/reports/opensafely-tpp-database-schema/).

We have agreed that we will use ehrQL to extract the data they need into a separate output file, which they will then be able to join to their existing outputs.

They have asked for the following fields:

    Participant_id (not new but to join to the other data)
    visit_num
    nhs_date_share
    last_linkage_dt or it could be called final_linkage_date

There may be multiple records per patient in the `ONS_CIS_New` table.  We will have to check how these records are aggregated to produce the one-row-per-patient output, and ensure that the output from ehrQL matches this logic.

If this is too fiddly, we shouldn't rule out adding the new fields they're asking for to cohort-extractor.  I'd initially understood that they wanted access to the new extra CIS data, which would have made this much harder.

I will create a related issue to update cohort-extractor to use `ONS_CIS_New` instead of `ONS_CIS`.

PR #973, opened 2023-01-25
Implement additional ONS_CIS columns

rebkwok:
See #972 

- Use ONS_CIS_New table instead of ONS_CIS (ONS_CIS_New should contain the same data as ONS_CIS, but with some additional columns)
- Add the columns that are needed for the ONS mental health study:
    - visit_date
    - visit_num
    - nhs_data_share
    - last_linkage_dt

`nhs_data_share` is an int column in the table, but ONS have confirmed that it can take one of two values, 0 ("opt-in") or 1("opt-out").  The `ons_cis` table is a QueryTable that converts the `nhs_data_share` column to its corresponding label.  This avoids issues with 0 values which actually represent a data value, and makes the output similar to cohort-extractor.  The ONS mental health study will combine ons cis outputs from cohort-extractor and databuilder, so we'll want them to be consistent in the way they output.

### DIFF
--- a/databuilder/tables/beta/tpp.py
+++ b/databuilder/tables/beta/tpp.py
@@ -17,6 +17,7 @@ __all__ = [
     "emergency_care_attendances",
     "hospital_admissions",
     "appointments",
+    "ons_cis",
 ]
 
 
@@ -193,3 +194,6 @@ class ons_cis(EventFrame):
     """
 
     visit_date = Series(datetime.date)
+    visit_num = Series(int)
+    nhs_data_share = Series(int)
+    last_linkage_dt = Series(datetime.date)
--- a/databuilder/tables/beta/tpp.py
+++ b/databuilder/tables/beta/tpp.py
@@ -195,5 +195,5 @@ class ons_cis(EventFrame):
 
     visit_date = Series(datetime.date)
     visit_num = Series(int)
-    nhs_data_share = Series(int)
+    is_opted_out_of_nhs_data_share = Series(bool)
     last_linkage_dt = Series(datetime.date)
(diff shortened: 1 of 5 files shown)


databuilder/tables/beta/tpp.py
```diff
@@ -195,5 +195,5 @@ class ons_cis(EventFrame):
 
     visit_date = Series(datetime.date)
     visit_num = Series(int)
-    nhs_data_share = Series(int)
+    nhs_data_share = Series(str)
```
evansd, line 198: This should have a categorical constraint if we're going to model it as a categorical. Something like:
```py
nhs_data_share = Series(
    str,
    constraints=[
        Constraint.Categorical(["opt-out", "opt-in"]),
    ],
)
```

But I wonder if this wouldn't be better modelled as a boolean. Something like `is_opted_in_to_nhs_data_share`?

evansd (approved):

databuilder/tables/beta/tpp.py
rebkwok: Yes, a boolean probably makes sense.  I wonder if that would make it confusing for the ONS team who are using a mix of cohort-extractor and databuilder?
I guess we can ask them

evansd: Well that's neat! (I think you meant "0=opt-in" above, right?)

rebkwok: yes :facepalm:

rebkwok: I will check the db to make sure there really are only 0, 1 or null in that column first though

rebkwok: Confirmed :point_up:

evansd (approved):

PR #980, opened 2023-01-26
Restructure docs directory

StevenMaude:
This is to keep the actual Markdown MkDocs content entirely separated from other resources (example: ehrQL tutorial dataset definitions).

One improvement here is that the `ehrql-tutorial-examples` README doesn't get included as an actual documentation site page, which [it currently does](https://docs.opensafely.org/data-builder/ehrql-tutorial-examples/).

It also changes `just generate-docs` to `just docs-generate` because the inconsistency with the other `just` recipes was bugging me.


StevenMaude:
I still need to check this against the documentation site build.

StevenMaude:
Will close this for now, and revisit in the future, as it would need extensive changes to merge, and isn't finished.

PR #983, opened 2023-01-27
Notify Slack on generative test failures

### LINKED ISSUE #957: Automatically create issue when generative tests fail in CI
Using [JasonEtco/create-an-issue](https://github.com/JasonEtco/create-an-issue) or another suitable action, we should create an issue against databuilder, whenever the generative tests fail in CI.

The issue should be called something like "Generative test failure on ${date}", and should contain a link to the failed CI run.

Bonus points:

* post a notification in #team-data on Slack
* add the issue to the Next column in the data team project board

In both cases, consider whether this is worth the effort before getting too bogged down here!

evansd:
I created a new Slack app for this purpose, Data Builder Bot, because I wanted far more limited privileges than Bennett Bot has. I just want to be able to write to channels and not read anything. It's possible that we'll have a more general need for a write-only bot, in which case we can (hopefully) rename this to something less repo-specific.

Closes #957

### DIFF
--- a/.github/workflows/generative-tests.yml
+++ b/.github/workflows/generative-tests.yml
@@ -22,3 +22,13 @@ jobs:
           GENTEST_DEBUG=t \
           GENTEST_CHECK_IGNORED_ERRORS=t \
           just test-generative
+
+      - name: "Notify Slack on Failure"
+        if: failure()
+        uses: voxmedia/github-action-slack-notify-build@3665186a8c1a022b28a1dbe0954e73aa9081ea9e # v1.6.0
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        with:
+          channel_id: C03FB777L1M
+          status: "Generative Test Failure"
+          color: danger


.github/workflows/generative-tests.yml
```diff
@@ -22,3 +22,13 @@ jobs:
           GENTEST_DEBUG=t \
           GENTEST_CHECK_IGNORED_ERRORS=t \
           just test-generative
+
+      - name: "Notify Slack on Failure"
+        if: failure()
+        uses: voxmedia/github-action-slack-notify-build@3665186a8c1a022b28a1dbe0954e73aa9081ea9e
```
ghickman, line 28: Might be useful to put the version number after this to save us searching in the future, we've done `  # vN.N.N` elsewhere and dependabot typically updates that too!

ghickman (approved):

.github/workflows/generative-tests.yml
evansd: Oh nice! I'll do that now

PR #985, opened 2023-01-27
Incorporate mkdocs-opensafely-databuilder into databuilder repo

### LINKED ISSUE #977: Improve how mkdocs-opensafely-databuilder plugin is managed
Relates to opensafely/documentation#898.

We have to manage this as a dependency, but as it's not published on PyPI, we have to manually bump the version both here and in the documentation repository.

rebkwok:
Fixes #977 

This removes the dependency on the mkdocs-opensafely-databuilder plugin, and moves the rendering code (and tests) from the plugin into this repo, with minimal changes.

the `generate-docs` command now generates the json as before, and then passes it through the renderers to generate the markdown files to be included in the docs.  The previous markers for the mkdocs-opensafely-databuilder plugin are just replaced with snippet markers.

Note that the public_docs.json file is not deleted yet, to avoid breaking the main docs when this PR is merged.  After this one is merged, the mkdocs-opensafely-databuilder can be removed from the main docs repo, and the public_docs.json file will no longer be required.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -229,7 +229,7 @@ update-external-studies: devenv
     $BIN/python -m tests.acceptance.update_external_studies
 
 docs-serve: devenv generate-docs
-    BACKEND_DOCS_FILE=docs/public_docs.json "$BIN"/mkdocs serve
+    "$BIN"/mkdocs serve
 
 # Run the snippet tests
 docs-test: devenv
--- a/Justfile
+++ b/Justfile
@@ -295,14 +295,16 @@ docs-build-dataset-definitions-outputs-docker-project:
 
 
 # Check the dataset public docs are current
-docs-check-public-docs-are-current: generate-docs
+docs-check-generated-docs-are-current: generate-docs
     #!/usr/bin/env bash
     set -euo pipefail
 
-    if [[ -z $(git diff ./docs/public_docs.json) ]]
+    # https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommitted-changes
+    # git diff --exit-code won't pick up untracked files, which we also want to check for.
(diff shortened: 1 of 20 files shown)


StevenMaude:
When we merge this, we should archive the [original repository](https://github.com/opensafely-core/mkdocs-opensafely-databuilder).

Justfile
```diff
@@ -221,14 +221,14 @@ databricks-test *ARGS: devenv databricks-env
...
+generate-docs OUTPUT_DIR="docs/includes/generated_docs": devenv
+    $BIN/python -m databuilder.docs {{ OUTPUT_DIR }}
+    echo "Generated data for documentation in {{ OUTPUT_DIR }}"
 
 update-external-studies: devenv
     $BIN/python -m tests.acceptance.update_external_studies
 
-docs-serve: devenv
+docs-serve: devenv generate-docs
     BACKEND_DOCS_FILE=docs/public_docs.json "$BIN"/mkdocs serve
```
StevenMaude, line 232: nit:

```suggestion
    "$BIN"/mkdocs serve
```

StevenMaude (approved):
This all looks good overall.

A few things:

* I've opened opensafely/documentation#1124 as a follow-on from this PR, to do the clean-up there.
* The `check` failure here is probably due to a recent Black upgrade since this PR was first opened.
* The `DEVELOPERS.md` should be [updated to remove the reference to `public_docs.json`](https://github.com/opensafely-core/databuilder/blob/825252c8f47d2fc60a596b9adaea3fddd96c5953/DEVELOPERS.md#L296). It possibly should also note there that this is currently a developer's responsibility in a PR. [^1]

[^1]: I'm not sure the "is documentation up-to-date?" check is a mandatory one, however. Maybe it should be? But I don't have permission to enable this.

rebkwok:
> I'm not sure the "is documentation up-to-date?" check is a mandatory one

It is, I've had it fail before

PR #987, opened 2023-01-27
Add Sort and PickOneRowPerPatient to generative tests

rebkwok:


tests/generative/variable_strategies.py
```diff
@@ -292,8 +290,11 @@ def many_rows_per_patient_frame(draw):
 
     @st.composite
     def sorted_frame(draw):
-        source = draw(many_rows_per_patient_frame())
-        return draw(sort(source))
+        source = draw(select_table())
+        # for _ in range(draw(st.integers(min_value=1, max_value=3))):
+        #     source = draw(sort(source))
+        source = draw(sort(source))
+        return source
```
rebkwok, line 297: @benbc I started trying to update this PR as we discussed this morning.  However, if I add the commented out lines here or the ones in `sort()` below, I start to see the too-many-filters error.  I also started to see it when removed those lines, and I could get it to go away by reducing the max number of filters in `many_rows_per_patient_frame`.

Are the for loops in `many_rows_per_patient_frame` and `ancestor_of` discarding examples?

benbc: No, they shouldn't be. I don't understand why that would be. It can't be using filtering to get integers that fall within the min/max bounds, can it?

rebkwok: The IntegerStrategy does do something with filtering, but I'm not sure I really understand what
https://github.com/HypothesisWorks/hypothesis/blob/3fc1f2a76b0104d70b7a0800a8edcb01ad0c3dab/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L111

benbc: That's the `IntegerStrategy`'s implementation of the `filter()` method so the consumer can apply further filtering. It doesn't look as though it uses it itself.

rebkwok: So, I've now updated this to how I _think_ the code should look for sorting/pick one row per patient.  But there is a weirdness that I've discovered (due to tedious and repeated trial and error, mostly):
With nothing cached, on a clean run of hypothesis, it always passes.  But on a second run, it always fails the health check.  (I found an old issue where someone was having unexpected health check failures, and David MacIver suggested trying to delete the `.hypothesis` folder and try again, although that did seem like a stab-in-the-dark suggestion)

Here's the hypothesis stats from a run with 100 examples that I just did, after deleting the .hypothesis folder:
```
============================= Hypothesis Statistics ==============================

tests/generative/test_data_setup.py::test_data_strategy_examples_round_trip:

  - during generate phase (0.05 seconds):
    - Typical runtimes: 3-44 ms, ~ 47% in data generation
    - 1 passing examples, 0 failing examples, 1 invalid examples

  - Stopped because settings.max_examples=1


tests/generative/test_query_model.py::test_query_model:

  - during generate phase (22.10 seconds):
    - Typical runtimes: 10-127 ms, ~ 84% in data generation
    - 100 passing examples, 0 failing examples, 448 invalid examples

  - Stopped because settings.max_examples=100
```

And then a second run, immediately afterwards:
```
==================================== FAILURES ====================================
________________________________ test_query_model ________________________________
  File "/home/becky/datalab/databuilder/tests/generative/test_query_model.py", line 85, in test_query_model
    @hyp.settings(**settings)
  File "/home/becky/datalab/databuilder/.venv/lib/python3.9/site-packages/hypothesis/core.py", line 1397, in wrapped_test
    raise the_error_hypothesis_found
  File "/home/becky/datalab/databuilder/.venv/lib/python3.9/site-packages/hypothesis/internal/healthcheck.py", line 27, in fail_health_check
    raise FailedHealthCheck(message)
hypothesis.errors.FailedHealthCheck: It looks like your strategy is filtering out a lot of data. Health check found 50 filtered examples but only 0 good ones. This will make your tests much slower, and also will probably distort the data generation quite a lot. You should adapt your strategy to filter less. This can also be caused by a low max_leaves parameter in recursive() calls
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.filter_too_much to the suppress_health_check settings for this test.
----------------------------------- Hypothesis -----------------------------------
You can add @seed(44756034410180094038173309016381335108) to this test or run pytest with --hypothesis-seed=44756034410180094038173309016381335108 to reproduce this failure.
```

It's not always 0 out of 50, but it does always seem to fail the second time.

rebkwok: It also will pass it I add `database=None` to the `@hyp.settings` decorator, but clearly we don't want to do that

benbc: This seems weird enough that it's worth engaging with the Hypothesis community to try to understand it. As well as the Github repo there are some active experts on StackOverflow.

benbc (approved):
This has the shape I was expecting now. Thanks Becky. I think it's worth getting a review from Dave as well so he is up-to-date on how it works. (And probably needs resolution of the filtering failure before merging.)

tests/generative/data_strategies.py
```diff
@@ -45,13 +44,15 @@ def patient_records(draw, class_, schema, value_strategies):
...
+    # somewhere between 0 and all of them will be used in the patient_records strategy
+    patient_ids = draw(st.permutations(list(range(1, max_patient_id + 1))))
 
     @st.composite
     def one_patient_record(draw_):
-        id_ = draw_(patient_ids)
-        hyp.assume(id_ not in used_ids)
-        used_ids.append(id_)
+        id_ = patient_ids.pop()
         return draw(record(class_, st.just(id_), schema, value_strategies))
```
benbc, line 56: This is a great idea.

I think you can simplify `one_patient_record()`:

```
def one_patient_record():
        id_ = patient_ids.pop()
        return record(class_, st.just(id_), schema, value_strategies)
```

rebkwok: Nope, when I tried this it started generating duplicate ids and getting SQL integrity errors

rebkwok: I did need to change `return draw(record(...` to `return draw_(record(...` to make it use the inner function's `draw_` and avoid the warning about using `st.composite` without using draw internally though

evansd (approved):
:+1: This all makes perfect sense to me

PR #999, opened 2023-01-31
fix: Fix abysmal performance of `get_setup_and_cleanup_queries`

evansd:
The previous algorithm had asymptotic complexity of order `O(no he didn't)`: on deeply nested query graphs, such as those produced by the "chained sorts" approach to retrieving sequential events, it would be slow enough that I never saw it complete.

Here we split the algorithm into two parts: first a recursive function works its way through the graph of GeneratedTables (taking care not to recurse into the same node twice) and yields pairs of dependecy relations between tables. We then feed the results of this into `TopologicalSorter` from the standard library which can sort the tables into an appropriate topological order given their dependencies.

I think this is a bit more comprehensible than what went before. But, more importantly, it's all over in milliseconds.

### DIFF
--- a/databuilder/utils/sqlalchemy_query_utils.py
+++ b/databuilder/utils/sqlalchemy_query_utils.py
@@ -1,3 +1,4 @@
+import graphlib
 import re
 
 import sqlalchemy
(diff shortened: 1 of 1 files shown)


databuilder/utils/sqlalchemy_query_utils.py
```diff
@@ -1,3 +1,4 @@
+import graphlib
```
inglesp, line 1: I can think of several places to rip out `import networkx as nx`...

databuilder/utils/sqlalchemy_query_utils.py
```diff
@@ -67,43 +68,31 @@ def get_setup_and_cleanup_queries(query):
...
-    #
-    queries = [(query, 0, set())]
-    table_levels = {}
-    while queries:
-        next_query, level, seen = queries.pop(0)
-        for table in get_generated_tables(next_query):
-            if table in seen:
-                continue
-            # We may end up setting this multiple times, but if so we'll set the highest
-            # level last, which is what we want
```
inglesp, line 93: "multiple times", lol

inglesp (approved):
This is so much more comprehensible than the code it's replacing!

PR #1003, opened 2023-01-31
Refactor query transformations to improve performance

evansd:
The big performance win here is the small commit towards the end which adds caching to `Node.__hash__()` in the query model. However to safely apply this caching we need to stop force-mutating the nodes when transforming query graphs.


tests/unit/query_model/test_transforms.py
```diff
@@ -301,5 +301,54 @@ def test_sorts_by_derived_value_handled_correctly():
     assert transform(variable) == expected
 
 
+def test_identical_operations_are_not_transformed_differently():
+    # Query model nodes are intended to be value objects: that is equality is determined
+    # by value, not identity and equal objects should be intersubstitutable. Approaches
```
bloodearnest, line 306: > intersubstitutable

Take a bow.

databuilder/query_model/nodes.py
```diff
@@ -106,12 +106,30 @@ class Node:
     def __init_subclass__(cls, **kwargs):
         # All nodes in the query model are frozen dataclasses
         dataclasses.dataclass(cls, frozen=True)
+        # Calculating the hash of an object requires recursively calculating the hashes
+        # of its children. In a deeply nested query graph this can take some time and,
+        # given how frequently `__hash__()` is called, this can end up completely
+        # dominating Data Builder's execution time. Given that these are immutable
+        # objects we can cache the hash value instead of recalcuting it each time.
+        cls.__hash__ = cached_method(cls.__hash__, "__hash__cache")
```
bloodearnest, line 114: > __hash__cache

He's on fire!

bloodearnest: Distant cousin of [ash cash](https://en.wiktionary.org/wiki/ash_cash) perhaps?

databuilder/query_model/transforms.py
```diff
@@ -41,39 +97,27 @@ class PickOneRowPerPatientWithColumns(PickOneRowPerPatient):
...
-def apply_transform(type_, transform, nodes, reverse_index):
+def apply_transform(rewriter, type_, transform, nodes, reverse_index):
     for node in nodes:
         if isinstance(node, type_):
-            transform(node, reverse_index)
+            transform(rewriter, node, reverse_index)
 
 
-def rewrite_sorts(node, reverse_index):
+def rewrite_sorts(rewriter, node, reverse_index):
```
inglesp, line 129: Do the the transforms need to know about the rewriter?  Or can the transforms return/yield nodes and their replacements?

Or, given there's currently a single transform, is this premature?

databuilder/query_model/transforms.py
```diff
@@ -27,9 +28,64 @@
...
+            return {
+                self.rewrite(k, replacing): self.rewrite(v, replacing)
+                for k, v in obj.items()
+            }
+        elif isinstance(obj, frozenset):
+            # As do frozensets
+            return frozenset(self.rewrite(v, replacing) for v in obj)
+        else:
+            # Any other values in the query graph we return unchanged
+            return obj
```
inglesp, line 63: Is it worth being explicit about the kinds of things that `obj` can be in this branch, and then adding an `else: assert False` to catch new kinds of object that we might want to handle differently?

databuilder/query_model/nodes.py
```diff
@@ -378,9 +378,15 @@ class Case(Series[T]):
...
-        # immutable once created so we're fine to make it hashable
-        return hash((tuple(self.cases.items()), self.default))
+        # `cases` is a dict and so not naturally hashable, but we treat it as immutable.
+        # The specfic hash used below is based on a recommendation by Raymond Hettinger.
+        # It's almost certainly unnecessary to include the values and default in the
+        # hash, but given that it's all cached anyway we may as well save ourselves the
+        # trouble of thinking about it. See:
+        # https://stackoverflow.com/a/16162138
+        return hash(
+            (frozenset(self.cases), frozenset(self.cases.values()), self.default)
```
inglesp, line 388: It might save a rontosecond of reading time if this was `(frozenset(self.cases.keys()), frozenset(self.cases.values()), self.default)`.

inglesp (approved):
Nice.  I'd like a second review from @benbc, since he's more familiar with the transformation code.

databuilder/query_model/transforms.py
```diff
@@ -41,39 +97,27 @@ class PickOneRowPerPatientWithColumns(PickOneRowPerPatient):
...
     ]
 
+    rewriter = QueryGraphRewriter()
     for type_, transform in transforms:
-        apply_transform(type_, transform, nodes, reverse_index)
+        apply_transform(rewriter, type_, transform, nodes, reverse_index)
 
-    variables = copy.deepcopy(variables)  # see comment above
-    return variables
+    return rewriter.rewrite(variables)
```
benbc, line 120: This won't be correct for non-commutative transformations. Obviously that's fine right now when we only have one transformation, but perhaps it deserves a comment to this effect on the `transforms` list.

databuilder/query_model/transforms.py
evansd: I think there's a fair bit of rejigging that could be done to the transform code, but I wanted to keep the changes as minimal as possible because I didn't really intend to be working on this at all.

databuilder/query_model/nodes.py
evansd: It wouldn't be Hettinger-approved though! I haven't looked enough at the C implementation of `dict` to know if it would make any difference but he does mention that fact the his approach allows `frozenset` to re-use the key hashes without having to recalculate them.

databuilder/utils/functools_utils.py
```diff
@@ -71,3 +71,23 @@ def __get__(self, obj, cls=None):
...
+    cache_attr = f"__{method.__name__}_cache"
+
+    def wrapper(self):
+        value = getattr(self, cache_attr, MISSING)
+        if value is not MISSING:
+            return value
+        value = method(self)
+        # We want to use this with frozen dataclasses which override `__setattr__` to
+        # trigger an error, so we have to use the method from `object` to do this
+        object.__setattr__(self, cache_attr, value)
```
benbc, line 90: I'm glad to see that you're keeping up the tradition of mutating these immutable objects.

databuilder/query_model/transforms.py
evansd: Yes, I originally didn't do this because I didn't want this code to know about every sort of thing we might wrap in a `Value` node. But now you raise this I think a better approach would be to handle `Value` nodes explicitly and then exhaustively list the other types of thing expect.

benbc (approved):

databuilder/query_model/transforms.py
evansd: Ah yes, my thought was that we might just want to do one rewrite per transform. They're relatively cheap now (and can short-circuit if there are no replacements) and that's probably easier than worrying about which of our transforms commute with which. Definitely worth a comment though.

databuilder/utils/functools_utils.py
bloodearnest: *mutatis mutandis*

databuilder/query_model/transforms.py
evansd: I've rebased a comment into the first commit.

databuilder/query_model/transforms.py
evansd: I've refactored it along these lines, and ended up pulling the rewriter out into its own module. This is all in additional commits; I haven't tried to rebase these.

PR #1004, opened 2023-01-31
Add a rudimentary ISARIC implementation

StevenMaude:
This facilitates access to the `ICNARC_New` table in TPP.


databuilder/backends/tpp.py
```diff
@@ -224,3 +224,28 @@ class TPPBackend(BaseBackend):
             last_linkage_dt="last_linkage_dt",
         ),
     )
+
+    isaric = QueryTable(
```
evansd, line 228: Given that you're not doing any transformations on this table, just renaming columns, you might be better off with a `MappedTable`. There's an example directly above of what that looks like.

databuilder/tables/beta/tpp.py
```diff
@@ -197,3 +198,69 @@ class ons_cis(EventFrame):
     visit_num = Series(int)
     is_opted_out_of_nhs_data_share = Series(bool)
     last_linkage_dt = Series(datetime.date)
+
+
+@table
+class isaric(EventFrame):
```
evansd, line 204: If we know this is all going to change then I might be tempted to give the table itself a name like `isaric_raw` or something. That would free you up to create an `isaric` table with the proper column types later without breaking anything written against the "raw" table.

databuilder/tables/beta/tpp.py
```diff
@@ -197,3 +198,69 @@ class ons_cis(EventFrame):
...
+    """
+
+    # Demographics
+    age = Series(str, description="Age")
+    age_factor = Series(str, description="TODO")
+    calc_age = Series(
+        str,
+        description="Calculated age (comparing date of birth with date of enrolment). May be inaccurate if a date of February 29 is used.",
+    )
+    sex = Series(str, description="Sex at birth.")
```
evansd, line 222: Stylistic nit, but if you put a trailing comma on these arguments then Black would consistently format them on a separate line (as it does with the longer descriptions) and I think overall that might be more readable.

evansd (approved):

databuilder/backends/tpp.py
StevenMaude: I was thinking of that, then ended up writing a query to cast various types and then just stuck with that.

No problem to switch to this.

databuilder/tables/beta/tpp.py
StevenMaude: Good idea; will do.

databuilder/backends/tpp.py
evansd: Ah yeah, it's annoying that we have to drop down to raw SQL just to cast columns. I think we could actually add support for casting (and other per-column transformations generally) to `MappedTable`, I just haven't got round to it.

databuilder/tables/beta/tpp.py
StevenMaude: That's a lot nicer. (Though this is the kind of thing I just expect Black should handle.)

PR #1005, opened 2023-01-31
fix: Faster iteration of deeply nested SQLAlchemy structures

evansd:
This is the final piece of the current performance puzzle. In #999 we addressed the performance issues created by a deeply nested stack of interdependent `GeneratedTable` objects. These are created when running against a query engine which uses temporary tables, like the MSSQL engine.

However, an equivalent problem arises when using a query engine that uses native SQL constructs like CTEs (as the SQLite engine does). Here the problem shows up when trying to iterate the structure using SQLAlchemy's supplied functions, which end up recursing back through the same sections of the graph over and over.

The fix is easy enough: we create a variant of `iterate()` which keeps track of what it's already seen and avoids re-recursing.

### DIFF
--- a/databuilder/utils/sqlalchemy_query_utils.py
+++ b/databuilder/utils/sqlalchemy_query_utils.py
@@ -1,3 +1,4 @@
+import collections
 import graphlib
 import re
 
@@ -9,7 +10,6 @@ from sqlalchemy.sql.elements import (
     UnaryExpression,
     operators,
 )
-from sqlalchemy.sql.visitors import iterate
 
 
 def is_predicate(clause):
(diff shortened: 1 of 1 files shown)


inglesp (approved):

databuilder/utils/sqlalchemy_query_utils.py
```diff
@@ -195,3 +195,40 @@ def yield_params_as_literals(compiled, dialect):
...
+    if not children:
+        return
+
+    stack = collections.deque([children])
+    seen = set()
+    while stack:
+        t_iterator = stack.popleft()
+        for t in t_iterator:
+            # Some SQLAlchemy objects overload equality so they aren't safe to use
+            # directly in sets but we can track them by ID
```
benbc, line 229: It's endlessly reassuring how often we come across wrinkles caused by SQLAlchemy's failure to separate the DSL from the semantic model.

PR #1017, opened 2023-02-07
Support passing parameters to dataset definition modules

inglesp:


evansd (commented):
I'm not convinced by this approach for a couple of reasons. The main one is that using `parse_known_kwargs()` means that incorrect arguments to databuilder now pass silently by default. If the user happens to implement strict parsing of the remaining args in their dataset definition then they'll get picked up there. But we can't assume that will happen: I imagine most users will do no parsing at all.

The other is that importing a shadow copy of `argparse` feels a bit heavyweight and leaves it obscure what things are just standard Python and what's Data Builder magic.

I wonder if instead we could use the convention that `--` marks the end of argument parsing and insist that user arguments go after that. So `databuilder.__main__` could split the args on the first `--` and use strict parsing on everything up to that. And then user arguments could be parsed using something like:
```py
from argparse import ArgumentParser

from databuilder import user_args


parser = ArgumentParser()
parser.add_argument("--year", type=int)
args = parser.parse_args(user_args)
```

databuilder/__main__.py
```diff
@@ -62,6 +62,9 @@ def show_help(**kwargs):
     add_generate_measures(subparsers, environ)
     add_test_connection(subparsers, environ)
 
+    if "--" in args:
```
evansd, line 65: If `--` is not in args then `user_args` will not be changed from `None` which means that `parser.parse_args(user_args)` will end up parsing the whole of `sys.argv[1:]` which I don't think is what we want.

databuilder/__init__.py
```diff
@@ -1,3 +1,7 @@
 from pathlib import Path
 
 __version__ = Path(__file__).parent.joinpath("VERSION").read_text().strip()
+
+# This is populated in databuilder.__main__.main with a list of any command line
+# arguments that follow "--", and can be used to parameterise a dataset definition.
+user_args = None
```
evansd, line 7: I think it would be useful to maintain the property that dataset definitions are executable without error directly in Python i.e. by running
```sh
python analyis/dataset_definition.py
```

As that doesn't go through `databuilder.__main__` it won't end up setting `user_args`.

I think it would be reasonable to populate `user_args` at import time, given that `sys.argv` is. But maybe it would be cleaner to make it a function like `get_user_args()` e.g.
```py
def split_cmd_and_user_args(args):
    if args is None:
        args = sys.argv[1:]
    if "--" in args:
        separator_index = args.index("--")
        cmd_args = args[:separator_index]
        user_args = args[separator_index + 1:]
    else:
        cmd_args = args
        user_args = []
    return cmd_args, user_args


def get_cmd_args(args=None):
    return split_cmd_and_user_args(args)[0]


def get_user_args(args=None):
    return split_cmd_and_user_args(args)[1]
```

inglesp: Thanks.  Would setting `user_args = []` in databuilder/__init__.py do the trick?

evansd: It would mean you had no way of passing arguments to the dataset definition other than by invoking it via databuilder. I think it would be nice if you could do:
```
python analysis/dataset_definition.py -- foo --bar=baz
```
And have it all work correctly.

I mean, that's obviously not essential. But given it's not hard to achieve either it seems worthwhile.

databuilder/main.py
```diff
@@ -236,6 +237,10 @@ def load_module(module_path):
     # library modules from that directory
     original_sys_path = sys.path.copy()
     sys.path.insert(0, str(module_path.parent.absolute()))
+    # Temporarily modify `sys.argv` so it contains any user-supplied arguments and
+    # generally looks as it would had you run: `python script.py some args --here`
+    original_sys_argv = sys.argv.copy()
+    sys.argv = [str(module_path), *user_args]
```
inglesp, line 243: The extra plumbing is worth it to get hold of `str(module_path)` here.

inglesp (commented):
I can't approve this, presumably because I opened it, but I would approve if I could!

evansd (approved):
Thanks! I'll approve my own code on your behalf then.

PR #1020, opened 2023-02-08
Default to terminal nodes after a max depth in generative tests

rebkwok:
This avoids hypothesis hitting the too-many-filters healthcheck by checking the current depth and returning a terminal node (select_column for series, select_patient_table or select_table for tables) if it's grown too big.


tests/generative/variable_strategies.py
```diff
@@ -1,6 +1,9 @@
 import datetime
+from functools import wraps
+from os import environ
 
 import hypothesis.strategies as st
+from hypothesis.control import current_build_context
```
evansd, line 6: Oh nice! I didn't realise there was a non-hacky way of accessing this state. How did you find this? I couldn't seem to find anything in the docs.

tests/generative/variable_strategies.py
```diff
@@ -58,6 +61,53 @@ def variable(patient_tables, event_tables, schema, value_strategies):
...
+    # hypothesis limit (100) and will be abandoned. This results in too many invalid examples,
+    # which triggers the too-many-filters healthcheck.
+    #
+    # If the max limit is set high - e.g. if we always let it go to 100 and then return our
+    # default terminating node, generating the examples takes a really long time.
+    #
+    # Setting the depth at which a table strategy bails too low means that hypothesis takes
+    # too long to shrink examples.
+    #
+    # The defaults are therefore set, somewhat arbitrarily, to 30 for series and 75 for tables.
```
evansd, line 79: I bet that was fun figuring out.

evansd (approved):
This is great.

The decorator approach is neat, but I wonder if we could get away with something even simpler for `limit_series_depth` given that everything[^1] goes through the `series` strategy. If you did the depth check here then it would be automatically applied to all series types.

And then give that `limit_table_depth` is only applied three times and requires an argument, it might be simpler implemented as a check in the functions themselves:
```py
if table_depth_exceeded():
    return select_table() # or return select_patient_table()
```

[^1]: Well, everything except `value` but then `value` doesn't need limiting.

That's a purely stylistic comment though. I think the behaviour here does exactly what we want.

tests/generative/variable_strategies.py
rebkwok: I googled something like "hypothesis test context" to try and get some way of tracking state via the frame, and found it as a happy coincidence

rebkwok:
> The decorator approach is neat, but I wonder if we could get away with something even simpler for `limit_series_depth` given that everything goes through the `series` strategy. If you did the depth check here then it would be automatically applied to all series types.

That's actually what I did to start with

evansd:
> That's actually what I did to start with

Ah, fine. We'll if you've looked at the alternative and are happier with this then that's fine by me.

tests/generative/variable_strategies.py
rebkwok: I'm not convinced about those numbers, but they can be modifed with env variables so we can try out different values if it turns out to take forever with the overnight ones or something

rebkwok:
> > That's actually what I did to start with
> 
> Ah, fine. We'll if you've looked at the alternative and are happier with this then that's fine by me.

Well, I actually meant I did it just in `series` and then when it seemed to work, moved it into a decorator so it could apply to all the various strategies, but of course it doesn't need to.  I've reworked it now :slightly_smiling_face:

tests/generative/variable_strategies.py
```diff
@@ -58,8 +60,37 @@ def variable(patient_tables, event_tables, schema, value_strategies):
...
+
+    def depth_exceeded(max_depth=None):
+        max_depth = max_depth or environ.get("GENTEST_SERIES_MAX_DEPTH", 30)
```
evansd, line 81: I think you need an `int` cast around the `environ.get()` here. But might be easier to do as module level constants? e.g. at the top of the module:
```py
SERIES_MAX_DEPTH = int(environ.get("GENTEST_SERIES_MAX_DEPTH", 30))
TABLE_MAX_DEPTH = int(environ.get("GENTEST_TABLE_MAX_DEPTH", 75))

rebkwok: I've done this, but I've also changed my mind about there needing to be 2 different settings, and after a bit more systematic checking, I don't think it makes a difference.  In fact, since the check is only used for a filtered/sorted table, and both of those strategies always use a series at least once, I think they'll always use the series depth anyway

PR #1021, opened 2023-02-08
Table-specific helper methods

evansd:
This adds three helper methods for common queries on certain ehrQL tables in TPP. They are:
 * `patients.age_on()`
 * `practice_registrations.for_patient_on()`
 * `addresses.for_patient_on()`

### DIFF
--- a/databuilder/tables/beta/tpp.py
+++ b/databuilder/tables/beta/tpp.py
@@ -2,6 +2,7 @@ import datetime
 
 from databuilder.codes import CTV3Code, DMDCode, ICD10Code, SNOMEDCTCode
 from databuilder.contracts.universal import patients
+from databuilder.ehrql import case, when
 from databuilder.tables import Constraint, EventFrame, PatientFrame, Series, table
 
 __all__ = [
@@ -131,6 +132,28 @@ class addresses(EventFrame):
     # they're represented in the data
     care_home_requires_nursing = Series(bool)
     care_home_does_not_require_nursing = Series(bool)
+
+    def for_patient_on(self, date):
+        """
+        Return each patient's registered address as it was on the supplied date.
+
+        Where there are multiple registered addresses we prefer any which have a known
+        postcode (though we never have access to this postcode) as this is used by TPP
+        to cross-reference other data associated with the address, such as the MSOA or
+        index of multiple deprevation.
+
+        Where there are multiple of these we prefer the most recently registered address
+        and then, if there are multiple of these, the one with the longest duration. If
+        there's stil an exact tie we choose arbitrarily based on the address ID.
+        """
+        spanning_addrs = self.take(self.start_date <= date).drop(self.end_date < date)
+        ordered_addrs = spanning_addrs.sort_by(
+            case(when(self.has_postcode).then(1), default=0),
+            self.start_date,
+            self.end_date,
+            self.address_id,
+        )
+        return ordered_addrs.last_for_patient()
 
 
 @table
--- a/databuilder/tables/beta/tpp.py
+++ b/databuilder/tables/beta/tpp.py
@@ -58,6 +58,23 @@ class practice_registrations(EventFrame):
         ],
     )
 
+    def for_patient_on(self, date):
+        """
+        Return each patient's practice registration as it was on the supplied date.
+
+        Where a patient is registered with multiple practices we prefer the most recent
+        registration and then, if there are multiple of these, the one with the longest
+        duration. If there's stil an exact tie we choose arbitrarily based on the
+        practice ID.
+        """
+        spanning_regs = self.take(self.start_date <= date).drop(self.end_date < date)
+        ordered_regs = spanning_regs.sort_by(
+            self.start_date,
+            self.end_date,
+            self.practice_pseudo_id,
+        )
+        return ordered_regs.last_for_patient()
+
 
 @table
 class ons_deaths(EventFrame):
--- a/tests/integration/tables/beta/test_tpp.py
+++ b/tests/integration/tables/beta/test_tpp.py
@@ -107,3 +107,130 @@ def test_practice_registrations_for_patient_on(in_memory_engine):
         {"patient_id": 4, "practice_pseudo_id": 456},
         {"patient_id": 5, "practice_pseudo_id": 789},
     ]
+
+
+def test_addresses_for_patient_on(in_memory_engine):
+    in_memory_engine.populate(
+        # Simple case: successive addresses
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=1,
+                    address_id=100,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2005, 1, 1),
+                ),
+                dict(
+                    patient_id=1,
+                    address_id=101,
+                    start_date=date(2005, 1, 1),
+                    end_date=date(2015, 1, 1),
+                ),
+                dict(
+                    patient_id=1,
+                    address_id=102,
+                    start_date=date(2015, 1, 1),
+                    end_date=date(2020, 1, 1),
+                ),
+            ]
+        },
+        # Address with NULL end date
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=2,
+                    address_id=103,
+                    start_date=date(2000, 1, 1),
+                    end_date=None,
+                ),
+            ]
+        },
+        # Overlapping: choose address with postcode
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=3,
+                    address_id=104,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2015, 1, 1),
+                    has_postcode=False,
+                ),
+                dict(
+                    patient_id=3,
+                    address_id=105,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2015, 1, 1),
+                    has_postcode=True,
+                ),
+            ]
+        },
+        # Overlapping: choose most recent
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=4,
+                    address_id=106,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2015, 1, 1),
+                ),
+                dict(
+                    patient_id=4,
+                    address_id=107,
+                    start_date=date(2005, 1, 1),
+                    end_date=date(2015, 1, 1),
+                ),
+            ]
+        },
+        # Overlapping: choose longest
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=5,
+                    address_id=108,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2012, 1, 1),
+                ),
+                dict(
+                    patient_id=5,
+                    address_id=109,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2015, 1, 1),
+                ),
+            ]
+        },
+        # Tie-break: choose largest address ID
+        {
+            tpp.addresses: [
+                dict(
+                    patient_id=6,
+                    address_id=110,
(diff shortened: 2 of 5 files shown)


rebkwok:
> make it possible to .take() after a .sort_by() (while retaining sortedness);

I closed [this PR](https://github.com/opensafely-core/databuilder/pull/981) but maybe I should open it again (and add some tests)

databuilder/tables/beta/tpp.py
```diff
@@ -87,6 +105,46 @@ class clinical_events(EventFrame):
...
+    def with_codes(
+        self,
+        codelist,
+        after=None,
+        before=None,
+        on_or_after=None,
+        on_or_before=None,
+        between=None,
+        on_or_between=None,
+    ):
```
inglesp, line 117: Would we want to provide some validation that users weren't providing eg both `before` and `on_or_before`, or `before` and `between`?

databuilder/tables/beta/tpp.py
```diff
@@ -87,6 +105,46 @@ class clinical_events(EventFrame):
...
+        events = self
+
+        # TODO: We should make `Codelist` iterable so we don't need this conditional
+        codes = codelist.codes if hasattr(codelist, "codes") else codelist
```
inglesp, line 121: Sounds like `Codelist` should have a `.system` attribute.

databuilder/tables/beta/tpp.py
```diff
@@ -58,6 +59,23 @@ class practice_registrations(EventFrame):
         ],
     )
 
+    def active_for_patient_at(self, date):
+        """
+        Return each patient's "active" practice registration as of the supplied date
+        """
+        spanning_registrations = self.take(
+            self.start_date <= date & ((self.end_date > date) | self.end_date.is_null())
```
inglesp, line 67: We should use `.drop` (or whatever we end up calling it) here.

databuilder/tables/beta/tpp.py
```diff
@@ -58,6 +59,23 @@ class practice_registrations(EventFrame):
         ],
     )
 
+    def active_for_patient_at(self, date):
```
inglesp, line 62: A clear improvement.

inglesp (approved):
Lots to like here!

databuilder/tables/beta/tpp.py
evansd: I did think about this, and there may be other good reasons for having that attribute. But in this specific context, given that we still want to handle plain old lists of codes, I don't think it would help us.

databuilder/tables/beta/tpp.py
evansd: Yes, you're right, we should. We end up generating some slightly silly SQL which is why I shied away from it. But I've got plans to fix the silly SQL and we shouldn't let that drive the kind of ehrQL we write.

evansd:
I'm withdrawing the `clinical_events.with_codes()` method from consideration. It was convenient but I'm not all convinced by the API and it all needs more thought.

I'm preserving the original PR description below for posterity:

---

This draft PR contains some table-specific helper functions I've found myself wanting while writing ehrQL. As discussed in Slack, I think that for these very general kinds of functions it works well to have them as methods on the table objects.

This is only intended as an initial draft in order to gather feedback. Before merging these would need testing, documenting, and ideally be integrated into the existing table documentation build.

A few points:

 1. I've used `at` instead of `as_of` because it means the same here but is shorter.

 2. I'm not quite sure about the term "active" for practice registrations and addresses, but couldn't immediately think of a better one. "Primary" maybe? It would be good to have a term we consistently use here.

 3. ehrQL has the convention that all methods which reduce from event- to patient-level data have `for_patient` in their name. I wasn't responsible for this, and I'm not sure it's explicitly documented as a convention anywhere, but I think it's a very good idea and all table-specific methods should adhere to it.

 4. The most controversial function is `clinical_events.with_codes()`. I ended up with this because it lets you do the most common thing you want to do with clinical events (find all matching a codelist within a specific time frame, ordered by date) with a minimum amount of boilerplate.

 5. I think we could do much better with than this if we made some relatively minor changes to ehrQL:
    - make it possible to `.take()` after a `.sort_by()` (while retaining sortedness);
    - make event-frame-returning methods return the same class of object they started with so that table-specific methods can be composed, rather than only being available on the base table object.

    With those in place we could add some convenience methods to tables which have a single (or an obvious primary) date column:
     - define methods like `.on_or_after(date)` as an alias for `table.take(table.date.is_on_or_after(date)`;
     - define a default sort order so that `first_for_patient()`/`last_for_patient()` work without having to do `table.sort_by(table.date)`.

    Putting all those together we could have something like:
    ```py
    dataset.hypertension = (
        clinical_events.with_codes(hypertension_codes)
        .on_or_before(index_date)
        .first_for_patient()
        .date
    )
    ```
    Which I think is pretty close to combining the simplicity of cohortextractor with the composability and well-defined semantics of ehrQL.

databuilder/tables/beta/tpp.py
```diff
@@ -115,6 +131,28 @@ class addresses(EventFrame):
...
+            # Prefer the most recently registered address
+            self.start_date,
+            # Prefer the address registered for longest
+            self.end_date,
+            # Prefer addresses with a postcode
+            case(when(self.has_postcode).then(1), default=0),
+            # Use the opaque ID as a tie-breaker for sort stability (we invert this simply
+            # so the order matches the original order defined in Cohort Extractor to
+            # facilitate direct comparison)
+            -self.address_id,
```
evansd, line 152: This logic was ported from Cohort Extractor but, looking at it now, I'm not convinced it's what we want. I think we might want to use `has_postcode` as the _most_ important sort condition (because I think that's what fields like `MSOA` or the care home fields hang off) and only after that use the dates.

databuilder/tables/beta/tpp.py
```diff
@@ -58,6 +59,21 @@ class practice_registrations(EventFrame):
         ],
     )
 
+    def active_for_patient_at(self, date):
+        """
+        Return each patient's "active" practice registration as of the supplied date
+        """
+        spanning_regs = self.take(self.start_date <= date).drop(self.end_date <= date)
```
milanwiedemann, line 66: I think this is identical to the logic that QOF uses to define registrations (this is good). The only difference is that QOF specifies another thing (GMS) which I don't entirely understand. 

> Select patients who meet either of the criteria below:
> - Registered for GMS prior to or on the achievement date and did not subsequently deregister from GMS (currently registered for GMS).
> - Registered for GMS prior to or on the achievement date and subsequently deregistered from GMS after the achievement date (previously registered for GMS).
(i.e. patients who were registered for GMS on the achievement date)
>
> Reject the remaining patients.

But I confused myself by thinking the smaller than in `.drop(self.end_date <= date)` is not doing what we want so I'm posting it and am moving on with my review.

```suggestion
        spanning_regs = self.take(self.start_date <= date).drop(self.end_date < date)
```

tests/integration/tables/beta/test_tpp.py
```diff
@@ -0,0 +1,236 @@
...
+                    patient_id=5,
+                    practice_pseudo_id=456,
+                    start_date=date(2000, 1, 1),
+                    end_date=date(2015, 1, 1),
+                ),
+            ]
+        },
+    )
+
+    active_reg = tpp.practice_registrations.active_for_patient_at("2010-01-01")
```
milanwiedemann, line 96: Just following from the comment above to help me think it through, if the test patient would have: `end_date=date(2010, 1, 1),` the current function `active_for_patient_at ()` would not include that patient, and I _think_ that's exactly what we want?

databuilder/tables/beta/tpp.py
```diff
@@ -115,6 +131,28 @@ class addresses(EventFrame):
     care_home_requires_nursing = Series(bool)
     care_home_does_not_require_nursing = Series(bool)
 
+    def active_for_patient_at(self, date):
```
milanwiedemann, line 134: we can name this function the same as in line 62 because databuilder knows whether the function gets called on address or registration data?

databuilder/tables/beta/tpp.py
milanwiedemann: > I'm not quite sure about the term "active" for practice registrations and addresses, but couldn't immediately think of a better one. "Primary" maybe? It would be good to have a term we consistently use here.

at first I thought "current" instead of "active" would be better, but then I realised that "current" is misleading if we look up an address or registration in the past. the only idea I had was to remove the prefix altogether and mention the table/data it refers to but then the functions have different names, e.g., 
- `practice_registrations.registration_for_patient_at()`
- `addresses.address_for_patient_at()`

milanwiedemann (approved):
It's really great to see helper methods coming to ehrQL 🎉 and very useful to see how you implemented (and tested) them. here some thoughts:
- what if the de-registration date is the same as the target date in the function (I think it's doing exactly what we want but I got confused)
- I agree, the word "active" seems odd but I also coudldn't think of anything better
- but I think changing the suffix `_as_of` (in cohortextractor) to `_at` is much clearer

tests/integration/tables/beta/test_tpp.py
evansd: To be honest, I don't actually know! My assumption was that a registration period includes the start date and all the days up to _but not including_ the end date. But quite possibly it includes the end date as well. We might be able to tell by looking at the data.

databuilder/tables/beta/tpp.py
milanwiedemann: I see, yes that makes sense and it'd be nice to keep the same name for these two methods

databuilder/tables/beta/tpp.py
evansd: I think we need something that conveys that a patient might have more than one of these things (address or practice registration) at a given time and we're somehow selecting the "best" one, which `address_for_patient_at()` doesn't quite do. I'm still not sold on "active" but I can't think of anything better.

milanwiedemann: or `pick_for_patient_at`

iaindillingham:
Seagulling in to ask, "Why `at` and not `on`?" The helper methods accept a date, not a time; _on_ is the preposition commonly used for dates and _at_ is the preposition commonly used for times.[^1] Are we trying to communicate an implicit time component? (e.g. the zeroth second).

[^1]: "At the first day of Christmas // my true love brought to me" sounds strange, no?

evansd:
@iaindillingham Ooh, this was exactly the ambiguity that "as of" painted over! Grammatically, we want `age_on(index_date)` and `age_at(study_start)`. I think you're right though: these _are_ dates and generally speaking people label the variables as such so "on" is the more appropriate.

databuilder/tables/beta/tpp.py
evansd: @inglesp and I agreed that "active" (and related terms like "primary") are confusing as they suggest this is somehow an official status recorded in the EHR, rather than just our best guess at picking something.

We've therefore gone for the minimalist `for_patient_on()` on the basis that the full explanation lives in the docs and the less we imply in the method name, the better.

databuilder/tables/beta/tpp.py
evansd: @inglesp and I decided that preferring addresses with postcodes was the right call because everything useful associated with an address (MSOA, IMD, care home status etc.) requires having a postcode. And now we're no longer trying to maintain exact parity with Cohort Extractor, we can drop the silly negation on `address_id`.

tests/integration/tables/beta/test_tpp.py
evansd: I have changed this to include the end date. If the end date overlaps with the start date of the next registration this makes no difference as we'll prefer the newer registration in any case.

PR #1029, opened 2023-02-09
Add `mean_for_patient()` aggregation

### LINKED ISSUE #1010: Add `mean_for_patient()` aggregation for numeric series
Defined as the sum of all non-null values divided by the count of all non-null values. The value is null if the count is zero.

Once we have [division support](https://github.com/opensafely-core/databuilder/issues/657) we could in theory bodge this together using:
```py
total = frame.series.sum_for_patient()
population = frame.take(frame.series.is_not_null()).count_for_patient()
mean = total.as_float() / population.as_float()
```

But not only is this clunky it will result in the database doing two aggregations where it need only do one; so I think we should add proper support for this.

This is needed to replace uses of the `mean_recorded_value()` function in Cohort Extractor.

evansd:
Closes #1010

### DIFF
--- a/databuilder/query_engines/in_memory.py
+++ b/databuilder/query_engines/in_memory.py
@@ -1,4 +1,5 @@
 import operator
+import statistics
 from collections import namedtuple
 
 from databuilder.query_engines.base import BaseQueryEngine
@@ -133,6 +134,10 @@ class InMemoryQueryEngine(BaseQueryEngine):
     def visit_Sum(self, node):
         col = self.visit(node.source)
         return col.aggregate_values(sum, default=None)
+
+    def visit_Mean(self, node):
+        col = self.visit(node.source)
+        return col.aggregate_values(statistics.fmean, default=None)
 
     def visit_CombineAsSet(self, node):
         assert False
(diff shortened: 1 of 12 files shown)


databuilder/query_engines/in_memory.py
```diff
@@ -134,6 +134,14 @@ def visit_Sum(self, node):
         col = self.visit(node.source)
         return col.aggregate_values(sum, default=None)
 
+    def visit_Mean(self, node):
+        def mean(values):
+            # NULLs are already handled by `aggregate_values`
+            return sum(values) / len(values)
```
iaindillingham, line 140: Why not use [`statistics.mean`](https://docs.python.org/3.10/library/statistics.html#statistics.mean) or [`statistics.fmean`](https://docs.python.org/3.10/library/statistics.html#statistics.fmean) instead?

iaindillingham (approved):

databuilder/query_engines/in_memory.py
evansd: Oh, you're quite right! When I first wrote this I thought I had to do all the NULL handling myself so I wrote my own `mean()` function. It was only when I got the coverage failure that I realised that `aggregate_values()` already does the NULL handling, so I ripped that out but didn't think to replace the function entirely.

`statistics.fmean` does what we want here because it gives a float result even if all the inputs are ints and happen to by divisible by the count.

PR #1039, opened 2023-02-14
Ignore arithmetic overflow errors in generative tests

### LINKED ISSUE #1034: Generative tests can trigger arithmetic overflow errors
Error:

```
  File "src/pymssql/_pymssql.pyx", line 478, in pymssql._pymssql.Cursor.execute
sqlalchemy.exc.OperationalError: (pymssql._pymssql.OperationalError) (8115, b'Arithmetic overflow error converting expression to data type int.DB-Lib error message 20018, severity 16:\nGeneral SQL Server error: Check messages from the SQL Server\n')
```

Failed test run:
https://github.com/opensafely-core/databuilder/actions/runs/4140532447/jobs/7159277900

The failed test example has a lot of nodes in it, but notably, a few `Multipy`s, which are generating very large numbers.  I could reproduce the error with this example:

```
@hyp.example(
    variable=Function.Multiply(lhs=Value(value=50000), rhs=Value(value=50000)),
    data=[]
)
```

rebkwok:
Fixes #1034

### DIFF
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -198,6 +198,23 @@ IGNORED_ERRORS = [
         sqlalchemy.exc.OperationalError,
         re.compile(".+Internal error: An expression services limit has been reached.+"),
     ),
+    # ARITHMETIC OVERFLOW ERRORS
+    # mssql raises this error if an operation results in an integer bigger than the max INT value
+    # or a float outside of the max range
+    # https://learn.microsoft.com/en-us/sql/t-sql/data-types/int-bigint-smallint-and-tinyint-transact-sql?view=sql-server-ver16
+    # https://learn.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver16#remarks
+    # https://github.com/opensafely-core/databuilder/issues/1034
+    (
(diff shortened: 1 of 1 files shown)


tests/generative/test_query_model.py
```diff
@@ -198,6 +198,23 @@ def run_error_test(query_engines, data, variable):
...
+    # https://github.com/opensafely-core/databuilder/issues/1034
+    (
+        sqlalchemy.exc.OperationalError,
+        re.compile(
+            ".+Arithmetic overflow error converting expression to data type [int|float].+"
+        ),
+    ),  # arithmetic operations that result in an out-of-range int or floar
+    (
+        sqlalchemy.exc.OperationalError,
+        re.compile(".+Arithmetic overflow error for type int.+"),
```
inglesp, line 215: Do we need to wrap each error string in `re.compile`?

inglesp (approved):

tests/generative/test_query_model.py
rebkwok: yes; or at least, we need to either wrap all of them or none.  The error ignoring code iterates through these and does a `.match()` on each one, so they all need to be compiled regexes otherwise that will error.

PR #1054, opened 2023-02-15
Automatic cast from int literals to floats

### LINKED ISSUE #540: ehrQL should automatically cast between ints and floats where necessary

### LINKED ISSUE #658: Support casting between ints and floats

milanwiedemann:
Closes #540 

This adds support for automatic casting from ints to floats.
We currently don't support IntSeries because we dont think we need it at the moment (and it'd be easier to add later than to remove). 
We also don't support casting from float to int for reasons described in the limitations section of #540

### DIFF
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -45,7 +45,7 @@ patients_schema = TableSchema(
     date_of_birth=Column(date), i=Column(int), f=Column(float)
 )
 patients = PatientFrame(SelectPatientTable("patients", patients_schema))
-events_schema = TableSchema(event_date=Column(date), i=Column(int), f=Column(float))
+events_schema = TableSchema(event_date=Column(date), f=Column(float))
 events = EventFrame(SelectTable("coded_events", events_schema))
 
 
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -188,13 +188,9 @@ class TestDateSeries:
 @pytest.mark.parametrize(
     "lhs,op,rhs,expected_type",
     [
-        (patients.i, "+", 10.0, IntPatientSeries),
         (patients.f, "-", 10, FloatPatientSeries),
-        (patients.i, ">", 10.0, BoolPatientSeries),
         (patients.f, "<", 10, BoolPatientSeries),
-        (events.i, "+", 10.0, IntEventSeries),
         (events.f, "-", 10, FloatEventSeries),
-        (events.i, ">", 10.0, BoolEventSeries),
         (events.f, "<", 10, BoolEventSeries),
     ],
 )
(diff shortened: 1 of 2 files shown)


tests/unit/test_query_language.py
```diff
@@ -179,6 +185,32 @@ def test_year(self):
...
+@pytest.mark.parametrize(
+    "lhs,op,rhs,expected_type",
+    [
+        (patients.f, "-", 10, FloatPatientSeries),
+        (patients.f, "+", 10, FloatPatientSeries),
+        (patients.f, "<", 10, BoolPatientSeries),
+        (events.f, "-", 10, FloatEventSeries),
+        (events.f, "<", 10, BoolEventSeries),
+        (events.f, ">", 10, BoolEventSeries),
+        (events.f, "<", 10.0, BoolEventSeries),
```
milanwiedemann, line 197: I'm not sure this covers all cases and am also not sure if the last one is needed.

evansd: I think the coverage here is fine.

Because of the way `_cast()` works we only have to test a single float operation with a single integer value to get 100% coverage. But I think it's useful to have a few more examples here, as you've done, because these tests are fast to run and the parameterisation means the examples themselves are very short and readable.

evansd (approved):
This is great, thanks! Hopefully it was a useful learning exercise even if it would be a stretch to describe it as fun :grinning: 

Can you use the "sqaush and merge" option when you come to merge this, as we don't need the full commit history here?

tests/unit/test_query_language.py
milanwiedemann: great, thanks for explaining

milanwiedemann:
Thanks for your help, it was an incredibly useful learning exercise!

PR #1056, opened 2023-02-15
Handle duplicates in the `Appointment` table

### LINKED ISSUE #1053: Handle duplicates in the `Appointment` table
There are duplicates in the `Appointment` table. According to Chris, we can handle them by selecting `DISTINCT`.

https://github.com/opensafely-core/databuilder/blob/42825b89c6777ef95bfe9aaf3721e1358f5f7e39/databuilder/backends/tpp.py#L196-L204

We think the duplicates will be removed from the DB, but we don't know when. It's pragmatic to remove them here. [Slack thread](https://bennettoxford.slack.com/archives/C04GVMBKST1/p1676292229600399).

iaindillingham:
Fixes #1053

### DIFF
--- a/databuilder/backends/tpp.py
+++ b/databuilder/backends/tpp.py
@@ -194,8 +194,12 @@ class TPPBackend(BaseBackend):
     )
 
     appointments = QueryTable(
+        # WARNING: There are duplicate rows in the Appointment table, so we add DISTINCT
+        # to remove them from this query. When they are removed from the Appointment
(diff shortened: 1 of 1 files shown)


evansd (approved):
I think if this wasn't intended to be temporary I would have been more interested in understanding the performance implications, and also might have wanted to include a duplicate in the backend test data to ensure it was removed.

But given this should be short-lived I'm happy with it as is.

PR #1061, opened 2023-02-16
fix: don't allow Values for boolean operations in generative tests

### LINKED ISSUE #1016: MSSQL error when trying to sort on a constant value
The PR to add Case to the generative tests has encountered a failing example: https://github.com/opensafely-core/databuilder/actions/runs/4104752821/jobs/7080653547#step:4:550

Translated into a spec test, the example looks like this:

```
def test_sort_by_case_column(spec_test):
    table_data = {
    e: """
          |  i1 | b1 | d1
        --+-----+----+---
        1 | 101 | T  | 1990-01-01
        1 | 102 | T  | 1991-01-01
        1 | 103 | F  | 1992-01-01
        2 | 203 | F  | 1993-01-01
        2 | 202 | T  | 1994-01-01
        2 | 201 | F  | 1995-01-01
        """,
}

    spec_test(
        table_data,
        e.sort_by(
            case(
                when(e.b1 & False).then(e.d1),  # always False
                default=None
            )
        ).first_for_patient().i1,
        {
            1: 101,
            2: 201,
        }
    )
```

The sqlite and in-memory engines return `{1: 101, 2: 201}` which implies they're sorting on `i1`; I'm not sure this is correct.
The mssql engine errors with:
```
sqlalchemy.exc.OperationalError: (pymssql._pymssql.OperationalError) (5309, b'Windowed functions, aggregates and NEXT VALUE FOR functions do not support constants as ORDER BY clause expressions.DB-Lib error message 20018, severity 16
```
If the case returns True for at least some cases, there are no errors e.g.
```
case(
        when(e.b1 & True).then(e.d1),
        default=None
    )
```

With a default value, we get the same error in mssql only, e.g.
```
case(
        when(e.b1 & False).then(e.d1),
        default=date(2000, 1, 1)
    )
```

rebkwok:
Fixes #1016

Boolean operators (And/Or/Not) take boolean series as their inputs.  A series can be a Value as well as a column or other operation.  It doesn't really make sense for an input to be e.g. `Value(False)`, and we wouldn't expect to ever see that in real ehrQL.  It results in some generative test errors where an example generates a `Value(False)` input to a `Case` statement, making the `Case` evaluate to a constant, and then uses that as the `sort_by` input to a `Sort` (see #1016).

This PR ensures that inputs to `And` and `Or` operations are never simple `Value` (`No`t already did this).  I've run the `@reproduce_failure` line here with this change and verified that it no longer fails.
https://github.com/opensafely-core/databuilder/actions/runs/4104752821/jobs/7080653547#step:4:561

### DIFF
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -225,10 +225,10 @@ def variable(patient_tables, event_tables, schema, value_strategies):
         return draw(binary_operation(type_, frame, Function.NE))
 
     def and_(type_, frame):
-        return binary_operation(type_, frame, Function.And)
+        return binary_operation(type_, frame, Function.And, allow_value=False)
 
     def or_(type_, frame):
-        return binary_operation(type_, frame, Function.Or)
+        return binary_operation(type_, frame, Function.Or, allow_value=False)
 
     @st.composite
     def lt(draw, _type, frame):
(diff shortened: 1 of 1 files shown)


evansd (approved):
Oh, this is a much neater solution; I wish I'd thought of it!

I think at some stage we will still want to do all the fancy query rewriting/simplification stuff, but we definitely don't want to do it now.

rebkwok:
> Oh, this is a much neater solution; I wish I'd thought of it!
> 
> I think at some stage we will still want to do all the fancy query rewriting/simplification stuff, but we definitely don't want to do it now.

:+1: I'm writing up an issue for the simplification and transforms after talking it over with @benbc today.  This is just the quick fix for the specific generative test issue.

Issue #974, opened 2023-01-25
databuilder Docker image is not being published


benbc:

StevenMaude:
There is a [more recent image build](https://github.com/opensafely-core/base-docker/pkgs/container/databuilder/67165270?tag=v0.68.1) available since this issue was opened, triggered by #999. 

Maybe this issue should be reworded as "Data Builder Docker image build is unreliable" if the image continues to be inconsistently published.

StevenMaude:
Hmm. This is still a little irregular: v0.69.2 didn't get published, but v0.69.3 did.

StevenMaude:
I'm not sure if this is the only source of unreliability, but it is at least one source.

I had another quick look at two recent corresponding workflow runs:

* `v0.69.2` that skipped publishing :x:
* `v0.69.3` that successfully published :heavy_check_mark: 

Both `v0.69.2` and `v0.69.3` seem to get correctly tagged, as intended:

`v0.69.2` tag workflow: https://github.com/opensafely-core/databuilder/actions/runs/4113680394/jobs/7100316254
`v0.69.3` tag workflow: https://github.com/opensafely-core/databuilder/actions/runs/4136327477/jobs/7150183536

But for `v0.69.2`, the initial check of tag in the Docker publication workflow apparently didn't find `v0.69.2` as a tag, and so skips publishing:

`v0.69.2` publish workflow: https://github.com/opensafely-core/databuilder/actions/runs/4113694025/jobs/7100126806
`v0.69.3` publish workflow: https://github.com/opensafely-core/databuilder/actions/runs/4136392906/jobs/7150191122

I looked around and can't find anything on what guarantees there are on tags showing up, having been submitted. You'd think that they'd be available following a push, but I'm not certain.

StevenMaude:
In that case, is there a reason why we couldn't just simplify the workflow to be triggered by a new tag, and then just publish unconditionally? (I may be missing something here.)

If there is some delay of tags showing, then that at least mitigates it by waiting for the tag to appear?

It's [possible to set such a tag-based trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs).

StevenMaude:
My proposal, which might help, is in #1062.

StevenMaude:
We could merge #1062, but it requires a token and waiting to see how we manage tokens for opensafely-core, which was discussed recently.

[...]

PR #1062, opened 2023-02-16
Trigger Docker image publication by tag on `main`

StevenMaude:
See #974: this is a possible way of improving the Docker image build reliability.

Details in individual commit messages.


StevenMaude:
> this is a possible way

*one* possible way; there may be better solutions, so just putting up for discussion.

Not properly tested out because that's tricky with these kind of workflows. I can evaluate this more if we think this is a good idea.

inglesp (approved):
Looks sensible to me.  Let's merge and see what happens?

StevenMaude:
Today I spent far too long trial-and-erroring this on a personal fork.

It seems to do the correct thing, as far as I can tell.

Questions/comments:

* The main thing I overlooked here is that a personal access token is required to trigger the "build image on new tag" workflow.
  * I think that's OK, at least in that it's no worse than what we have, because the existing token, I think, must have a repository read/write permission to be able to write tags.
  * This can be a fine-grained personal access token with:
    * read access to metadata
    * read and write access to repository contents
* I have tested pushing a Docker image to the GitHub registry.
  * It seems to work in my personal fork, when I set the `contents: packages: write` in the build workflow. This is necessary because forks of public repositories default the `GITHUB_TOKEN` permissions to read only. I don't think this needs setting here, because I think the default is read/write.
  * I think then, that the permissions on the PAT used in the *triggering* workflow are irrespective to the permissions of `GITHUB_TOKEN` in the *triggered* workflow.
* If we wanted to limit the exposure of the token more, we could just use the tag action to figure out the tag, but not push it, and then push it ourselves. (Though I'm not entirely sure if the tag action actually does anything clever in trying to prevent race conditions from two commits on `main` getting tagged simultaneously. That situation is unlikely in any case, and is potentially a current risk.)
* Checking the history of tags, I think only one in the past few months never got built as an image; do we still need this set of changes?
* If we go ahead, what's the best approach to add a new token? What's the suggested expiration? (maybe better to answer this via Slack)

.github/workflows/main.yml
inglesp, line 1: It would've made reviewing slightly easier if this commit had been split in two: one to move the action in the file, and one to make the substantive change.

inglesp (approved):
All looks good, and your testing sounds thorough!

> If we wanted to limit the exposure of the token more, we could just use the tag action to figure out the tag, but not push it, and then push it ourselves.

Do you mean a human person pushing it?  I'd like to avoid that if possible.

> Checking the history of tags, I think only one tag in the past few months never got built as an image; do we still need this set of changes? 

I think this is worth it for the simplified workflow (and the remove code!).

> If we go ahead, what's the best approach to add a new token? What's the suggested expiration? (maybe better to answer this via Slack)

Discussion in #tech?  Can you make sure that whatever we decide is documented.

.github/workflows/main.yml
StevenMaude: > Discussion in #tech? Can you make sure that whatever we decide is documented.

Asked here, with no reply yet: https://bennettoxford.slack.com/archives/C63UXGB8E/p1678974536982929

StevenMaude:
So, all this needs is an appropriately named fine-grained token with access to this repository adding as a secret with the appropriate permissions:

* read access to metadata
* read and write access to repository contents

PR #1074, opened 2023-02-22
Limit range of dates in generative tests

### LINKED ISSUE #912: Limit range of dates in generative tests
We define the range of possible generated dates here:

https://github.com/opensafely-core/databuilder/blob/3791a3a753831efc26e941813da7146faac3fcf6/tests/generative/test_query_model.py#L32-L34

This wide range means that we're very unlikely to generate colliding dates and so never explore date equality in the generative tests.

For integers we deliberately kept the range very small for this reason:

https://github.com/opensafely-core/databuilder/blob/3791a3a753831efc26e941813da7146faac3fcf6/tests/generative/test_query_model.py#L30

iaindillingham:
Closes #912

I changed the lower and upper bounds, and reran the generative tests with `GENTEST_EXAMPLES=200`, to see what happened, but the tests passed without issue. What have I missed?

### DIFF
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -45,7 +45,7 @@ value_strategies = {
     int: st.integers(min_value=0, max_value=10),
     bool: st.booleans(),
     datetime.date: st.dates(
-        min_value=datetime.date(1900, 1, 1), max_value=datetime.date(2100, 12, 31)
+        min_value=datetime.date(2010, 1, 1), max_value=datetime.date(2020, 12, 31)
     ),
     float: st.floats(min_value=0.0, max_value=11.0, width=16, allow_infinity=False),
 }


rebkwok (approved):
I don't think you've missed anything, I think the point was just that we were allowing such a wide range of dates that the tests are unllkely to check the cases where there are two dates in an example and the strategies generate the same ones.  I don't think we necessarily expected them to fail with a smaller range.  That said, the range might want to be reduced more, it still allows for a lot of dates, compared to e.g. the int strategy (only 1-10).

benbc:
I think it probably needs to be narrower if we're going to reduce it at all. Just a few years.

I _think_ the only benefit we get from narrowing it is that we sometimes hit equality conditions with dates. For the range you've got that's still very unlikley.

iaindillingham:
Thanks both. I've narrowed to the range from 2010-01-01 to 2020-12-31.

Issue #793, opened 2022-10-19
Support creating a temporary patient table from a file


evansd:
This is to match the `patients.with_value_from_file()` and `patients.which_exist_in_file()` functions in cohortextractor:
https://github.com/opensafely-core/cohort-extractor/blob/0acb74ee40a318a9a232c7d82a7b150a09a58bf6/cohortextractor/patients.py#L2787-L2819

Given a file (CSV, Arrow, whatever) with a `patient_id` column and some other columns, you should be able to reference it in ehrQL using a function like (strawman name) `table_from_file()`. This would return a PatientFrame with columns matching the columns in the file, which could then be used like any other patient frame.

Example:
```py
matched_cases = table_from_file("outputs/matched_cases.arrow")

dataset.year_of_birth = patients.date_of_birth.year
dataset.case_year_of_birth = matched_cases.date_of_birth.year
```

The query engines would have to load the contents of this file and upload them, somehow, into a temporary table of the appropriate sort.

Note, this feature was created to support complex case-control matching logic be done in Python and then having the results of that matching available in downstream study definitions. It's in theory possible that if the logical constructs available in Data Builder were powerful enough then this wouldn't be necessary. However the kind of logic used in case-control matching is most naturally expressed imperatively and I suspect would be very difficult to achieve in SQL.

evansd:
Implementing this will require various new bits of machinery in place.


## Query Model

At the core will be a new query model node:
```py
class InlinePatientTable(OneRowPerPatientFrame):
    rows: Iterable[tuple]
    schema: TableSchema
```

This looks just like a `SelectPatientTable` node, but instead of `name: str` attribute holding the table name it has a `rows: Iterable[tuple]` attribute holding the actual table data.

Unfortunately, the type-checking code is going to choke on this. Exactly how it chokes will depend on what sort of value we supply for `rows`, but there's no value we can supply which will satisfy it as things stand. So it will need some thought as to exactly what we do here.

But assuming we have a satisfactory resolution for that, the query model itself changes very little. I think the only change is that we'll need to register `InlinePatientTable` on the `get_root_frame` implementation here:
https://github.com/opensafely-core/databuilder/blob/09bc161caa813e42354bf444cd5fe350d5201581/databuilder/query_model/nodes.py#L745-L748


## Query Engines

[...]

PR #1076, opened 2023-02-22
Jongmassey/table from file

Jongmassey:
addresses #793

### DIFF
--- a/databuilder/file_formats/__init__.py
+++ b/databuilder/file_formats/__init__.py
@@ -1,5 +1,13 @@
-from databuilder.file_formats.arrow import validate_dataset_arrow, write_dataset_arrow
+import os
+
+from databuilder.file_formats.arrow import (
+    read_dataset_arrow,
+    validate_dataset_arrow,
+    write_dataset_arrow,
+)
 from databuilder.file_formats.csv import (
+    read_dataset_csv,
+    read_dataset_csv_gz,
     validate_dataset_csv,
     validate_dataset_csv_gz,
     write_dataset_csv,
@@ -8,15 +16,15 @@ from databuilder.file_formats.csv import (
 from databuilder.file_formats.validation import ValidationError
 
 FILE_FORMATS = {
-    ".arrow": (write_dataset_arrow, validate_dataset_arrow),
-    ".csv": (write_dataset_csv, validate_dataset_csv),
-    ".csv.gz": (write_dataset_csv_gz, validate_dataset_csv_gz),
+    ".arrow": (write_dataset_arrow, validate_dataset_arrow, read_dataset_arrow),
+    ".csv": (write_dataset_csv, validate_dataset_csv, read_dataset_csv),
+    ".csv.gz": (write_dataset_csv_gz, validate_dataset_csv_gz, read_dataset_csv_gz),
 }
 
 
 def write_dataset(filename, results, column_specs):
     extension = get_file_extension(filename)
-    writer = FILE_FORMATS[extension][0]
+    writer = get_methods(extension)[0]
     # We use None for stdout
     if filename is not None:
         filename.parent.mkdir(parents=True, exist_ok=True)
@@ -24,9 +32,17 @@ def write_dataset(filename, results, column_specs):
 
 
 def validate_dataset(filename, column_specs):
+    if not os.path.isfile(filename):
+        raise FileNotFoundError(f"{filename} not found")
     extension = get_file_extension(filename)
-    validator = FILE_FORMATS[extension][1]
+    validator = get_methods(extension)[1]
     validator(filename, column_specs)
+
+
+def read_dataset(filename, column_specs):
+    extension = get_file_extension(filename)
+    reader = get_methods(extension)[2]
+    return reader(filename, column_specs)
 
 
 def validate_file_types_match(dummy_filename, output_filename):
@@ -47,3 +63,11 @@ def get_file_extension(filename):
         return "".join(filename.suffixes[-2:])
     else:
         return filename.suffix
+
+
+def get_methods(extension):
+    if extension == ".feather":
+        extension = ".arrow"
+    if extension not in FILE_FORMATS:
+        raise ValueError(f"Loading from {extension} files not supported")
+    return FILE_FORMATS[extension]
--- a/databuilder/file_formats/__init__.py
+++ b/databuilder/file_formats/__init__.py
@@ -24,7 +24,7 @@ FILE_FORMATS = {
 
 def write_dataset(filename, results, column_specs):
     extension = get_file_extension(filename)
-    writer = get_methods(extension)[0]
+    writer = FILE_FORMATS[extension][0]
     # We use None for stdout
     if filename is not None:
         filename.parent.mkdir(parents=True, exist_ok=True)
@@ -35,13 +35,15 @@ def validate_dataset(filename, column_specs):
     if not os.path.isfile(filename):
         raise FileNotFoundError(f"{filename} not found")
     extension = get_file_extension(filename)
-    validator = get_methods(extension)[1]
+    if extension not in FILE_FORMATS:
+        raise ValueError(f"Loading from {extension} files not supported")
+    validator = FILE_FORMATS[extension][1]
     validator(filename, column_specs)
 
 
 def read_dataset(filename, column_specs):
     extension = get_file_extension(filename)
-    reader = get_methods(extension)[2]
+    reader = FILE_FORMATS[extension][2]
     return reader(filename, column_specs)
 
 
@@ -63,11 +65,3 @@ def get_file_extension(filename):
         return "".join(filename.suffixes[-2:])
     else:
         return filename.suffix
-
-
-def get_methods(extension):
-    if extension == ".feather":
-        extension = ".arrow"
-    if extension not in FILE_FORMATS:
-        raise ValueError(f"Loading from {extension} files not supported")
-    return FILE_FORMATS[extension]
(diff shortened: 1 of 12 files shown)


databuilder/file_formats/arrow.py
```diff
@@ -196,3 +197,9 @@ def get_first_record_batch_from_file(filename):
     with pyarrow.OSFile(str(filename), "rb") as f:
         with pyarrow.ipc.open_file(f) as reader:
             return reader.get_batch(0)
+
+
+def get_table_from_file(filename):
+    with pyarrow.OSFile(str(filename), "rb") as f:
+        with pyarrow.ipc.open_file(f) as reader:
+            return reader.read_all()
```
Jongmassey, line 205: would streaming the data rather than reading all be preferable here?

Jongmassey:
One potential flaw with this strategy - could a user define multiple `@table_from_file`-annotated tables in a dataset definition since a unique table name is generated and would this cause problems elsewhere?

databuilder/query_language.py
```diff
@@ -752,13 +753,18 @@ def decorator(cls):
...
+        column_specs = {
+            name: ColumnSpec(type=col_type)
+            for name, col_type in dict(
+                [("patient_id", int)] + schema.column_types
+            ).items()
+        }
+
+        if path.suffix == ".csv":
+            rows = csv.read_dataset_csv(path, column_specs)
+        elif path.suffix == ".csv.gz":
```
iaindillingham: It would have been good to rebase this change into 56d1fe56. I don't understand why we now pass `column_specs` to `read_dataset_csv` when we passed `schema` before.

databuilder/query_language.py
```diff
@@ -767,10 +773,7 @@ def decorator(cls):
         else:
             raise ValueError("only csv and arrow files may be loaded into tables")
 
-        qm_node = qm.InlinePatientTable(
-            rows=qm.IterWrapper(rows),
-            schema=schema,
-        )
+        qm_node = _inline_patient_table_from_rows(cls, rows)
```
iaindillingham, line 772: It would have been good to rebase this change into e01eb8c0.

tests/spec/table_from_file/test_table_from_file.py
```diff
@@ -18,10 +16,18 @@
...
 
-@pytest.mark.xfail
-def test_table_from_file(spec_test):
-    path = Path("")
+def test_table_from_file_csv(spec_test, tmp_path):
+    path = tmp_path / "test_table_from_file.csv"
+    with path.open("w") as f:
+        writer = csv.writer(f)
+        writer.writerow(["patient_id", "n"])
+        writer.writerows(csv_data)
```
iaindillingham: Alternatively, pass a string to `path.write_text`. For example, see:

https://github.com/opensafely-core/databuilder/blob/6bb616acb9187ff8f85e8c6fcdbe29420f27ef15/tests/unit/test_codes.py#L18-L27

databuilder/query_language.py
```diff
@@ -762,8 +762,8 @@ def decorator(cls):
 
         if path.suffix == ".csv":
             rows = csv.read_dataset_csv(path, column_specs)
-        elif path.suffix == ".csv.gz":
-            rows = csv.read_dataset_csv_gz(path, schema)
+        elif path.suffix == ".gz" and path.stem.endswith("csv"):
```
iaindillingham, line 763: Would I be right in thinking that `table_from_file` didn't work with `.csv.gz` files before this change?

```suggestion
        elif p.suffixes == ['.csv', '.gz']
```

databuilder/file_formats/csv.py
```diff
@@ -79,7 +79,7 @@ def read_dataset_csv(filename, column_specs):
 
 def read_dataset_csv_gz(filename, column_specs):
     with gzip.open(filename, "rt", newline="") as f:
-        return read_dataset_csv_lines(f, column_specs)
+        yield from read_dataset_csv_lines(f, column_specs)
```
iaindillingham, line 82: As before, I don't understand why we use `yield from` now, but we used `return` before. If we should have used `yield from` before, then a rebase would have been helpful.

databuilder/file_formats/arrow.py
```diff
@@ -4,7 +4,6 @@
 import pyarrow
 
 from databuilder.file_formats.validation import ValidationError, validate_headers
-from databuilder.query_model.table_schema import Column, TableSchema
```
iaindillingham: This change is part of 9e477846. What should I expect from this commit? How should I review it?

databuilder/query_engines/base_sql.py
```diff
@@ -34,8 +34,8 @@
 from databuilder.sqlalchemy_types import TYPE_MAP
 from databuilder.utils.functools_utils import singledispatchmethod_with_cache
 from databuilder.utils.sqlalchemy_query_utils import (
-    expr_has_type,
     GeneratedTable,
+    expr_has_type,
```
iaindillingham, line 40: Let's not leave this commit in the log.

databuilder/main.py
```diff
@@ -161,7 +161,11 @@ def get_sql_strings(query_engine, variable_definitions):
     sql = clause_as_str(results_query, dialect)
     sql_strings.append(f"-- Results query\n{sql}")
 
-    assert not cleanup_queries, "Support these once tests exercise them"
+    for n, query in enumerate(cleanup_queries, start=1):
```
iaindillingham, line 164: The `n` confused me. Surely `i`?

databuilder/file_formats/arrow.py
```diff
@@ -200,6 +200,6 @@ def get_first_record_batch_from_file(filename):
 
 
 def get_table_from_file(filename):
-    with pyarrow.OSFile(str(filename), "rb") as f:
+    with pyarrow.memory_map(str(filename), "rb") as f:
```
iaindillingham, line 203: Again, we don't need to preserve this commit in the log. We should rebase it into an earlier commit.

iaindillingham (commented):
This PR would benefit from @evansd taking a look, as I'm not nearly as familiar with the problem or the codebase.

I've made several comments on the solution, but if I had to emphasize one thing, it would be that a tidy commit log and good commit messages make it much, much easier to review a PR. You might find [_How to Write a Git Commit Message_](https://cbea.ms/git-commit/) useful and interesting.

databuilder/main.py
Jongmassey: I was following the pattern used elsewhere, but I agree with your instinct

databuilder/file_formats/csv.py
Jongmassey: yes, sorry I thought I'd squashed those together in a rebase but clearly not!

databuilder/main.py
evansd: Oh yeah, that was me, sorry! And the pattern's probably replicated to a few different files now. Worth changing all these loop variables to `i` in one commit I think.

databuilder/query_language.py
Jongmassey: Correct. The initial failing test was written as more of a sketch but probably shouldn't have been committed as just caused confusion

databuilder/main.py
Jongmassey: done

tests/unit/test_query_language.py
```diff
@@ -320,6 +323,67 @@ class some_table(EventFrame):
...
+        SchemaError,
+        match="`@table_from_file` can only be used with `PatientFrame`",
+    ):
+
+        @table_from_file([])
+        class some_table(EventFrame):
+            some_int = Series(int)
+
+
+def test_table_from_file_only_accepts_known_filetypes(tmp_path):
```
evansd, line 353: Once the file handling logic is moved to into `file_formats/__init__.py` it would make sense to move some of these tests to `test_file_formats.py`.

evansd (approved):
Thanks @Jongmassey. I think all the functionality looks good here. I've made some suggestions about how we could reorganise some of the code so it fits better into the existing structures.

databuilder/file_formats/__init__.py
```diff
@@ -47,3 +63,11 @@ def get_file_extension(filename):
         return "".join(filename.suffixes[-2:])
     else:
         return filename.suffix
+
+
+def get_methods(extension):
+    if extension == ".feather":
+        extension = ".arrow"
```
evansd, line 70: I think we were planning to use `.arrow` consistently as that seems to be the de-facto standard now. It's quite possible some of the old Github issues still talk about `.feather` though, in which case: apologies!

databuilder/file_formats/arrow.py
```diff
@@ -198,7 +198,12 @@ def get_first_record_batch_from_file(filename):
...
-def get_table_from_file(filename):
+def read_dataset_arrow(filename, column_specs=None):
     with pyarrow.memory_map(str(filename), "rb") as f:
         with pyarrow.ipc.open_file(f) as reader:
+            for i in range(0, reader.num_record_batches):
+                rb = reader.get_record_batch(i)
+                for j in range(0, rb.num_rows):
+                    yield tuple(rb.take([j]).to_pylist()[0].values())
+
             return reader.read_all()
```
evansd, line 209: This `return` won't do anything meaningful inside a generator function, so it would be better to just get rid of it.

evansd (approved):
Thanks, I think this makes a lot more sense structurally now.

A couple of smaller comments below.

databuilder/file_formats/__init__.py
Jongmassey: The current docs use `.feather` (e.g. https://docs.opensafely.org/actions-scripts/#large-highly_sensitive-output-files https://docs.opensafely.org/measures/#calculate-the-measures) so I figured I'd cover both bases.

evansd: Yeah, it's true that's what Cohort Extractor uses. But I think we're going for a clean break, fix-everything-at-once approach with ehrQL so I'd avoid carrying over legacy stuff unless we've got a demonstrable need for it.

Jongmassey: That seems fair, I'll yank out `.feather` and remove that method

Issue #837, opened 2022-11-23
Add sequence operations to generative tests


benbc:
See `tests.generative.variable_strategies.known_missing_operations` for the missing operations.

At the time of writing they are:

```
        AggregateByPatient.CombineAsSet,
        Function.In,
        Function.StringContains,
```

(Note that there may be other non-sequence-related operations mentioned there which are not covered by this issue.)

The operations need adding to `tests.generative.variable_strategies`. If you're not familiar with the generative tests then you should get an intro from (probably) Ben before starting this work.

There is a verification step to ensure that we're including all the operations:

```
GENTEST_COMPREHENSIVE=t GENTEST_EXAMPLES=5000 just test-generative
```

Once you've added the operations you'll find that this fails. You will then need to remove the newly added operations from tests.generative.test_query_model.recorder and it should pass. If you find you've removed the last of the known_missing operations then please remove this exception mechanism altogether.this exception mechanism altogether.

evansd:
`AggregateByPatient.CombineAsSet` isn't actually usable in any query engines at the moment. It was added while we were working out how we could model certain kinds of query and at one point I had proof-of-concept support for it in the SQL engines. But it won't be used until we close #518.

iaindillingham:
See:

* #1077
* #1086

I don't think I need to add a generative test for `AggregateByPatient.CombineAsSet`, because we haven't implemented it yet (see Dave's comment).

PR #1077, opened 2023-02-23
Add StringContains to generative tests

iaindillingham:
See #837

### DIFF
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -32,6 +32,8 @@ schema = TableSchema(
     d2=Column(datetime.date),
     f1=Column(float),
     f2=Column(float),
+    s1=Column(str),
+    s2=Column(str),
 )
 (
     patient_classes,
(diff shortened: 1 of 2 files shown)


tests/generative/test_query_model.py
```diff
@@ -48,6 +50,7 @@
         min_value=datetime.date(1900, 1, 1), max_value=datetime.date(2100, 12, 31)
     ),
     float: st.floats(min_value=0.0, max_value=11.0, width=16, allow_infinity=False),
+    str: st.text(alphabet=["a", "b", "c"], min_size=1, max_size=3),
```
rebkwok, line 53: This wouldn't allow an empty string.  Do we want to?  It'd probably be a likely value in a string column.

rebkwok (commented):
Looks good, and remarkably straightforward.  Just one question about str value strategy

tests/generative/test_query_model.py
iaindillingham: Probably! I'll add the empty string and force push.

iaindillingham: I've added the empty string by setting `min_size=0`:

https://github.com/opensafely-core/databuilder/blob/8f8e23645b4b5aac0f210cdf1ab14fdf5a7dc5fd/tests/generative/test_query_model.py#L53

This is because [`alphabet` should be a collection of length one strings][1] and the empty string is length zero. Although `min_size=0` is the default, setting it explicitly is consistent with the other strategies.

[1]: https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.text

rebkwok: yes, that's what I'd expected, thanks

rebkwok (approved):

PR #1081, opened 2023-02-27
Automatically cast strings to codes

evansd:
This adds validation to the `BaseCode` constructor so that it's no longer possible to create, e.g. `SNOMEDCTCode` instances containing strings which could never be valid SNOMED-CT codes.

With this validation in place we can then use the same mechanism as for handling ISO format date strings to cast strings to the appropriate coding system of the column they're being used with.

This means there's no longer a need for users to import the coding system types, which we don't want them doing anyway as we hadn't made them part of the `databuilder.ehrql` namespace.

### DIFF
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -51,10 +51,6 @@ class OPCS4Code(BaseCode, system_id="opcs4"):
     "OPCS-4"
 
 
-class Read2Code(BaseCode, system_id="readv2"):
-    "Read V2"
-
-
 class SNOMEDCTCode(BaseCode, system_id="snomedct"):
     "SNOMED CT"
 
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -4,6 +4,7 @@ enforce that queries use the appropriate coding system for a given column.
 """
 import csv
 import dataclasses
+import re
 from pathlib import Path
 
 # Populated by `BaseCode.__init_subclass__` below
@@ -21,6 +22,10 @@ class BaseCode:
     def __init_subclass__(cls, system_id, **kwargs):
         REGISTRY[system_id] = cls
 
+    def __post_init__(self):
+        if not self.regex.fullmatch(self.value):
+            raise ValueError(f"Invalid {self.__class__.__name__}: {self.value}")
+
     @classmethod
     def _primitive_type(cls):
         return str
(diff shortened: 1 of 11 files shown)


iaindillingham (approved):
As ever, I really appreciate the time you've taken to sequence the commits 🙂

databuilder/codes.py
```diff
@@ -34,30 +39,84 @@ def _to_primitive_type(self):
...
+    )
+
 
 class CTV3Code(BaseCode, system_id="ctv3"):
     "CTV3 (Read V3)"
 
-
-class DMDCode(BaseCode, system_id="dmd"):
-    "Dictionary of Medicines and Devices"
+    # Some of the CTV3 codes in OpenCodelists violate the below, either by having a
```
inglesp, line 62: Clarify that this is in the OpenCodelists coding system data, and not in any codelists?

databuilder/codes.py
```diff
@@ -34,30 +39,84 @@ def _to_primitive_type(self):
...
+        """,
+        re.VERBOSE,
+    )
+
 
 class Read2Code(BaseCode, system_id="readv2"):
     "Read V2"
 
+    # Syntactically equivalent to CTV3
+    regex = CTV3Code.regex
```
inglesp, line 103: Don't think this is right -- I'm pretty sure they can contain things like `XXXXX.00`.  But do we ever expose V2 codes anywhere?

databuilder/query_language.py
```diff
@@ -535,7 +536,9 @@ def _wrap(qm_node):
         ]
         assert len(matches) == 1
         cls = matches[0]
-    return cls(qm_node)
+        wrapped = cls(qm_node)
+        wrapped._type = type_
```
inglesp, line 546: It's a bit confusing having both `_type` and `type_` floating around.  Is it worth standardising?  Or is there some meaning that I'm missing -- and if so, could we have a comment at the top of the module?

inglesp (approved):

databuilder/codes.py
evansd: Thanks, good suggestion.

databuilder/query_language.py
inglesp: Ah OK, that sounds pretty standard.  If there's not project-specific semantics here, then it's probably not worth documenting.

databuilder/codes.py
evansd: OK, I was just basing this off the Wikipedia page!
https://en.wikipedia.org/wiki/Read_code

But I think the only reason I added this as a type originally was that I just copied all the systems included in OpenCodelists. I don't think these are used anywhere in OpenSAFELY, so maybe the right thing to do is to remove the type?

inglesp: Let's remove the type completely!

PR #1083, opened 2023-02-27
Prepare for removal of `Codelist` class

evansd:
This is a "make things worse before making them better" PR: it adds some conditional behaviour to `codelist_from_csv()` to change the return type based on the arguments supplied.

The end goal is to remove both the `system` argument and the `Codelist` type entirely and replace them with lists and dicts containing strings. But to facilitate a smooth transition we first support both lists/dicts and `Codelist`. Then, once we've updated all extant ehrQL projects, we can remove the old code and simplify.

The aim here is both to simplify things for end users and to reduce the API surface of ehrQL. Now that we can [validate](https://github.com/opensafely-core/databuilder/pull/1081) strings against the coding system they're being used with we no longer need users to specify the system for each codelist. Of course, ideally we'd get metadata from OpenCodelists which tells us the system automatically. But until we have this, making the user specify it doesn't add any meaningful validation and complicates the API.

As for removing the `Codelist` class, I think the more we can use standard Python types the better. If we later find a use for a specialised container type for codes we can add one, but driven by concrete needs rather than speculatively.

One final thought: conceptually codelists should probably be represented by `set` rather than `list`. But I couldn't really think of any concrete advantages to this, and using `list` felt  like a simpler and more friendly type for Python newbies.

### DIFF
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -140,35 +140,42 @@ class Codelist:
                 setattr(self, name, category_map)
 
 
-def codelist_from_csv(filename, column, system):
+def codelist_from_csv(filename, column, system=None):
     filename = Path(filename)
     if not filename.exists():
         raise CodelistError(f"No CSV file at {filename}")
     with filename.open("r") as f:
-        return codelist_from_csv_lines(f, column, system)
+        return codelist_from_csv_lines(f, column, system=system)
 
 
-def codelist_from_csv_lines(lines, column, system):
-    try:
-        code_class = REGISTRY[system]
-    except KeyError:
-        raise CodelistError(
-            f"No system matching '{system}', allowed are: "
-            f"{', '.join(REGISTRY.keys())}"
-        )
+def codelist_from_csv_lines(lines, column, system=None):
+    if system is None:
+        code_class = str
+    else:
+        try:
+            code_class = REGISTRY[system]
+        except KeyError:
+            raise CodelistError(
+                f"No system matching '{system}', allowed are: "
+                f"{', '.join(REGISTRY.keys())}"
+            )
     # `restval` ensures we never get None instead of string, so `.strip()` will
     # never blow up
     reader = csv.DictReader(iter(lines), restval="")
     if column not in reader.fieldnames:
         raise CodelistError(f"No column '{column}' in CSV")
-    codes = set()
+    codes = []
     # We treat every other column in the CSV as being a mapping from codes to categories
     category_maps = {name: {} for name in reader.fieldnames if name != column}
     for row in reader:
         code_str = row[column].strip()
         if code_str:
             code = code_class(code_str)
-            codes.add(code)
+            codes.append(code)
             for name, category_map in category_maps.items():
                 category_map[code] = row[name].strip()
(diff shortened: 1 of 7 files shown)


databuilder/codes.py
```diff
@@ -140,35 +140,53 @@ def __post_init__(self):
...
     # We treat every other column in the CSV as being a mapping from codes to categories
     category_maps = {name: {} for name in reader.fieldnames if name != column}
+    if category_column is not None and category_column not in category_maps:
+        raise CodelistError(f"No column '{category_column}' in CSV")
     for row in reader:
         code_str = row[column].strip()
         if code_str:
             code = code_class(code_str)
-            codes.add(code)
+            codes.append(code)
```
rebkwok, line 182: If there are duplicate codes in the csv that's passed in, we'll keep them here (and currently get rid of them at the end, when we pass a set into `Codelist`).  If the plan is that we'll get rid of the `Codelist` class and just return a list of codes, would it be better to use a set here, and (once the `Codelist` class is gone) convert to a list when returning?  Otherwise we'll make a list, (probably) turn it into a set to get rid of dupliclates, and then back into a list to return it to the user.

rebkwok (approved):
One qn, not a blocker

databuilder/codes.py
evansd: I wanted to avoid a set here because I thought it would be a bit more user-friendly to preserve the order. And I've got a separate PR (which I can push as a draft but we can't merge until all the studies have been updated) which refactors all this stuff away in any case. I agree it's not the nicest way of doing things, but I didn't think it was worth sweating over given the code will be short lived.

tests/acceptance/test_external_studies.py
```diff
@@ -53,6 +55,18 @@ def test_external_study(name):
     # think this is sufficient for these tests which are intended to ensure we don't
     # accidentally break the API. If we're unable to execute a valid query, that's a
     # separate class of problem for which we need separate tests.
-    dataset = load_dataset_definition(dataset_def_path, user_args=())
+    with chdir(study_path):
```
iaindillingham, line 58: Could we not replace `chdir` with [pytest's `monkeypatch` fixture][1]?

```suggestion
    with monkeypatch.context() as m:
        m.chdir(study_path)
```

[1]: https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-monkeypatch

iaindillingham (approved):
You asked for a second pair of eyes on the API changes. I've reviewed this PR, as well as several of those linked to from #1087.

I agree with your point about removing the `system` argument.

I think that modelling codelists as lists/dicts of strings will make it easier for users to reason about codelists. If you've done an introductory Python course, then `c1 + c2` is easier to reason about than `combine_codelists(c1, c2)`. The first is "Two sequences of elements, joined"; the second is "A function does something with two sequences of elements".

Whilst I'd also prefer `set` to `list`, set literals/comprehensions still surprise me; list literals/comprehensions don't.

tests/acceptance/test_external_studies.py
evansd: Oh yes, that saves some code. Thanks!

PR #1084, opened 2023-02-27
fix: Use pysqlite-binary instead of system SQLite

evansd:
The specific problem this is intended to resolve is that some developers have a version of SQLite compiled without the optional math functions library, and this means they don't have the `FLOOR()` function which we use.

But in general it's better to be able to pin our dependencies in requirements.txt rather than relying on whatever the system version happens to be.

We use [pysqlite-binary][1] rather than [pysqlite3-binary][2] because that has pre-built macOS wheels. Any reservations we might have about the anonymous forker behind this package are less significant given that we only use SQLite as a development and testing aid.

[1]: https://github.com/pysqlite3/pysqlite3
[2]: https://github.com/coleifer/pysqlite3


iaindillingham (approved):
🚀

PR #1088, opened 2023-02-28
Ensure MSSQL engine treats all float-like values as float, not decimal

### LINKED ISSUE #1065: MSSQL engine treats a float literal as a decimal and can truncate its precision
TL;DR: we should add a custom Float type to mssql_dialect, as we do with dates so that float literals are interpreted as floats and not decimal. 
 
Seen in generative tests:
https://github.com/opensafely-core/databuilder/actions/runs/4200368991/jobs/7286312970

As a spec test:
```
table_data = {
        p: """
              | b1 
            --+----
            1 | T    
            """,    
        e: """
              | b1 | f1
            --+----+---
            1 | F  | 0.1 
            """,
    }
    spec_test(
        table_data,
        case(
            when(p.b1).then(  # always False
                case(
                    when(e.b1).then(0.1),
                    default=None
                ).sum_for_patient()            
            ),
            default=0.75
        ),
        {1: 0.75,}
    )
```
This gives a result of 0.8 with the mssql engine, rather than the expected default 0.75.

It seems this is due to literals that we intended to be floats being treated as decimals in mssql, which do not have fixed precision, and so the `sum_for_patient()` in the case statement causes mssql to choose the precision it thinks is most appropriate.  In this case, it's truncating to the same prevision as the `then` clause.
https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?redirectedfrom=MSDN&view=sql-server-ver16

This returns the correct result:
```
case(
            when(p.b1).then(  # always False
                case(
                    when(e.b1).then(0.01),  # same precision as default
                    default=None
                ).sum_for_patient()            
            ),
            default=0.75
        ),
```

as does 
```
case(
            when(p.b1).then(  # always False
                case(
                    when(e.b1).then(e.f1),  # non-literal
                    default=None
                ).sum_for_patient()            
            ),
            default=0.75
        ),
```

It's also fine if there is at least one other condition with a non-literal:
```
case(
            when(p.b1).then(  # always False
                case(
                    when(e.b1).then(0.1),
                    when(e.f1 > 0).then(e.f1),
                    default=None
                ).sum_for_patient()            
            ),
            default=0.75
        ),
        {1: 0.75,}
```
#1063 would help this, by simplifying sole constants in case statements.

However, as @evansd [said](https://bennettoxford.slack.com/archives/C03FB777L1M/p1676629911887419?thread_ts=1676608640.640799&cid=C03FB777L1M):
> I think we might need to add a custom Float type to mssql_dialect, as we do with dates. It looks like there's a custom way of formatting literals that ensures they are interpreted as floats rather than decimals:
https://stackoverflow.com/questions/3024306/tsql-make-a-literal-float-value/54974960#54974960

rebkwok:
Fixes #1065 

Adds a custom Float type for the mssql engine, which ensures that all bound parameters are cast to float.

I haven't added the spec test that fails in #1065, because it doesn't seem like a useful test to be included in the documentation, but I have verified that it passes with this change.

### DIFF
--- a/databuilder/query_engines/mssql_dialect.py
+++ b/databuilder/query_engines/mssql_dialect.py
@@ -69,12 +69,29 @@ class MSSQLDateTime(_MSSQLDateTimeBase, sqlalchemy.types.TypeDecorator):
     format_str = "%Y-%m-%dT%H:%M:%S"
 
 
+# MS-SQL can interpret values that we intend to be floats as Decimals, which
+# can result in results with unexpectedly truncated precision.
+# https://github.com/opensafely-core/databuilder/issues/1065
+# https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql
+class MSSQLFloat(sqlalchemy.types.TypeDecorator):
+    impl = sqlalchemy.types.Float
+    cache_ok = True
+
+    def bind_expression(self, bindvalue):
+        # Wrap any bound parameters in an explicit CAST to their intended type.
+        # This ensures that any float-like numbers are always cast to
+        # floats, otherwise MSSQL will sometimes make a best-guess and return a
+        # Decimal type, which behaves differently to other engine.
+        return cast(bindvalue, type_=self)
+
+
 class MSSQLDialect(MSDialect_pymssql):
     supports_statement_cache = True
 
     colspecs = MSDialect_pymssql.colspecs | {
         sqlalchemy.types.Date: MSSQLDate,
         sqlalchemy.types.DateTime: MSSQLDateTime,
+        sqlalchemy.types.Float: MSSQLFloat,
     }
 
     # The base MSSQL dialect generates different SQL depending on the version of SQL
--- a/tests/integration/query_engines/test_mssql_dialect.py
+++ b/tests/integration/query_engines/test_mssql_dialect.py
@@ -33,3 +33,11 @@ def test_enforces_minimum_server_version(mssql_engine, monkeypatch):
     monkeypatch.setattr(MSSQLDialect, "minimum_server_version", (999999,))
     with pytest.raises(RuntimeError, match=r"we require at least \(999999,\)"):
         mssql_engine.sqlalchemy_engine().connect()
+
+
+def test_float_literals_have_correct_type(mssql_engine):
+    sum_literal = sqlalchemy.func.sum(0.5)
+    query = sqlalchemy.select(sum_literal + 0.25)
+    with mssql_engine.sqlalchemy_engine().connect() as conn:
+        results = list(conn.execute(query))
+    assert results[0][0] == 0.75
--- a/tests/unit/query_engines/test_mssql_dialect.py
+++ b/tests/unit/query_engines/test_mssql_dialect.py
@@ -74,3 +74,14 @@ def _str(expression):
         compile_kwargs={"literal_binds": True},
     )
     return str(compiled).strip()
+
+
+def test_mssql_float_type():
+    float_col = sqlalchemy.Column("float_col", sqlalchemy_types.Float())
+    # explicitly casts floats
+    assert _str(float_col == 0.75) == "float_col = CAST(0.75 AS FLOAT)"
+    assert _str(float_col == None) == "float_col IS NULL"  # noqa: E711
+    assert (
+        _str(sqlalchemy.sql.case((float_col > 0.5, 0.1), else_=0.75))
+        == "CASE WHEN (float_col > CAST(0.5 AS FLOAT)) THEN CAST(0.1 AS FLOAT) ELSE CAST(0.75 AS FLOAT) END"
+    )


databuilder/query_engines/mssql_dialect.py
```diff
@@ -69,12 +69,29 @@ class MSSQLDateTime(_MSSQLDateTimeBase, sqlalchemy.types.TypeDecorator):
...
+class MSSQLFloat(sqlalchemy.types.TypeDecorator):
+    impl = sqlalchemy.types.Float
+    cache_ok = True
+
+    def bind_expression(self, bindvalue):
+        # Wrap any bound parameters in an explicit CAST to their intended type.
+        # This ensures that any float-like numbers are always cast to
+        # floats, otherwise MSSQL will sometimes make a best-guess and return a
+        # Decimal type, which behaves differently to other engine.
+        return cast(bindvalue, type_=self)
```
evansd, line 85: I didn't initially think of doing this via CAST, although this ought to work fine.

For what it's worth, MSSQL has a way of expressing float literals using scientific notation:
https://learn.microsoft.com/en-us/sql/t-sql/data-types/constants-transact-sql?view=sql-server-2017#float-and-real-constants

So I think we could do this by adding a `process_bind_param()` method which formats the Python float as a string and then sticks `E0` on the end (if the string doesn't contain `E` already).

I think this would result in slightly cleaner SQL, but there's probably not much practical difference.

evansd (approved):
I agree that this shouldn't have a spec test, but I think it might be worth an integration test (along the lines of [the one we added](https://github.com/opensafely-core/databuilder/blob/63fce6e12a716aee82c9e5d77508bd2e489e763f/tests/integration/query_engines/test_mssql_dialect.py#L9-L19) to test the date type issue earlier) to enforce this behaviour.

rebkwok:
> I agree that this shouldn't have a spec test, but I think it might be worth an integration test

@dave I tried to add an integration test similar to the date one you pointed at, but I don't think I'm writing it correctly.  I thought this was doing the equivalent of the spec test, but it's passing without the new Float type:

```
def test_float_literals_have_correct_type(mssql_engine):
    case_statement = sqlalchemy.case(
        (sqlalchemy.literal(2) == 1, 0.1), else_=0.75
    )
    query = sqlalchemy.select(case_statement.label("output"))
    with mssql_engine.sqlalchemy_engine().connect() as conn:
        results = list(conn.execute(query))
    assert results[0].output == 0.75
```

evansd:
@rebkwok After a spectacular amount of wild-goose-chasing I have reduced the failing test case down to this:
```py
def test_float_literals_have_correct_type(mssql_engine):
    sum_literal = sqlalchemy.func.sum(0.5)
    query = sqlalchemy.select(sum_literal + 0.25)
    with mssql_engine.sqlalchemy_engine().connect() as conn:
        results = list(conn.execute(query))
    assert results[0][0] == 0.75
```

It was `SUM` that was doing the work all along here, not `CASE`.

databuilder/query_engines/mssql_dialect.py
rebkwok: @evansd I've tried this, and I couldn't get it to work.  If I format it as a string (in `process_bind_param()`), it complains about types:
```
sqlalchemy.exc.OperationalError: (pymssql._pymssql.OperationalError) (8117, b'Operand data type nvarchar is invalid for sum operator.DB-Lib error message 20018, severity 16:\nGeneral SQL Server error: Check messages from the SQL Server\n')
[SQL: SELECT sum(%(sum_1)s) + %(sum_2)s AS anon_1]
[parameters: {'sum_1': '0.5E0', 'sum_2': '0.25E0'}]
```
If I return it as the literal `0.5E0` (couldn't actually figure out how to do that, but I made it return the constants as a test), it still ends up with the truncated value (I guess because it means the same as `0.5`, so it hasn't actually been converted to anything different).

evansd: Ah, yeah I see. I guess there's no way round that then given the limitations of `pymssql`. Thanks for investigating.

tests/integration/query_engines/test_mssql_dialect.py
```diff
@@ -33,3 +33,11 @@ def test_enforces_minimum_server_version(mssql_engine, monkeypatch):
...
     with pytest.raises(RuntimeError, match=r"we require at least \(999999,\)"):
         mssql_engine.sqlalchemy_engine().connect()
+
+
+def test_float_literals_have_correct_type(mssql_engine):
+    sum_literal = sqlalchemy.func.sum(0.5)
+    query = sqlalchemy.select(sum_literal + 0.25)
+    with mssql_engine.sqlalchemy_engine().connect() as conn:
+        results = list(conn.execute(query))
+    assert results[0][0] == 0.75
```
evansd, line 43: Just based on the confusion we had recently when doing some test archaeology, I think we should add some more comments here to explain what this otherwise very weird looking test is doing. Something like:
 ```suggestion
def test_float_literals_have_correct_type(mssql_engine):
    # When using the `pymssql` driver without special float handling the "0.5" below
    # gets typed as a decimal and then the result of SUM gets typed as fixed precision
    # decimal.
    sum_literal = sqlalchemy.func.sum(0.5)
    # When added to a decimal of greater precision, the result gets rounded and ends up
    # being 0.8
    query = sqlalchemy.select(sum_literal + 0.25)
    with mssql_engine.sqlalchemy_engine().connect() as conn:
        results = list(conn.execute(query))
    # By explicitly casting floats in our custom dialect we can get the correct result
    assert results[0][0] == 0.75

```

evansd (approved):

tests/integration/query_engines/test_mssql_dialect.py
rebkwok: Ah, I merged too soon.  I'll add this

Issue #1087, opened 2023-02-28
Simplify codelist API


evansd:
Steps:

 1. **Remove requirement to use `*Code` classes:**
    https://github.com/opensafely-core/databuilder/pull/1081

 2. **Remove requirement to pass `system` argument and return basic Python types rather than `Codelist`:**
    https://github.com/opensafely-core/databuilder/pull/1083

 3. **Update existing studies to use new API:**
    (These are marked as Draft until the PR above is merged, at which point the tests should start passing)

    - https://github.com/opensafely/comparative-booster-ehrql-poc/pull/1
    - https://github.com/opensafely/cis-pop-validation-ehrql/pull/21
    - https://github.com/opensafely/openprompt-vaccine-long-covid/pull/3
    - https://github.com/opensafely/isaric-exploration/pull/8
    - https://github.com/opensafely/openprompt_health_utilisation/pull/4
    - https://github.com/opensafely/deaths-at-home-covid19/pull/3
    - https://github.com/opensafely/qof-diabetes/pull/13

    This project required no changes:
    - https://github.com/opensafely/MH_pandemic/

 5. **Remove `Codelist` class and associated code:**
    https://github.com/opensafely-core/databuilder/pull/1089

PR #1089, opened 2023-02-28
Remove `Codelist` class

evansd:
This is the successor to the work started in #1083 and can only be merged once all in-progess studies have been updated, as described in:
 - #1087


inglesp (approved):
Great!

PR #1090, opened 2023-02-28
Use `InMemoryQueryEngine` for population validation

### LINKED ISSUE #1019: Replace (most of) `population_validation` module with in-memory query engine
The `population_validation` module attempts to determine what a series expression would evaluate to when run against an empty database. This allows us to reject a problematic class of population definitions:
https://github.com/opensafely-core/databuilder/blob/825252c8f47d2fc60a596b9adaea3fddd96c5953/databuilder/query_model/population_validation.py

Most of the logic in here is already implemented by the in-memory query engine but, when it was added, the in-memory query engine existed only in the tests and wasn't part of the published Data Builder package so this duplication was unavoidable.

Now that the in-memory query engine is part of Data Builder we should use it, and remove this duplication.

evansd:
The previous `population_validation` module ended up duplicating logic from the in-memory engine. The total amount of code wasn't huge, but it meant that anyone adding an operation to the query model had _yet another_ set of changes to worry about (alongside a Python implementation, three flavours of SQL and a generative test strategy). Re-using the in-memory engine avoids this.

Closes #1019

### DIFF
--- a/tests/lib/query_model_utils.py
+++ b/tests/lib/query_model_utils.py
@@ -2,13 +2,16 @@ import dataclasses
 
 from databuilder.query_model import nodes as query_model
 
+# These are only exercised during the long-running generative tests when
+# GENTEST_COMPREHENSIVE is enabled
 
-def get_all_operations():
+
+def get_all_operations():  # pragma: no cover
     "Return every operation defined in the query model"
     return [cls for cls in iterate_query_model_namespace() if is_operation(cls)]
 
 
-def is_operation(cls):
+def is_operation(cls):  # pragma: no cover
     "Return whether an arbitrary value is a query model operation class"
     # We need to check this first or the `issubclass` check can fail
     if not isinstance(cls, type):
@@ -21,7 +24,7 @@ def is_operation(cls):
     return len(dataclasses.fields(cls)) > 0
 
 
-def iterate_query_model_namespace():
+def iterate_query_model_namespace():  # pragma: no cover
     "Yield every public thing in the query_model module"
     yield from [getattr(query_model, name) for name in query_model.__all__]
     yield from vars(query_model.Function).values()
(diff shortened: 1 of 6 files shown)


tests/unit/query_model/test_population_validation.py
```diff
@@ -157,7 +157,7 @@ def test_rejects_basic_unreasonable_population():
     ),
     (
         True,
-        Function.In(Value(10), Value({30, 20, 10})),
+        Function.In(Value(10), Value(frozenset([30, 20, 10]))),
```
inglesp, line 158: Can we validate this in a constructor?

tests/lib/query_model_utils.py
```diff
@@ -2,13 +2,16 @@
 
 from databuilder.query_model import nodes as query_model
 
+# These are only exercised during the long-running generative tests when
+# GENTEST_COMPREHENSIVE is enabled
```
inglesp, line 6: Is it worth adding a custom coverage pragma for things that are only hit in generative tests?  Or E_LIFETOOSHORT?

inglesp (approved):
Nice.

tests/unit/query_model/test_population_validation.py
evansd: I was going to say that that's fiddly to do and low value, but then I realised it's really easy! Will add a commit.

tests/lib/query_model_utils.py
evansd: I don't think it's worth it when we're just talking about coverage of test code. There are all sorts of reasons that not every bit of test code gets exercised on every test run and so we're generally more liberal with no-cover here.

tests/unit/query_model/test_population_validation.py
evansd: Commit here:
https://github.com/opensafely-core/databuilder/pull/1090/commits/d061c7e3b7e6263848ce3a11ba90a2b5add9f4e6

PR #1092, opened 2023-03-01
Add comments to explain what the float literal test is doing

rebkwok:


evansd (approved):
Thanks! Sorry I didn't get there sooner with this, but I think future-us will be grateful.

PR #1098, opened 2023-03-01
Specify ethnicity variable that actually exists

StevenMaude:
`ethnic` doesn't exist as a column, though is included as an item in the relevant Data Dictionary.

It is actually represented as ten columns, and patients can belong to multiple groups denoted by these columns.

These are still left as "TODO" values in tests, and generic strings as represented in Data Builder for now, until we better understand them.


StevenMaude:
Reference: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#ISARIC_New

wjchulme:
For info, these take values `"Checked"` or `"Unchecked"`.  I don't think `NULL`s are possible. So these could be reasonably represented as standard booleans

StevenMaude:
> For info, these take values "Checked" or "Unchecked". I don't think NULLs are possible. So these could be reasonably represented as standard booleans

If you're fairly certain, I can amend this (probably tomorrow morning).

Otherwise I'm also fine being conservative now, checking that assumption is correct on a successful run, and then adjust it accordingly.

I'm going to have to add/change other variables anyway.

wjchulme:
Yes, happy to wait until we know what's going on for sure

evansd (approved):

PR #1105, opened 2023-03-02
feat: Add more ISARIC clinical variables

### LINKED ISSUE #1071: Updates to ISARIC variables: add clinical characteristics
In support of the ISARIC validation project, the following variables in the `isaric_raw` table need to be added to the tpp backend:

`chrincard`
`hypertension_mhyn`
`chronicpul_mhyn`
`asthma_mhyn`
`renal_mhyn`
`modliv`
`mildliver`
`chronicneu_mhyn`
`malignantneo_mhyn`
`chronichaemo_mhyn`
`aidshiv_mhyn`
`obesity_mhyn`
`diabetescom_mhyn`
`diabetes_mhyn`
`rheumatologic_mhyn`
`dementia_mhyn`
`malnutrition_mhyn`
`diabetes_type_mhyn` (taking values `"No", "1", "2", "N/K"`)
`smoking_mhyn` (taking values `"Yes", "Never Smoked", "Former Smoker", "N/K"`)

~All but the last two variables take values `"YES", "NO", "Unknown"`. Null values will occur if `comborb_none!=""` (indicating no comorbidities) according to the data dictionary. This implies null values are semantically equivalent to `"No"`. Information wise, these variables are essentially nullable booleans, but I think it's preferable to retain the string values due to issues with nulls for booleans in pandas. So:~

~`"Yes"` -> `"Yes"`~
~`"No", NULL"` -> `"No"`~
~`"Unknown"` -> `"Unknown"`~

~But I need to confirm whether that's an approrpriate interpretation of nulls.~

UPDATE:

All but the last two variables take values `"YES", "NO", "Unknown", "NA"`. `"NA"` values occur if `comborb_none=="No comorbidities"`. This implies null values are semantically equivalent to `"No"`. So:

`"YES"` -> `"Yes"`
`"NO", "NA"` -> `"No"`
`"Unknown"` -> `"Unknown"`

`diabetes_type_mhyn` and `smoking_mhyn` also currently return values as is. There may be an advantage to standardising "unknown"/"N/K"/"Don't know" type values across the dataset, but it would be good to get input from others about this.

StevenMaude:
Fixes #1071.


databuilder/backends/tpp.py
```diff
@@ -257,6 +257,101 @@ class TPPBackend(BaseBackend):
...
+                    WHEN dementia_mhyn IN ('NO', 'NA') THEN 'No'
+                    ELSE 'Unknown'
+                END AS dementia_mhyn,
+                CASE
+                    WHEN malnutrition_mhyn = 'YES' THEN 'Yes'
+                    WHEN malnutrition_mhyn IN ('NO', 'NA') THEN 'No'
+                    ELSE 'Unknown'
+                END AS malnutrition_mhyn,
+                CASE
+                    WHEN smoking_mhyn IN ('Never Smoked', 'Former Smoker', 'Yes') THEN smoking_mhyn
```
StevenMaude, line 351: question: Do we want to be defensive here and lower-case the strings to match? Or are we reasonably assured that we'll never get anything other than the case given?

databuilder/backends/tpp.py
```diff
@@ -257,6 +257,101 @@ class TPPBackend(BaseBackend):
...
+                END AS rheumatologic_mhyn,
+                CASE
+                    WHEN dementia_mhyn = 'YES' THEN 'Yes'
+                    WHEN dementia_mhyn IN ('NO', 'NA') THEN 'No'
+                    ELSE 'Unknown'
+                END AS dementia_mhyn,
+                CASE
+                    WHEN malnutrition_mhyn = 'YES' THEN 'Yes'
+                    WHEN malnutrition_mhyn IN ('NO', 'NA') THEN 'No'
+                    ELSE 'Unknown'
```
StevenMaude, line 348: question: The `ELSE` is setting the value to `Unknown`, because that's the only string that should be left as a column value are matching.

Is it preferable to check explicitly, and result in a `null` if no match at all?

iaindillingham: Would a researcher ever wish to distinguish "no data" (`null`) from "not known" (`Unknown`)? I don't think they would, so I think this is the correct implementation.

databuilder/backends/tpp.py
iaindillingham: Are you able to check by querying L3? If not, then I will 🙂 We haven't been defensive elsewhere (e.g. `Patient.DateOfBirth`, `PatientAddress.MSOACode`) so unless we have reason to, then I wouldn't be defensive here.

iaindillingham (approved):
I like the two-step approach to your PRs, where you add the columns first and then add the values/types second.

databuilder/tables/beta/tpp.py
```diff
@@ -314,6 +314,84 @@ class isaric_raw(EventFrame):
         description="Onset date of first/earliest symptom.",
     )
 
+    # Clinical characteristics
+    chrincard = Series(
+        str,
+        description="Chronic cardiac disease, including congenital heart disease (not hypertension).",
+    )
```
evansd, line 321: These need a categorical constraint on them so they get typed correctly as categorical variables. Something like:
```suggestion
    chrincard = Series(
        str,
        description="Chronic cardiac disease, including congenital heart disease (not hypertension).",
        constraints=[
            Constraint.Categorical(["Yes", "No", "Unknown"])
        ],
    )
```

databuilder/backends/tpp.py
evansd: I'm not sure this is right, following @wjchulme's comment about [structural nulls](https://github.com/opensafely-core/databuilder/issues/1071#issuecomment-1441450857). I think _either_ NULL means something different from `Unknown`, in which case the logic above needs to distinguish them. Or, if it doesn't mean anything different, then these should be modelled as nullable booleans rather than three-valued categoricals.

I don't have a view as to which is the right answer here. But I think it needs to be one or the other and the status quo sits halfway between them.

iaindillingham: Ah, I'd not seen Will's comment. I'd also not considered nullable booleans as an alternative to three-valued categoricals. Thanks, Dave 👍🏻.

wjchulme: We should use [the data dictionary](https://raw.githubusercontent.com/isaric4c/wiki/master/ISARIC/CCP_REDCap_ISARIC_data_dictionary_codebook.pdf) to guide us here. 

If the "cormob_none" button is clicked, then none of the remaining comorbidity fields are shown, and all will take value `"NA"` in the table. (I think originally this would have been represented as `null` in redcap, but then I think the data went to R, and then it went as a csv to tpp and ended up as `"NA"`). Otherwise, `comorb_none` is not clicked ( -> `null` -> `"NA"`), and so all other comorbidity fields are shown and _must_ be answered (becasue they are "required" fields).

`"Unknown"` in this case means "the question has been considered and the answer is 'we don't know'". This is distinct from "the question has not been considered or has been left unanswered". Iain is probably right that, epistemologically, this makes no difference to researchers, but I still think an explicit `"Unknown"` category here is helpful as it removes all ambiguity. It also distinguishes it from " code is present / code is absent" booleans derived from coded event data, where missingness/unknownness is not a meaningful concept.

evansd: OK, then I think the answer is that we need to explicitly handle "Unknown" as a case and let the default be NULL. Something like:
```
  WHEN malnutrition_mhyn = 'Unknown' THEN 'Unknown'
END AS malnutrition_mhyn
```

And then the ehrQL columns should become three valued categoricals.

databuilder/backends/tpp.py
evansd: > Can we just put in a test that says "we expect these columns to only take values ["X", "Y", "Z", "NA"] and if not then fail"?

Yeah, I've actually had a thought about this for categorical handling more generally. I'm going to write it up separately, but the basic idea is that the default for all CASE statements like this should be to return the original value unchanged. And then we can check while we're retrieving the data that we only ever get the expected categories.

StevenMaude: @evansd:

> the default for all CASE statements like this should be to return the original value unchanged

To clarify/summarise everything here and in the other thread, then that means for most of the examples in this query (NB: there are a couple of other columns that don't fit this pattern), then we have:

```
CASE
   WHEN column_name IN ('YES', 'NO', 'Unknown') THEN column_name
   WHEN column_name = 'NA' THEN 'No'
END AS column_name
```

StevenMaude: Or, is it actually:

```
CASE
    WHEN column_name = 'NA' THEN 'No'
    ELSE column_name
END as column_name
```

and the categorical constraint, when enforced, will validate that the data only contains the values known to us ("YES", "NO", "Unknown")?

evansd: So, you're completely correct in your understanding of what I was suggesting. But I'm now less convinced (following discussion with BenBC) that what I was suggesting is the right approach!

I think there's a wider question we need to resolve here that it's not sensible to block your PR on. So I'd actually be inclined to stick with what you originally had and we'll  tackle the wider question separately. Sorry for all the noise!

StevenMaude: @evansd: I'll do a bit of refining of this PR in response to your feedback anyway before merging, because I think it's sensible.

We can still change this on a whim if we decide otherwise without breaking anything much for users, while it's still Will who's the primary consumer.

StevenMaude:
To summarise, for now, I've gone with categorical variables where the ISARIC values are left intact, except for `NA` in columns that have them. These `NA` are instead changed to `NO`.

This does mean there is some inconsistency in text between variables: `YES` versus `Yes`, `Unknown` versus `N/K`, but these are as the original data dictionary defines. If we want to harmonise this type of untidiness in Data Builder generally, that feels like a separate discussion to just implementing these variables in the short-term.

evansd (approved):
I think this looks good.

You're right that it's awkward trying to test this stuff. I wonder if it would help to build that SQL string programatically, given that it's mostly repeated CASE expressions with the variable name changing. That would make the code shorter and make typos/copy-paste errors less likely.

And then you could just test one CASE expression exhaustively and have reasonable confidence that the rest were OK.

StevenMaude:
> I think this looks good.
> 
> You're right that it's awkward trying to test this stuff. I wonder if it would help to build that SQL string programatically, given that it's mostly repeated CASE expressions with the variable name changing. That would make the code shorter and make typos/copy-paste errors less likely.
> 
> And then you could just test one CASE expression exhaustively and have reasonable confidence that the rest were OK.

I agree.

For now, I'll merge to unblock Will, but we should revisit this when the query is actually finalised.

Issue #875, opened 2022-12-14
Add acceptance tests for in-development studies


benbc:
For each of our [in-development studies](https://docs.google.com/spreadsheets/d/1-lw0qZSRdZwPY2HrQh_wDkifvJKvtqR6WKie_WLQNJI/edit#gid=0), add a test case in [test_external_studies.py](https://github.com/opensafely-core/databuilder/blob/d03f997aef3769b916aa3bfd33e9b306c9e0a39d/tests/acceptance/test_external_studies.py) to let us know if we break backwards-compatibility with the studies that have been written so far.

rebkwok:
@benbc 2 questions:
1) Should we add acceptance tests for all of the external studies, even if they're not done yet?
2) For the [openprompt long covid vaccine study](https://github.com/opensafely/openprompt-vaccine-long-covid), is analysis/dataset_definition.py going to be used?  It's not in the project.yaml, and it's failing the external study test because it uses an undefined `study_end_date` variable [here](https://github.com/opensafely/openprompt-vaccine-long-covid/blob/main/analysis/dataset_definition.py#L44).  I'm not sure if this is an error, or the study is waiting on paramterisable dataset defs.

rebkwok:
If I include the studies that aren't complete yet, a couple will need to use branches other than main, and will need to be updated later:
- bias study: currently using `evansd/ehrql-translation` branch
- Nuffield medication validation study: currently using `validation` branch

evansd:
Can I add another question here? Some of the in development studies are private. But vendoring them into Data Builder effectively makes them public. Do we know whether this acceptable?

rebkwok:
Hmm, that's a good point. I had to add auth headers so that the tests could run on private repos, but maybe they just shouldn't be included until they're public

benbc:
@rebkwok  I can get an answer to the question about the OpenPROMPT study, but I'd rather you, @inglesp and @evansd took whatever route you think best for the rest of this.

benbc:
I'm pretty sure that `analysis/dataset_definition.py` in that study is left around from some earlier development. You can ignore it.

benbc:
I'm pretty sure that `analysis/dataset_definition.py` in that study is left around from some earlier development. You can ignore it.

rebkwok:
See https://bennettoxford.slack.com/archives/C03FB777L1M/p1677835053960809

PR #1106, opened 2023-03-02
Acceptance tests for external studies

rebkwok:
See #875 
Adds acceptance tests for all the studies in [this list](https://docs.google.com/spreadsheets/d/1-lw0qZSRdZwPY2HrQh_wDkifvJKvtqR6WKie_WLQNJI/edit#gid=0), except the Winter Pressures (needs event data, not using databuilder yet) and 3 private repos.
- ONS CIS validation
- OpenPROMPT long covid vaccines
- Diabetes in QOF
- ONS mental health
- OpenPROMPT long covid economics

(Private repos not included: ISARIC data exploration, Nuffield medication validation, Bias study)

Some other updates needed along the way:
- ignore the files pulled in from external studies in pre-commit hooks and `just check`
- Allow the external study test to be run on multiple dataset definitions


.pre-commit-config.yaml
```diff
@@ -1,6 +1,8 @@
 default_language_version:
   python: python3.9
 
+exclude: tests/acceptance/external_studies/
```
evansd, line 4: Ah, I wish I'd known this was possible earlier! I ended up taking the nuclear option to work around this elsewhere:
https://github.com/opensafely-core/databuilder/blob/5adc1735960748d3334979f5437a5c1da1b2f9dc/.github/workflows/update-external-studies.yml#L26-L27

tests/acceptance/test_external_studies.py
```diff
@@ -64,8 +65,22 @@
 STUDY_DIR = Path(__file__).parent / "external_studies"
 
 
+@pytest.fixture
+def reset_module_namespace():
+    """
+    Studies often use the same names for modules (e.g. codelists.py, variables_lib.py)
+    Ensure that we clean up the module namespace after each external study test.
```
evansd, line 119: Oh, of course. I should have realised this would be problem. Your workaround looks like it will do the trick. I suppose there are other side-effects that dataset definitions could have (mucking about with `sys.path`, for instance) which might end up causing us problems down the line. But let's worry about those when they arise.

The "proper" way to do this would be to in a subprocess where none of this stuff can leak back to the caller. But I don't think it's worth changing anything until we hit actual problems.

Justfile
```diff
@@ -95,7 +95,7 @@ check: devenv
     $BIN/flake8
     $BIN/pyupgrade --py39-plus --keep-percent-format \
         $(find databuilder -name "*.py" -type f) \
-        $(find tests -name "*.py" -type f)
+        $(find tests -not -path 'tests/acceptance/external_studies/*' -name "*.py" -type f)
```
evansd, line 98: :+1: Thanks for catching this!

evansd (approved):
Great, all looks good.

One thing that might help future code archaeologists is if you separated out the commits which update the external test specifications for those which actually pull in the vendored code. Otherwise the diffs get overwhelmingly noisy.

I think you could reasonably just have one commit at the end which pulls in all the code for all the studies you've added, rather than adding them one by one, and then that commit can be easily ignored.

PR #1109, opened 2023-03-02
fix: Handle ISARIC date columns as strings

### LINKED ISSUE #1108: Updates to ISARIC variables: remove NA values before casting to dates
Following #1075 which uses the correct typing for date variables, it turns out the column contain quite a few `"NA"` values, which `CAST(... AS DATE)` doesn't like. I think replacing the casting with `CAST((CASE WHEN hostdat = 'NA' THEN null ELSE hostdat END) AS DATE) AS hostdat` will fix this.

`

StevenMaude:
And then cast them to dates, or `null`.

Fixes #1108.


StevenMaude:
I think this PR in itself is fine, but I also discussed thoughts on checking this end-to-end, including the "invalid" `NA` strings:

https://bennettoxford.slack.com/archives/C03FB777L1M/p1677780277280009

iaindillingham (approved):

databuilder/backends/tpp.py
```diff
@@ -249,21 +249,42 @@ class TPPBackend(BaseBackend):
                 ethnic___9,
                 ethnic___10,
                 covid19_vaccine,
-                CAST(covid19_vaccined AS DATE) as covid19_vaccined,
-                CAST(covid19_vaccine2d AS DATE) as covid19_vaccine2d,
+                CASE
+                    WHEN covid19_vaccined = 'NA' THEN null
+                    ELSE CAST(covid19_vaccined AS DATE)
```
evansd, line 254: Sadly, and there's no way you could have known this, you can't just cast strings to dates because the way MSSQL choose to parse them is locale dependent. You might think that ISO dates are unambiguous, and they are. But (utterly bizarrely) MSSQL can sometimes parse them as `YYYY-DD-MM` — a sort of ungodly hybrid US/ISO style which exists nowhere else in the world.

The safe way to do this is to use `CONVERT` and explicitly specify the format (which for ISO format happens to be number 23).

More gory details here:
https://learn.microsoft.com/en-us/sql/t-sql/data-types/nondeterministic-convert-date-literals?view=sql-server-ver16

```suggestion
                    ELSE CONVERT(DATE, covid19_vaccined, 23)
```

databuilder/backends/tpp.py
```diff
@@ -249,21 +249,42 @@ class TPPBackend(BaseBackend):
...
+                END AS covid19_vaccined,
+                CASE
+                    WHEN covid19_vaccine2d = 'NA' THEN null
+                    ELSE CAST(covid19_vaccine2d AS DATE)
+                END AS covid19_vaccine2d,
```
evansd, line 259: You could write this logic slightly more neatly as:
```suggestion
                    WHEN covid19_vaccine2d != 'NA' THEN CONVERT(DATE, covid19_vaccined, 23)
                END AS covid19_vaccine2d,
```

i.e. you don't need the explicit NULL clause.

databuilder/backends/tpp.py
iaindillingham: FFS! Thanks, Dave.

StevenMaude: ouch! I'll fix this. (I'm fortunate enough to have not used MSSQL before.)

PR #1110, opened 2023-03-03
Keep databuilder image version in tutorial up to date

### LINKED ISSUE #966: Handle updating of the Docker image in the tutorial `project.yaml`

rebkwok:
Adds a just command to find the latest (major) databuilder image version and update it in the ehrql tutorial's project.yaml.

A new github action runs after merges to main to run the update command, build the tutorial dataset definitions with the OpenSAFELY CLI, which will pull in any new updates to the image after the merge to main, and create a PR if anything has changed. 

Fixes #966

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -314,16 +314,23 @@ docs-update-tutorial-databuilder-version:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    if [ -z ${TOKEN+x}]
+    package="opensafely-core/databuilder"
+
+    if [ -z ${GITHUB_TOKEN+x} ]
     then
-      echo "TOKEN env variable is required (a PAT with the packages:read permission)"
-      exit 1
+        token="$(
+        curl --silent \
+            "https://ghcr.io/token?scope=repository:$package:pull&service=ghcr.io" \
+        | jq -r '.token'
+        )"
+    else
+        token="$(echo $GITHUB_TOKEN | base64)"
     fi
 
-    # find latest major version
-    latest_version=$(curl -H "Authorization: Bearer $TOKEN" \
-    https://api.github.com/orgs/opensafely-core/packages/container/databuilder/versions?per_page=1 \
-    | jq -r '.[0].metadata.container.tags[-1]')
+    latest_version=$(curl --silent --header "Authorization: Bearer $token" \
+      "https://ghcr.io/v2/$package/tags/list" \
+      | jq -r '.tags | map(select(test("^v\\d+$"))) | max_by(. | ltrimstr("v") | tonumber)'
+    )
 
     # replace latest version in tutorial project.yaml
     sed -Ei "s/v[0-9]\b/${latest_version}/g" docs/ehrql-tutorial-examples/project.yaml
--- a/Justfile
+++ b/Justfile
@@ -308,3 +308,22 @@ docs-check-generated-docs-are-current: generate-docs
       echo "Generated docs directory contains files/directories not in the repository."
       exit 1
     fi
+
+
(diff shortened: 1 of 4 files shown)


Justfile
```diff
@@ -308,3 +308,22 @@ docs-check-generated-docs-are-current: generate-docs
...
+
+docs-update-tutorial-databuilder-version:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    if [ -z ${TOKEN+x}]
+    then
+      echo "TOKEN env variable is required (a PAT with the packages:read permission)"
+      exit 1
+    fi
```
evansd, line 321: It's a really annoying fact of the Docker registry API that you need to get a token before accessing public resources. But you can get a token even when unauthenticated, and I think it's worth doing it this way to avoid having yet another PAT to maintain.

The code looks like:
```sh
package="opensafely-core/databuilder"
token="$(
  curl --silent \
    "https://ghcr.io/token?scope=repository:$package:pull&service=ghcr.io" \
  | jq -r '.token'
)"
latest_version="$(
  curl --silent --header "Authorization: Bearer $token" \ 
    "https://ghcr.io/v2/$package/tags/list" \
  | jq -r '.tags | map(select(test("^v\\d+$"))) | max_by(. | ltrimstr("v") | tonumber)'
)"
```

evansd (approved):

Justfile
rebkwok: The `TOKEN` it uses in CI is the usual `GITHUB_TOKEN`; is it sensible to continue to use it there, and just do this for the if the env var isn't there, so you don't have to supply one locally?

rebkwok: Actually, it looks like you can use GITHUB_TOKEN with Docker registry now
https://github.com/orgs/community/discussions/26279#discussioncomment-3251172

evansd: I _think_ you still have to do the authentication dance because you have to exchange the GITHUB_TOKEN for a Bearer token. I'm not sure just using the GITHUB_TOKEN as is will work. But I'd be delighted to be wrong about that.

rebkwok: It seems to be fine:
https://github.com/opensafely-core/databuilder/actions/runs/4343258472/jobs/7585083367

(now I'll get rid of that temporary commit!)

evansd: Oh nice! How did you know to base64 encode it?

rebkwok: It was in that link I posted :arrow_up:

evansd: That'll teach me! I didn't actually read the link because I assumed it was the same Github discussions link I had seen on this topic earlier, but it wasn't :smile:

PR #1112, opened 2023-03-03
Throw error for unexpected values

evansd:
We use a CASE expression to map the values in the `Patients.Sex` column to our user facing categories. Previously we would treat any expected values (i.e. anything other than `M`, `F`, `I`, `U`) as `unknown`. Now we pass unexpected values through unmodified where they will trigger an error when we try to write them to disk.

Note this is not a perfect solution for at least two reasons:

 * We can't do a similar trick when mapping strings to non-strings e.g. mapping `Y/N` to a boolean. If we get the string `?` there's nothing we can return to indicate a problem.
 
 * If the unexpected value happens to match one of our user-facing categories then that won't be treated as an error e.g. if we encounter the string `male` instead of `M` in the `Patients.Sex` column that's a violation of the data contract and should cause us to investigate, but it won't trigger an error here.

The proper solution to this, and many other issues, is an "offline" data consistency check. But until we have that, this is a low effort way of improving the situation.


evansd:
Converting back to draft because I'm not quite convinced this is the right approach.

evansd:
There was some discussion of this PR here:
https://bennettoxford.slack.com/archives/C01SE1WB6AW/p1677854142947179

The tentative conclusion was that the consistent thing to do here would be to raise a database error when encountering an unexpected case. This parallels the behaviour we get when doing other forms of casting, e.g a string to a date, if we encounter an invalid value.

However, triggering an error in SQL is not straightforward – especially if you want to know what the invalid value was as part of the error. It is possible to do this, but the workaround is not particularly pleasant:
```sql
CASE
  WHEN some_column = 'Cat1' THEN 'category_a'
  WHEN some_column = 'Cat2' THEN 'category_b'
  WHEN some_column = 'Cat3' THEN 'category_c'
  -- Trigger error for unexpected values
  WHEN CAST('invalid: ' + some_column AS INT) = 0 THEN ''
END
```

To trigger the error we form a string which we know can never be a valid int (because it has the prefix `invalid: `) and then attempt to cast it to an int. As this condition is only evaluated if none of the previous conditions were true then the error is only triggered if none of the other cases match.

evansd:
Closing in favour of doing this properly as described in #1550

PR #1118, opened 2023-03-06
fix: Raise helpful error if backslash used in paths

evansd:
A small UX improvement based on observed user errors.

### DIFF
--- a/databuilder/codes.py
+++ b/databuilder/codes.py
@@ -110,7 +110,18 @@ class DMDCode(BaseCode):
 def codelist_from_csv(filename, *, column, category_column=None):
     filename = Path(filename)
     if not filename.exists():
-        raise CodelistError(f"No CSV file at {filename}")
+        # If the character which comes after the backslash in the string literal happens
+        # to form a valid escape sequence then no backslash will appear in the compiled
+        # string. Checking the repr for backslashes has false positives (e.g. a tab
+        # character will trigger it) but that seems OK in this context.
+        if "\\" in repr(filename):
+            hint = (
+                "\n\n"
+                "HINT: Use forward slash (/) instead of backslash (\\) in file paths"
+            )
+        else:
+            hint = ""
+        raise CodelistError(f"No CSV file at {filename}{hint}")
     with filename.open("r") as f:
         return codelist_from_csv_lines(
             f, column=column, category_column=category_column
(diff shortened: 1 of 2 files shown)


databuilder/codes.py
```diff
@@ -110,7 +110,17 @@ class DMDCode(BaseCode):
 def codelist_from_csv(filename, *, column, category_column=None):
     filename = Path(filename)
     if not filename.exists():
-        raise CodelistError(f"No CSV file at {filename}")
+        filename = str(filename)
+        if "\\" in filename:
```
rebkwok, line 114: I feel like I'm missing something.  Why are we looking for double backslashes?
Errors like [this one](https://github.com/opensafely/openprompt_health_utilisation/blob/c862c96a5d9216c7376f4e75eebb135082c104f2/analysis/codelists.py#L52) result from a user trying to pass a string with a single backslash, like `codelists\opensafely-lung-cancer.csv`.

I'd assumed that calling Path on the filename was escaping the `\` into `\\`, but in a shell I get:
```
>>> filename = "foo\foo.csv"
>>> Path(filename)
PosixPath('foo\x0coo.csv')
>>> str(Path(filename))
'foo\x0coo.csv'
```

tests/unit/test_codes.py
```diff
@@ -32,6 +32,11 @@ def test_codelist_from_csv_missing_file(tmp_path):
         codelist_from_csv(tmp_path / "no_file_here.csv", column="CodeID")
 
 
+def test_codelist_from_csv_missing_file_hint(tmp_path):
+    with pytest.raises(CodelistError, match="backslash"):
+        codelist_from_csv(tmp_path / "bad\\separator\\file.csv", column="CodeID")
```
rebkwok, line 37: `codelist_from_csv()` would usually be called with a string which is then turns into a path.  I think it would make sense for the tests to do the same.

databuilder/codes.py
evansd: :facepalm: I was using double-backslash as I've got used to that being how you spell backslash in Python e.g.
```
>>> print("\\")
\
```
But of course the problem is the user's backslashes aren't even being interpreted as backslashes — they're treated as escape sequences.

Thanks for catching this. Back to the drawing board!

tests/unit/test_codes.py
evansd: Yes, good point. I think I should explicitly check that they accept both.

databuilder/codes.py
rebkwok: Aah! good detective work!

rebkwok (approved):

PR #1124, opened 2023-03-07
Check links in databuilder docs

### LINKED ISSUE #1123: Add link checking for documentation
See #808

rebkwok:
Adds a workflow that checks out the documentation repo, builds it, checks the links in the built databuilder docs only, and posts to slack if any failures are found.

This does duplicate the checks on the main docs repo (which run on all the docs, including the imported databuilder ones), but allows us to notify slack specifically for databuilder links, which are trickier to check locally (any references to the main docs are relative links which work when the main site is built, but not when running just the local docs).  Broken links in databuilder docs will need to be fixed here in this repo.

Closes #1123

### DIFF
--- a/.github/workflows/check-docs-links.yml
+++ b/.github/workflows/check-docs-links.yml
@@ -43,3 +43,13 @@ jobs:
           fail: true
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: "Notify Slack on Failure"
+        if: failure()
+        uses: voxmedia/github-action-slack-notify-build@3665186a8c1a022b28a1dbe0954e73aa9081ea9e # v1.6.0
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+        with:
+          channel_id: C03FB777L1M
+          status: "Databuilder Check Docs Links Failure"
+          color: danger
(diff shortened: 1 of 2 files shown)


StevenMaude (approved):
Great; thanks for adding this!

> This does duplicate the checks on the main docs repo (which run on all the docs, including the imported databuilder ones), but allows us to notify slack specifically for databuilder links, which are trickier to check locally (any references to the main docs are relative links which work when the main site is built, but not when running just the local docs). Broken links in databuilder docs will need to be fixed here in this repo.

Couple of general thoughts:

* We could use Lychee's `--exclude-path` in the parent documentation to exclude the redundant checks on Data Builder, maybe? That's a separate task elsewhere, in any case. https://github.com/opensafely/documentation/issues/1159
* We may well need an `.lycheeignore` adding, but that will become apparent when the check is run, and we can add it as needed. I don't think it's possible to run the workflow in this repository until it's merged anyway.

rebkwok:
I've run the workflow locally with `act` (also ran it with a deliberate error to make sure it actually fails), and the lack of .lycheeignore doesn't seem to cause problems.  

The main docs workflow runs once a week - I think we might as well just leave it doing the checks on the data-builder files too.

StevenMaude:
ah, of course! :smile:

StevenMaude:
> of course it doesn't need a .lycheeignore - it already has one

So one quirk is that exceptions need to be put in the parent documentation for now, unless we have a local `.lycheeignore` here, and overwrite the parent one.

PR #1125, opened 2023-03-07
Make databuilder docs look more like the main docs

rebkwok:
This does some minimal styling improvements to make the local docs (run with `just docs-serve`) look a bit more like the main docs.

It doesn't use the [mkdocs-multirepo-plugin feature](https://github.com/jdoiro3/mkdocs-multirepo-plugin#%CE%B2-development-in-imported-repos-beta), for [reasons mentioned here](https://github.com/opensafely-core/databuilder/issues/978#issuecomment-1423033126).  Having had another attempt at getting that feature to work, I've come to the conclusion that our docs may just be a bit too complex for it, and it's intended more for a skeleton parent repo that pulls in docs from other repos, so doing the reverse for a single imported repo doesn't involve pulling in a lot of complicated files.

Mainly, this PR adds the same markdown extensions as the main repo (with the exception of the configuration of the snippets paths).  This gets us, amongst other things, the nice formatting of notes and warnings:
Screenshot from 2023-03-07 15-24-09

In addition, there's a small hook to replace the parent_snippet notation with a "note" box to make it clear that these snippets will be replaced in the main build:
Screenshot from 2023-03-07 15-24-20

There are also a few minor fixes to the docs themselves:
1) Fix the indentation for a `!!! note` block
2) Add a missing snippet (which I'll remove from the main docs repo once this PR is merged)
3) Update the navigation for local docs to include the index page (the page at /data-builder on the main docs site) and the toc

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,8 @@ repos:
     - id: check-json
     - id: check-toml
     - id: check-yaml
+      # --unsafe is a workaround for the use of !! in mkdocs.yml.
+      args: [--unsafe]
     - id: detect-private-key
 
   - repo: local
--- a/docs/includes/data-builder-danger-header.md
+++ b/docs/includes/data-builder-danger-header.md
@@ -1,0 +1,13 @@
+!!! danger
+
+    This page discusses the new [OpenSAFELY Data
+    Builder](https://github.com/opensafely-core/databuilder) for
+    accessing OpenSAFELY data sources.
+
+    **Use OpenSAFELY [cohort-extractor](https://github.com/opensafely-core/cohort-extractor), unless you are
+    specifically involved in the development or testing of Data
+    Builder.**
+
+    OpenSAFELY Data Builder and its documentation are still undergoing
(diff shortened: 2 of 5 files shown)


StevenMaude (approved):
All looks good; thanks for all these recent improvements @rebkwok (and the ones from last week)!

Only thought: maybe it's worth adding a comment to the `mkdocs.yml` that this is a partial duplication of the configuration in the main documentation? It's possibly worth us doing the same in the main documentation's `mkdocs.yml` as a reminder that any changes there should be moved here too.

It's also maybe worth us retrying the multirepo plugin again in a few months if there's a newer version and seeing if it better supports the configuration we have. Not having to maintain the duplication manually would be one less chore, but it's probably not that bad really. In practice, we don't change `mkdocs.yml` that often though, and it only means that this preview doesn't look quite as representative.

rebkwok:
> Only thought: maybe it's worth adding a comment to the mkdocs.yml that this is a partial duplication of the configuration in the main documentation? 

Will do - it's also probably worth noting that it gets ignored when it's pulled in and built in the main docs

PR #1132, opened 2023-03-08
Ensure all QM nodes are covered by generative tests in CI

### LINKED ISSUE #1055: Ensure all QM node types are generated in standard CI run
In a [standard CI run](https://github.com/opensafely-core/databuilder/blob/main/.github/workflows/main.yml#L23-L35) (which uses `just test-all`) we want to verify that all QM node types have been generated by the generative tests.  @rebkwok has verified that 200 examples is enough here, but this takes up to 60s, which is much longer than the rest of the non-Spark test suite.

So by default, `pytest` should run the generative tests with a handful of examples, and without checking that all types of node have been generated.

This means that `just test-all` should set `GENTEST_EXAMPLES` and `GENTEST_COMPREHENSIVE`, but that [by default the number of examples that hypothesis generates](https://github.com/opensafely-core/databuilder/blob/main/tests/generative/test_query_model.py#L69) should be low, eg 10.

rebkwok:
Fixes #1055 

- Sets the default generative tests examples to 10, which means they take a couple of seconds to run.
- Makes `just test-all` use GENTEST_COMPREHENSIVE=t GENTEST_EXAMPLES=200 (which is sufficient to cover all the QM nodes).  GENTEST_EXAMPLES can still be defined as something else by doing e.g. `GENTEST_EXAMPLES=500 just test-all`
- Updates the recording fixture scope so that comprehensive check failures are reported at a more logical point in the test run (after the generative tests, rather than at the end of the entire run)

### DIFF
--- a/tests/generative/recording.py
+++ b/tests/generative/recording.py
@@ -47,7 +47,7 @@ class Recorder:
         return tuple(copy.values())
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def recorder(request):  # pragma: no cover
     recorder_ = Recorder()
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -34,7 +34,8 @@ example `just test-unit`).
 
 ### Running tests
 
-To run all tests, as they're run in CI (with code coverage):
+To run all tests, as they're run in CI (with code coverage, and the comprehensive
+check for [generative tests](#generative-tests)):
 ```
 just test-all
 ```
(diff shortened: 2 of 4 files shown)


tests/generative/recording.py
```diff
@@ -47,7 +47,7 @@ def _hashable(item):
         return tuple(copy.values())
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
```
evansd, line 50: Oh, this is such a neat solution :pinched_fingers:

evansd (approved):

PR #1133, opened 2023-03-08
Re-enable logging in tests with an environment variable

### LINKED ISSUE #907: Support logging SQL in tests with an environment variable
Until [this commit](https://github.com/opensafely-core/databuilder/commit/defe0bbe0a6d30fee9392cd931053d8f31e4c697), we could log the SQL used in a test by setting `LOG_SQL` in the environment.  This was useful and should be reinstated, without reintroducing the excessive logging that was suppressed by that commit.

[Slack thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1672400296293549).

(For now, logging can be re-enabled by passing `echo=True` to [`create_engine()`](https://github.com/opensafely-core/databuilder/blob/3ef8f00e54ae474c306b3fd6d78e7cb33d7b6fbd/tests/lib/databases.py#L62).)

rebkwok:
Adds a new environment variable `LOG_TESTS` that turns on logging in tests, and re-enables the `LOG_SQL` env variable to log SQL queries in tests.

Fixes #907


tests/lib/databases.py
```diff
@@ -63,6 +63,7 @@ def host_url(self):
     def engine(self, dialect=None, **kwargs):
         url = self._url(self.host_from_host, self.port_from_host, include_driver=True)
         engine_url = sqlalchemy.engine.make_url(url)
+        kwargs["echo"] = bool(os.getenv("LOG_SQL"))
```
evansd, line 66: Enabling `echo` here will only effect the test setup queries, not the queries executed by the Query Engine which uses it's own connection object. You're better off relying on the behaviour of `init_logging()` and just making sure you call that whenever `LOG_SQL` is set.

rebkwok: Another thing I misunderstood!  I thought what we wanted was to be able to log the sql queries without all the other logging. 
That's done now.

evansd: Ah, really sorry about this Becky! I think this was one of those "quickly note something down to record the bug" issues which then got turned in to a ticket in Next without making sure it would make sense to someone who didn't have all the original context.

I would say it's all my fault but actually I think it's all Peter's :grinning: 

Let's try and make sure we don't do this again.

iaindillingham (approved):
🚀

tests/lib/databases.py
rebkwok: It's actually fine, I've learned lots along the way about what pytest does with logging :)

PR #1135, opened 2023-03-10
fix: Fix bug in raw SQL compilation

evansd:
I spotted that `dump-dataset-sql` was blowing up for certain studies and bisected the issue down to this.

I think this might have been exposed by a change in SQLAlchemy as I'm pretty sure these studies used to compile to raw SQL without problem. But in any case, the only reason we have to do all this hand-rolled parameter interpolation is to work aroud bugs in SQLAlchemy which have since been fixed in v2.0. Once we get rid of Spark, and thus PyHive, we can upgrade to SQLAlchemy 2.0 and hopefully throw away a lot of this code.

### DIFF
--- a/databuilder/utils/sqlalchemy_query_utils.py
+++ b/databuilder/utils/sqlalchemy_query_utils.py
@@ -186,15 +186,21 @@ def clause_as_str(clause, dialect):
 
 
 def yield_params_as_literals(compiled, dialect):
-    for key, value in compiled.params.items():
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Can't say I fully understand this, but it looks reasonable.  Would it have been caught sooner if we'd been dumping SQL in the acceptance tests?

evansd:
Yes it would. Or indeed in the generative tests.

PR #1136, opened 2023-03-10
Remove support for Spark

### LINKED ISSUE #1120: Remove Spark from codebase
We have no need for a Spark backend (for now) and its presence in the codebase adds friction to development.  We should remove it.

@evansd has already done the work of identifying which parts of the codebase only exist to support Spark.  These are marked by `pragma: cover-spark-only`.

inglesp:
We won't need it any time soon, and its presence adds friction to development and slows down the tests.

I have removed all references to "spark", "graphnet", "databricks", and "dbx".

Fixes #1120.


tests/lib/databases.py
```diff
@@ -76,9 +70,10 @@ def _url(self, host, port, include_driver=False):
         else:
             protocol = self.protocol
         url = f"{protocol}://{auth}{host}:{port}/{self.db_name}"
-        # only used for databricks atm
-        if self.query:  # pragma: no cover
-            url += "?" + "&".join(f"{k}={v}" for k, v in self.query.items())
+        # TODO
+        # # only used for databricks atm
+        # if self.query:  # pragma: no cover
+        #     url += "?" + "&".join(f"{k}={v}" for k, v in self.query.items())
```
benbc, line 76: Missed removing this?

inglesp: Fixed, thanks.

evansd (approved):
I think the only remaining question is: in which diatonic mode should we write our Holy Song of Thanksgiving?

PR #1145, opened 2023-03-13
Update glossary

### LINKED ISSUE #804: Update glossary
[The glossary](https://github.com/opensafely-core/databuilder/blob/main/GLOSSARY.md) would benefit from an update.  The audience of the glossary is developers working on the databuilder codebase.

There are two pieces of work here:

- [x] Collect terms that require definitions (update the checkboxes below)
- [x] Write the definitions!
   - [x]  Backend
   - [x]  Contract
   - [x]  Dummy data
   - [x]  ehrQL/query language (an update to the Domain Specific Language entry)
   - [x]  Measure
   - [x]  Query engine
   - [x]  Table
   - [x]  Variable definition (although leaning towards the implementation, an entry would be useful as `variable_definitions` is found throughout the codebase)

[This thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1675174832155669) contains some helpful discussion.

iaindillingham:
Closes #804

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -51,13 +51,13 @@ When a user writes ehrQL this constructs a Query that is then passed to a Query 
 This provides a clear boundary between the syntactic surface of ehrQL (where expressiveness and user convenience are paramount) and the underlying semantics.
 
 **ehrQL Table**
-Describes the structure and types of some EHR data.
-User's can write queries against this data using ehrQL.
+The table structure we present to the user.
+User's can write queries against this table structure using ehrQL.
 
 **Database Table**
-Table in the SQL sense as stored in the database of an EHR provider.
-These are not directly exposed to users.
-Users interact with ehrQL Tables and their queries are translated into queries against Database Tables by a Backend.
+The table structure in an EHR provider's database.
+We do not present this table structure to the user.
+Instead, the user interacts with ehrQL Tables; their queries are translated into queries against this table structure by a Backend.
 
 **`variable_definitions`.**
 Found throughout the Data Builder codebase, the Query Model equivalent of a Dataset Definition.
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -3,12 +3,12 @@
 **Backend.**
 A class that translates between the table structure we present to the user (the ehrQL Tables) and the table structure in an EHR provider's database (the Database Tables).
 
-Each Backend is specific to an EHR provider and a Query Engine, which tells Data Builder how to talk to the kind of database used by the EHR provider.
-For instance, the TPP backend uses the MSSQLQueryEngine, which knows how to connect to Microsoft SQL Server and speak Transact-SQL to it.
+Each Backend is specific to an EHR provider and a Query Engine.
+For example, the TPP backend is specific to TPP (an EHR provider) and the MSSQLQueryEngine.
 
-A Backend is not required in order to connect to a database.
-If the structure of the database tables exactly matches the structure of the ehrQL tables, then there is no translation to be done.
-In that case, we can query the database just be specifying the appropriate Query Engine.
+A Backend is not required in order to connect to an EHR provider's database.
+If the structure of the Database Tables exactly matches the structure of the ehrQL Tables, then no translation is required.
+In this case, we query the EHR provider's database by specifying the appropriate Query Engine.
 
 **Dataset Definition.**
 Describes, using ehrQL, the EHR data to extract from a Backend.
@@ -36,9 +36,8 @@ A quotient (a numerator divided by a denominator) for a given time interval and 
 A graph of connected Query Model objects.
 
 **Query Engine.**
-Takes a Query, connects to a database, and returns the corresponding data.
-Usually this will be a relational database speaking some dialect of SQL.
-But it doesn't have to be.
+Takes a Query, connects to an EHR provider's database, and returns the corresponding data.
+Usually, the database will be relational and will speak SQL.
 
 **Query Language.**
 A programming language designed for querying a data store.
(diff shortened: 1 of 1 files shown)


GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
 # Glossary
 
-* **Domain Specific Language** (DSL)
-  * The functions and classes that are used by researchers to define a dataset.
-* **Query Model** (QM)
-  * The objects that are created by the DSL.
+**Backend.**
+The logical representation of a collection of EHR data.
+A Backend is associated with a provider of EHR data, a data store, and a Query Language.
+For example, the TPP Backend is associated with TPP, Microsoft SQL Server, and Transact-SQL.
```
evansd, line 6:
```suggestion
A class which translates between the table structure we present to the user (the ehrQL tables) and the table structure in an EHR provider's database (the database tables).

Each Backend is specific to an EHR provider and a Query Engine, which tells Data Builder how to talk to the kind of database used by the EHR provider. For instance the TPP backend uses the MSSQLQueryEngine which knows how to connect to Microsoft SQL Server and speak Transact-SQL to it.

A Backend is not required in order to connect to a database. If the structure of the database tables exactly matches the structure of the ehrQL tables then there is no translation to be done. In that case, we can query the database just be specifying the appropriate Query Engine.
```

GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
...
+**Dataset Definition.**
+Describes, using ehrQL, the EHR data to extract from a Backend.
+
+**Domain-Specific Language (DSL).**
+A programming language designed for a specific application domain.
+For example, ehrQL is a DSL;
+it is a programming language designed for the specific application domain of EHR data.
+
+**Dummy Data.**
+Randomly-generated data used as a substitute for the EHR data to extract from a Backend.
```
evansd, line 21:
```suggestion
Randomly-generated data used as a substitute for the EHR data when generating a dataset from a dataset definition.
```

GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
...
+
+**Electronic Health Records (EHR) data.**
+Data about the health of the patients in a population, stored electronically (i.e. digitally).
+
+**Measure.**
+A quotient (a numerator divided by a denominator) for a given time interval and period (e.g. each month for a year).
+
+**Query Engine.**
+Translates the Query Model into the Query Language associated with a Backend.
+For example, a Query Engine translates the Query Model into Transact-SQL, which is associated with the TPP Backend.
```
evansd, line 36: Takes a query (a graph of connected Query Model objects), connects to a database and returns the corresponding data. Usually this will be a relational database speaking some dialect of SQL. But it doesn't have to be.

GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
...
+
+**Query Language.**
+A programming language designed for querying a data store.
+For example, ehrQL.
+
+**Query Model.**
+It consists of the objects that are created by ehrQL and passed to the Query Engines.
```
evansd, line 43:
```suggestion
The core data structure by which Data Builder represents a query. It is responsible for defining the semantics of all the operations we support on the data, and for ensuring that only semantically valid queries can be constructed. When a user writes ehrQL this constructs a graph of Query Model objects which are then passed to a Query Engine for execution. This provides a clear boundary between the syntactic surface of ehrQL (where expressiveness and user convenience are paramount) and the underlying semantics.
```

GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
...
-* **Query Model** (QM)
-  * The objects that are created by the DSL.
+**Backend.**
+The logical representation of a collection of EHR data.
+A Backend is associated with a provider of EHR data, a data store, and a Query Language.
+For example, the TPP Backend is associated with TPP, Microsoft SQL Server, and Transact-SQL.
+
+**Contract.**
+A specification for a collection of EHR data.
+A Contract is fulfilled by a Table.
```
evansd, line 10: We don't _really_ have contracts as entity any more. Obviously there is a `contracts/` directory, but all it's got in it are tables just like those in `tables/`.

I think the idea is that a contract would be something like "a table definition which is implemented by more than one EHR provider". But I don't we've fully settled what we're doing here yet.

GLOSSARY.md
```diff
@@ -1,6 +1,50 @@
...
+**Query Language.**
+A programming language designed for querying a data store.
+For example, ehrQL.
+
+**Query Model.**
+It consists of the objects that are created by ehrQL and passed to the Query Engines.
+
+**Table.**
+The conceptual representation of a collection of EHR data.
+A Table fulfils a Contract.
```
evansd, line 47: I think it's worth distinguishing two kinds of table here:

**ehrQL table**
Describes the structure and types of some EHR data. User's can write queries against this data using ehrQL.

**Database table**
Table in the SQL sense as stored in the database of an EHR provider. These are not directly exposed to users. Users interact with ehrQL tables and their queries are translated into queries against database tables by a Backend.

evansd (approved):
First off apologies! I've dumped a whole load of thoughts into here. Part of the reason I've been putting off tackling this is there are quite a few different concepts floating around and working out the right way to describe them and their interrelations is not a straightforward task. Feel free to rework anything I've written if you can express it in a way which makes more sense to non-Daves.

iaindillingham:
I've incorporated your changes, @evansd, but have reworked them because I am a non-Dave. I've tried to separate pedantry from reworking-proper. I'll pause before merging, to give you time to revisit, should you wish.

evansd (approved):
Thanks for doing this Iain, it all looks good.

PR #1148, opened 2023-03-14
feat: Add place of death to ONS deaths table

inglesp:


iaindillingham (changes requested):
I think that for consistency, the column name in the backend should match that in the database.

iaindillingham (approved):

PR #1150, opened 2023-03-14
Support `where()` after `sort_by()` in ehrQL

evansd:
It turns out to be quite natural to want to do this. I have seen several of the data team assume that you can and try to refactor things along these lines. This includes, on occasion, myself.

Becky had a PR which used a [clever trick](https://github.com/opensafely-core/databuilder/pull/981) to support this without changing the query model. But I think it's better if remove the restriction from the query model rather than working around it.


tests/generative/variable_strategies.py
```diff
@@ -408,11 +408,16 @@ def filtered_table(draw):
 
     @st.composite
     def sorted_frame(draw):
-        # select a table which may already have been filtered
-        source = draw(many_rows_per_patient_frame())
-        # Now apply 1-3 sorts
-        for _ in range(draw(st.integers(min_value=1, max_value=3))):
-            source = draw(sort(source))
+        # Decide how many Filters and Sorts we're going to apply
+        num_filters = draw(st.integers(min_value=1, max_value=6))
```
rebkwok, line 412: Do we want a sorted frame to also always be filtered at least once?  The previous `many_rows_per_patient_frame()` source could be a simple `select_table()` with no filters as well as a filtered table

tests/generative/variable_strategies.py
```diff
@@ -408,11 +408,16 @@ def filtered_table(draw):
...
+        num_sorts = draw(st.integers(min_value=1, max_value=3))
+        # Mix up the order of operations
+        operations = [filter_] * num_filters + [sort] * num_sorts
+        shufffled_operations = draw(st.permutations(operations))
```
rebkwok, line 416: nit: extra f in shuffled

rebkwok (approved):

tests/generative/variable_strategies.py
evansd: Oh, good point. I guess just reducing `min_value` to 0 here will do what we want, right?

tests/generative/variable_strategies.py
evansd: :joy: The perils of autocomplete propagating my typos! That said, I kind of like it: I think it implies that the list is _even more_ shuffled than normal.

inglesp: image

tests/unit/query_model/test_nodes.py
```diff
@@ -443,3 +437,22 @@ def test_comparisons_between_value_nodes_are_strict():
...
+        PickOneRowPerPatient(events, Position.FIRST)
+
+
+def test_can_pick_row_from_sorted_and_filtered_table():
+    events = SelectTable("events", EVENTS_SCHEMA)
+    date = SelectColumn(events, "date")
+    code = SelectColumn(events, "code")
+    sorted_by_date = Sort(events, date)
+    filtered_by_code = Filter(sorted_by_date, Function.EQ(code, Value("abc123")))
+    assert PickOneRowPerPatient(filtered_by_code, Position.FIRST)
```
inglesp, line 458: Is it worth having a test for the other way round too -- filter then sort?

inglesp (approved):

tests/unit/query_model/test_nodes.py
evansd: We cover that in a few places already e.g.
https://github.com/opensafely-core/databuilder/blob/fd05ad711afbb0edaffe5d80c3c494d39b6c932c/tests/unit/query_model/test_nodes.py#L299-L303

And
https://github.com/opensafely-core/databuilder/blob/fd05ad711afbb0edaffe5d80c3c494d39b6c932c/tests/unit/query_model/test_nodes.py#L51-L58

And possibly others, so I don't think we'd gain much here.

PR #1153, opened 2023-03-15
Jongmassey/table from rows

Jongmassey:
Groundwork for #793 commits cherry-picked from #1076 and rebased accordingly due to effects of changes in `main`.

This PR now includes the re-introduction of code for setup and cleanup queries removed in #1136

### DIFF
--- a/databuilder/main.py
+++ b/databuilder/main.py
@@ -154,14 +154,18 @@ def get_sql_strings(query_engine, variable_definitions):
     dialect = query_engine.sqlalchemy_dialect()
     sql_strings = []
 
-    for n, query in enumerate(setup_queries, start=1):
+    for i, query in enumerate(setup_queries, start=1):
         sql = clause_as_str(query, dialect)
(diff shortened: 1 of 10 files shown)


evansd (commented):
Thanks Jon, I think you've nailed the core functionality here. I've put some comments/questions about specific changes below. But I also think there's some different ways of factoring some of this which I find it easier to explain in code than prose so I've made a [separate branch](https://github.com/opensafely-core/databuilder/pull/1158) to cherry-pick your work into so I can explain what I'm talking about.

Maybe we could compare the two approaches and see what you think? It's also possible there are specific reasons you did things a certain way that I haven't taken account of here.

Separately to all that, I've realised that due to an [omission](https://github.com/opensafely-core/databuilder/pull/1157) in our coverage check we weren't being warned that `InlinePatientTable` wasn't been covered by the generative tests. I've fixed the coverage check but now there'll be further work integrating this new type of table into the generative tests.

PR #1154, opened 2023-03-16
fix: Remove deprecated ehrQL methods

rebkwok:
Remove the take, drop and set_population aliases now that all external studies have been updated.

Also required the external studies to be updated, and some changes to the update_external_studies script to accommodate them.

### DIFF
--- a/databuilder/query_language.py
+++ b/databuilder/query_language.py
@@ -2,7 +2,6 @@ import dataclasses
 import datetime
 import enum
 import functools
-import warnings
 from collections import ChainMap
 from typing import Union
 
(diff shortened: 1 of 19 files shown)


iaindillingham (approved):
Could I check my understanding? I think that (some) external studies exist in the opensafely org, and that their dataset definitions are copied here as acceptance tests. I think that we copy them here whenever the need arises, which is why many of the changes introduced by 3ab63edb4c4d49a4b03f7be1ee2447fb5159bc85 are unrelated to the deprecated ehrQL methods. To what extent is that correct?

evansd:
Yes, that's basically correct except it's not just when the need arises that we copy them in. There's a nightly scheduled action which creates a PR to update them. That's failing at the moment though, I think due to one of the test dataset definitions being deleted:
https://github.com/opensafely-core/databuilder/pull/1146

rebkwok:
@iaindillingham we have `tests/acceptance/test_external_studies.py`, which makes sure that dataset definitions in external studies work (in the sense of not erroring, at least).  It also defines the minimal external study files that are needed in order to test the dataset definitions. Then there's a script `update_external_studies` that vendors in those external study files.

We have a github action that runs on a schedule to fetch the external study code and open a PR if any of it has changed, but I needed to update them here so that the tests would pass - partly because I'd missed the test-age-distribution study, so it still had a `take` method in it.  Also, one of the openprompt studies has since been updated so that it relies on an output file, and that was causing the external study tests to fail too.

Issue #964, opened 2023-01-24
Production dependencies not getting updated by Dependabot


StevenMaude:
Relates to the past issue #579.

We currently have quite outdated dependencies.

Dependabot is currently failing on `pyproject.toml`:

```
updater | ERROR <job_585486053> Error processing pydocstyle[toml] (Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | ERROR <job_585486053> Using indexes:
updater | <job_585486053>   https://pypi.org/simple
updater | <job_585486053> 
updater | <job_585486053>                           ROUND 1                           
updater | <job_585486053> /usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/pip/_internal/req/req_install.py:866: PipDeprecationWarning: DEPRECATION: Constraints are only allowed to take the form of a package name and a version specifier. Other forms were originally permitted as an accident of the implementation, but were undocumented. The new implementation of the resolver no longer supports these forms. A possible replacement is replacing the constraint with a requirement. Discussion can be found at https://github.com/pypa/pip/issues/8210
updater | <job_585486053>   deprecated(
updater | <job_585486053> Traceback (most recent call last):
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/bin/pip-compile", line 8, in <module>
updater | <job_585486053>     sys.exit(cli())
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
updater | <job_585486053>     return self.main(*args, **kwargs)
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/click/core.py", line 1055, in main
updater | <job_585486053>     rv = self.invoke(ctx)
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
updater | <job_585486053>     return ctx.invoke(self.callback, **ctx.params)
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/click/core.py", line 760, in invoke
updater | <job_585486053>     return __callback(*args, **kwargs)
updater | <job_585486053>   File "/usr/local/.pyenv/versions/3.9.16/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
[...]

PR #1156, opened 2023-03-16
chore: Update all dependencies with `pip-compile`

StevenMaude:
Because Dependabot isn't quite working properly: see #964.


StevenMaude:
The only issue here is with a deprecation warning from structlog.

From reading around the [structlog documentation](https://www.structlog.org/en/stable/thread-local.html) and hynek/structlog#487, and searching around the Data Builder source, I *think* this is safe to remove as mentioned in the commit message in 139e98cfe4bbe906a828985f1810f1b01d27ff8a, but please correct me if not.

iaindillingham (approved):
I'm going to merge this, as I believe you're away until the 27th @StevenMaude.

Issue #954, opened 2023-01-18
Add Sort operation to new generative tests


rebkwok:

rebkwok:
Fixed by #987

PR #1157, opened 2023-03-17
Fix comprehensive gentests coverage check

evansd:
Previously, we only picked up operations added to the module's `__all__` list. As an example, we forgot to add the new `InlinePatientTable` node to this list and as a result we didn't realise it wasn't covered by the generative tests.

### DIFF
--- a/databuilder/query_model/nodes.py
+++ b/databuilder/query_model/nodes.py
@@ -19,6 +19,7 @@ __all__ = [
     "Value",
     "SelectTable",
     "SelectPatientTable",
+    "InlinePatientTable",
     "SelectColumn",
     "Filter",
     "Sort",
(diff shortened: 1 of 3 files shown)


iaindillingham (approved):
Should there be an issue for adding generative tests for `InlinePatientTable`? Similar to #954, for example? (If the answer is "yes", then I'll create one 🙂. I wanted to check my understanding first.)

evansd:
I don't think it wants a separate issue. As `main` currently stands `InlinePatientTable` is, like `CombineAsSet`, an operation that exists in the query model but is not yet supported by any query engines so it's correctly excluded from the tests.

There's an open PR which adds support for it, and so the gentest changes should go in as part of that rather than as a separate piece of work.

PR #1158, opened 2023-03-17
Support `table_from_rows`

evansd:
This is the first part of addressing:
 * https://github.com/opensafely-core/databuilder/issues/793


Jongmassey:
The LHS and RHS tables have been created with different collations on one or both of the `string`/`varchar` columns. If the collation has not been explicitly specified in the `CREATE TABLE` statement, then it will inherit the default collation of the database in which it was created. I suspect one of them is in TempDB and the other is in a "normal" database which has a collation which differs from the _server_ default collation (which TempDB inherits unless overridden). 

If one or more of the tables is not explicitly created but is instead created via a `SELECT [...] INTO [...]` statement then the created table columns will adopt the collation of the source `FROM` table in the query. If the source is not another table but literals in, most commonly, a `VALUES()` clause then... I don't know!

Solutions that spring to mind: 
* explicit collations where tables are created
* re-collation in `WHERE` clauses and `ON` of `JOIN`s
* reconfiguring TempDB/the other DB (do we _really_ want to do case-sensitive comparisons?) and documenting this for when it inevitably reappears in the future
* some variation of the above whereby collations are compared by looking at the table metadata and applying the required transformations

Jongmassey:
Hmmm: server, tempdb, master, and OPENCorona default collations are all  `Latin1_General_CI_AS` which is the default collation for an English(UK) installation. 

```sql

CREATE TABLE #foo(bar varchar(100) not null);

SELECT c.collation_name
FROM tempdb.sys.columns c 
JOIN tempdb.sys.tables t
  ON c.object_id = t.object_id
WHERE t.name like N'#foo';
```

also gives `Latin1_General_CI_AS` as the collation for `bar`.


I'm now wondering where `SQL_Latin1_General_CP1_CI_AS` (US English default installation) and `SQL_Latin1_General_CP1_CS_AS` are coming into play.

evansd:
There are some defaults set here:
https://github.com/opensafely-core/databuilder/blob/18336aeab4e1d7c124a4ab88e90052b85e882bcf/tests/support/mssql/setup.sql#L1-L7

I don't know where these specific values came from, but they may be related to the defaults on the TPP server.

Jongmassey:
Right! That makes a lot of sense. I suspect that `SQL_Latin1_General_CP1_CI_AS` will be the default server collation for the mssql docker container and this is where the mismatch is arising. I'll just do some quick checks to be sure.

Jongmassey:
from `just connect-to-mssql`

```
1> select name, collation_name from sys.databases;
2> go
name                                                                                                                             collation_name                                                                                                                  
-------------------------------------------------------------------------------------------------------------------------------- --------------------------------------------------------------------------------------------------------------------------------
master                                                                                                                           SQL_Latin1_General_CP1_CI_AS                                                                                                    
tempdb                                                                                                                           SQL_Latin1_General_CP1_CI_AS                                                                                                    
model                                                                                                                            SQL_Latin1_General_CP1_CI_AS                                                                                                    
msdb                                                                                                                             SQL_Latin1_General_CP1_CI_AS                                                                                                    
test                                                                                                                             SQL_Latin1_General_CP1_CS_AS                                                                                                    
temp_tables                                                                                                                      SQL_Latin1_General_CP1_CS_AS    
```



Why are the collations configured in the mssql test setup different from those on TPP?

evansd:
I don't know where that specific collation came from, other than I introduced it here:
https://github.com/opensafely-core/databuilder/pull/582

In TPP, different columns have different collations, e.g. see:
https://github.com/opensafely-core/cohort-extractor/blob/43cbbd19f2b888d463c9dce1b7a36707f268890f/tests/tpp_backend_setup.py#L122-L133
```py
class MedicationDictionary(Base):
    ...
    DMD_ID = Column(String(collation="Latin1_General_CI_AS"))


class CodedEvent(Base):
    ...
    CTV3Code = Column(String(collation="Latin1_General_BIN"))
```

I still haven't quite worked out how to deal with this. When using string literals and `IN` clauses it doesn't make any difference. But if we start creating temporary tables with user uploaded data in then it becomes relevant.

rebkwok (approved):

PR #1159, opened 2023-03-17
Add an introduction to the generative tests

rebkwok:
This is an attempt to describe the generative tests and how they work. The intention is that it serves both as an introduction for people encountering them for the first time, and a reference for later, when we've all forgotten why they do what they do.

### DIFF
--- a/tests/generative/README.md
+++ b/tests/generative/README.md
@@ -19,8 +19,9 @@ from the same inputs.
 ## Terminology
 
 We are used to thinking about query model things such as `Series` as representing a concrete column
-in a table. The generative tests require a mental shift from thinking in terms of `strategies` - i.e.
-a `series` strategy is a recipe for defining a `Series`, rather than a `Series` itself.
+in a table. The generative tests require a mental shift from thinking in terms of concrete objects to
+thinking in terms of  `strategies` - i.e. a `value` strategy is a recipe for generating a `Value`,
+rather than a `Value` itself.
 
 ### Building strategies
 
--- a/tests/generative/README.md
+++ b/tests/generative/README.md
@@ -198,17 +198,16 @@ def variable(patient_tables, event_tables, schema, value_strategies):
 
 Let's take a closer look at the elements of the `valid_variable`.
 
-`type_` is a strategy for choosing a type, sampled from the types used as the keys for
-`value_strategies`:
+`type_` is a type, drawn using the `any_type()` strategy, that chooses one of the types used as the keys for `value_strategies`:
 ```
     def any_type():
         return st.sampled_from(list(value_strategies.keys()))
 ```
 
-`frame` is a strategy for choosing a frame with one row per patient. That could be a
-simple patient table (picked from one of the two `SelectPatientTable` nodes defined by our data
-strategies), or a `PickOneRowPerPatient` node, which is the result of a number of sorting and
-filtering operations.
+`frame` is a frame with one row per patient, drawn using the `one_row_per_patient_frame()`
+strategy. That could be a simple patient table (picked from one of the two
+`SelectPatientTable` nodes defined by our data strategies), or a `PickOneRowPerPatient`
+node, which is the result of a number of sorting and filtering operations.
 
 Each of the sort and filter operations are themselves defined by strategies which require a
 series to sort/filter on, so a frame strategy can quickly become deeply nested. The nesting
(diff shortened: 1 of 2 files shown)


tests/generative/README.md
```diff
@@ -0,0 +1,399 @@
...
+## variable strategies
+
+The variable strategies are the most complex part of databuilder's generative test strategies.
+
+A variable is defined by calling `variable()` in [`variable_strategies.py`](variable_strategies.py),
+with the tables, schema and value strategies as described above.
+
+`variable` defines lots of inner functions, each of which returns a strategy for creating the
+thing it is named for, not the thing itself.
+
```
iaindillingham, line 154: An example would be useful.

```suggestion
For example, the `value` inner function returns a strategy for creating `Value` objects, and not a `Value` object itself.

```

iaindillingham (approved):
This is great; thank you for writing it.

tests/generative/README.md
```diff
@@ -0,0 +1,399 @@
...
+        # define contraints for possible series strategies
+        series_constraints = {
+            exists: ({bool}, DomainConstraint.PATIENT),
+            count: ({int}, DomainConstraint.PATIENT),
+        }
+        ...
+        def constraints_match(s):
+            ...
+
+        # find possible series stratgeies that match contraints, and choose one
```
sebbacon, line 246:
```suggestion
        # find possible series strategies that match constraints, and choose one
```

evansd (approved):
This is really good. I like the way it starts with the gentle stuff and slowly ramps up to `binary_operation_with_types()`! The examples really help to make things clear here.

I've made one single word suggested change (although it's possible that I've misunderstood the sentence here).

I've also written a bit of text below which is my way of justifying to myself why this stuff is so complicated. I'm not sure if it's helpful and, if it is, I'm not sure where it would best go. But feel free to make use of it or not as you see fit.

> The most difficult part of the generative tests is teaching Hypothesis how to construct queries out of query model objects. The query model has various constraints or validation rules which tell you, given a particular object, what sort of operations you can apply to it and what other kinds of object you can combine it with.
> 
> One way of trying to produce valid queries is just to generate lots of different objects, try sticking them together and reject any structures that are invalid. This is what we did at first. It has the benefit of being simple, but as the number of different types in the query model grows it becomes harder and harder to randomly generate valid examples and eventually it stops working altogether.
> 
> The approach we take now involves, effectively, applying the query model validation rules _backwards_. We start by deciding on the sort of object we want to end up with e.g. a Series of one-row-per-patient integers. Then we ask, given we've got this object, what sorts of operation could validly produce an object like that? We then get Hypothesis to pick one. That operation in turn requires other inputs and so, again, we ask what sorts of operation could validly produce it and get Hypothesis to pick one. This process repeats until we reach a "terminal node" i.e. an operation which doesn't require any inputs. (And if Hypothesis doesn't naturally give us a terminal node after reaching a certain depth, we force it to choose one.)
> 
> This is very efficient at generating large and complex queries for testing. But applying the validation rules backwards is fundamentally quite a mind-stretching exercise so don't be surprised if it takes a little while for everything to fall into place.

tests/generative/README.md
```diff
@@ -0,0 +1,400 @@
...
+Note that because the inputs are automatically generated, we can't test that the results are
+*correct*, but we rely on other (unit, integration, acceptance) tests to check the exact outputs
+produced by the query engines, and we assume that the engines are sufficiently different in
+implementation that it is a sufficient test to check that they always produce the same results
+from the same inputs.
+
+## Terminology
+
+We are used to thinking about query model things such as `Series` as representing a concrete column
+in a table. The generative tests require a mental shift from thinking in terms of `strategies` - i.e.
```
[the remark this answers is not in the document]
rebkwok: I've reworded this.  I only wrote half the sentence I'd intended to write :)

tests/generative/README.md
```diff
@@ -0,0 +1,425 @@
...
+        type_ = draw(any_type())
+        frame = draw(one_row_per_patient_frame())
+        return draw(series(type_, frame))
+
+    return valid_variable()
+```
+
+Let's take a closer look at the elements of the `valid_variable`.
+
+`type_` is a strategy for choosing a type, sampled from the types used as the keys for
```
benbc, line 201: This isn't quite right. `any_type()` returns the strategy you describe. `type_` is an actual type, drawn from the strategy. (The same applies to `frame` below.)

tests/generative/README.md
```diff
@@ -0,0 +1,425 @@
...
+
+`count` returns an int series with one row per patient.
+
+The `frame` we've passed in matches the domain constriants for both strategies (it's a
+one-row-per-patient frame). However, `exists` produces a bool series, and we need a strategy
+that produces an int series, so `count` is the only possible strategy that matches. `series()`
+will draw from the `count` strategy.
+
+Note that the `type_` and `frame` are always passed on as arguments to the selected series
+strategy, and each individual series operation strategy has the same function signature.
```
benbc, line 289:
```suggestion
strategy, so that all series operation strategies have the same function signature.
```

Maybe slightly clearer?

benbc (approved):
This is really good, Becky. Thanks for going to the effort to write it all up (and to understand it in the first place :-)).

PR #1163, opened 2023-03-23
Add sketch of potential measures API

evansd:
Included is a sketch of what I think an ehrQL [measures API](https://github.com/opensafely-core/databuilder/issues/816) could look like. It's a translation of the [vaccine-declined measures](https://github.com/opensafely/covid-vaccine-declined-variation/blob/46281b8a01cfec375fb49a3a9dfd9e7cab14bcf2/analysis/study_definition.py) which happened to be the first simple example I stumbled across.

I've done enough proof-of-concept hacking to convince myself that we can implement this. And I've got some ideas of varying levels of complexity and sophistication as to how we could implement it [efficiently](https://github.com/opensafely-core/databuilder/issues/42) (at least, more efficiently than we currently do – which isn't hard).

You'd invoke this using something like:
```
databuilder generate-measures measure_definitions.py --output measures.arrow
```

Notice that there's only a single output file here. I think it would be quite sensible to have some way of outputting multiple files, one per measure, by specifying an output file pattern or something. But I'm very keen for that not to be _required_. Indeed, I'd like to follow `generate-dataset` here and just write CSV to stdout if no output file is specified at all.

I think reducing to the bare minimum the ceremony required to get _something_ out of a command is an important part of minimising the learning curve.

This does mean the resulting output file structure is a little odd in that it may have group-by columns which aren't relevant for some of the measures (because they are not grouped by that value). But these can just take NULL values and I don't think there's any practical problem here.

For example, a small part of the output for these measures might look like:

measure_id | interval_start | interval_end | value | numerator | denominator | age_band | practice | event_code | region
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
total | 2020-12-01 | 2020-12-31 | 0.05 | 50 | 1000 | 50-55 |   |   |  
total | 2021-01-01 | 2021-01-31 | 0.075 | 60 | 800 | 50-55 |   |   |  
total | 2020-12-01 | 2020-12-31 | 0.025 | 40 | 1600 | 55-60 |   |   |  
total | 2021-01-01 | 2021-01-31 | 0.08 | 80 | 1000 | 55-60 |   |   |  
practice | 2020-12-01 | 2020-12-31 | 0.05 | 50 | 1000 | 40-45 | 123456 |   |  
practice | 2020-12-01 | 2020-12-31 | 0.075 | 60 | 800 | 40-45 | 789101 |   |  
... |   |   |   |   |   |   |   |   |  

I haven't thought about small number suppression at this stage.

### DIFF
--- a/measure_definitions.py
+++ b/measure_definitions.py
@@ -1,19 +1,11 @@
-from databuilder.ehrql import case, when
-
-# I'm importing from a separate `measures` module here, but I think there's an argument
-# that this should all be importable from the `ehrql` namespace
-from databuilder.measures import (
-    INTERVAL_END_DATE,
-    INTERVAL_START_DATE,
-    Intervals,
-    Measure,
-)
+from databuilder.ehrql import Measures, case, when
 from databuilder.tables.beta.tpp import (
     clinical_events,
     ons_deaths,
     patients,
     practice_registrations,
 )
+
 
 # Note that the INTERVAL_START_DATE/INTERVAL_END_DATE values are ehrQL parameters (or
 # placeholders). From the point of view of the rest of ehrQL they are just
@@ -31,11 +23,13 @@ from databuilder.tables.beta.tpp import (
 # DEMOGRAPHICS
 #
 
-practice = practice_registrations.for_patient_on(INTERVAL_START_DATE)
+practice = practice_registrations.for_patient_on(Measures.INTERVAL_START_DATE)
 
-has_died = ons_deaths.where(ons_deaths.date <= INTERVAL_START_DATE).exists_for_patient()
+has_died = ons_deaths.where(
+    ons_deaths.date <= Measures.INTERVAL_START_DATE
+).exists_for_patient()
 
-age = patients.age_on(INTERVAL_START_DATE)
+age = patients.age_on(Measures.INTERVAL_START_DATE)
 
 age_band = case(
     when((age >= 0) & (age <= 4)).then("0-4"),
@@ -86,7 +80,11 @@ codelist = [
 e = clinical_events
 event = (
     e.where(e.snomedct_code.is_in(codelist))
-    .where(e.date.is_on_or_between(INTERVAL_START_DATE, INTERVAL_END_DATE))
+    .where(
+        e.date.is_on_or_between(
+            Measures.INTERVAL_START_DATE, Measures.INTERVAL_END_DATE
+        )
+    )
     .sort_by(e.date)
     .first_for_patient()
 )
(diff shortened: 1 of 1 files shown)


measure_definitions.py
```diff
@@ -0,0 +1,165 @@
...
+#         (date(2020,1,15), date(2020,1,21)),
+#     ]
+#
+# `Intervals` is just acting as a namespace here; it felt cleaner to me to keep these
+# together rather than having them as free floating functions. But I'm open to other
+# suggestions.
+
+intervals = Intervals.months_between("2020-12-01", "2021-04-01")
+
+
```
sebbacon, line 110: By analogy to how we define `dataset` for export, perhaps we should do 

```
measures = Measures()
measures.total = Measure(
                     numerator=had_matching_event, 
                     ...)
```


Perhaps `group_by` could have a similar flavour

total = Measure("

measure_definitions.py
```diff
@@ -0,0 +1,165 @@
...
+#         (date(2020,1,1), date(2020,1,7)),
+#         (date(2020,1,8), date(2020,1,14)),
+#         (date(2020,1,15), date(2020,1,21)),
+#     ]
+#
+# `Intervals` is just acting as a namespace here; it felt cleaner to me to keep these
+# together rather than having them as free floating functions. But I'm open to other
+# suggestions.
+
+intervals = Intervals.months_between("2020-12-01", "2021-04-01")
```
sebbacon, line 108: I worry slightly about inclusive/exclusive ambiguity. Is there a way to resolve that? I can't think of any!

measure_definitions.py
evansd: I wondered about that too, but I felt like having both `Measures` and `Measure` floating around was liable to be confusing.

milanwiedemann:
I like the option to have (1) one output file with all measures and (2) multiple files. just wanted to add some thoughts to the output file structure. here's an example structure that I started to use for my measures (I use the measures framework from cohort extractor and then restructure the data to match the format below). it's the same underlying data (just a bit _longer_ but can easily be pivoted to the example above by taking the column names from `variable` and the values from `value`. 

the main advantage for me was that it's easier to join measures files from different study definitions (or even research projects) into one file that I can continue working with (analyses, plots, ...) 

measure_id | interval_start | interval_end | group | variable | value |
-- | -- | -- | -- | :-- | --: |
age_band | 2020-12-01 | 2020-12-31 | 50-55 | numerator | 50 
age_band | 2020-12-01 | 2020-12-31 | 50-55 | denominator |  1000
age_band | 2020-12-01 | 2020-12-31 | 50-55 | percentage |  0.05
age_band | 2021-01-01 | 2021-01-31 | 50-55 | numerator | 60 
age_band | 2021-01-01 | 2021-01-31 | 50-55 | denominator |  800
age_band | 2021-01-01 | 2021-01-31 | 50-55 | percentage |  0.075
practice | 2020-12-01 | 2020-12-31 | 111 | numerator | 50 
practice | 2020-12-01 | 2020-12-31 | 111 | denominator |  1000
practice | 2020-12-01 | 2020-12-31 | 111 | percentage |  0.05
practice | 2021-01-01 | 2021-01-31 | 111 | numerator | 60 
practice | 2021-01-01 | 2021-01-31 | 111 | denominator |  800
practice | 2021-01-01 | 2021-01-31 | 111 | percentage |  0.075
practice | 2020-12-01 | 2020-12-31 | 222 | numerator | 50 
practice | 2020-12-01 | 2020-12-31 | 222 | denominator |  1000
practice | 2020-12-01 | 2020-12-31 | 222 | percentage |  0.05
practice | 2021-01-01 | 2021-01-31 | 222 | numerator | 60 
practice | 2021-01-01 | 2021-01-31 | 222 | denominator |  800
practice | 2021-01-01 | 2021-01-31 | 222 | percentage |  0.075

milanwiedemann:
oh sorry for late reply, yes all the rest made sense to me! the only variable name that I think could be clearer is `value` – I would use something more descriptive, maybe `percentage`

inglesp:
This all looks sensible.  I wonder if we can do something further if we force users to always use the same intervals and denominator (which I would guess is quite likely).

Here, we make use of a `Dataset`, and then the numerators and denominator are defined in terms of columns on that `Dataset`.

```
dataset = Dataset()

has_died = ons_deaths.where(ons_deaths.date <= INTERVAL_START_DATE).exists_for_patient()
registration = practice_registrations.for_patient_on(INTERVAL_START_DATE)
dataset.age_band = patients.age_band_on(INTERVAL_START_DATE)

dataset.set_population(
    registration.exists_for_patient()
    & ~has_died
    & patients.sex.is_in(["male", "female"])
    & (dataset.age_band != "missing")
)

dataset.practice = registration.practice
dataset.region = registration.nuts1_region_name

e = clinical_events
dataset.event_code = (
    e.where(e.snomedct_code.is_in(codelist))
    .where(e.date.is_on_or_between(INTERVAL_START_DATE, INTERVAL_END_DATE))
    .sort_by(e.date)
    .first_for_patient()
    .snomedct_code
)


measures = Measures(
    dataset=dataset,
    intervals=intervals,
    # this could be a dict mapping measure name to numerator columns
    numerators = [
        ["age_band"],
        ["age_band", "event_code"],
        ["age_band", "practice"],
        ["age_band", "region"],
    ],
    # could also have numerator=[...] for single measure
    denominator=["population"],
)
```

What do you think?

evansd:
Thanks @milanwiedemann 

> the only variable name that I think could be clearer is `value` – I would use something more descriptive, maybe `percentage`

Yes, I realised afterwards this was a bad name. I was thinking of using `ratio`. I think we'd need to multiply by 100 if we wanted to use `percentage`.

evansd:
Thanks @inglesp.  I'm afraid I don't quite understand your example though:
```
    numerators = [
        ["age_band"],
        ["age_band", "event_code"],
        ["age_band", "practice"],
        ["age_band", "region"],
    ],
```

These are grouping columns – aren't they? – not numerator columns. The numerator is the `had_matching_event` variable in my example, but I don't see the equivalent anywhere in yours. Apologies if I'm missing something obvious.

More generally, I'd be reluctant to see a return to the days of defining variables in one place and referring to them using strings elsewhere, even where that makes things syntactically neater. And defining a dataset in order to define measures feels to me like Stockholm syndrome with respect to the old API, rather than something that makes sense in its own right.

evansd:
I've pushed a new version with changes prompted by some of Seb's suggestions, and I think there's some convergent evolution with Peter's proposal as well. The changes are:

 * We now have an explicit `Measures` object, rather than using a list.

 * To avoid having both `Measures` and `Measure` floating around, we create new measures using a `define_measure()` methods on the `Measures()` object.

 * We could bikeshed the name here, but I went with "define" for consistency with `define_population()`.

 * I've moved the import to be `from databuilder.ehrql import Measures`.

 * I've moved `INTERVAL_START_DATE` and `INTERVAL_END_DATE` to be class attributes of `Measures` purely to keep the namespace clean.

 * Likewise I've moved the interval generating methods onto `Measures` for the same reason.

 * I've changed `months_between("2020-12-01", "2021-04-01")` to `by_month(start="2020-12-01", end="2021-04-01")` which I felt was clearer.

evansd:
I've pushed another version which uses default values in the `Measures` constructor to avoid repetition in the individual measure definitions, while still allowing variation if that's needed.

I'm not totally against requiring measures in the same file to have common denominators and/or intervals, if that makes the API significantly nicer. But my preference is for it to be possible to define heterogeneous measures together in a single place because that gives us the option later to find clever ways of exploiting commonalities between them so as to execute them more efficiently. If users are forced to define them in different files then that will never be an option.

But as I say, that takes second place to making the API "nice", for some value of "nice".

evansd:
Actual API implemented here:
 * https://github.com/opensafely-core/ehrql/pull/1276

PR #1169, opened 2023-03-29
Upgrade to Python 3.11

evansd:
The main user facing benefit here are the new fine-grained error locations:
https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep657

And I can see us maybe wanting to make use of the exception "notes" feature:
https://docs.python.org/3/library/exceptions.html#BaseException.add_note

The compounded performance improvements from Python 3.10 and 3.11 are also quite significant, although that's unlikely to make much difference to users other than in the case of generating dummy data.

Beyond that, we get the opportunity to clean up some craft and generally be all modern and forward looking.


rebkwok (commented):
Does the pages-deployment.yml workflows also need 3.9 -> 3.11?  It's building only the databuilder docs, which involves also running some python scripts in this repo.  (The check-docs-links.yml should presumably stay as 3.9 because they build the main docs in order to check all the links)

evansd:
Oh, thanks for spotting these. I did a find and replace on all the workflows but these got missed because they use single quotes rather than double!

I'll fix the `pages-deployment` workflow. The `check-docs-links` workflow feels a bit like it's in the wrong repo to me. But I'm happy to just leave it as it is for now.

rebkwok (approved):

StevenMaude:
@rebkwok:

> the check-docs-links one does feel like it could just go with the rest in the docs repo

The main reason that I had: it seems a little odd for a failing link check to be due to documentation source Markdown files that are actually in another repository. You then have the fuss of opening a PR in another repository, getting it merged, and then rechecking in the current repository. But I don't think that's a huge problem.

Another reason to retain a check would be if we want to ever implement checking new/modified links on PRs before merge. I don't know if that's more trouble than it's worth though (because you may get a failing PR due to a temporary site outage solely because a link fails, which could be transitory).

***

Also, I *think* that the main documentation repository can just be moved to Python 3.11. The only reason to keep it Python 3.9 was to tie it to the version of Python that Data Builder was using. But neither Data Builder nor cohort-extractor are installed in the documentation repository anyway now, following all the move.

PR #1187, opened 2023-04-05
Jongmassey/table_from_file_docs

Jongmassey:
Include previously-omitted `@table_from_file` in table of contents,
and make supporting changes to its tests and the doc generation.

The argument capturing used to generate docs from spec tests
is incompatible with the pytest `tmp_file` fixture as used previously.
Instead, a `TemporaryDirectory` from `tempfile` is used.

`build_paragraph()` changed to use regexes to handle variable levels
on indentation from the required `with` clause of this context manager.

Spec tests for `@table_from_file` renamed to look better in docs


Jongmassey:
closed in favour of #1203

PR #1203, opened 2023-04-12
Improve tests for `@table_from_file`

Jongmassey:
The tests in  #1158 didn't offer much in the way of integration testing and were poorly organised; this PR fixes both of those and a couple of bugs revealed by these novel tests


tests/unit/test_query_language.py
```diff
@@ -322,20 +323,36 @@ class some_table(EventFrame):
...
 
 
-def test_table_from_file(tmp_path):
-    rows = [(1, 100), (2, 200)]
-    csv_path = tmp_path / "test.csv"
-    with csv_path.open("w") as f:
-        writer = csv.writer(f)
-        writer.writerow(("patient_id", "n"))
-        writer.writerows(rows)
+@pytest.mark.parametrize("file_extension", ["csv", "csv.gz", ".arrow"])
```
evansd, line 326: Thinking about it further, you could actually do:
```py
from databuilder.file_formats import FILE_FORMATS

...

@pytest.mark.parametrize("file_extension", FILE_FORMATS.keys())
```
Which would mean that it's automatically parameterised over all formats we support (including any new ones we add).

The only thing to watch out for is that these extensions already include the leading dot so you'd need to change the name string to `f"test_file{file_extension}"`.

evansd (approved):
:+1: Thanks Jon!

tests/unit/test_query_language.py
Jongmassey: ooh, I like this.

PR #1204, opened 2023-04-13
Fix dummy data generation for InlinePatientTables

evansd:
As well as fixing the specific issue here this PR extends the generative tests so that they attempt to generate a minimal amount of dummy data for each query. As CI ensures that these tests produce at least one instance of every node type this gives us some assurance that the `QueryInfo` class which drives the dummy data generation can handle every node type.

### DIFF
--- a/databuilder/dummy_data/query_info.py
+++ b/databuilder/dummy_data/query_info.py
@@ -6,6 +6,7 @@ from databuilder.query_model.nodes import (
     Column,
     Constraint,
     Function,
+    InlinePatientTable,
     SelectColumn,
     SelectPatientTable,
     SelectTable,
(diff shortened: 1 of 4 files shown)


databuilder/dummy_data/query_info.py
```diff
@@ -118,6 +119,13 @@ def from_variable_definitions(cls, variable_definitions):
         # For every column used in the query (sorted for consistency) …
         for column in sort_by_name(by_type[SelectColumn]):
             table = get_root_frame(column.source)
+            # We're only interested in "standard" tables here
```
Jongmassey, line 122: are we, though? Does this mean that if a user defines a dataset drawing from both one or more "standard" and one or more Inline tables, the fields from the latter will be omitted from the dummy data?

Jongmassey (approved):

PR #1214, opened 2023-04-19
feat: Tag newly built images as ehrql as well as databuilder

rebkwok:
This means that we can use `ehrql:v0` in project.yaml and `opensafely exec` in advance of the full renaming work.

Will also require (in this order): 
- update to job-runner: https://github.com/opensafely-core/job-runner/pull/606
- update to the vendored jobrunner in https://github.com/opensafely-core/opensafely-cli

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -16,6 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       image: ghcr.io/opensafely-core/databuilder
+      aliased_image: ghcr.io/opensafely-core/ehrql
     steps:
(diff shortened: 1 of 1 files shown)


evansd (approved):
Great, I think this is overall a much simpler approach.

As well as updating the vendored jobrunner in opensafely-cli, we'll need to update [this function](https://github.com/opensafely-core/opensafely-cli/blob/368de251114b2c4d410563ec451442cbd0fe285d/opensafely/pull.py#L131-L135) as well:
```py
def get_default_version_for_image(name):
    if name == "databuilder":
        return "v0"
    else:
        return "latest"
```

PR #1221, opened 2023-04-20
Rebrand: change module name

### LINKED ISSUE #1191: Rebrand to ehrQL: Python packaging changes
After this change, users will be able to write `from ehrql import Dataset`, `from ehrql.tables.beta.tpp import clinical_events` etc.

Ideally, this would not be a breaking change, and `from databuilder.ehrql import Dataset` would continue to work.  A separate ticket (#1192) covers updating study code and removing all remaining references to databuilder from the codebase.

Backwards compatibility can be achieved through retaining some `__init__.py` files in the `databuilder` directory, which import and re-export only the necessary names.

inglesp:
This renames the `databuilder` module to `ehrql` and moves the primary import point to the top level (to avoid users having to write `from ehrql.ehrql import ...`). This means that dataset definitions should now look like:
```py
from ehrql import Dataset  # ... and case, when, codelist_from_csv etc
from ehrql.tables.beta import tpp
```

Module shims are provided so that the old imports still work (as evidenced by the unmodified acceptance tests still passing).

I have done my best to expunge the phrase "Data Builder" entirely from the codebase with a few exceptions:
 * the references needed to maintain the shims (both the aliased Docker image name and Python imports);
 * the name of the Cloudflare Pages project the docs are published to (which I don't think I can change);
 * the `docs/` directory, which I have ignored as it is in the process of being rewritten (though some of the global find-and-replace operations will have fixed these in any case).

Closes #1191


evansd (commented):
Thanks for taking this on. I've made some comments below regarding the CLI entrypoint which also needs renaming to ehrQL.

The other thing that really needs doing here is that we want users to be able to write:
```py
from ehrql import Dataset
```
Rather than:
```py
from ehrql.ehrql import Dataset
```

There might be some trickiness there as it's possible we'll get cyclical import problems. Or it might just work! In either case, it needs doing I think.

The other thing is that I think it's worth doing a clean sweep here rather than leaving references to Data Builder around which we will never fix and will be a slight source of confusion forever.

There aren't _that_ many:
```
rg -i 'data\s*builder' -g '!docs' -g '!tests/acceptance/external_studies
```

(I'm ignoring the docs there as that's all changing anyway.)

If you don't have time for this I'm happy to take it on.

evansd (approved):

inglesp (commented):
Thanks for finishing the job @evansd!

PR #1226, opened 2023-04-21
fix: Move `patient_id` check to TableSchema

evansd:
It's not valid to use `patient_id` as a column on any sort of table, so it makes sense to validate it at the TableSchema level rather than on a particular node type.

And given that this is potentially a user-facing, rather than just a devloper-facing, error we should raise an exception rather than using `assert`.

### DIFF
--- a/databuilder/query_model/nodes.py
+++ b/databuilder/query_model/nodes.py
@@ -203,12 +203,6 @@ class InlinePatientTable(OneRowPerPatientFrame):
     rows: Iterable[tuple]
     schema: TableSchema
 
-    def __post_init__(self):
-        assert (
-            "patient_id" not in self.schema.column_names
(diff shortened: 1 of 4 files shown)


rebkwok:
Should we also prevent users from using `patient_id` in a dataset definition? I tried this locally:
```
from databuilder.ehrql import Dataset
from databuilder.tables.beta.tpp import patients

dataset = Dataset()
age = patients.age_on("2023-01-01")
dataset.define_population(age > 16)
dataset.age = age
dataset.patient_id = age
```
And in my output, patient_id has been overwritten with age (i.e. I get 2 identical columns)

evansd:
Oh yes, very good point. That should be on our banned names list along with `population`. I think that's an issue for a separate PR though. This was a check added specifically for users of the `table_from_file` feature, and I realised it could do with being tweaked slightly.

PR #1232, opened 2023-04-24
Fix: support first_/last_for_patient in sandbox

inglesp:
Most sandbox queries exercise code that is already well tested. However, because sandbox queries aren't always attached to a dataset, QM transformations are not applied.  In the case of picking one row per patient, this means that a PickOneRowPerPatient node won't have been transformed to a PickOneRowPerPatientWithColumns node.

### DIFF
--- a/tests/unit/query_engines/test_in_memory.py
+++ b/tests/unit/query_engines/test_in_memory.py
@@ -13,7 +13,7 @@ class events(EventFrame):
 
 def test_pick_one_row_per_patient():
     # This test verifies that picking one row per patient works without first having
-    # applied QM transformations to all variables ina dataset.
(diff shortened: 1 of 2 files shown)


tests/unit/query_engines/test_in_memory.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@table
+class events(EventFrame):
+    date = Series(date)
+
+
+def test_pick_one_row_per_patient():
+    # This test verifies that picking one row per patient works without first having
+    # applied QM transformations to all variables ina dataset.
```
evansd, line 16:
```suggestion
    # applied QM transformations to all variables in a dataset.
```

PR #1236, opened 2023-04-26
minimum_of() and maximum_of() series

Jongmassey:


ehrql/query_language.py
```diff
@@ -11,7 +11,9 @@
 from ehrql.query_model import nodes as qm
 from ehrql.query_model.column_specs import get_column_specs_from_schema
 from ehrql.query_model.nodes import get_series_type, has_one_row_per_patient
-from ehrql.query_model.population_validation import validate_population_definition
+from ehrql.query_model.population_validation import (
```
Jongmassey, line 14: black noise, to fix before finish

ehrql/query_language.py
```diff
@@ -553,7 +563,10 @@ def _wrap(qm_node):
         # superclass
         matches = [
             cls
-            for ((target_type, target_dimension), cls) in REGISTERED_TYPES.items()
```
Jongmassey, line 556: more `black` noise

ehrql/query_model/nodes.py
```diff
@@ -388,6 +388,17 @@ class In(Series[bool]):
         lhs: Series[T]
         rhs: Series[Set[T]]
 
+    # Horizontal Aggregation
+    class MinimumOf(Series[Comparable]):
+        # def __init__(self, *args):
```
Jongmassey, line 393: remove these commented bits

ehrql/query_engines/mssql_dialect.py
```diff
@@ -125,7 +132,42 @@ def get_children(self):
...
+    return sqlalchemy.select(aggregate_function(v.c.aggregate))
+
+
+# N.B SQL Server 2019+ supports GREATEST() and LEAST() so
+# could check version and pass to native function if we
+# ever upgrade
+@compiles(least, "mssql")
+def visit_least(element, compiler, **kw):
+    stmt = get_aggregate_subquery(element, sqlalchemy.func.min).label("least")
+    return compiler.process(stmt, **kw)
```
evansd, line 167: I appreciate that using custom `compiles` constructs is the way the SQLAlchemy documentation will point you, but I actually think it's not what we want to do here because it ends up adding an additional layer of abstraction to the query engine system, which is _already_ an abstraction layer over DBMSs.

So I think the way to handle this would be to have this logic in the MSSQLQueryEngine.

ehrql/query_engines/base_sql.py
```diff
@@ -411,6 +413,21 @@ def get_sql_to_first_of_month(self, node):
...
+
+    @get_sql.register(Function.MinimumOf)
+    def get_sql_minimum_of(self, node):
+        return self.get_nary_function(node, least)
+
+    def get_nary_function(self, node, function):
+        sources = [*node.sources]
+        columns = [self.get_expr(s) for s in sources]
+        types = {c.type for c in columns}
+        function.type = types.pop()
```
evansd, line 428: Does passing the type as keyword argument work? e.g.
```py
return function(*columns, type=types.pop())
```
Mutating the shared object feels a bit icky, but probably won't cause any actual problems.

ehrql/query_language.py
```diff
@@ -343,6 +355,17 @@ def parse_date_if_str(value):
         return value
 
 
+def harmonise_datelike_series(series):
```
evansd, line 358: I think the trick here is use the `_cast` method because that will give us support for casting more generally (e.g. using int literals with floats). Something like:
```py
def cast_all_arguments(args):
    series_args = [arg for arg in args if isinstance(arg, BaseSeries)]
    if series_args:
        # Choose the first series arbitrarily, the type checker will enforce that they
        # are all the same type in any case
        cast = series_args[0]._cast
        return [cast(arg) for arg in args]
    else:
        # This would only be the case if all the arguments were literals, which would be
        # an unusual and pointless bit of ehrQL, but we should handle it without error
        return args
```

ehrql/query_language.py
```diff
@@ -152,6 +154,16 @@ def if_null_then(self, other):
             default=self._cast(other),
         )
 
+    @staticmethod
```
evansd, line 157: I don't think it makes sense to have these as static methods. We either want them as standalone functions in the `ehrql` namespace or as methods on the `ComparableFunctions` mixin.

I'm not quite sure what the best option is here though. Having them as methods would be nice from the point of view of not needing an extra import. But I can't quite think of a way of naming them where the resulting ehrQL makes sense. Maybe something like:
```
baseline_date = vaccination_date.or_lesser_of(infection_date, "2022-01-02")
```

But we can bikeshed that separately in Slack with a wider group I think.

evansd (commented):
This is such great progress. I feel like (fingers crossed!) you've whacked your head on all the major hard problems here and solved them and so it's really a case of now of filing off rough edges and getting the pieces structured in the right way.

I'm afraid there's a real hodgepodge of comments below. Some minor micro-level comments that I didn't want to forget, and them some bigger structural stuff.

Happy to talk about any of it as needed.

ehrql/query_engines/base_sql.py
Jongmassey: it doesn't, alas. This was the ickiness I was referring to on Slack.

ehrql/query_engines/sqlite_dialect.py
```diff
@@ -11,3 +21,33 @@ def do_on_connect(self, connection):
...
+
+
+# sqlite's aggregate functions return NULL if any of the inputs are NULL
+# this produces a list of coalescences of all columns with the first
+# input to coalesce at each index being the column at the index in
+# the input columns
+def get_cyclic_coalescence(columns):
+    len_cols = len(columns)
+    column_cycles = [
+        list(islice(cycle(columns), i, i + len_cols)) for i in range(len_cols)
```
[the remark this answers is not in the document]
Jongmassey: ah yes that does work but I don't entirely understand how it wraps round the end of the list?

ehrql/utils/typing_utils.py
```diff
@@ -140,6 +140,7 @@ def get_typespec(value):
         return type_
 
 
+@get_typespec.register(tuple)
 @get_typespec.register(Set)
 def get_typespec_for_set(value):
```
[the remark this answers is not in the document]
Jongmassey: https://github.com/opensafely-core/ehrql/pull/1236/files#diff-ff058d49ebf995e041f48164834956c45223046094c42af129e5fc6c81c58d76R151 ?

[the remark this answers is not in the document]
Jongmassey:
```python
raise TypeError(
            f"{type(value).__name__}s must be of homogeneous type: {value!r}"
        )
```

on  L151 not sufficient?

ehrql/query_engines/mssql_dialect.py
```diff
@@ -125,7 +132,42 @@ def get_children(self):
...
 def visit_select_star_into(element, compiler, **kw):
+    source_query = compiler.process(element.selectable.element, asfrom=True, **kw)
+    target = compiler.process(element.table, asfrom=True, **kw)
+
+    # special handling of CTEs
+    if source_query.lower().lstrip().startswith("with"):
+        last_from_ix = source_query.lower().rfind("from")
+        return (
+            f"{source_query[:last_from_ix]} INTO {target} {source_query[last_from_ix:]}"
+        )
```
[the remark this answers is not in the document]
Jongmassey: I've tested this and you're right we don't to use CTEs to solve the aggregate-of-aggregate problem so I'll back this out

ehrql/query_engines/mssql.py
```diff
@@ -23,6 +23,18 @@
...
     sqlalchemy_dialect = MSSQLDialect
 
+    # Use a CTE as the source for the aggregate query rather than a
+    # subquery in order to avoid the "Cannot perform an aggregate function
+    # on an expression containing an aggregate or a subquery" error
+    def aggregate_series_by_patient(self, source_node, aggregation_func):
+        query = self.get_select_query_for_node_domain(source_node)
+        cte_query = query.add_columns(self.get_expr(source_node))
+        cte_query = apply_patient_joins(cte_query)
+        cte_query = cte_query.cte("cte")
```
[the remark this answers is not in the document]
Jongmassey: you're right, I've changed this over

ehrql/query_engines/mssql.py
```diff
@@ -23,6 +23,18 @@
 class MSSQLQueryEngine(BaseSQLQueryEngine):
     sqlalchemy_dialect = MSSQLDialect
 
+    # Use a CTE as the source for the aggregate query rather than a
+    # subquery in order to avoid the "Cannot perform an aggregate function
+    # on an expression containing an aggregate or a subquery" error
+    def aggregate_series_by_patient(self, source_node, aggregation_func):
```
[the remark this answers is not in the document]
Jongmassey: I couldn't find a clean way of doing that, hence why it ended up here :/

ehrql/query_language.py
Jongmassey: You're right, my thinking with making them static methods was that they are things that you can do to collections of PatientSeries or EventSeries but it's currently syntactically _meh_ to call them from an instance of one of those series. I now read that there's no prohibition on doing such a thing ( C#-brain strikes again!). I'll start up some bikeshedding in Slack once I've addressed some of the more obvious revisions

ehrql/query_language.py
Jongmassey: Ah yes I had started down that path but got too focused on dates, good catch!

ehrql/utils/sqlalchemy_query_utils.py
```diff
@@ -232,3 +234,28 @@ def compile(self, *args, **kwargs):  # NOQA: A003
...
+    return compiler.visit_function(element, **kw)
+
+
+def check_and_get_function_args_common_type(function_element):
+    columns = function_element.clauses.clauses
+    types = {c.type for c in columns}
+    inner_types = {type(t) for t in types}
+    if len(inner_types) > 1 and not inner_types == {String, Text}:
+        raise ValueError("columns to aggregate must be of same type")
+    return types.pop()
```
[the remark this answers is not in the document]
Jongmassey: I've factored this out, resolving

ehrql/query_engines/sqlite_dialect.py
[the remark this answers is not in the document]
evansd: It's taking every item from index `i` to the end of the list, followed by every item from the beginning of the list up to (but not including) index `i`.

So I guess the wrapping is handled by the presence of that second list.

ehrql/query_engines/mssql_dialect.py
Jongmassey: I think my latest commit achieves this if you'd be so kind as to take a look

ehrql/utils/sqlalchemy_query_utils.py
```diff
@@ -232,3 +232,13 @@ def compile(self, *args, **kwargs):  # NOQA: A003
...
+
+
+class greatest(sqlalchemy.sql.expression.FunctionElement):
+    name = "greatest"
+    inherit_cache = True
+
+
+class least(sqlalchemy.sql.expression.FunctionElement):
+    name = "least"
+    inherit_cache = True
```
evansd, line 244: These are redundant now, right?

ehrql/query_engines/base_sql.py
```diff
@@ -411,6 +411,25 @@ def get_sql_to_first_of_month(self, node):
     def to_first_of_month(self, date):
         raise NotImplementedError()
 
+    @get_sql.register(Function.MaximumOf)
+    def get_sql_maximum_of(self, node):
+        args = self.get_nary_function_args(node)
+        return self.get_aggregate_subquery(sqlalchemy.func.max, *args).label("greatest")
```
evansd, line 417: What do the labels do here? Is this just about creating more readable SQL?

ehrql/query_engines/mssql.py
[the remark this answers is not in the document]
evansd: I think now you've refactored things you might be able to move this logic into 
the MSSQL implementation of `get_aggregate_subquery()`.

evansd (commented):
This is really coming together now, great work.

I realised this morning (because of a bug Peter) hit there's another tiny change needed in the query rewriter in order to support tuples in the query model. These lines here;
https://github.com/opensafely-core/ehrql/blob/d91128080e4e1a5e0dc77883ced078e47e5e9fc6/ehrql/query_model/query_graph_rewriter.py#L44-L46

Will need changing to something like:
```py
        elif isinstance(obj, frozenset | tuple):
            # As do frozensets and tuples
            return obj.__class__(self._rewrite(v, replacements) for v in obj)
```

ehrql/utils/sqlalchemy_query_utils.py
Jongmassey: ah yes, good catch

ehrql/query_engines/base_sql.py
Jongmassey: yeah, just to give whoever might have to debug some SQL later a helping hand. Now that we're not following the convention of ANSI SQL `greatest` and `least` then these could maybe be changed to  something else helpful

ehrql/query_engines/mssql.py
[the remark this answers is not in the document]
Jongmassey: I don't think it's that simple. The required changes are outside of the scope of the "horizontal aggregation function" (i.e. maximum_of) but are instead in the scope of the aggregation upon the output of that (i.e. when aggregating across event series then aggregating those up to a one-row-per-patient). I could inspect the `source_node` to see if it's a `Function.MaximumOf` or `Function.MinimumOf` and only apply the query transformation then.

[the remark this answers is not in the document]
evansd: I think it is possible, but you're right that it's going to need some thought. I just got the tests passing by getting rid of the custom `aggregate_series_by_patient()` method and updating `get_aggregate_subquery` like this:
```py
    def get_aggregate_subquery(self, node, aggregate_function, columns, return_type):
        v = values(column("aggregate", return_type), name="aggregate_values").data(
            [(c,) for c in columns]
        )
        aggregated = sqlalchemy.select(aggregate_function(v.c.aggregate))
        if has_many_rows_per_patient(node):
            query = self.get_select_query_for_node_domain(node)
            from_subquery = query.add_columns(aggregated.label("value"))
            from_subquery = apply_patient_joins(from_subquery)
            from_subquery = from_subquery.alias("aggregate_subquery")
            return from_subquery.c.value
        else:
            return aggregated
```

That required passing `node` through from `get_sql_maximum_of`/`get_sql_minimum_of` so I had to update the SQLite implementations to accept and ignore that argument. It still needs work, but I think something like will be possible.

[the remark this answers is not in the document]
Jongmassey: I've just pushed an alternative way which I think might actually be messier, thoughts?

tests/generative/variable_strategies.py
```diff
@@ -362,6 +364,33 @@ def binary_operation_with_types(
...
+        # first decide on a non-zero value for n within a reasonable range
+        n = draw(st.integers(1, 20))
+
+        # draw _n_ args of same type
+        series_options = [value, series]
+        args = []
+        for _ in range(n):
+            arg_frame = draw(st.one_of(any_frame(), st.just(frame)))
+            arg_series = draw(st.sampled_from(series_options))
+            args.append(draw(arg_series(series_type, arg_frame)))
```
rebkwok, line 385: I think this could be generated using a hypothesis `lists` strategy; the contents of the loop would be a strategy for generating a nary arg, and the `lists` strategy has a min/max size arg that would deal with picking the number of input args.

Something like:
```
@st.composite
def narg_strategy():
    series_options = [value, series]
    arg_frame = draw(st.one_of(any_frame(), st.just(frame))
    return draw(st.sampled_from(series_options))

args = st.lists(narg_strategy(), min_size=1, max_size=20)
```

(not sure if that'd actually work, I always seem to need to experiment with where `composite` and `draw` go)

Jongmassey: that looks cleaner, I'll have a poke at it

evansd (commented):
Great, this is getting very close to the line I think.

I've added a few comments on specifics below.

ehrql/query_engines/mssql.py
```diff
@@ -23,6 +24,18 @@
...
+        query = self.get_select_query_for_node_domain(source_node)
+        from_subquery = query.add_columns(self.get_expr(source_node))
+        from_subquery = apply_patient_joins(from_subquery)
+        from_subquery = from_subquery.alias("aggregate_subquery")
```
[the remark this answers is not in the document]
Jongmassey: good point, will replace with `.subquery()`

ehrql/query_engines/mssql.py
```diff
@@ -226,6 +239,19 @@ def get_results(self, variable_definitions):
...
+    # https://stackoverflow.com/questions/71022/sql-max-of-multiple-columns/6871572#6871572
+    def get_aggregate_subquery(self, aggregate_function, columns, return_type):
+        v = values(column("aggregate", return_type), name="aggregate_values").data(
+            [(c,) for c in columns]
+        )
+        froms = set().union(*[c._from_objects for c in columns])
+        aggregated = sqlalchemy.select(
+            aggregate_function(v.c.aggregate)
+        ).scalar_subquery()
+        aggregated._from_objects = list(froms)
```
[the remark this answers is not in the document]
Jongmassey: Agreed, it was horribly gnarly

ehrql/query_engines/mssql.py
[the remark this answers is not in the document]
Jongmassey: I think even though when compiled by mssql it would end up being essentially a no-op I agree as it'll result in clearer sql for debugging. Done.

Jongmassey:
@evansd  I've made the changes as per https://github.com/opensafely-core/ehrql/commit/2ff2fbbbb7d9e0c5d185885b08d41cb9bbca20c7 but hypothesis seems to have found a case where it still triggers the ''Cannot perform an aggregate function on an expression containing an aggregate or a subquery." error. I've spent ages trying to decipher it (the hypothesis output is as gnarly as ever) but can't seem to work out why it's happening.

https://github.com/opensafely-core/ehrql/actions/runs/5577115921/jobs/10189368921

ehrql/query_engines/mssql.py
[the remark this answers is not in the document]
Jongmassey: not done as discussed [on slack](https://bennettoxford.slack.com/archives/C03FB777L1M/p1689634007373319)

evansd (approved):
:tada: :champagne: :partying_face: 

The size of the Python diff utterly obscures the enormous submerged iceberg of effort it took to get here. And as is the case in all good sagas, it's only when you reach the end of the journey that you discover the real treasure was the weird and arbitrary MSSQL limits you discovered along the way...

Great work, and thanks for persevering with it.

I'll write up some thoughts on the nested aggregation issue as we discussed, but I'm not suggesting we need to do anything about it in the short term.

Issue #976, opened 2023-01-25
Remove use of `set-output` from workflows


StevenMaude:
It's [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

https://github.com/opensafely-core/databuilder/blob/308a74a954a515964625a083821a1d0f82ff06a2/.github/workflows/build-and-deploy.yml#L77-L82

StevenMaude:
It's also in the `tag-new-version` workflow because the action we use, uses it.

We're currently stuck on an old version: #716.

Issue #716, opened 2022-09-14
Solve the problem of upgrading github-tag-action 


StevenMaude:
The recent upgrade to github-tag-action v6.0 (#689) broke repository version tagging (#709). The revert in #714 fixed this, but now we are stuck on an older version of the action.

We should either look into this more (bisect it? which is possibly a pain for workflows) or consider moving to something else, whether third-party or our own code.

PR #1247, opened 2023-05-04
Replace use of `set-output` in workflow

StevenMaude:
It has been deprecated for some time.

See #976.


StevenMaude:
This should work, but probably best to try out end-to-end to ensure it still behaves properly.

StevenMaude:
This doesn't fix the Actions warnings entirely, because of #716, where the github-tag-action version we're still on uses `set-output`, but at least resolves the workflows we've written.

Issue #1248, opened 2023-05-05
Possible confusion over contents of `APCS_Der.Spell_Primary_Diagnosis`


evansd:
Previous [investigation](https://github.com/opensafely-core/ehrql/issues/1066#issuecomment-1441614133) by @milanwiedemann suggests that the `APCS_Der.Spell_Primary_Diagnosis` column can contain multiple diagnosis codes, separate by semi-colons.

But the schema description says that this column has a maximum length of 5, which doesn't seem big enough if this column is expected to contain multiple diagnosis codes:
https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#APCS_Der

ICD-10 codes can be as short as 2 characters, so technically you could fit two of them plus a separator. But if the intention is that this column can contain multiple codes then this seems far too small.

We should get to the bottom of what's going on here, as possibly this indicates some confusion elsewhere.

StevenMaude:
The original intent, at least, seems to be that `der_diagnosis_all` can contain multiple codes:

> The field "der_diagnosis_all" is a concatenation of all diagnosis fields in all
episodes within the spell.

https://www.england.nhs.uk/wp-content/uploads/2018/06/04-b-pb-04-07-2018-ebi-consultation-document.pdf

evansd:
@StevenMaude Yes, that's `Der_Diagnosis_All` which is different from `Spell_Primary_Diagnosis`.

Further investigation confirms:

 * the schema description is correct and we never have more than 5 characters in this column:
   ```sql
   SELECT MAX(LEN(Spell_Primary_Diagnosis)) FROM APCS_Der;
   -- 5
   ```
 * the character `;` occurs nowhere in this column:
   ```sql
   SELECT COUNT(*) FROM APCS_Der WHERE Spell_Primary_Diagnosis LIKE '%;%';
   -- 0
   ```

@milanwiedemann Do you remember what you saw that made you think there were multiple diagnoses here? It's always possible that something has changed since the time you checked, but it seems unlikely.

milanwiedemann:
I vaguely remember seeing something like this `A1;B1;C1`

evansd:
I think the next steps here are:

1. We should add a singular `primary_diagnosis` field to the `hospitalisations` table:
   https://github.com/opensafely-core/ehrql/blob/2e09a533261a3d3ebdc059afcc46d56f56221bec/ehrql/tables/beta/tpp.py#L191-L201
   Along the lines of:
   ```py
[...]

PR #1249, opened 2023-05-05
Automate TPP schema build

### LINKED ISSUE #1104: Automate (as far as possible) a mirror of the TPP schema for testing purposes
### Problem

In order to test that the `QueryTable` and `MappedTable` definitions in [`databuilder/backends/tpp.py`](https://github.com/opensafely-core/databuilder/blob/22ae890c46f38b5a17272eb125d20d59e8070689/databuilder/backends/tpp.py) are correctly defined we have a bunch of integration tests in [`tests/integration/backends/test_tpp.py`](https://github.com/opensafely-core/databuilder/blob/22ae890c46f38b5a17272eb125d20d59e8070689/tests/integration/backends/test_tpp.py).

These work by inserting rows into a database whose schema is intended to match that of the TPP database, and is defined using the SQLAlchemy ORM in [`tests/lib/tpp_schema.py`](https://github.com/opensafely-core/databuilder/blob/22ae890c46f38b5a17272eb125d20d59e8070689/tests/lib/tpp_schema.py).

But nothing ensures that this schema _does_ match that of the TPP database. Often, when adding a new data source, the same person defines the ehrQL table, the backend table mapping, _and_ the `tpp_schema.py` ORM table against which it is tested. So it's very easy for the same mistake (typo, misunderstanding, whatever) to make it into all three. Then the tests are all green but the job blows up when trying to run it in production.

Additionally, the ORM models in `tpp_schema.py` are much more complex than they need to be as they define a whole load of foreign key relationships between the different models. This is because they were copied directly from Cohort Extractor where they formed the primary means of constructing test fixtures. It was therefore important to be able easily to construct complex sets of interrelated objects. This need isn't so great in Data Builder as the models exist purely to test the `QueryTable` and `MappedTable` definitions.

### Proposal

Ideally we'd have some mechanism for automatically creating `tpp_schema.py` based on the actual schema of the TPP database, using something like the mechanism that powers the [Database Schema Report](https://reports.opensafely.org/reports/opensafely-tpp-database-schema/).

This wouldn't need to model any foreign-key relationships, just the column names and types. We could add a surrogate auto-increment primary key to every table (`_pk = Column(Integer, primary_key=True)` or something like that) to keep the ORM happy.

This would give us more confidence that the tests we write against that schema would be meaningful.

evansd:
This automates the process of downloading the published TPP database schema and building a test schema which matches it. This runs on a nightly scheduled action which will open a PR if it finds any changes.

In the process of doing this I discovered another DATE/DATETIME confusion which I have fixed.

I also discovered some other issues/oddities which I have created separate issues for:
 * https://github.com/opensafely-core/ehrql/issues/1248
 * https://github.com/opensafely/tpp-database-schema/issues/49
 * https://github.com/opensafely/tpp-database-schema/issues/50

Closes #1104


tests/lib/update_tpp_schema.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+if __name__ == "__main__":
+    command = sys.argv[1] if len(sys.argv) > 1 else None
+    if command == "fetch":
+        fetch_schema()
+    elif command == "build":
+        build_schema()
+    else:
+        raise RuntimeError(f"Unknown command: {command}")
```
StevenMaude, line 210: suggestion: maybe have a more helpful error here, so someone doesn't have to read the source to figure out the script?

tests/lib/tpp_schema.py
```diff
@@ -1,1057 +1,3261 @@
...
+    _45to49 = mapped_column(t.Integer)
+    _50to54 = mapped_column(t.Integer)
+    _55to59 = mapped_column(t.Integer)
+    _5to9 = mapped_column(t.Integer)
+    _60to64 = mapped_column(t.Integer)
+    _65to69 = mapped_column(t.Integer)
+    _70to74 = mapped_column(t.Integer)
+    _75to79 = mapped_column(t.Integer)
+    _80to84 = mapped_column(t.Integer)
+    _85Plus = mapped_column(t.Integer)
```
StevenMaude, line 3094: thought: These columns, and maybe others, start with a single underscore which has a specific meaning when used as class attributes. I don't know if it warrants removing or replacing them.

tests/lib/update_tpp_schema.py
```diff
@@ -0,0 +1,210 @@
...
+#   python tests/lib/update_tpp_schema.py build
+#
+
+from sqlalchemy import types as t
+from sqlalchemy.orm import DeclarativeBase, mapped_column
+
+
+class Base(DeclarativeBase):
+    "Common base class to signal that models below belong to the same database"
+"""
```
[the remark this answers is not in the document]
evansd: Ruff and Black check the file _generated_ from this template, so that's still subject to all the same requirements as hand-written code.

tests/lib/tpp_schema.py
```diff
@@ -1,1057 +1,3261 @@
...
-    Patient = relationship("Patient", back_populates="MedicationIssues")
-    Consultation_ID = Column(Integer)
-    MedicationIssue_ID = Column(Integer, primary_key=True)
-    RepeatMedication_ID = Column(Integer)
-    MultilexDrug_ID = Column(
-        NVARCHAR(length=20), ForeignKey("MedicationDictionary.MultilexDrug_ID")
-    )
-    MedicationDictionary = relationship(
-        "MedicationDictionary", back_populates="MedicationIssues", cascade="all, delete"
-    )
```
StevenMaude, line 40: question: Why can we remove the use of some of the explicitly specified primary and foreign keys, and relationships? Is this just because SQLAlchemy 2 lets us do this?

I'm sure this is fine; just for my own understanding :smiley:

StevenMaude (approved):
summary: This looks fine to me, and the related issues that have been opened are well documented.

The PR might warrant checking over by someone who understands the underlying SQLAlchemy schema changes better than I do though.

tests/lib/tpp_schema.py
evansd: A fair question; I hadn't noticed these leading underscore column names. The approach I've generally taken here is to mirror the table and column names exactly without regard to any of the normal Python naming conventions. So we have class names like `ONS_Deaths` rather than `ONSDeaths`, and attribute names like `Patient_ID` rather than `patient_id`. And for names which aren't valid class or attribute names (e.g. `class`) I've just made the minimal adjustment to make them valid (e.g. `class_`).

Given that these are purely internal test classes I think the benefits of matching the database as closely as possible outweigh the downsides of having unpythonic names.

Given that the meaning of single leading underscores is just a convention I'm inclined to leave these as they are.

tests/lib/tpp_schema.py
evansd: This is discussed a bit in the original issue but I didn't reference this at in the PR:

> Additionally, the ORM models in `tpp_schema.py` are much more complex than they need to be as they define a whole load of foreign key relationships between the different models. This is because they were copied directly from Cohort Extractor where they formed the primary means of constructing test fixtures. It was therefore important to be able easily to construct complex sets of interrelated objects. This need isn't so great in Data Builder as the models exist purely to test the `QueryTable` and `MappedTable` definitions.

So the answer is that the foreign key relationships were things we added for our own convenience, not things represented in the database itself.

tests/lib/tpp_schema.py
StevenMaude: Yep, that sounds fine; was just flagging in case you weren't aware and was important.

tests/lib/tpp_schema.py
StevenMaude: Thanks :+1:

tests/lib/update_tpp_schema.py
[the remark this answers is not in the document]
StevenMaude: ah, true!

tests/lib/update_tpp_schema.py
StevenMaude: Maybe even something terse like:

```suggestion
        raise RuntimeError(f"Unknown command: {command}; valid commands: fetch, build")
```

PR #1273, opened 2023-05-17
Upgrade to github-tag-action v6.1

### LINKED ISSUE #716: Solve the problem of upgrading github-tag-action

### LINKED ISSUE #976: Remove use of `set-output` from workflows

StevenMaude:
Fixes #716.
Fixes #976.
Supersedes #851.

This was previously updated in #689 and reverted in #714, due to it breaking tagging.

Having looked around the code, issues and PRs for the action, it looks as if the README is not in line with the current behaviour (mathieudutour/github-tag-action#150).

The author's suggestion there is to just not run the action of PRs, which is a reasonable fix for now. If the tagging behaviour is more controllable as we need in the future, we could revisit this to remove the conditional check in our workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: hadolint/hadolint-action@54c9adbab1582c2ef04b2016b760714a4bfde3cf
+      - uses: hadolint/hadolint-action@54c9adbab1582c2ef04b2016b760714a4bfde3cf #v3.1.0
         with:
           failure-threshold: error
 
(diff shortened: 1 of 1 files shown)


StevenMaude:
This worked for me when testing in a fork of this repository. However, we should keep an eye on version tagging and image release once merged, to check all is well.

iaindillingham (approved):

Issue #1274, opened 2023-05-18
Commits from "update" workflows aren't signed 


StevenMaude:
Not sure if this is a serious problem. However, if a goal/policy is to sign all our commits, then we should be doing so.

Example unverified commit: 70e658b0bada5ff65c3ee74c869926b9aa4aa77a

Relevant information:

https://github.com/actions/runner/issues/667
https://github.blog/2019-08-15-commit-signing-support-for-bots-and-other-github-apps/
https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c

evansd:
I believe this is closed by #1275

PR #1275, opened 2023-05-18
Author automated commits as GitHub Actions bot

StevenMaude:
This uses the name and email as shown in the create-pull-request update guide for an old version:

https://github.com/peter-evans/create-pull-request/blob/efbb6ba75d718110dc7d3bd08cb98ebd247adc68/docs/updating.md
  
By default, the create-pull-request action specifies the person who triggered the workflow as the author.

If this is an automated scheduled run, then the person who is counted as triggering the workflow is the last person who edited the workflow file.

With #1274, it means that individual authors, may have unsigned and therefore unverified commits showing up in their commit history.

(I am perhaps fussier about this than other people.)

This is not very obvious behaviour.

By using the GitHub Actions name and email address, we can ascribe the
commits to that, which seems reasonable, as it's entirely automated.

### DIFF
--- a/.github/workflows/update-external-studies.yml
+++ b/.github/workflows/update-external-studies.yml
@@ -33,6 +33,8 @@ jobs:
           add-paths: tests/acceptance/external_studies/*
           branch: bot/update-external-studies
           base: main
+          author: "github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>"
+          committer: "github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>"
           commit-message: "chore: Update `external_studies` test code"
           title: "Update `external_studies` test code"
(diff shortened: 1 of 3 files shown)


evansd:
It does seem a _bit_ arbitrary to be using that particular user (and do we have any idea where that big number comes from?).

How about using one of existing bot accounts e.g.
https://github.com/orgs/opensafely/people/opensafely-github-bot

So something like:
```
opensafely-github-bot <opensafely-github-bot@users.noreply.github.com>
```

StevenMaude:
The ID 41898282 is taken from https://api.github.com/users/github-actions[bot]

https://github.com/orgs/community/discussions/26560#discussioncomment-3252340

But it's fine to change it.

Issue #42, opened 2021-07-09
New efficient measures implementation


benbc:

evansd:
Adding a note here because I'm assuming this issue will be looked at by whoever designs the new measures API.

It would be good to enforce that numerator and denominator definitions are such that a patient can only be a member of the numerator if they are also a member of the denominator.

See Slack threads here for where our inability to enforce this in cohortextractor has lead to confusion:
https://bennettoxford.slack.com/archives/C03FB777L1M/p1665483592932119
https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1664978145256269

evansd:
Closing as we have now (for better or worse) implemented _a_ measures API that we need to live with.

Making the implementation of that API efficient is a separate issue, tracked as:
 * #2294

PR #1276, opened 2023-05-19
Add measures support

### LINKED ISSUE #816: Design and implement measures API
We need some equivalent to Cohort Extractor's [measures framework](https://docs.opensafely.org/measures/) in Data Builder. It's OK if the first version of this uses the same inefficient implementation as Cohort Extractor. But the API needs to be such it doesn't preclude us swapping this out for a more [efficient implementation](https://github.com/opensafely-core/databuilder/issues/42) further down the road.

evansd:
This adds measures support to ehrQL. Specifically it adds a new command `generate-measures`:
```
usage: ehrql generate-measures [-h] [--output OUTPUT_FILE] [--dsn DSN] [--dummy-tables DUMMY_TABLES_PATH]
                               [--dummy-data-file DUMMY_DATA_FILE] [--query-engine QUERY_ENGINE_CLASS]
                               [--backend BACKEND_CLASS]
                               definition_file

positional arguments:
  definition_file       Path of the file where measures are defined
```

This has identical arguments to `generate-dataset` except that it takes a measures definition file rather than a dataset definition.

Below is a toy example measure definition. It defines a single measure which counts the proportion of registered patients who received a vaccination in each month in 2021, broken down by sex and age band:
```py
from ehrql import INTERVAL, Measures, case, months, when
from ehrql.tables.beta.tpp import (
    patients,
    practice_registrations,
    vaccinations,
)


has_vaxx_during_interval = (
    vaccinations.where(vaccinations.date.is_during(INTERVAL))
    .exists_for_patient()
)
is_registered_during_interval = (
    practice_registrations.for_patient_on(INTERVAL.start_date)
    .exists_for_patient()
)

age = patients.age_on(INTERVAL.start_date)
age_band = case(
    when(age < 30).then("0-30"),
    when(age < 60).then("30-60"),
    when(age >= 60).then("60+"),
)

measures = Measures()

measures.define_measure(
    name="vaccinations",
    numerator=has_vaxx_during_interval,
    denominator=is_registered_during_interval,
    group_by={
        "sex": patients.sex,
        "age_band": age_band,
    },
    intervals=months(12).starting_on("2021-01-01"),
)
```

This produces output with the following structure:
measure | interval_start | interval_end | ratio | numerator | denominator | sex | age_band
-- | -- | -- | -- | -- | -- | -- | --
vaccinations | 2021-01-01 | 2021-01-31 | 0.0008 | 10 | 12345 | male | 0-30
vaccinations | 2021-01-01 | 2021-01-31 | 0.0013 | 20 | 15432 | male | 30-60
vaccinations | 2021-01-01 | 2021-01-31 | 0.0018 | 30 | 16789 | male | 60+
vaccinations | 2021-01-01 | 2021-01-31 | 0.0015 | 15 | 10123 | female | 0-30
vaccinations | 2021-01-01 | 2021-01-31 | 0.0023 | 25 | 10987 | female | 30-60
vaccinations | 2021-01-01 | 2021-01-31 | 0.0019 | 35 | 18765 | female | 60+

## Implementation

The current implementation is not any cleverer than the old Cohort Extractor implementation in that it still involves running a separate query for each interval and pulling down the full set of patient level results each time.

However, despite this there should still be some significant efficiency gains because we no longer need to write each full cohort to disk, copy it out of the volume, copy it back to _different_ volume, read it all into memory and finally calculate the measures. Instead we calculate the measures in a streaming fashion without ever needing to hold the full patient-level results in memory or write them to disk.

Longer term, the API is designed so that we can improve the [efficiency](https://github.com/opensafely-core/ehrql/issues/42) of the implementation incrementally without requiring any changes from users.

Closes #816

### DIFF
--- a/ehrql/measures/calculate.py
+++ b/ehrql/measures/calculate.py
@@ -1,0 +1,156 @@
+from collections import defaultdict
+
+from ehrql.measures.measures import get_all_group_by_columns
+from ehrql.query_model.nodes import Case, Function, Value, get_series_type
+from ehrql.query_model.transforms import substitute_parameters
+
+
+def get_measure_results(query_engine, measures):
+    # Group measures by denominator and intervals as we'll handle them together
+    grouped = defaultdict(list)
+    for measure in measures:
+        group_id = (measure.denominator, measure.intervals)
+        grouped[group_id].append(measure)
+
+    all_group_by_columns = get_all_group_by_columns(measures).keys()
+
+    for measure_group in grouped.values():
+        calculator = MeasureCalculator(measure_group)
+        results = calculator.get_results(query_engine)
+        for measure, interval, numerator, denominator, group_dict in results:
+            ratio = numerator / denominator if denominator else None
+            yield (
+                measure.name,
+                interval[0],
+                interval[1],
+                ratio,
+                numerator,
+                denominator,
+                # Return a column for every group used across all measures, even if this
+                # particular measure doesn't use that group
+                *(group_dict.get(name) for name in all_group_by_columns),
+            )
+
+
+class MeasureCalculator:
+    def __init__(self, measures):
+        self.denominator = None
+        self.intervals = None
+        self.variables = {}
+        self.measures = []
+        self.fetchers = []
+        for measure in measures:
+            self.add_measure(measure)
+
+    def get_results(self, query_engine):
+        for interval in self.intervals:
+            results = self.get_results_for_interval(query_engine, interval)
+            for measure, numerator, denominator, group in results:
+                group_dict = dict(zip(measure.group_by.keys(), group))
+                yield measure, interval, numerator, denominator, group_dict
+
+    def get_results_for_interval(self, query_engine, interval):
+        # Build the query for this interval by replacing the interval start/end
+        # placeholders with actual dates
+        query = substitute_interval_parameters(self.variables, interval)
+
+        # "fetchers" are functions which take a row and return the values relevant to a
+        # given measure (numerator, denominator and groups); "accumulators" are dicts
+        # storing the cumulative numerator and denominator totals for each group in the
+        # measure
+        fetcher_accumulator_pairs = [
+            (fetcher, defaultdict(lambda: [0, 0])) for fetcher in self.fetchers
+        ]
+
+        for row in query_engine.get_results(query):
+            for fetcher, accumulator in fetcher_accumulator_pairs:
+                numerator, denominator, group = fetcher(row)
+                totals = accumulator[group]
+                totals[0] += numerator
+                totals[1] += denominator
+
+        for measure, (_, accumulator) in zip(self.measures, fetcher_accumulator_pairs):
+            for group, (numerator, denominator) in accumulator.items():
+                yield measure, numerator, denominator, group
+
+    def add_measure(self, measure):
+        # Record denominator and intervals from first measure
+        if self.denominator is None:
+            self.denominator = measure.denominator
+            self.intervals = measure.intervals
+            self.variables["population"] = series_as_bool(self.denominator)
+        else:
+            assert measure.denominator == self.denominator
+            assert measure.intervals == self.intervals
+
+        numerator_index = self.add_variable(series_as_int(measure.numerator))
+        denominator_index = self.add_variable(series_as_int(measure.denominator))
+        group_indexes = [
+            self.add_variable(column) for column in measure.group_by.values()
+        ]
+
+        # Create a function which takes a results row and returns a numerator,
+        # denominator, and tuple of group values, based on their indices
+        fetcher = self.create_fetcher(numerator_index, denominator_index, group_indexes)
+
+        self.measures.append(measure)
+        self.fetchers.append(fetcher)
+
+    def add_variable(self, variable):
+        # Return the position of `variable` in the variables dict, adding it if not
+        # already present
+        try:
+            return list(self.variables.values()).index(variable)
+        except ValueError:
+            index = len(self.variables)
+            self.variables[f"column_{index}"] = variable
+            return index
+
+    @staticmethod
+    def create_fetcher(numerator_index, denominator_index, group_indexes):
+        # Given a bunch of indices we want a function which extracts just those indices
+        # from a tuple. This is going to be called frequently, so the fastest way to do
+        # this is to build a function definition and then eval it.
+        group_items = [f"row[{i}]" for i in group_indexes]
+        items = [
+            f"row[{numerator_index}]",
+            f"row[{denominator_index}]",
+            f"({', '.join(group_items)},)",
+        ]
+        code = f"lambda row: ({', '.join(items)})"
+        return eval(code)
+
+
+def series_as_bool(series):
+    series_type = get_series_type(series)
+    if series_type is bool:
+        return series
+    elif series_type is int:
+        return Function.GT(series, Value(0))
+    else:
+        assert False
+
+
+def series_as_int(series):
+    series_type = get_series_type(series)
+    if series_type is int:
+        return series
+    elif series_type is bool:
+        # TODO: This is definitely not the most efficient way to do this. We should
+        # extend the `CastToInt` operation to apply to boolean as well.
+        return Case(
+            {
+                Function.EQ(series, Value(True)): Value(1),
+                Function.EQ(series, Value(False)): Value(0),
+            }
+        )
+    else:
+        assert False
+
+
+def substitute_interval_parameters(variable_definitions, interval):
+    return substitute_parameters(
+        variable_definitions,
+        interval_start_date=interval[0],
+        interval_end_date=interval[1],
+    )
--- a/ehrql/measures/calculate.py
+++ b/ehrql/measures/calculate.py
@@ -1,6 +1,8 @@
+import datetime
 from collections import defaultdict
 
 from ehrql.measures.measures import get_all_group_by_columns
+from ehrql.query_model.column_specs import ColumnSpec, get_column_spec_from_series
 from ehrql.query_model.nodes import Case, Function, Value, get_series_type
 from ehrql.query_model.transforms import substitute_parameters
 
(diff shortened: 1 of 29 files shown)


rebkwok (approved):
This all looks good to me. I think it might be good for someone who's more familiar with the old cohort-extractor measures to review too, though. I felt like I spent a good bit of this PR learning what measures actually are.

I had a couple of very minor comments, that's all.  Thank you for the logical commits, it made it way easier for reviewing.

It might just be me, but I found the `Intervals.START_DATE, Intervals.END_DATE` notation non-intuitive.  i.e. in
 ```
numerator=sixtieth.is_on_or_between(Intervals.START_DATE, Intervals.END_DATE)
``` 

I kept expecting it to be referring to the start and end date of all of the intervals, not each individual interval.  I wonder if simply naming the class `Interval` instead of `Intervals` would help with that.

**Comment by evansd** [comment]
Thanks @rebkwok. You're completely right about the plural `Intervals.START_DATE` being confusing. I actually did have it as singular initially and then changed it when I added the `weeks/months/years` method because it made more sense in that context to have it as plural, but didn't properly consider the consequences.

In general, I'm not all that happy with the specifics of the API I've got here but I didn't want to spend too much time going in circles on it by myself. I'd  be very happy to take other suggestions here.

The key pieces we need are:

 * Something to construct a container for several measures.
 * A method on this container to set default values for the measures.
 * A method on this container to define a new measure.
 * A pair of START_DATE/END_DATE parameter constants to use when constructing queries for measures.
 * Functions for generating sequences of intervals (e.g. weeks, months and years).

Exactly what these should all be called and where they should live I'm not quite sure.

File: tests/unit/measures/test_dummy_data.py [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+@table
+class events(EventFrame):
+    date = Series(date)
+    code = Series(
+        str,
+        constraints=[Constraint.Categorical(["abc", "def", "foo"])],
+    )
+
+
+def test_duumy_measures_data_generator():
```
sebbacon, line 30:
```suggestion
def test_dummy_measures_data_generator():
```

sebbacon:
I also found the Intervals API hard to interpret.

I appreciate this will have side-effects I've not thought through, but spitballing, is this any better?

```py
sixtieth = patients.date_of_birth + years(60)

intervals = Intervals.months(12, start="2022-01-02")

measures.define_measure(
    name="sixtieth_birthdays_by_sex",
    numerator=sixtieth.is_during(intervals),
    denominator=patients.exists_for_patient(),
    group_by={"sex": patients.sex},
    intervals=intervals,
)
```

evansd:
@sebbacon Your suggestion inspired me to think new thoughts. See what you make of the updated PR description.

I'm now reusing the `years`, `months`, `weeks` functions which already exist in ehrQL to provide the API for generating the intervals. And that means the placeholder itself can just be start/end date tuple.

I'm slightly in two minds as to whether the placeholder should be `interval` or `INTERVAL`. The uppercase version might help to indicate that it's functioning as a placeholder rather than an actual value (and it would be more gooder Python style). On the other hand it might seem weird and arbitrary to users as it would be the only uppercase thing in ehrQL.

sebbacon:
I think it's loads clearer!

To resolve the "what is `interval`" question, I still like my suggestion, which with the current syntax would be something like:

```py

intervals=months(12).starting_on("2022-01-02")

measures.define_measure(
    name="sixtieth_birthdays_by_sex",
    numerator=sixtieth.is_during(intervals),
    denominator=patients.exists_for_patient(),
    group_by={"sex": patients.sex},
    intervals=intervals,
)
```

What do you think?

evansd:
We could make that work, though it would take a bit of magic to do so.

But my worry is that this obscures, rather than clarifies the semantics here (which I agree are a bit opaque). In particular, having the plural "intervals" as the placeholder makes some constructions read quite strangely. It sort of works OK with something like:
```py
clinical_events.date.is_during(intervals)
```
But even then it suggests to me that we're asking "does this event happen during _any_ of these intervals?" – which is not what we're asking.

And if you need to specifically reference one of the boundary dates you end up with this odd mixture of singular and plural. For instance, if you want as your denominator everyone between aged over 40 at the start of the month you get:
```
denominator = patients.age_on(intervals.start_date) > 40
```

The way we avoided some of this confusion previously was that we just had a single date parameter: `index_date`. And because "index date" as a term is well understood as having this kind of placeholder role I think it made sense to people (or maybe it didn't, and people just blindly copied and pasted stuff that worked!).

I think it's significantly easier and more flexible to have both the start and end dates parameterised because it avoids having to manually calculate the end date yourself and allows us to support nice syntactic sugar like `is_during()`. But it does mean we don't have a natural term for this placeholder variable.

sebbacon:
Makes sense. 

Yes, I agree. So how about:

```py

intervals = months(12).starting_on("2022-01-02")

measures.define_measure(
    intervals=intervals,
    name="sixtieth_birthdays_by_sex",
    numerator=sixtieth.is_during(intervals.current_interval()),
    denominator=patients.exists_for_patient(),
    group_by={"sex": patients.sex},
)
```

evansd:
I've had some helpful discussion with @wjchulme in which I provided him some example measure definitions using different APIs but without any explanation or comments. To summarise his response (Will, correct me if I'm wrong on any of this):

 * He was able to understand what was going on in the measure definitions.

 * He understood the placeholder role (and used that term, though I hadn't introduced it) of the interval object.

 * He disliked the fact that there is this kind of implicit loop over the interval placeholders going on, mostly because it violates a mental model of ehrQL where you think of it as working directly with the actual data (like a Pandas DataFrame) rather than constructing a query graph to be executed later. This mental model is incorrect, but quite convenient and helpful and the measures API forces you to abandon it.

 * He didn't have a particularly strong feeling between what I'll call the Evans and Bacon APIs. The Bacon API was slightly easier to understand on the first read through, but he'd probably prefer to work with the Evans API longer term.

 * With the Evans API he thought that making the interval placeholder uppercase would actually be helpful because its syntactic strangeness would help signal its semantic strangeness, and using uppercase is a reasonably common convention for placeholders generally (e.g. "This document was edited by: YOUR_NAME_HERE").

 * Overall he understood, and would be happy working with either API.
 
 * (He also had some suggestions for extending the measures system to support different kinds of statistic aggregation. I'll record these in a separate ticket.)

I've therefore updated this PR to make the placeholder uppercase, and also changed the example in the PR description to the slightly richer one I shared with Will.

I haven't implemented the API Seb suggested. I'm not absolutely opposed to it, but I think it's a less accurate representation of the actual semantics here and would also involve a more complex and highly coupled API. So I think we'd need to be convinced it was quite a lot more comprehensible for the trade-off to be worth it.

wjchulme:
Thanks, @evansd, very clear summary of our conversation. No quibbles from me.

sebbacon:
RIP Bacon API  - it's lovely to see some hard thinking and testing going into the design

ehrql/measures/calculate.py
```diff
@@ -0,0 +1,177 @@
...
+        assert False
+
+
+def series_as_int(series):
+    series_type = get_series_type(series)
+    if series_type is int:
+        return series
+    elif series_type is bool:
+        # TODO: This is definitely not the most efficient way to do this. We should
+        # extend the `CastToInt` operation to apply to boolean as well.
```
inglesp, line 161: Can you add a ticket for this so we don't lose sight of it?

tests/integration/test_main.py
inglesp, line 1: Does the addition of this file in this PR mean that we're missing comparable tests for `generate_dataset`?

ehrql/measures/dummy_data.py
inglesp, line 1: Why `ehrql/measures/dummy_data.py` and not `ehrql/dummy_data/measures.py`?

inglesp (approved):
This is great -- some very minor questions but nothing to delay merging.

tests/unit/query_model/test_transforms.py
```diff
@@ -352,3 +354,9 @@ def test_identical_operations_are_not_transformed_differently():
 
 def transform(variable):
     return apply_transforms({"v": variable})["v"]
+
+
+def test_substitute_parameters():
+    graph = {"value": Function.Add(Value(10), Parameter("i", int))}
+    transformed = substitute_parameters(graph, i=20)
+    assert transformed == {"value": Function.Add(Value(10), Value(20))}
```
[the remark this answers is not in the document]
evansd: Parameters are identified by their name _and_ type (as just a natural consequence of their dataclass definition) so if the types don't match then no substitution occurs and you'll get an error if you try to execute the resulting query model.

ehrql/measures/measures.py
```diff
@@ -0,0 +1,194 @@
...
+    def _validate_interval(self, interval):
+        if (
+            not isinstance(interval, tuple)
+            or len(interval) != 2
+            or not all(isinstance(i, datetime.date) for i in interval)
+        ):
+            raise ValidationError(
+                f"`intervals` must contain pairs of dates, got: {interval!r}"
+            )
+        if interval[0] > interval[1]:
```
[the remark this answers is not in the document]
evansd: I think that's OK . The intervals are closed so that represents just a single date, and if you want to evaluate a measure on a single date then who are we to say you can't?

ehrql/measures/measures.py
```diff
@@ -0,0 +1,194 @@
...
+
+def get_all_group_by_columns(measures):
+    """
+    Merge all `group_by` dictionaries into a single, flat dictionary
+    """
+    return {
+        name: column
+        for measure in measures
+        for name, column in measure.group_by.items()
+    }
```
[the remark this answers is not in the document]
evansd: That's right, yes. I'll add an appropriate comment.

ehrql/measures/dummy_data.py
evansd: It made sense to me to keep the measures stuff encapsulated together as the measures module "knows about" the rest of the system, but not vice versa. I can see a future it which that stops being the case and we decide to move it, but I quite like keeping the measures code reasonably contained at the moment.

ehrql/query_language.py
```diff
@@ -386,6 +386,10 @@ def is_between(self, start, end):
     def is_on_or_between(self, start, end):
         return (self >= start) & (self <= end)
 
+    def is_during(self, interval):
+        start, end = interval
+        return self.is_on_or_between(start, end)
```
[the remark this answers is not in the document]
evansd: Reluctantly, I agree. I wonder if, at some point, we could do something clever so that where we have multiple spec tests which produce identical query model we only run them once against real databases? But that's something to tackle as and when the speed of the spec tests becomes problematic.

ehrql/measures/calculate.py
evansd: Done:
https://github.com/opensafely-core/ehrql/issues/1313

PR #1285, opened 2023-05-24
Avoid the pitfalls of system-SQLite on macOS

iaindillingham:
If only Apple took Rishi Sunak's advice and gave system-SQLite the maths skills it needs to pass.


ghickman (approved):
This looks great and it's good to have this knowledge written down somewhere too 👍

Do you want someone from the data team to review this as well?

PR #1287, opened 2023-05-25
Expose OpenPROMPT

### LINKED ISSUE #1246: Expose OpenPROMPT questionnaire data in ehrQL
OpenPROMPT questionnaire data is now be available in TPP; it needs to be exposed in ehrQL so the OpenPROMPT studies can use it. This is PROMS data which patients have entered via Airmid, TPP's mobile app.

The schema is now available [here](https://github.com/opensafely-core/ehrql/pull/1267) as part of the ehrQL test suite.

The data will include some non-TPP patients. Obviously these can't be linked to other data, but they will be used in some studies nevertheless. I _think_ that this data will be in a separate table, but I'm not certain of that. (In which case two tables will need exposing.)

This data needs to be included in the [list of restricted tables](https://github.com/opensafely-core/opensafely-cli/blob/main/opensafely/check.py#L14-L19) that can only be accessed with specific permission.  (See https://github.com/opensafely-core/opensafely-cli/issues/203.)

We will need to do some data exploration (via SQL Runner) to convince ourselves that the data is useful.  We should bring in domain experts from LSHTM to help interpret the data.

Acceptance criteria:

- [x] A study using the OpenPROMPT data has generated dummy data
- [x] A study using the OpenPROMPT data has been run against patient data

iaindillingham:
This PR exposes [the `OpenPROMPT` table][1].

First, some terminology. Participants answer questions in questionnaires. Their answers are responses. We expect one questionnaire to be administered once, on day 0:

1. Baseline questionnaire

We expect five questionnaires to be administered four times, on days 0, 30, 60, and 90:

1. Covid and vaccination questionnaire
1. Eq-5D questionnaire
1. Productivity questionnaire
1. Fatigue questionnaire
1. Breathing problems questionnaire

A survey comprises the responses to the questions in between five and six questionnaires on one day. There are four days, so we expect each participant to take part in four surveys.

Based on [airmid-short-data-report][2], I'm confident that:

* Participants are mostly TPP patients. However, there are some non-TPP patients (see the `patient_id` action)
* Each row represents one response (see the `consultation_id_num_rows` action)
* Each row has a valid CTV3 code (`CodedEvent_ID`). We know these codes are valid, because they have corresponding entries in the `CTV3Dictionary` table.
* For those CTV3 codes that are "Y" codes, then for each row `CodeSystemId = 2` and `ConceptId` repeats the CTV3 code (the "Y" code)
* For those CTV3 codes that aren't "Y" codes, then for each row `CodeSystemId = 0` and `ConceptId` is a SNOMED CT code. We don't know these codes are valid, because there isn't a `SNOMEDCTDictionary` table

There are some data quality issues, but these shouldn't prevent us from exposing the `OpenPROMPT` table. For example, several values of `ConsultationDate` pre-date the OpenPROMPT project (see the `consultation_date` action); many participants took part in more than four surveys and many participants took part in less than four surveys (see the `consultation_id` action).

Closes #1246 

[1]: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#OpenPROMPT
[2]: https://github.com/opensafely/airmid-short-data-report

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -360,3 +360,18 @@ class TPPBackend(BaseBackend):
             FROM ISARIC_New
         """
     )
+
+    open_prompt = QueryTable(
+        """
+        SELECT
+            Patient_ID AS patient_id,
+            CodedEvent_ID AS ctv3_code,
+            CASE
+                WHEN CodeSystemId = 0 THEN ConceptId
+            END AS snomedct_code,
+            CAST(ConsultationDate AS date) AS consultation_date,
+            Consultation_ID AS consultation_id,
+            NumericValue AS numeric_value
+        FROM OpenPROMPT
+    """
+    )
--- a/ehrql/tables/beta/tpp.py
+++ b/ehrql/tables/beta/tpp.py
@@ -1,7 +1,7 @@
 import datetime
 
 from ehrql import case, when
-from ehrql.codes import SNOMEDCTCode
+from ehrql.codes import CTV3Code, SNOMEDCTCode
 from ehrql.tables import Constraint, EventFrame, PatientFrame, Series, table
 from ehrql.tables.beta.core import clinical_events, medications, ons_deaths, patients
 
@@ -21,6 +21,7 @@ __all__ = [
     "appointments",
     "ons_cis",
     "isaric_raw",
+    "open_prompt",
 ]
 
 
@@ -458,4 +459,28 @@ class isaric_raw(EventFrame):
     dsstdtc = Series(
         datetime.date,
         description="Outcome date.",
+    )
+
+
+@table
+class open_prompt(EventFrame):
+    ctv3_code = Series(
+        CTV3Code,
+        description="The question, as a CTV3 code",
+    )
+    snomedct_code = Series(
+        SNOMEDCTCode,
+        description="The question, as a SNOMED CT code or None",
+    )
+    consultation_date = Series(
+        datetime.date,
+        description="The date the survey was administered",
+    )
+    consultation_id = Series(
+        int,
+        description="The ID of the survey",
+    )
+    numeric_value = Series(
+        float,
+        description="The response to the question",
     )
--- a/tests/integration/backends/test_tpp.py
+++ b/tests/integration/backends/test_tpp.py
@@ -22,6 +22,7 @@ from tests.lib.tpp_schema import (
     MedicationIssue,
     ONS_CIS_New,
     ONS_Deaths,
+    OpenPROMPT,
     Organisation,
     Patient,
     PatientAddress,
@@ -940,6 +941,48 @@ def test_isaric_raw_clinical_variables(select_all):
         patient_7_results,
         patient_8_results,
         patient_9_results,
+    ]
+
+
+@register_test_for(tpp.open_prompt)
+def test_open_prompt(select_all):
+    results = select_all(
+        OpenPROMPT(
+            Patient_ID=1,
+            CodedEvent_ID="00000",
+            CodeSystemId=0,  # SNOMED CT
+            ConceptId="100000",
+            ConsultationDate="2023-01-01",
+            Consultation_ID=1,
+            NumericValue=1.0,
+        ),
+        OpenPROMPT(
+            Patient_ID=2,
+            CodedEvent_ID="Y0000",
+            CodeSystemId=2,  # CTV3 "Y"
+            ConceptId="Y0000",
+            ConsultationDate="2023-01-01",
+            Consultation_ID=2,
+            NumericValue=1.0,
+        ),
+    )
+    assert results == [
+        {
+            "patient_id": 1,
+            "ctv3_code": "00000",
+            "snomedct_code": "100000",
+            "consultation_date": date(2023, 1, 1),
+            "consultation_id": 1,
+            "numeric_value": 1.0,
+        },
+        {
+            "patient_id": 2,
+            "ctv3_code": "Y0000",
+            "snomedct_code": None,
+            "consultation_date": date(2023, 1, 1),
+            "consultation_id": 2,
+            "numeric_value": 1.0,
+        },
     ]
 
 


ehrql/codes.py
```diff
@@ -100,6 +100,12 @@ class SNOMEDCTCode(BaseCode):
     regex = re.compile(r"[1-9][0-9]{5,17}")
 
 
+class SNOMEDYCode(BaseCode):
```
evansd, line 103: With the caveat the almost anything I say about clinical coding systems could turn out to be wrong, I'm _pretty_ sure that there's no such thing as SNOMED-Y Codes.

SNOMED codes are always integers (they're designed to fit into 64 bit int columns).

I think Y-codes are _something_ like a TPP-specific subset of CTV3 in which case we can use the existing `CTV3Code` type for this – but we'd need to confirm that.

ehrql/tables/beta/tpp.py
```diff
@@ -459,3 +460,48 @@ class isaric_raw(EventFrame):
...
+
+    Each time is a survey. That is, a survey comprises the responses to the questions in
+    between five and six questionnaires.
+    """
+
+    coded_event_id = Series(
+        CTV3Code,
+        description="[TBC] The question, as a CTV3 code",
+    )
+    snomedct_concept_id = Series(
```
evansd, line 488: We've tended to use the phrase "code" rather than "ID" or "concept ID" elsewhere. But we have kept the coding system as part of the column name (exactly as you've done here) where there are multiple coding systems in play. So for this specific column, maybe `snomedct_code`?

ehrql/tables/beta/tpp.py
```diff
@@ -459,3 +460,48 @@ class isaric_raw(EventFrame):
...
+    * Eq-5D questionnaire
+    * Productivity questionnaire
+    * Fatigue questionnaire
+    * Breathing problems questionnaire
+
+    Each time is a survey. That is, a survey comprises the responses to the questions in
+    between five and six questionnaires.
+    """
+
+    coded_event_id = Series(
```
evansd, line 484: This is an odd one because the column name TPP have given it is very much at odds with what it actually seems to contain. If our understanding of what it does contain is correct I'd be inclined to call it `ctv3_code`, but the disparity makes me a bit nervous.

ehrql/backends/tpp.py
```diff
@@ -360,3 +360,21 @@ class TPPBackend(BaseBackend):
...
+        """
+        SELECT
+            Patient_ID AS patient_id,
+            CodedEvent_ID AS coded_event_id,
+            CASE
+                WHEN CodeSystemId = 0 THEN ConceptId
+            END AS snomedct_concept_id,
+            CASE
+                WHEN CodeSystemId = 2 THEN ConceptId
+            END AS snomedy_concept_id,
```
evansd, line 374: (Not a comment on your SQL, this is just a natural place to hang the question!)

Do we know if each survey answer appears as two rows, once with a ConceptId from each system? Or do some questions always use one system and other questions always use another?

evansd (commented):
Thanks for doing this. I think you've taken the right approach here, although I've included a few comments/questions about specifics.

Do you already have a collection somewhere of Questions We Have about this data that we can try to get answers to?

ehrql/codes.py
iaindillingham: Ah! I have probably misunderstood/misremembered. There are such things as "Y" codes, and they are TPP-specific, but they may well be CTV3 codes. I know @inglesp knows more. Could he confirm?

ehrql/tables/beta/tpp.py
iaindillingham: Good to know, thank you. 8b4e506 and b5ee243 update the properties based on `ConceptId`, although I need to investigate that column in more detail.

ehrql/tables/beta/tpp.py
iaindillingham: Yes, me too.

The values in `CodedEvent_ID`[^1] match (by eye) the values in the _CTV3 codes_ column of [the codelist that LSHTM provided][1], which is why I assumed they were CTV3 codes. This could be a mis-labelling on TPP's part or on LSHTM's part. I'm leaning towards the latter.

The `ConceptId` column contains CTV3 codes. I updated the property name to reflect this in 8b4e506. So, I'll not replace `coded_event_id` with `ctv3_code` just yet.

[1]: https://docs.google.com/spreadsheets/d/1bmiQH2Y4EH_PQ4ecaadq_0jWvTHnK0s9/edit#gid=155161282

[^1]: See <https://github.com/opensafely/airmid-short-data-report> and <https://jobs.opensafely.org/datalab/opensafely-internal/airmid-short-data-report/>

ehrql/backends/tpp.py
iaindillingham: A response doesn't have a PK. Nevertheless, we can infer that one response is represented by one row -- rather than, say, two rows; one for each coding system -- by grouping by `Consultation_ID` (a survey's PK) and counting the number of rows. If one response were represented by two rows, we'd expect the number of rows to be even in all cases. It isn't.

iaindillingham: See opensafely/airmid-short-data-report#11

iaindillingham: See opensafely/airmid-short-data-report#12

iaindillingham:
I'm marking this as ready for review. Once reviewed, I'll squash before merging.

ehrql/tables/beta/tpp.py
iaindillingham: We now have `ctv3_code` and `snomedct_code`. Does ehrQL have a mechanism for communicating that the former shouldn't contain null-values, but the latter could?

ehrql/tables/beta/tpp.py
iaindillingham: See above. We now have `ctv3_code` and `snomedct_code`.

tests/integration/backends/test_tpp.py
```diff
@@ -943,6 +944,48 @@ def test_isaric_raw_clinical_variables(select_all):
...
+            CodeSystemId=0,  # SNOMED CT
+            ConceptId="100000",
+            ConsultationDate="2023-01-01",
+            Consultation_ID=1,
+            NumericValue=1.0,
+        ),
+        OpenPROMPT(
+            Patient_ID=2,
+            CodedEvent_ID="00000",
+            CodeSystemId=2,  # CTV3 "Y"
```
inglesp, line 962: As above.

tests/integration/backends/test_tpp.py
```diff
@@ -943,6 +944,48 @@ def test_isaric_raw_clinical_variables(select_all):
...
+            ConceptId="Y0000",
```
inglesp, line 963: According to my reading of the PR description, `ConceptId` should match `CodedEvent_Id` here.

inglesp (changes requested):
Thanks, nearly there!

Do we have a list of the SNOMED codes?  We can cross-reference them against what's in OpenCodelists.

Question: should we be exposing a SNOMED ID at all, if some rows are missing them?

inglesp (approved):
Didn't mean to hit request changes.  I mean, I am requesting changes, but I don't think I need to see this again!

tests/integration/backends/test_tpp.py
```diff
@@ -943,6 +944,48 @@ def test_isaric_raw_clinical_variables(select_all):
     ]
 
 
+@register_test_for(tpp.open_prompt)
+def test_open_prompt(select_all):
+    results = select_all(
+        OpenPROMPT(
+            Patient_ID=1,
+            CodedEvent_ID="00000",
+            CodeSystemId=0,  # SNOMED CT
```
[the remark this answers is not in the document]
iaindillingham: Sorry, that's an error in the description. I've updated it. To clarify: SNOMED CT codes are `0` and CTV3 codes are `2`.

tests/integration/backends/test_tpp.py
iaindillingham: That's an error in the description. Sorry!

tests/integration/backends/test_tpp.py
iaindillingham: It should, thank you for spotting this. Because we're testing that there isn't a `snomedct_code` when `ctv3_code` is a "Y" code, `CodedEvent_ID` matches `ConceptId`. See c772727.

ehrql/backends/tpp.py
```diff
@@ -360,3 +360,18 @@ class TPPBackend(BaseBackend):
...
         """
     )
+
+    open_prompt = QueryTable(
+        """
+        SELECT
+            Patient_ID AS patient_id,
+            CodedEvent_ID AS ctv3_code,
+            CASE
+                WHEN CodeSystemId = 0 THEN ConceptId
```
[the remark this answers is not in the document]
inglesp: > However, I think the best place to check for this is in [airmid-short-data-report](https://github.com/opensafely/airmid-short-data-report).

Are we going to be running this regularly?  And would we notice if it started finding new values for `CodingSystemId`?

iaindillingham:
> Do we have a list of the SNOMED codes? We can cross-reference them against what's in OpenCodelists.

Yes, we have one list [from the codebook](https://docs.google.com/spreadsheets/d/1bmiQH2Y4EH_PQ4ecaadq_0jWvTHnK0s9/edit?usp=share_link&ouid=117227935465371333291&rtpof=true&sd=true) and one list from [airmid-short-data-report](https://github.com/opensafely/airmid-short-data-report). Whilst we could cross-reference the codes in the codebook against OpenCodelists, we may find that the codes in airmid-short-data-report are not a subset of the codes in the codebook. It would be better, then, to request the release of the codes in airmid-short-data-report. This feels like something that a researcher should check, though.

> should we be exposing a SNOMED ID at all, if some rows are missing them?

Possibly not, but I thought that it would be useful for a researcher to compare CTV3 to SNOMED CT, where CTV3 codes are not "Y" codes.

ehrql/tables/beta/tpp.py
evansd: > Does ehrQL have a mechanism for communicating that the former shouldn't contain null-values, but the latter could?

It does indeed! By default, everything is considered nullable, but a `NotNull` constraint can be applied:
https://github.com/opensafely-core/ehrql/blob/bdd093584a7560cf91d4da586e9a1e13b6362e24/ehrql/tables/beta/core.py#L20-L24

iaindillingham: Great, thanks Dave. I should have spotted instances of `Constraint`. I've added them in #1296.

PR #1289, opened 2023-05-27
Add generated docs for all backends and schemas

### LINKED ISSUE #1184: Include full table schema reference in docs
We want a place in the documentation where users can see all the tables and columns available to them, plus as much useful metadata about those as we can include.

We actually already have most of the machinery in place for doing this, but it's currently applied to "contracts": the [`docs.contracts`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/contracts.py) module introspects the codebase and produces metadata; this metadata is consumed by [`docs.render_includes.contracts`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/render_includes/contracts.py) and turned into markdown; and this markdown is then rendered into [nicely formatted HTML](https://docs.opensafely.org/data-builder/contracts/reference/).

But we don't (currently) want to talk about contracts as a user-facing concept. We haven't abandoned the aspiration behind them – which is an abstraction over the data in multiple EHR providers so that a single query can be run against all of them – but we don't have any examples yet and when we do I think we'll end up presenting them differently.

So we should remove contracts from the docs and instead use this machinery to document the entities we _do_ have, which are ...

### Tables

Objects built using the `@table` decorator which can be used to build ehrQL queries.

For example:
https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/tables/beta/tpp.py#L27-L32

### Schemas

Modules which contain one or more tables e.g. [`databuilder.tables.beta.tpp`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/tables/beta/tpp.py).

We don't expose this term to users currently (for instance, users import from `databuilder.tables` not `databuilder.schemas`) but we need some way of referring to this concept internally.

### Backends

Classes for querying a specific EHR providers database.

Backends can declare that they **implement** one or more schemas, which means that the backend defines all the necessary tables and columns to execute queries defined using those schemas. These declarations are checked and enforced by the code.

For example the TPP backend implements the `databuilder.tables.beta.tpp` schema (naturally) and also the `databuilder.tables.beta.smoketest` schema:
https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/backends/tpp.py#L7-L12

The "[smoketest](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/tables/beta/smoketest.py)" schema is an ultra-minimal schema that we ought to be able to implement in any backend and so can be used as part of a basic end-to-end smoketest for any new backend – something we can use to confirm that we can connect to and query the backend. In a sense, it's our first "contract". But as it has no research value I don't think there's any point in calling it that.

## New Pages

None of the below is gospel, just my thoughts on one way of structuring this stuff. So whoever implements this should feel free to make changes if there turns out to be better ways of doing it once we see it in situ.

### 1. Backends index page

We actually have the basics of this already. The metadata is generated by [`docs.backends`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/backends.py), converted to markdown by [`docs.render_includes.backends`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/render_includes/backends.py) and rendered [here](https://docs.opensafely.org/data-builder/contracts/data-backends/).

 * We need to drop the references to "contracts" here and include the list of schemas the backend implements (with each schema linked to the appropriate page – on which more below).
 * I think we should pull in the backend class docstring and display it here as well.
 * Each backend should be anchor-linkable (obviously a bit pointless while we only have one, but I'd like to get the basics in place here).

### 2. Schemas index page

We already collect the necessary metadata for this in [`docs.schemas`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/schemas.py) but don't display it anywhere.

 * Display each schema together with its module docstring and a list of the backends which implement it (again, kind of pointless right now, but let's get the right structures in place while it's easy to do so).
 * Link to the relevant Schema page (as below).
 * Link each backend to the appropriate anchor on the Backends index page.

### 3. Individual Schema pages

For each schema we want a dedicated page showing all the tables it contains and all the columns in those tables. We already collect the metadata needed in [`docs.schemas`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/schemas.py) and most of the formatting logic is already present in [`docs.render_includes.contracts`](https://github.com/opensafely-core/databuilder/blob/3ea4c88a9c6a68a71e94b326b76128a34ae39b56/databuilder/docs/render_includes/contracts.py). I think the current [Contracts Reference](https://docs.opensafely.org/data-builder/contracts/reference/) page is actually pretty close to what we want here. But with a few changes:

 * Tables are already anchor linkable, but I think it would be good if individual columns in tables were linkable too. These tables are likely to get pretty big, especially as we start documenting the columns better and including more constraints and I think it will be very helpful to be able to link directly to a column e.g. when answering questions in Slack.
 * It would be good to repeat the schema information from the index page at the top of the page here as well i.e. schema name, docstring and the list of backends it's implemented by.

## General remarks

 * We should make a decision about whether we expect user-facing docstrings (i.e. those that feature in the documentation) should be plain text or markdown formatted. I think there's an argument for having them markdown formatted because I think that will improve the quality of documentation we can offer.
 * Once we start displaying docstrings, description fields etc on the site I think we'll realise how patchy and inadequate they currently are. But that's fine! I think the fact of having them public will motivate us to improve them.
 * As you can tell, I'm a big fan of making things anchor-linkable. It's really useful to be able to point people to specific bits of a page without resorting to [text-fragment](https://developer.mozilla.org/en-US/docs/Web/Text_fragments) hacks.

evansd:
This adds some pages to the documentation:

 * `/backends/`
 * `/schemas/`
 * a page for each individual schema (i.e. each module in `ehrql.tables`) 

Apart from some introductory text, these are auto-generated from the Python code.

Displaying our docstrings in this way exposed some obvious areas for improvement which I have implemented. But most of the TPP schema remains undocumented so it would be good to start filling this out.

Closes #1184

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.core.md
+++ b/docs/includes/generated_docs/schemas/beta.core.md
@@ -1,0 +1,357 @@
+# <strong>beta.core</strong> schema
+
+Available on backends: [**TPP**](../../backends#tpp), [**EMIS**](../../backends#emis)
+
+This schema defines the core tables and columns which should be available in any backend
+providing primary care data, allowing dataset definitions written using this schema to
+run across multiple backends.
+
+!!! warning
+    This schema is very much still a work-in-progress while the EMIS backend remains
+    under developemnt.
+
+``` {.python .copy title='To use this schema in an ehrQL file:'}
+from ehrql.tables.beta.core import (
+    clinical_events,
+    medications,
+    ons_deaths,
+    patients,
+)
+```
+
+<p class="dimension-indicator"><code>many rows per patient</code></p>
+## clinical_events
+
+
+
+<dl markdown="block" class="schema-column-list">
+<div markdown="block">
+  <dt id="clinical_events.date">
+    <strong>date</strong>
+    <a class="headerlink" href="#clinical_events.date" title="Permanent link">🔗</a>
+    <code>date</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="clinical_events.snomedct_code">
+    <strong>snomedct_code</strong>
+    <a class="headerlink" href="#clinical_events.snomedct_code" title="Permanent link">🔗</a>
+    <code>SNOMED-CT code</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="clinical_events.numeric_value">
+    <strong>numeric_value</strong>
+    <a class="headerlink" href="#clinical_events.numeric_value" title="Permanent link">🔗</a>
+    <code>float</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+</dl>
+
+<p class="dimension-indicator"><code>many rows per patient</code></p>
+## medications
+
+
+
+<dl markdown="block" class="schema-column-list">
+<div markdown="block">
+  <dt id="medications.date">
+    <strong>date</strong>
+    <a class="headerlink" href="#medications.date" title="Permanent link">🔗</a>
+    <code>date</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="medications.dmd_code">
+    <strong>dmd_code</strong>
+    <a class="headerlink" href="#medications.dmd_code" title="Permanent link">🔗</a>
+    <code>dm+d code</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+</dl>
+
+<p class="dimension-indicator"><code>many rows per patient</code></p>
+## ons_deaths
+
+
+
+<dl markdown="block" class="schema-column-list">
+<div markdown="block">
+  <dt id="ons_deaths.date">
+    <strong>date</strong>
+    <a class="headerlink" href="#ons_deaths.date" title="Permanent link">🔗</a>
+    <code>date</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="ons_deaths.place">
+    <strong>place</strong>
+    <a class="headerlink" href="#ons_deaths.place" title="Permanent link">🔗</a>
+    <code>string</code>
+  </dt>
+  <dd markdown="block">
+
+
+ * Possible values: `Care Home`, `Elsewhere`, `Home`, `Hospice`, `Hospital`, `Other communal establishment`
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="ons_deaths.cause_of_death_01">
+    <strong>cause_of_death_01</strong>
+    <a class="headerlink" href="#ons_deaths.cause_of_death_01" title="Permanent link">🔗</a>
+    <code>ICD-10 code</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="ons_deaths.cause_of_death_02">
+    <strong>cause_of_death_02</strong>
+    <a class="headerlink" href="#ons_deaths.cause_of_death_02" title="Permanent link">🔗</a>
+    <code>ICD-10 code</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="ons_deaths.cause_of_death_03">
+    <strong>cause_of_death_03</strong>
+    <a class="headerlink" href="#ons_deaths.cause_of_death_03" title="Permanent link">🔗</a>
+    <code>ICD-10 code</code>
+  </dt>
+  <dd markdown="block">
+
+
+  </dd>
+</div>
+
+<div markdown="block">
+  <dt id="ons_deaths.cause_of_death_04">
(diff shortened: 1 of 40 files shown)


docs/includes/generated_docs/schemas/beta.core.md
```diff
@@ -0,0 +1,357 @@
...
+
+This schema defines the core tables and columns which should be available in any backend
+providing primary care data, allowing dataset definitions written using this schema to
+run across multiple backends.
+
+!!! warning
+    This schema is very much still a work-in-progress while the EMIS backed remains
+    under developemnt.
+
+``` {.python .copy title='To use this schema in an ehrQL file:'}
```
StevenMaude, line 13: thought: In the main documentation, we have `content.code.copy` enabled in `mkdocs.yml`. We don't in this repository. Maybe we should? (I don't think having the attribute matters here particularly.)

docs/includes/generated_docs/schemas/beta.core.md
```diff
@@ -0,0 +1,357 @@
+# <strong>beta.core</strong> schema
+
+Available on backends: [**TPP**](../../backends#tpp), [**EMIS**](../../backends#emis)
+
+This schema defines the core tables and columns which should be available in any backend
+providing primary care data, allowing dataset definitions written using this schema to
+run across multiple backends.
+
+!!! warning
+    This schema is very much still a work-in-progress while the EMIS backed remains
```
StevenMaude, line 10: nit:

```suggestion
    This schema is very much still a work-in-progress while the EMIS backend remains
```

docs/includes/generated_docs/schemas/beta.core.md
```diff
@@ -0,0 +1,357 @@
...
+
+
+<dl markdown="block" class="schema-column-list">
+<div markdown="block">
+  <dt id="clinical_events.date">
+    <strong>date</strong>
+    <a class="headerlink" href="#clinical_events.date" title="Permanent link">🔗</a>
+    <code>date</code>
+  </dt>
+  <dd markdown="block">
```
StevenMaude, line 34: thought: This HTML is needed throughout because Python's Markdown support isn't flexible to write it in Markdown. So it's slightly clunky and repetitive, but it's fine: we're not hand-editing this form of the content. At worst, it would be the template that's edited.

There are:

* [definition lists](https://python-markdown.github.io/extensions/definition_lists/)
* [attribute lists](https://python-markdown.github.io/extensions/attr_list/), but these don't support code blocks, so don't know if they would even work

docs/stylesheets/extra.css
```diff
@@ -0,0 +1,42 @@
...
+dl.schema-column-list {
+  font-size: .64rem;
+  border: .05rem solid var(--md-typeset-table-color);
+  border-radius: .1rem;
+  border-bottom: none;
+}
+
+dl.schema-column-list > div {
+  border-bottom: .05rem solid var(--md-typeset-table-color);
+  padding: .7720588235em 1.1764705882em;
```
StevenMaude, line 15: question: Where do these numbers come from? They look like they've been calculated.

ehrql/docs/schemas.py
```diff
@@ -1,15 +1,21 @@
...
 from ehrql import tables
-from ehrql.query_language import BaseFrame, Series
+from ehrql.codes import BaseCode
+from ehrql.query_language import BaseFrame, PatientFrame, Series
 from ehrql.utils.module_utils import get_submodules
 
 from .common import reformat_docstring
 
 
+SORT_ORDER = {k: i for i, k in enumerate(["beta.tpp", "beta.core"])}
```
StevenMaude, line 12: thought: Eventually the core schema should probably be first, then the more specific schemas. But seeing as TPP is more fully featured at this time, I see why this is not the case now.

StevenMaude (approved):

docs/includes/generated_docs/schemas/beta.core.md
evansd: Yes, I went through exactly this thought process myself. Initially it felt like a hack, but I think for auto-generated stuff like this it's fine to resort to raw HTML to get the layout or structure we want because it's easy to change later.

docs/stylesheets/extra.css
evansd: I just used DevTools to grab styles from existing bits of the theme. In this particular case, I wanted the definition list to have the same styling as a table so I just found an example table and borrowed the styles from there.

ehrql/docs/schemas.py
evansd: Yes, I think that's exactly right.

docs/includes/generated_docs/schemas/beta.core.md
evansd: That sounds sensible, but happy to leave that call to you.

PR #1294, opened 2023-05-30
chore: Upgrade pymdown-extensions to 10.0.1

StevenMaude:
This fixes a bug which stops include files from being loaded in the documentation.


StevenMaude:
This fixes the issue that @evansd spotted when building the docs.

StevenMaude:
The fix is also in #1289, but it seems good to fix this now since it's broken.

iaindillingham (approved):

Issue #1181, opened 2023-04-04
Ensure dummy data contains missing values


iaindillingham:
If dummy data doesn't contain missing values, but real data does, then downstream actions that don't expect missing values, such as [matching][1], will succeed when run against dummy data but fail when run against real data.

I think this issue underlies [this tech support request][2], although the tech support request is ongoing.

[1]: https://github.com/opensafely-core/matching
[2]: https://bennettoxford.slack.com/archives/C010SJ89SA3/p1680513917587309

PR #1296, opened 2023-05-31
Add `NotNull` constraints to `open_prompt`

iaindillingham:
We know these properties are not nullable from the database schema [1] and from the results of airmid-short-data-report [2].

[1]: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#OpenPROMPT
[2]: https://github.com/opensafely/airmid-short-data-report

### DIFF
--- a/ehrql/tables/beta/tpp.py
+++ b/ehrql/tables/beta/tpp.py
@@ -474,6 +474,7 @@ class isaric_raw(EventFrame):
 class open_prompt(EventFrame):
     ctv3_code = Series(
         CTV3Code,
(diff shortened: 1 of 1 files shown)


evansd (approved):
Excellent!

We haven't made much use of this constraint yet so I'm not surprised you hadn't come across it. Partly this is because it doesn't _do_ very much right now. But given that we'll shortly be including these in the docs (when #1289 lands) and we plan to use them to influence dummy data (#1181) I think it's good to start getting the constraints in place.

Issue #1219, opened 2023-04-20
Replace existing Data Builder/ehrQL documentation with the wiki rewrite


StevenMaude:
We're reworking the documentation from scratch.

When we're happy that it's in the right direction, we should move it into the repository and remove the previously written content.

StevenMaude:
Things to consider:
* Internal wiki links will need rewriting to point to appropriate Markdown
* We should check that dataset definitions, `project.yaml` files behave as expected.
  * In some cases, we give examples of error-causing dataset definitions. We may also want to validate that these give the appropriate error. We don't yet have an easy mechanism for doing so. 
* We may want to use code line highlighting in examples that we cannot currently do so (due to note being a part of GitHub's Markdown implementation).

StevenMaude:
Any other ehrQL documentation work probably should be held until this is done.

StevenMaude:
Related issues:

* #1049

Issues that are likely deprecated following this move:

* #1041
* #1051
* #1052

StevenMaude:
:hammer: WIP

# Remove from the existing documentation

## Documentation pages

* `/comparison-with-cohort-extractor`
* `/dataset-definition`
* `/dataset-sql-explanation`
* `/dummy-data`
* `/`
* `/patient-event-tables-explanation`
* `/query-engine-explanation`
* `/quick-start`
* `/versioning`

## Code snippets

* `/snippets`
* `/ehrql-tutorial-examples`

## Includes

* `/includes/data-builder-danger-header`

## Other miscellany

* `../Justfile` entries
* `../pre-commit-config.yaml` entries that refer to generating docs
* Some of the `../.github/workflows/check-docs.yml` configuration (for now, until we reintroduce checks of examples)

# Retain in the existing documentation

* `/backends`
* `/schemas`
* `/toc` (for now, until the URL migration gets done; see below on "Redirect")

# Redirects from the existing documentation URLs

* Some of the existing URLs should get redirected.
  * This has two aspects:
    * Adding redirects for `data-builder` to `/ehrql` if the remainder of the URL remains the same
    * Adding redirects for the old URL to the new if the remainder of the URL changes, but there is a similar page.  
  * Relates to https://github.com/opensafely/documentation/issues/1091 and https://github.com/opensafely/documentation/issues/1242.

# Update from the existing documentation

* `../DEVELOPERS.md`

PR #1298, opened 2023-06-01
Remove "Contracts" related documentation

evansd:
To be very clear here (particularly for Seb's benefit!) we are not abandoning the idea, or even necessarily the terminology, behind contracts. But I think it's important to maintain conceptual and terminological consistency in the codebase and documentation; and what we had previously contained a mixture of asirational and evolving ideas that didn't directly connect to the reality of how things worked.

I've updated parts of the existing documentation which I think are shortly going to be replaced anyway (see #1219) because it was easier to do everything than to worry about what was and wasn't going to be kept.

Depends on #1289


StevenMaude (approved):
This looks fine from a glance through.

I couldn't find any other references to `[Cc]ontracts?`, nor any obvious links to the removed pages.

PR #1302, opened 2023-06-05
Add support for count distinct

### LINKED ISSUE #546: Add `count_distinct` series aggregation
This method would go on the `EventSeries` type so as to be available to series of all types:
https://github.com/opensafely-core/databuilder/blob/2ce31449d6e35b21d45a0e96c746c410be6b7ac9/databuilder/query_language.py#L87-L94

There's no support anywhere for it at the moment, so a new query model operation would need to be added, and then support for it added to the query engines.

Creating appropriate spec tests for these methods should drive out all the work that needs doing.

I have a particular use case in mind for the combination of `count_distinct` and [`to_first_of_month`](https://github.com/opensafely-core/databuilder/issues/545) because it will allow us to count the number of distinct months in which a patient was prescribed a medication. This seems to be used as the basis for several variable definitions and we currently have to deal with this in a clunky way by defining a bunch of intermediate binary variables (one for each month) and then using logical expressions to count them.

milanwiedemann:
Closes #546

### DIFF
--- a/ehrql/query_model/nodes.py
+++ b/ehrql/query_model/nodes.py
@@ -238,6 +238,9 @@ class AggregateByPatient:
     class Count(OneRowPerPatientSeries[int]):
         source: Frame
 
+    class CountDistinct(AggregatedSeries[int]):
+        source: Series[Any]
+
     class Min(AggregatedSeries[Comparable]):
         source: Series[Comparable]
 
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -99,6 +99,7 @@ def variable(patient_tables, event_tables, schema, value_strategies):
             select_column: (value_strategies.keys(), DomainConstraint.ANY),
             exists: ({bool}, DomainConstraint.PATIENT),
             count: ({int}, DomainConstraint.PATIENT),
+            count_distinct: ({int}, DomainConstraint.PATIENT),
             min_: (comparable_types(), DomainConstraint.PATIENT),
             max_: (comparable_types(), DomainConstraint.PATIENT),
             sum_: ({int, float}, DomainConstraint.PATIENT),
@@ -163,6 +164,12 @@ def variable(patient_tables, event_tables, schema, value_strategies):
 
     def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
+
+    @st.composite
+    def count_distinct(draw, _type, _frame):
+        type_ = draw(any_type())
+        frame = draw(many_rows_per_patient_frame())
+        return AggregateByPatient.CountDistinct(draw(series(type_, frame)))
 
     def min_(type_, _frame):
         return aggregation_operation(type_, AggregateByPatient.Min)
--- a/tests/spec/aggregate_series/test_count_distinct.py
+++ b/tests/spec/aggregate_series/test_count_distinct.py
@@ -1,0 +1,74 @@
+from ..tables import e
+
+
+title = "Count distinct aggregation"
+
+table_data = {
+    e: """
+          |  i1 |  f1 | s1 |     d1
+        --+-----+-----+----+------------
+        1 | 101 | 1.1 |  a | 2020-01-01
+        1 | 102 | 1.2 |  b | 2020-01-02
+        1 | 103 | 1.5 |  c | 2020-01-03
+        2 | 201 | 2.1 |  a | 2020-02-01
+        2 | 201 | 2.1 |  a | 2020-02-01
+        2 | 203 | 2.5 |  b | 2020-02-02
+        3 | 301 | 3.1 |  a | 2020-03-01
+        3 | 301 | 3.1 |  a | 2020-03-01
+        3 |     |     |    |
+        3 |     |     |    |
+        4 |     |     |    |
+        """,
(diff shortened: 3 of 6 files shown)


ehrql/query_model/nodes.py
```diff
@@ -238,6 +238,9 @@ class Exists(OneRowPerPatientSeries[bool]):
     class Count(OneRowPerPatientSeries[int]):
         source: Frame
 
+    class CountDistinct(AggregatedSeries[int]):
+        source: Series[Comparable]
```
evansd, line 242: This can actually be `Series[Any]`. The `Comparable` class (which I don't think is a great name but it fits with conventions used elsewhere) is for types which have `>` and `<` defined. But here we only need types which can be compared as equal or not, and all types support that.

tests/generative/variable_strategies.py
```diff
@@ -164,6 +165,9 @@ def exists(_type, _frame):
     def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
 
+    def count_distinct(type_, _frame):
+        return aggregation_operation(type_, AggregateByPatient.CountDistinct)
```
evansd, line 169: This is very close to being correct, but not quite! `CountDistinct` always returns an `int`, but it can take series of any type. So we can ignore the type we're passed (which will always be `int`) and instead pick a new type to use as the input.
```suggestion
    def count_distinct(_type, _frame):
        return aggregation_operation(any_type(), AggregateByPatient.CountDistinct)
```

(I'm following the convention here that ignored arguments have a leading underscore.)

tests/spec/aggregate_series/test_count_distinct.py
```diff
@@ -0,0 +1,66 @@
...
+    e: """
+          |  i1 |  f1 | s1 |     d1
+        --+-----+-----+----+------------
+        1 | 101 | 1.1 |  a | 2020-01-01
+        1 | 102 | 1.2 |  b | 2020-01-02
+        1 | 103 | 1.5 |  c | 2020-01-03
+        2 | 201 | 1.1 |  a | 2020-02-01
+        2 | 201 | 1.1 |  a | 2020-02-01
+        2 | 203 | 1.5 |  b | 2020-02-02
+        3 |     |     |    |
```
evansd, line 16: I think the test cases need to include a patient who has a mixture of NULL and non-NULL values. We want to ensure that, e.g., `101`, `101`, `NULL`, `NULL` is counted as 1 distinct value not 3 (or even 4, some databases regard every NULL value as unique).

ehrql/utils/math_utils.py
```diff
@@ -16,3 +16,10 @@ def floordiv(lhs, rhs):
         return None
     else:
         return int(lhs // rhs)
+
+
+def count_distinct(series):
+    """
+    Count distinct values in a series.
+    """
+    return len(set(series))
```
evansd, line 25: You asked whether this was the right place for this function and I'm not sure it is. I think it would actually be simpler and clearer to define it inline in `InMemoryQueryEngine`. There are already a few examples of this e.g.:
https://github.com/opensafely-core/ehrql/blob/ad6ac4a389d0be2a2a0faf918e4736aa324df2ed/ehrql/query_engines/in_memory.py#L292-L296

In fact, I think we should probably move the other math_utils functions inline as well, but that's a job for a different PR.

The reason we initially had them separate like this was that they got used in both the in-memory engine and in the population validation module. But I've since simplified the population validation code to re-use the logic from the in-memory engine, so we no longer need to define behaviour in two places.

evansd (approved):
Looks great! There are a couple of small changes to make, but nothing huge.

You'll also need to add the prefix `feat:` to one of the commit message to get it to generate a new release (this still makes me sad).

ehrql/query_model/nodes.py
milanwiedemann: thanks!

tests/generative/variable_strategies.py
milanwiedemann: thanks :)

tests/spec/aggregate_series/test_count_distinct.py
milanwiedemann: good idea!

tests/generative/variable_strategies.py
```diff
@@ -164,6 +165,9 @@ def exists(_type, _frame):
     def count(_type, _frame):
         return st.builds(AggregateByPatient.Count, any_frame())
 
+    def count_distinct(_type, _frame):
+        return aggregation_operation(any_type(), AggregateByPatient.CountDistinct)
```
evansd, line 169: Ughh, so my "correction" was itself incorrect. This Hypothesis code can be really hard to reason about because it's easy to forget whether you're working with a value of a certain type, or a _strategy_ for generating values of that type.

In any case, I've actually tested the below this time and I'm pretty sure it's what we want.
```suggestion
    @st.composite
    def count_distinct(draw, _type, _frame):
        type_ = draw(any_type())
        frame = draw(many_rows_per_patient_frame())
        return AggregateByPatient.CountDistinct(draw(series(type_, frame)))
```

evansd (approved):

PR #1303, opened 2023-06-07
Don't treat `values_used` as categories

### LINKED ISSUE #1266: Dummy data should not treat "values used" as equivalent to categories
## Context

As part of dummy data generation the `query_info` module finds every comparison made with every column used in a dataset definition and records them in a `values_used` set for each column:
https://github.com/opensafely-core/ehrql/blob/c58fbd8bf3e35983dd98944d8ea7222c143f1fc3/ehrql/dummy_data/query_info.py#L41-L44

This is all well and good, but when we come to generate dummy data we treat this `values_used` set just as if it were a categorical constraint 
https://github.com/opensafely-core/ehrql/blob/c58fbd8bf3e35983dd98944d8ea7222c143f1fc3/ehrql/dummy_data/query_info.py#L47-L52

And we end up saying that only these values can appear in the column.
https://github.com/opensafely-core/ehrql/blob/c58fbd8bf3e35983dd98944d8ea7222c143f1fc3/ehrql/dummy_data/generator.py#L212-L214

This is particularly silly when only a single comparison has been made because then we only have one value to work with so we just use it every time (see #1265).

I think the logic we want is:
 * If there is a categorical constraint on the column then that determines the set of valid values and we should always choose one of those.
 * Otherwise, if there are `values_used` then we should randomly choose one of these _or_ choose to fall through to whatever random generation approach we would have used otherwise.
 * The probability distribution according to which we choose to fall through or not should depend on the number of values we have. I don't know the exact logic we want here, but clearly if we've only got one value then we want to be falling through certainly more than 50% of the time. Whereas if we've got 30 values then I think we want to be mostly using those and only occasionally falling through.

## Implementation notes

Dave's sketch of how this could work:

 * Get rid of the `choices` property on `ColumnInfo`.

 * Rename the `values_used` attribute to `_values_used` to make it internal.

 * Add a public `values_used` property which returns the sorted values:
   ```py
   @cached_property
   def values_used(self):
       return sorted(self._values_used)
   ```
   (It's worth caching this because we'll be accessing this list frequently.)

 * The first thing that `DummyPatientGenerator.get_random_value()` should do is check for a categorical constraint:
   ```py
   if cat_constraint := column_info.get_constraint(Constraint.Categorical):
       return self.rnd.choice(cat_constraint.values)
   ```

 * The next thing is to check whether `column_info.values_used` is empty, and if it's not we need to decide whether we're going to use one of these values or not. @inglesp suggested doing this with probability `1/(N+1)` for `N` values, so something like (untested code):
   ```py
   if column_info.values_used:
       if self.rnd.randint(0, len(column_info.values_used)) != 0:
           return self.rnd.choice(column_info.values_used)
   ```

 * Finally we fall back to the `if/elif` chain based on `column_info.type`.

 * The dummy data system has 100% unit test coverage so it supports a fast testing cycle while working on it:
   ```
   pytest --cov ehrql.dummy_data --cov-report=html tests/unit/dummy_data/
   ```

iaindillingham:
A column may admit more values than those stored within `values_used`. Consequently, we fall through to other values with a probability of `1/(N+1)`.

Closes #1266

### DIFF
--- a/ehrql/dummy_data/generator.py
+++ b/ehrql/dummy_data/generator.py
@@ -209,9 +209,12 @@ class DummyPatientGenerator:
 
     def get_random_value(self, column_info):
         # TODO: This never returns None although for realism it sometimes should
-        if column_info.choices:
+        if cat_constraint := column_info.get_constraint(Constraint.Categorical):
             # TODO: It's obviously not true in general that categories are equiprobable
-            return self.rnd.choice(column_info.choices)
+            return self.rnd.choice(cat_constraint.values)
+        elif column_info.values_used:
+            if self.rnd.randint(0, len(column_info.values_used)) != 0:
+                return self.rnd.choice(column_info.values_used)
         elif column_info.type is bool:
             return self.rnd.choice((True, False))
         elif column_info.type is int:
--- a/ehrql/dummy_data/query_info.py
+++ b/ehrql/dummy_data/query_info.py
@@ -4,7 +4,6 @@ from functools import cached_property
 
 from ehrql.query_model.nodes import (
     Column,
-    Constraint,
     Function,
     InlinePatientTable,
     SelectColumn,
@@ -26,7 +25,7 @@ class ColumnInfo:
     name: str
     type: type  # NOQA: A003
     constraints: tuple = ()
-    values_used: set = dataclasses.field(default_factory=set)
+    _values_used: set = dataclasses.field(default_factory=set)
 
     @classmethod
     def from_column(cls, name, column):
@@ -41,15 +40,11 @@ class ColumnInfo:
     def record_value(self, value):
         if hasattr(value, "_to_primitive_type"):
             value = value._to_primitive_type()
-        self.values_used.add(value)
+        self._values_used.add(value)
 
     @cached_property
-    def choices(self):
-        cat_constraint = self.get_constraint(Constraint.Categorical)
-        if cat_constraint is not None:
-            return cat_constraint.values
-        else:
-            return sorted(self.values_used)
+    def values_used(self):
+        return sorted(self._values_used)
 
     def get_constraint(self, type_):
         return self._constraints_by_type.get(type_)
--- a/tests/unit/dummy_data/test_query_info.py
+++ b/tests/unit/dummy_data/test_query_info.py
@@ -54,7 +54,7 @@ def test_query_info_from_variable_definitions():
                     "code": ColumnInfo(
                         name="code",
                         type=str,
-                        values_used={"abc00"},
+                        _values_used={"abc00"},
                     )
                 },
             ),
@@ -106,7 +106,7 @@ def test_query_info_records_values():
     assert column_info == ColumnInfo(
         name="value",
         type=str,
-        values_used={"a", "b", "c", "d"},
+        _values_used={"a", "b", "c", "d"},
     )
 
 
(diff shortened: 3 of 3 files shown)


tests/unit/dummy_data/test_query_info.py
```diff
@@ -54,7 +54,7 @@ def test_query_info_from_variable_definitions():
                     "code": ColumnInfo(
                         name="code",
                         type=str,
-                        values_used={"abc00"},
+                        _values_used={"abc00"},
```
iaindillingham, line 57: I questioned whether there was a better approach in [this Slack thread][1]. Outside tests, we'd call `ColumnInfo.record_value` rather than `ColumnInfo._values_used.add`. However, passing the value of an internal attribute like this seems a bit unpleasant.

[1]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1686143435147369

ehrql/dummy_data/query_info.py
```diff
@@ -41,15 +40,11 @@ def __post_init__(self):
...
 
     @cached_property
-    def choices(self):
-        cat_constraint = self.get_constraint(Constraint.Categorical)
-        if cat_constraint is not None:
-            return cat_constraint.values
-        else:
-            return sorted(self.values_used)
+    def values_used(self):
+        return set(sorted(self._values_used))
```
evansd, line 47: I don't think you want the `set()` wrapper here: `_values_used` is already a set, so this is just undoing the affect of the sorting which we want to give us deterministic dummy data.

ehrql/dummy_data/generator.py
```diff
@@ -209,9 +209,12 @@ def populate_row(self, table_info, row):
...
     def get_random_value(self, column_info):
         # TODO: This never returns None although for realism it sometimes should
-        if column_info.choices:
+        if cat_constraint := column_info.get_constraint(Constraint.Categorical):
             # TODO: It's obviously not true in general that categories are equiprobable
-            return self.rnd.choice(column_info.choices)
+            return self.rnd.choice(cat_constraint.values)
+        elif column_info.values_used:
+            if self.rnd.randint(0, len(column_info.values_used)) != 0:
+                return self.rnd.choice(list(column_info.values_used))
```
evansd, line 217: If you remove the call to `set()` from the `values_used` property then you don't need to transform it to a list here.

tests/unit/dummy_data/test_query_info.py
evansd: Ah I see. I don't think it's unreasonable for test code to use internal things as part of its setup. But an alternative would be to leave the attribute with its old name, `values_used` and call the property something like `all_values_used`.

evansd (approved):
Thanks for this!

One small comment about the use of `set()`. And you'll also need a pedantic commit prefix if you want to create a new release.

tests/unit/dummy_data/test_query_info.py
iaindillingham: I think I'm going to leave this as-is, because it's test code. Thank you for highlighting an alternative.

ehrql/dummy_data/query_info.py
iaindillingham: I added this to ensure the return type was consistent, but on reflection I was wrong: whilst `_values_used` is a set, `values_used` need not be. I'd also forgotten that `set`s don't maintain insertion order. Thanks for your comment!

PR #1305, opened 2023-06-08
Include "table helper" methods in docs

evansd:
This means things like the `patients.age_on()` method or `practice_registrations.for_patient_on()`.


StevenMaude:
Looks good to me!

PR #1317, opened 2023-06-13
Include wiki content as docs

### LINKED ISSUE #1219: Replace existing Data Builder/ehrQL documentation with the wiki rewrite

StevenMaude:
Fixes #1219.

Roughly, this PR:

* adds the wiki documentation content that the Bennett Institute data team collectively wrote
* fixes it up so that it works better in MkDocs
* re-organises the pages slightly along the lines of the [Diataxis framework](https://diataxis.fr); this may not be absolute perfect or final, but seems to work quite well here.
* :warning: this will break existing URLs; @inglesp suggested it wasn't worth redirecting them

This PR does not:

* include the Python examples as separate files; they are all inline
* test any of those Python console examples or dataset definitions

Testing will be added in a future PR, as it alone could be a moderate task, and this PR is big enough.

opensafely/documentation#1251 should be merged first, because this has configuration changes relating to the changes in this PR.


StevenMaude:
Question: is there too much or too few mentions of "ehrQL" in the table of contents/page titles? It can look a little bit repetitive.

StevenMaude:
Later:

- [ ] consider using admonitions in text where appropriate
- [ ] delete the wiki pages after merge of this PR
- [ ] add a pointer from the wiki pages to the MkDocs-based documentation
- [ ] separate out example Python console prompts, dataset definitions and outputs 
- [ ] edit backends page introduction (sentences are quite long)
- [ ] consider including schema pages as actual pages instead of symlink (if you click on one, you lose the table of contents navigation)
- [ ] test all the examples for success/failure as appropriate

Not doing:

- [ ] ~add suitable redirects for Data Builder URLs where necessary~

inglesp (approved):
Thanks for addressing feedback.

Issue #1270, opened 2023-05-17
Add support for querying medications without dm+d codes in TPP


inglesp:
In TPP, medications in the `MedicationIssue` table are recorded with Multilex codes.  The `MedicationDictionary` table maps from Multilex to dm+d codes, allowing users to look up medication events with a dm+d codelist.

Some medication events have Multilex codes that do not have a mapping to a dm+d code.  Our clinical informatics team have investigated this and in most cases this is not a problem.  However, there are some Multiliex codes whose mapping to a dm+d code is missing from `MedicationDictionary`.  In particular, a significant number of morphine prescriptions do not have dm+d codes, and so are missed.  See discussion at https://github.com/ebmdatalab/sensitive-issues/issues/15.

In some cases the `MedicationDictionary` may be updated (eventually) but this is not guaranteed, so we need to be able to use our own mappings (provided by the clinical informatics team).  The mappings can be stored in a CSV file in this repo.

We should ensure that duplicates between our own mapping and the mapping in `MedicationDictionary` do not result in double counting medication events, and we should prefer mappings from `MedicationDictionary`.  If this is hard to do efficiently on each query, we should consider some other way of checking that there are no duplicates.

Acceptance criteria:

- [x] A test dataset definition using the [morphine-subcutaneous-dmd codelist](https://www.opencodelists.org/codelist/opensafely/morphine-subcutaneous-dmd/1185fc5b/) captures more events after this fix is deployed than before it is deployed.

evansd:
At one level, I think this is fairly simple. It's a case of changing the QueryTable definition from:
https://github.com/opensafely-core/ehrql/blob/06c19588d14f5f5aa2d821239f6d9e2d0b2ddbe6/ehrql/backends/tpp.py#L98-L109

To something which includes an extra join on `OurCustomDictionary` (but prefers results from `MedicationDictionary` where they exist):
```sql
SELECT
    meds.Patient_ID AS patient_id,
    CAST(meds.ConsultationDate AS date) AS date,
    COALESCE(dict.DMD_ID, our_dict.DMD_ID) AS dmd_code,
    meds.MultilexDrug_ID AS multilex_code
FROM MedicationIssue AS meds
LEFT JOIN MedicationDictionary AS dict
ON meds.MultilexDrug_ID = dict.MultilexDrug_ID
LEFT JOIN OurCustomDictionary AS our_dict
ON meds.MultilexDrug_ID = our_dict.MultilexDrug_ID
```

[...]

PR #1337, opened 2023-06-20
Add support for querying medications without dm+d codes in TPP

iaindillingham:
I'd like some help with two aspects of this PR, which addresses #1270.

---

**First**, the `CustomMedicationDictionary` table contains our (Bennett Institute) mapping. I think this table should live within a temporary database; and I think the name of this temporary database is the value of `query_engine_instance.config.get("TEMP_DATABASE_NAME")`. However, I'm unsure how to access this value, here:

https://github.com/opensafely-core/ehrql/blob/de94fb15fc5fad1dbaf83678b40962b6add81623/ehrql/backends/tpp.py#L113

To put that another way, how do I access a query engine instance from a backend class?

I think I should make `TPPBackend.medications` a property, but how do I access a query engine instance from a backend instance?

---

**Second**, I've separated the management of `CustomMedicationDictionary` (i.e. dropping it, creating it, and inserting into it) from the use of this table. The management of this table is done here:

https://github.com/opensafely-core/ehrql/blob/de94fb15fc5fad1dbaf83678b40962b6add81623/ehrql/main.py#L279-L314

However, the implementation is tightly-coupled to TPP. Originally, I'd located the SQL statements (`text(...)`) within `TPPBackend`, reasoning that we'd have different SQL statements for `EMISBackend`. Doing so ensures the implementation is loosely coupled to TPP, but the separation between the data on the one hand and ehrQL on the other isn't as clean. Is there an approach that affords loose coupling _and_ separation between data and ehrQL?

### DIFF
--- a/ehrql/__main__.py
+++ b/ehrql/__main__.py
@@ -16,6 +16,7 @@ from .main import (
     generate_measures,
     run_sandbox,
     test_connection,
+    update_custom_medication_dictionary,
 )
 
 
@@ -66,6 +67,7 @@ def main(args, environ=None):
     add_generate_measures(subparsers, environ, user_args)
     add_run_sandbox(subparsers, environ, user_args)
     add_test_connection(subparsers, environ, user_args)
+    add_update_custom_medication_dictionary(subparsers, environ, user_args)
 
     kwargs = vars(parser.parse_args(args))
     function = kwargs.pop("function")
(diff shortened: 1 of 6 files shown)


iaindillingham:
opensafely-core/cohort-extractor#961 covers managing `CustomMedicationDictionary`; #1341 adds dynamically constructed queries. Both mean that the approach in this PR needs reworking, so I'm going to close it without merging and open another.

PR #1341, opened 2023-06-22
Support QueryTables with dynamically constructed queries

evansd:
This refactors the backed definition code in two ways:
 * Backend instances now receive a copy of the environment via the `config` keyword argument in the same way that QueryEngines do.
 * Methods on SQLTable objects are now passed a reference to the backend instances.

These allow us to provide a new way of constructing a QueryTable by using `@QueryTable.from_function` as a decorator. This means that the query returned can now depend on the contents of the `config` dictionary, for example:
```py
class SomeBackend:
    ...

    @QueryTable.from_function
    def positive_tests(self):
        table_name = self.config.get("table_name", "some_table")
        return f"SELECT patient_id, date FROM {table_name}"

```

### DIFF
--- a/ehrql/backends/base.py
+++ b/ehrql/backends/base.py
@@ -17,6 +17,9 @@ class BaseBackend:
     tables = None
     implements = ()
 
+    def __init__(self, config=None):
+        self.config = config or {}
+
     def __init_subclass__(cls, **kwargs):
         assert cls.display_name is not None
         assert cls.query_engine_class is not None
@@ -140,6 +143,9 @@ class QueryTable(SQLTable):
 
 
 class DefaultBackend:
+    def __init__(self, config=None):
+        pass
+
     def get_table_expression(self, table_name, schema):
         """
         Returns a SQLAlchemy Table object matching the supplied name and schema
--- a/ehrql/backends/base.py
+++ b/ehrql/backends/base.py
@@ -32,8 +32,10 @@ class BaseBackend:
                 cls._init_table(value)
                 cls.tables[name] = value
 
-        for table_namespace in cls.implements:
-            cls.validate_against_table_namespace(table_namespace)
+        # Construct an instance in order to validate it
+        instance = cls()
+        for table_namespace in instance.implements:
+            instance.validate_against_table_namespace(table_namespace)
 
     @classmethod
     def _init_table(cls, table):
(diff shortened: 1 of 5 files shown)


ehrql/backends/base.py
```diff
@@ -44,23 +46,22 @@ def _init_table(cls, table):
...
-                    f"{cls} does not implement table '{table.name}' from "
+                    f"{self.__class__} does not implement table '{table.name}' from "
                     f"{table_namespace}.{attr}"
                 )
 
-            table_def = cls.tables[table.name]
+            table_def = self.tables[table.name]
             try:
-                table_def.validate_against_table_schema(table.schema)
+                table_def.validate_against_table_schema(self, table.schema)
```
iaindillingham, line 60: I was puzzled by `self` here - I thought this `self` would replace the `table_def.validate_against_table_schema(self, ...)` (possibly a half-remembered piece of JavaScript?). Of course, it doesn't: this `self` is an instance of `BaseBackend`, which is the second positional argument to `MappedTable.validate_against_table_schema`.

To my inexperienced eye, passing `self` like this looks unusual. Instead, how about:

```suggestion
                table_def.validate_against_table_schema(backend=self, schema=table.schema)
```

ehrql/backends/base.py
```diff
@@ -74,7 +75,7 @@ def get_table_expression(self, table_name, schema):
         Raises:
             ValueError: If unknown table passed in
         """
-        return self.tables[table_name].get_expression(table_name, schema)
+        return self.tables[table_name].get_expression(self, table_name, schema)
```
iaindillingham, line 78: Naturally, I'm no longer puzzled by `self`.

iaindillingham (approved):
Goodness, this looks nothing like my attempt! I started out with something like...

```py
class TPPBackend(BaseBackend):
    @property
    def medications(self):
        return "SELECT..."
```

I understand the changes introduced by this PR, and I've made a couple of comments in places where I had to stop and think about what I was reading. I've one question, which I can't seem to add as a comment because the code hasn't changed. Why not...

```py
class SQLTable:
    def validate_against_table_schema(self, backend, schema):
        raise NotImplementedError()
```

(That is, why not pass `backend` as an argument to this method, as this method seems to be an interface.)

ehrql/backends/base.py
evansd: Good suggestion, thanks.

PR #1347, opened 2023-06-22
Make `weeks` importable like other time durations

StevenMaude:
Spotted this inconsistency while updating imports for ehrQL example code in the documentation.

The current import used is:

```
from databuilder.query_language import weeks
```

https://github.com/opensafely-core/ehrql/blob/2ab0e357aa86ecff6103666d274794eed73c7c06/docs/explanation/examples.md#L554

I wanted to update this to `from ehrql import weeks` and noticed that this wasn't possible.


evansd (approved):
Thanks Steve, great catch!

PR #1348, opened 2023-06-23
feat: Query `CustomMedicationDictionary`

iaindillingham:
In this PR, we add support for querying medications that don't have dm+d codes stored in the `MedicationDictionary` table (#1270). The dm+d codes for these medications are now stored in the `CustomMedicationDictionary` table (it exists!), which is managed by cohort-extractor (opensafely-core/cohort-extractor#961).

If `CustomMedicationDictionary` doesn't exist -- because it hasn't been created, for example -- then the query will fail. I think this is desired behaviour: If we're not querying both `MedicationDictionary` and `CustomMedicationDictionary`, then we want to know ASAP.

As required by #1270, we prefer `MedicationDictionary` to `CustomMedicationDictionary`.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -101,17 +101,22 @@ class TPPBackend(BaseBackend):
         """
     )
 
-    medications = QueryTable(
-        """
+    @QueryTable.from_function
+    def medications(self):
+        temp_database_name = self.config.get(
+            "TEMP_DATABASE_NAME", "PLACEHOLDER_FOR_TEMP_DATABASE_NAME"
+        )
+        return f"""
             SELECT
                 meds.Patient_ID AS patient_id,
                 CAST(meds.ConsultationDate AS date) AS date,
-                dict.DMD_ID AS dmd_code
+                COALESCE(dict.DMD_ID, custom_dict.DMD_ID) AS dmd_code
             FROM MedicationIssue AS meds
             LEFT JOIN MedicationDictionary AS dict
             ON meds.MultilexDrug_ID = dict.MultilexDrug_ID
+            LEFT JOIN {temp_database_name}..CustomMedicationDictionary AS custom_dict
+            ON meds.MultilexDrug_ID = custom_dict.MultilexDrug_ID
         """
-    )
 
     addresses = QueryTable(
         """
--- a/tests/integration/backends/test_tpp.py
+++ b/tests/integration/backends/test_tpp.py
@@ -14,6 +14,7 @@ from tests.lib.tpp_schema import (
     Appointment,
     CodedEvent,
     CodedEvent_SNOMED,
+    CustomMedicationDictionary,
     EC_Cost,
     EC_Diagnosis,
     HealthCareWorker,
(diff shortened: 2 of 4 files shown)


tests/integration/backends/test_tpp.py
```diff
@@ -278,10 +281,32 @@ def test_clinical_events(select_all):
 
 
 @register_test_for(tpp.medications)
-def test_medications(select_all):
+@pytest.mark.parametrize(
+    "dictionary_class", [MedicationDictionary, CustomMedicationDictionary]
+)
```
evansd, line 286: This is purely a matter of stylistic preference but, to my eyes, using parameterisation here adds an unnecessary layer of abstraction. I think it would be clearer to just add two cases (one with a MedicationDictionary entry and one with a CustomMedicationDictionary entry) and confirm they're both as expected.

In fact, you could then collapse the other separate test into this as well as a third case.

evansd (approved):
Looks good!

I've made a few suggestions around clarifying some things, but nothing fundamental.

ehrql/backends/tpp.py
```diff
@@ -101,17 +101,20 @@ class TPPBackend(BaseBackend):
         """
     )
 
-    medications = QueryTable(
-        """
+    @QueryTable.from_function
+    def medications(self):
+        temp_database_name = self.config.get("TEMP_DATABASE_NAME")
```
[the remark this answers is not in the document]
iaindillingham: Yes, that would be confusing. I've set the default to `"PLACEHOLDER_FOR_TEMP_DATABASE_NAME"`, so the string incorporates a sub-string that can be greped.

tests/integration/backends/test_tpp.py
evansd: No, the opposite! Three variations but as examples in a single test.

evansd: Might be easier to show rather than tell ...
```py
@register_test_for(tpp.medications)
def test_medications(select_all):
    results = select_all(
        Patient(Patient_ID=1),
        # Just MedicationDictionary entry
        MedicationDictionary(MultilexDrug_ID="abc", DMD_ID="xyz"),
        MedicationIssue(
            Patient_ID=1, ConsultationDate="2020-05-15T10:10:10", MultilexDrug_ID="abc"
        ),
        # Just CustomMedicationDictionary entry
        CustomMedicationDictionary(MultilexDrug_ID="def", DMD_ID="uvw"),
        MedicationIssue(
            Patient_ID=1, ConsultationDate="2020-05-16T10:10:10", MultilexDrug_ID="def"
        ),
        # Both entries present: MedicationDictionary should be preferred
        MedicationDictionary(MultilexDrug_ID="hij", DMD_ID="rst"),
        CustomMedicationDictionary(MultilexDrug_ID="hij", DMD_ID="opq"),
        MedicationIssue(
            Patient_ID=1, ConsultationDate="2020-05-17T10:10:10", MultilexDrug_ID="hij"
        ),
    )
    assert results == [
        {
            "patient_id": 1,
            "date": date(2020, 5, 15),
            "dmd_code": "xyz",
        },
        {
            "patient_id": 1,
            "date": date(2020, 5, 16),
            "dmd_code": "uvw",
        },
        {
            "patient_id": 1,
            "date": date(2020, 5, 17),
            "dmd_code": "rst",
        },
    ]
```

iaindillingham: I see, thanks! I think we've traded a bit less abstraction for a bit more code; but a bit more code is okay, in tests.

iaindillingham:
Thanks for your comments, @evansd 👍🏻

PR #1364, opened 2023-06-30
Replace string operators with functions

iaindillingham:
If we replace string operators, such as `"+"`, with functions, such as `operator.add`, then we don't need long and error-prone `if`/`elif`/`else` blocks. I'm sure there's a word for this - or maybe it's a case of DRY?

### DIFF
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -1,3 +1,4 @@
+import operator
 from datetime import date
 from inspect import signature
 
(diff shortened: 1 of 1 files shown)


evansd (approved):
Lovely, thanks!

I'm not sure of the word either, but I'd describe it as a more functional style i.e. passing functions which do the thing you want, rather than passing values which tell some other thing what you want to do.

PR #1365, opened 2023-06-30
Add language reference docs for ehrQL

### LINKED ISSUE #1046: Include a full ehrQL API reference
Moved from opensafely/documentation#906.

evansd:
This PR ensures that everything in the public `ehrQL` namespace has basic reference documentation.

Because I wanted to be able to enforce the presence of docstrings on all public functions, classes etc I have done a single pass over them and added _some_ kind of content for all them. I don't pretend this is particularly good content. I don't think it's necessarily consistent with the terminology we use elsewhere in the docs, and I'm not even sure it's self consistent. But I do think it's strictly better than nothing. I'd happily accept fixes to it on this PR. But it might be easier to merge as is (or with obvious typo fixes) and then to treat making the docs more consistent and expansive as a separate ticket.

One aspect of the docstrings which feels particularly clunky is trying to explain the NULL behaviour of the various operations in prose, rather than a few nicely formatted examples. I'll write up a separate proposal on how we might better integrate the spec tests with the reference docs.

I'd also like to auto-generate docs for the various CLI commands and their options too, but I'll address this separately.

Closes #1046


inglesp (approved):
This is really good to see.

I'm going to hold off reviewing the text in detail until this is integrated with examples from spec tests.

One small thing: the "boolean patient series" link is broken.

StevenMaude:
This relates to, or maybe entirely fixes #1046.

It *does* look really good on preview :star_struck:

StevenMaude (approved):
> One small thing: the "boolean patient series" link is broken.

Also for "integer patient series"; the links are lower case `#boolpatientseries` but the fragment IDs actually follow the class name `#BoolPatientSeries`.

***
I read through the entire output page on preview and it looks very good for a first attempt. Overall, it's a great improvement, so I'd be for merging this sooner.

(I very quickly skimmed the code to get a very rough feel for how it worked, but haven't reviewed it in detail.)

PR #1366, opened 2023-06-30
Add helper for querying `cause_of_death` fields

### LINKED ISSUE #1363: Add helper method for querying multiple `cause_of_death` fields
The `ons_deaths` table records multiple causes of death in separate, numbered fields (apparently 15 is the most number of things you can die of):
https://github.com/opensafely-core/ehrql/blob/27f548e744abd259fae4bb2d36070cd988b3c6a1/ehrql/tables/beta/core.py#L70-L86

At some stage we'd like to have proper support for [multi-valued columns](https://github.com/opensafely-core/ehrql/issues/574) in ehrQL, but until we do it would be helpful to provide a method for querying across all these fields in one go. I'm imagining something like:
```py
import operator
from functools import reduce

...

def any_cause_of_death_is_in(self, codelist):
    conditions = [
        getattr(self, f"cause_of_death_{i:02d}").is_in(codelist)
        for i in range(1, 16)
    ]
    return reduce(operator.or_, conditions)
```

To test this we should first move the misplaced `test_patients_age_on` function from [`tests/integration/tables/beta/test_core.py`](https://github.com/opensafely-core/ehrql/blob/27f548e744abd259fae4bb2d36070cd988b3c6a1/tests/integration/tables/beta/test_tpp.py#L7-L29) to `tests/integration/tables/beta/test_core.py` and replace references to `tpp.patients` with `core.patients`.

Then we can add a `test_ons_deaths_any_cause_of_death_is_in` function along similar lines.

iaindillingham:
Closes #1363

### DIFF
--- a/tests/integration/tables/beta/test_core.py
+++ b/tests/integration/tables/beta/test_core.py
@@ -1,0 +1,29 @@
+from datetime import date
+
+from ehrql import Dataset
+from ehrql.tables.beta import core
+
+
+def test_patients_age_on(in_memory_engine):
+    in_memory_engine.populate(
+        {
+            core.patients: [
+                dict(patient_id=1, date_of_birth=date(1980, 1, 1)),
+                dict(patient_id=2, date_of_birth=date(1800, 1, 1)),
+                dict(patient_id=3, date_of_birth=date(2010, 1, 1)),
+                dict(patient_id=4, date_of_birth=date(2010, 1, 2)),
+            ]
+        }
+    )
+
+    dataset = Dataset()
+    dataset.define_population(core.patients.exists_for_patient())
+    dataset.age_2010 = core.patients.age_on("2010-01-01")
+    results = in_memory_engine.extract(dataset)
+
+    assert results == [
+        {"patient_id": 1, "age_2010": 30},
+        {"patient_id": 2, "age_2010": 210},
+        {"patient_id": 3, "age_2010": 0},
+        {"patient_id": 4, "age_2010": -1},
+    ]
--- a/tests/integration/tables/beta/test_core.py
+++ b/tests/integration/tables/beta/test_core.py
@@ -27,3 +27,31 @@ def test_patients_age_on(in_memory_engine):
         {"patient_id": 3, "age_2010": 0},
         {"patient_id": 4, "age_2010": -1},
     ]
+
+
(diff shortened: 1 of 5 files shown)


tests/integration/tables/beta/test_core.py
```diff
@@ -0,0 +1,57 @@
...
+    dataset.cause_of_death_is_in_codelist = (
+        core.ons_deaths.any_cause_of_death_is_in(["A000"]).count_distinct_for_patient()
+        > 0
+    )
+
+    results = in_memory_engine.extract(dataset)
+
+    assert results == [
+        {"patient_id": 1, "cause_of_death_is_in_codelist": True},
+        {"patient_id": 2, "cause_of_death_is_in_codelist": False},
```
iaindillingham, line 55: This should cause the test to fail.

evansd: The call to `any_cause_of_death_is_in()` is giving you a boolean event series like this:
```py
{
    1: [True],
    2: [False],
    3: [True],
}
```

And then you're counting the number of distinct values in each series which gives you:
```py
{
    1: 1,
    2: 1,
    3: 1,
}
```

And then you're asking which of those is greater than zero, which gives you:
```py
{
    1: True,
    2: True,
    3: True,
}
```

There's no straightforward way to write the test you want to write without sticking that boolean event series back into a `where()` clause and then using `exists_for_patient()` on the whole frame.

I think this is a deficiency in ehrQL, as you've mentioned elsewhere.

iaindillingham: Thanks @evansd. I'm probably missing something obvious here, but as you say, `any_cause_of_death_is_in()` gives

```py
{
    1: [True],
    2: [False],
    3: [True],
}
```

and `count_distinct_for_patient()` gives

```py
{
    1: 1,
    2: 1,
    3: 1,
}
```

and `> 0` gives

```py
{
    1: True,
    2: True,
    3: True,
}
```

but the test passes with

```py
{
    1: True,
    2: False,  # <- ?
    3: True,
}
```

Returning to #1363, you mention that it would be good to move `test_patients_age_on`, which is covered by 08ad174 and f43aed8. I'll open a separate PR for these commits, and close this PR and the associated issue.

iaindillingham: Not at all, my comment was terse; it's helpful to have the steps repeated back 🙂

PR #1369, opened 2023-07-03
Improve table metadata

iaindillingham:
A couple of improvements to `appointments` and `open_prompt` table metadata, following [this Slack thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1688133357818489).

In both cases, I've documented what a user needs to do to access "the report" (which, in the case of the `open_prompt` table, is currently a set of results); where they can find information about when the report was run; and where they can find the code that comprises the report.

* [Preview of `appointments`](https://ae8e7a2d.databuilder.pages.dev/reference/schemas/beta.tpp/#appointments)
* [Preview of `open_prompt`](https://ae8e7a2d.databuilder.pages.dev/reference/schemas/beta.tpp/#open_prompt)

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -2120,8 +2120,15 @@ TODO
 
 This table contains responses to questions from the OpenPROMPT project.
 
-You can find out more about this table in the associated short data report:
-<https://github.com/opensafely/airmid-short-data-report>.
+You can find out more about this table in the associated short data report. To view
+it, you will need a login for [Level 4][open_prompt_1]. The
+[workspace][open_prompt_2] shows when the code that comprises the report was run;
+the code itself is in the [airmid-short-data-report][open_prompt_3] repository on
+GitHub.
+
+[open_prompt_1]: https://docs.opensafely.org/security-levels/#level-4-nhs-england-are-data-controllers-of-the-data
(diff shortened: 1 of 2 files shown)


evansd (approved):
Great, very happy with the text here.

The only annoying wrinkle is that you can't use numeric URL references in the markdown here because they are shared across the whole page, so they need to be unique.

If you look at the preview here you'll see all the links are pointing to the wrong places:
https://iaindillingham-improve-table.databuilder.pages.dev/reference/schemas/beta.tpp/#appointments

You have to either come up with short names that you can still be reasonably confident will be unique e.g.
```
The [workspace][appts_tbl_1] shows when the code that comprises the report was
run; the code itself is in the
[appointments-short-data-report][appts_tbl_3] repository on GitHub.

[appts_tbl_1]: https://jobs.opensafely.org/datalab/opensafely-internal/appointments-short-data-report/outputs/latest/tpp/output/reports/report.html
[appts_tbl_2]: https://jobs.opensafely.org/datalab/opensafely-internal/appointments-short-data-report/
[appts_tbl_3]: https://github.com/opensafely/appointments-short-data-report
```

Or embed the full URLs in the text itself.

iaindillingham:
Good spot, @evansd, thank you. I've ironed out the wrinkle by using better names for the link labels -- I've opted for `<table name>_x` -- because I find link labels easier on the eye than inline links.

PR #1370, opened 2023-07-03
fix: Drop `is_between()` date series method

### LINKED ISSUE #1188: Drop `is_between()` DateSeries method
The `is_between()` method is the exclusive complement to the inclusive `is_on_or_between()` and was added (I think) just for the sake of consistency because the other inclusive methods (`is_on_or_before()` and `is_on_or_after()`) have exclusive complements.

But as @wjchulme has pointed out, the potential for confusion here is quite high as it would be very easy for a casual reader to assume that it was inclusive. And on top of that, an exclusive date range function doesn't seem all that useful.

Slack thread here:
https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1680707203503939

milanwiedemann:
Drop `is_between()` as described in #1188. For acceptance test studies in this repo suggest an identical alternative using `is_on_or_between()` together with adding and subtracting one day to/from the interval. 

Closes #1188

### DIFF
--- a/tests/acceptance/external_studies/ons-mental-health/analysis/dataset_definition_ons_cis_new.py
+++ b/tests/acceptance/external_studies/ons-mental-health/analysis/dataset_definition_ons_cis_new.py
@@ -1,4 +1,4 @@
-from databuilder.ehrql import Dataset
+from databuilder.ehrql import Dataset, days
 from databuilder.tables.beta import tpp as schema
 
 
--- a/tests/acceptance/external_studies/openprompt-long-covid-vaccines/analysis/dataset_definition_longcovid_prevaccine.py
+++ b/tests/acceptance/external_studies/openprompt-long-covid-vaccines/analysis/dataset_definition_longcovid_prevaccine.py
@@ -115,16 +115,16 @@ dataset.define_population(population)
 # covid tests ------------------------------------------------------------
 dataset.first_test_post_lc = sgss_covid_all_tests \
     .where(sgss_covid_all_tests.is_positive) \
-    .where(sgss_covid_all_tests.specimen_taken_date.is_between(dataset.pt_start_date, dataset.pt_end_date)) \
+    .where(sgss_covid_all_tests.specimen_taken_date.is_on_or_between(dataset.pt_start_date + days(1), dataset.pt_end_date - days(1))) \
     .sort_by(sgss_covid_all_tests.specimen_taken_date).first_for_patient().specimen_taken_date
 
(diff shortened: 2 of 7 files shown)


tests/acceptance/external_studies/openprompt-long-covid-vaccines/analysis/dataset_definition_longcovid_prevaccine.py
```diff
@@ -115,16 +115,16 @@
 # covid tests ------------------------------------------------------------
 dataset.first_test_post_lc = sgss_covid_all_tests \
     .where(sgss_covid_all_tests.is_positive) \
-    .where(sgss_covid_all_tests.specimen_taken_date.is_between(dataset.pt_start_date, dataset.pt_end_date)) \
```
evansd, line 118: This is a correct translation of the semantics of the original query, but I wonder if it's what the author actually intended? I think we'll need to confirm with them in each case.

evansd (commented):
As mentioned in [this comment](https://github.com/opensafely-core/ehrql/issues/1188#issuecomment-1618101179) I think we'll need to check the intention with the study authors of the few uses of `is_between()` that we want to remove.

tests/acceptance/external_studies/ons-mental-health/analysis/dataset_definition_ons_cis_new.py
```diff
@@ -19,7 +19,7 @@
 # range
 #######################################################################################
 
-ons_cis = schema.ons_cis.where(schema.ons_cis.visit_date.is_on_or_between(start_date, end_date))
+ons_cis = schema.ons_cis.where(schema.ons_cis.visit_date.is_on_or_between(start_date + days(1), end_date - days(1)))
```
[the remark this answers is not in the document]
milanwiedemann: ahhh it doesnt! thanks for spotting

tests/acceptance/external_studies/openprompt-long-covid-vaccines/analysis/dataset_definition_longcovid_prevaccine.py
milanwiedemann: I checked with the study author, they did intend to use the `is_between()` as it was implemented, i.e., excluding the start and end date

milanwiedemann:
we decided to keep this functionality because it was the intended behaviour in a research study. but we will rename it to `is_between_but_not_on()` to avoid confusion and follow the naming convention of similar methods, work continues here #1405

PR #1386, opened 2023-07-03
Fix `medications` error

iaindillingham:


Jongmassey (commented):
Should we be guarding against there being a row for a given multilex id in both the standard and custom dictionaries generating duplicate rows here?

Jongmassey (approved):
Assuming this scenario has been encountered in the real data then this seems like the correct fix

PR #1392, opened 2023-07-04
Add some fixes to documentation

LFISHER7:
This adds various fixes as I've been working through the docs. The biggest changes are the addition of instructions for using the category column with `codelist_from_csv()` and an explainer on extracting ethnicity, making use of the category column.

### DIFF
--- a/docs/explanation/examples.md
+++ b/docs/explanation/examples.md
@@ -13,7 +13,7 @@ and then jump to a specific example of interest.
 and to make it possible to use [backend-specific tables in addition to core tables](backend-tables.md),
 the example dataset definitions here all use the TPP backend.
 
-### Some examples use `codelist_from_csv()`
+### Some examples using `codelist_from_csv()`
 
 :construction: This guidance should be improved in future.
 
@@ -39,10 +39,15 @@ asthma_codelist = codelist_from_csv("your-asthma-codelist.csv", column="code")
 which provides the filename `your-asthma-codelist.csv`
 and the name of the CSV column with codes.
 
-:notepad_spiral: It is also possible to use a category column
-with `codelist_from_csv()`.
-Until this documentation is improved,
-refer to the [ehrQL source code](https://github.com/opensafely-core/ehrql/blob/main/ehrql/codes.py).
+Some codelists will have a category column that groups individual codes into categories. For example, [this codelist](https://www.opencodelists.org/codelist/opensafely/ethnicity-snomed-0removed/2e641f61/) for ethnicity has 2 category columns, which represent categories at both 6 and 16 levels.To make use of these categories you can do this:
+
+```python
+ethnicity_codelist = codelist_from_csv("ethnicity_codelist_with_categories", column="snomedcode", category_column="Grouping_6")
+```
+
+If you include an argument for `category_column`, the codelist returned will be a dictionary mapping individual codes to their respective categories, instead of simply a list of codes, which is returned if you don't specify this argument.
+
+You can see an example of how to access these categories within your dataset definition [below](#finding-each-patients-ethnicity).
 
 ## Finding patient demographics
 
@@ -150,7 +155,34 @@ The possible values are "female", "male", "intersex", and "unknown".
 
 ### Finding each patient's ethnicity
 
-:construction: This is more complicated and will be added in future.
+:construction: Ethnicity is incompletely coded within primary care. More detail will be added in the future on how to improve ethnicity capture by incorporating data from secondary care.
+
+Ethnicity can be defined using a codelist. There are a lot of individual codes that can used to indicate a patients' fine-grained ethnicity. To make analysis more manageable, ethnicity is therefore commonly grouped into higher level categories. Above, we described how you can [import codelists that have a category column](#some-examples-using-codelist_from_csv). You can use a codelist like this to map clinical event codes for ethnicity to higher level categories like this:
+
+```python
+from ehrql import Dataset
+from ehrql.tables.beta.core import clinical_events
+from ehrql.codes import codelist_from_csv
+
+dataset = Dataset()
+
+ethnicity_codelist = codelist_from_csv(
+    "ethnicity_codelist_with_categories",
+    column="snomedcode",
+    category_column="Grouping_6",
+)
+
+dataset.latest_ethnicity_code = (
+    clinical_events.where(clinical_events.snomedct_code.is_in(ethnicity_codelist))
+    .where(clinical_events.date.is_on_or_before("2023-01-01"))
+    .sort_by(clinical_events.date)
+    .last_for_patient()
+    .snomedct_code
+)
+latest_ethnicity_group = latest_ethnicity_code.to_category(
+    ethnicity_codelist
+)
+```
 
 ## Finding attributes related to each patient's address as of a given date
 
--- a/docs/explanation/examples.md
+++ b/docs/explanation/examples.md
@@ -39,15 +39,17 @@ asthma_codelist = codelist_from_csv("your-asthma-codelist.csv", column="code")
 which provides the filename `your-asthma-codelist.csv`
 and the name of the CSV column with codes.
 
-Some codelists will have a category column that groups individual codes into categories. For example, [this codelist](https://www.opencodelists.org/codelist/opensafely/ethnicity-snomed-0removed/2e641f61/) for ethnicity has 2 category columns, which represent categories at both 6 and 16 levels.To make use of these categories you can do this:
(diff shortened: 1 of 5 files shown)


docs/tutorial/dataset-definition-concepts.md
```diff
@@ -267,12 +271,14 @@ from ehrql.tables.beta.core import patients, medications
 
 Next, we create a dataset object.
 Don't worry about exactly what this line does!
+# I'd like to know what this is here. I don't think it is explained anywhere
+# e.g. to this we can add the patient series (variables) we are interested in. These form a patient frame that can then be filtered to the population of interest.
```
StevenMaude, line 275: nit:

This and the prior sentences — which weren't your work! — need rewriting.

We do actually explain a little of `Dataset()` above now, which may have not been the case at the time those sentences were written.

> A dataset is a table containing one row per patient, and one column per feature of interest…

I agree that we should probably have something add something along the lines of your suggestion. That is, something like the following points that you've made:

* `Dataset()` represents our output dataset. 
* We add patient "features" (what we've termed variables in ehrQL) as columns to it, by assigning to attributes of the dataset, that is, `dataset.age = … — these have one row per patient
* We define our output dataset's population of interest by using the `define_population()` method on the dataset.
* In ehrQL, methods and attributes are both accessed via Python's "dot" notation.

(Markdown uses `#` as a heading, so this will appear in text, unlike many languages where `#` indicates a comment[^1].)

[^1]: Here's a couple of ways you could actually add comments in Markdown: https://stackoverflow.com/questions/4823468/comments-in-markdown

StevenMaude: thought:

> e.g. to this we can add the patient series (variables) we are interested in. These form a patient frame that can then be filtered to the population of interest.

I'm not sure how important this is to clarify, but I'll discuss it anyway. And maybe it needs a clearer distinction on the page.

From my (limited) understanding, there is a distinction between `Dataset` and `PatientFrame`. The internal Python representation of [`Dataset`](https://github.com/opensafely-core/ehrql/blob/7eb6bd3f7f66b00d03832c674cc8cccdb259e4b9/ehrql/query_language.py#L39) *is not* a [`PatientFrame`](https://github.com/opensafely-core/ehrql/blob/7eb6bd3f7f66b00d03832c674cc8cccdb259e4b9/ehrql/query_language.py#L961
), nor related to `PatientFrame` by class inheritance.

So, things you can do with a `PatientFrame`, like  `.exists_for_patient()`, you cannot do with a `Dataset`.

:exclamation: **However, `Dataset` does represent a *similar kind of table* as you'd get in a `PatientFrame`**.

Though, `Dataset` is a one row per patient table that you're constructing for the purpose of generating a final output table only. The only filtering operation you can do is to define the population to include with `define_population()`.

I will tag @evansd to correct me if I'm wrong here though. I feel like it's maybe a point of confusion that we should help users to understand, so I need to be correct in what I'm saying too! :smile:

docs/tutorial/running-ehrql.md
```diff
@@ -58,7 +58,7 @@ For example if you type `1 + 1` and press the return key, you should see:
 
 To use ehrQL, you'll first need to import the tables that you want to interact with:
 
-    >>> from ehrql.tables.beta.core import patients, medications
+    from ehrql.tables.beta.core import patients, medications
```
StevenMaude, line 61: thought:

I'm guessing you've removed the `>>>` prompt to make it easier to copy-paste the code. That makes sense.

But I'm not sure if we're still best to keep the prompt indicators in, to show that this to be done in a Python interactive console.

docs/explanation/examples.md
```diff
@@ -150,7 +155,34 @@ The possible values are "female", "male", "intersex", and "unknown".
 
 ### Finding each patient's ethnicity
 
-:construction: This is more complicated and will be added in future.
+:construction: Ethnicity is incompletely coded within primary care. More detail will be added in the future on how to improve ethnicity capture by incorporating data from secondary care.
+
+Ethnicity can be defined using a codelist. There are a lot of individual codes that can used to indicate a patients' fine-grained ethnicity. To make analysis more manageable, ethnicity is therefore commonly grouped into higher level categories. Above, we described how you can [import codelists that have a category column](#some-examples-using-codelist_from_csv). You can use a codelist like this to map clinical event codes for ethnicity to higher level categories like this:
```
StevenMaude, line 160: nit:

```suggestion
Ethnicity can be defined using a codelist. There are a lot of individual codes that can used to indicate a patients' fine-grained ethnicity. To make analysis more manageable, ethnicity is therefore commonly grouped into higher level categories. Above, we described how you can [import codelists that have a category column](#some-examples-using-codelist_from_csv). You can use a codelist with a category column to map clinical event codes for ethnicity to higher level categories as in this example:
```

docs/explanation/examples.md
```diff
@@ -150,7 +155,34 @@ The possible values are "female", "male", "intersex", and "unknown".
 
 ### Finding each patient's ethnicity
```
StevenMaude, line 156: This section looks great; thanks for filling it out :+1:

docs/tutorial/running-ehrql.md
```diff
@@ -240,6 +240,8 @@ The file will contain the following CSV data:
     8,,
     9,,
 
+:notepad_spiral: The bottom 4 rows in the generated dataset show that there are 4 patients in the provided example data that do not have any record for the medications specified in the dataset definition.
```
StevenMaude, line 243: nit:

I like this clarification, but maybe we can be even clearer:

```suggestion
:notepad_spiral: The bottom 4 rows in the generated dataset show that there are 4 patients in the defined population that do not have any record for the medications specified in the dataset definition.
```

StevenMaude (commented):
@LFISHER7: This is *great*; thanks for your contribution! :smiley:

I think it's the first substantial changes to the ehrQL docs that have come from outside of the Bennett data team.

I'll leave it as unapproved for now, just because I'd like to take a final look over before it's merged in.

docs/explanation/examples.md
```diff
@@ -39,10 +39,15 @@ asthma_codelist = codelist_from_csv("your-asthma-codelist.csv", column="code")
 which provides the filename `your-asthma-codelist.csv`
 and the name of the CSV column with codes.
 
-:notepad_spiral: It is also possible to use a category column
-with `codelist_from_csv()`.
-Until this documentation is improved,
-refer to the [ehrQL source code](https://github.com/opensafely-core/ehrql/blob/main/ehrql/codes.py).
+Some codelists will have a category column that groups individual codes into categories. For example, [this codelist](https://www.opencodelists.org/codelist/opensafely/ethnicity-snomed-0removed/2e641f61/) for ethnicity has 2 category columns, which represent categories at both 6 and 16 levels.To make use of these categories you can do this:
```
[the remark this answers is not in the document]
LFISHER7: Agree that having this on a separate page would be good!

docs/tutorial/running-ehrql.md
LFISHER7: That's fine! Having that context as a [code block title](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-a-title) would be nice!

docs/tutorial/dataset-definition-concepts.md
LFISHER7: Thanks, I didn't know about the comment feature!

I came across [this reference](https://docs.opensafely.org/ehrql/reference/language/#Dataset) for `Dataset`, that I missed and think would have been enough of an explanation. Perhaps just a link to that for now?

LFISHER7 (commented):
Thanks @StevenMaude! I've pulled in all of your suggestions. I'll just remove my incorrectly added comments

docs/tutorial/dataset-definition-concepts.md
StevenMaude: Yes, that reference basically explains what we've said here. A link to that would be sufficient, I agree :+1:

docs/tutorial/running-ehrql.md
StevenMaude: Ah, that's actually a use of that titling I hadn't considered!

I'll look into how we might use that.

StevenMaude (approved):
@LFISHER7: All that's left I think are:

* re-add the `>>>` prompt back in, where it was removed
* add the link to the `Dataset` in the ehrQL reference

But I'll approve now and leave up to you to merge when ready.

Thanks again for this; really nice to see helpful contributions based on real use!

PR #1393, opened 2023-07-05
Expose additional `ONS_CIS_New` columns

### LINKED ISSUE #1358: Column exposure ONS_CIS_New
Currently, only seven columns of the `ONS_CIS_New` table are exposed in ehrql. 
For [approved project 117](https://www.opensafely.org/approved-projects/#project-117), we would like to have more columns available. 

Please could all columns indicated in [this spreadsheet](https://docs.google.com/spreadsheets/d/1lCcTGpcz9gUI7Lewv06WInoO-c2AL70Z/edit#gid=828721221) be added? 
The requested columns are indicated with 1's in the Want? column.

iaindillingham:
This adds the columns requested in #1358, with the exception of `pseudo_visit_id`, which is a binary column. I've checked that columns of type `bool` have values of `0`, `1`, or `NULL` in the TPP backend.

I've used information from the codebook to map from database column names to ehrQL column names. To see the mappings side-by-side, see _ehrql/backends/tpp.py_.

It might be easier to parse column names, descriptions, and types from [the docs preview](https://439e1fb4.databuilder.pages.dev/reference/schemas/beta.tpp/#ons_cis).

Closes #1358


LindaNab:
Thanks @iaindillingham, this looks great! Some comments:
- wrt column names and descriptions, seeing it in one document now makes me realise it be good to add suffices like `_nhs` and `_cis` to make a clear distinction between the swab results from nhs (based on recall) vs cis. We don't want people to use the patient's recall data as if it was the 'official' cis result. My suggestion:
covid_was_admitted --> covid_was_admitted_nhs
covid_was_swab_tested --> covid_was_swab_tested_nhs
covid_date_last_negative_swab_test --> covid_date_last_negative_swab_test_nhs
covid_date_first_positive_swab_test -->  covid_date_first_positive_swab_test_nhs
covid_swab_test_result --> covid_swab_test_result_nhs
swab_test_result --> swab_test_result_cis (or swab_test_cis_result)
swab_test_date --> swab_test_date_cis (or swab_test_cis_date)
ct_sgene_result --> ct_sgene_result_cis (or ct_sgene_cis_result)
Let me know what you think.
- Sorry for ignorance! Please could you explain why `pseudo_visit_id` has not been added? I'm not sure if we need it but wrt duplicate rows we might need it as an identifier? 
- Would it be useful to check for variable `work_in_patient_facing_healthcare_status` if it could be mapped to a `boolean` as well? I'm thinking that the 'Not working in healthcare' category (2) could be `NA` (for all those of whom `is_work_in_healthcare_status` is `FALSE`).

iaindillingham:
Thanks for your comments, @LindaNab. I've just pushed a commit that adds categorical values to all categorical columns, with the exception of `work_in_patient_facing_healthcare_status`. I shall convert this to a Boolean, with values of `2` or `Not working in healthcare` mapped to `None`.

Sorry for not explaining why I haven't yet added `pseudo_visit_id`. It's a binary column, and I haven't figured out how to convert it to something more useful (i.e. not `bytes`). When I have, I'll add it!

I will add the suffixes, as you suggest. Is it the case that all columns are either NHS or CIS columns? And does the original `covid_` prefix (i.e. on the database columns, as you have them in the codebook) indicate NHS? I'd like to be extra sure that I've not unintentionally changed the meanings of the columns.

LindaNab:
Thanks 👍 
I think adding the suffices should do the trick. The covid variables are not necessarily nhs variables, but all are based on questions from a questionnaire.
For context, all ONS CIS (covid infection survey) participants were sent a questionnaire (first every week, after 4 weeks every month) along with a nose/throat swab thing to be sent off to be PCR tested for SARS-CoV-2. Some of the participants might already have had symptoms and went for regular NHS testing between their checks for CIS. 

For variable `covid_swab_test_result`, could we change `Any tests negative, but none positive` to `All tests negative`?

LindaNab:
Oh and wrt `pseudo_visit_id`, not sure if you can check this somehow, but if the combination (patient_id and visit_date) is a unique identifier, don't think we need it. If (patient_id and visit_date) is not a unique identifier, we could use `pseudo_visit_id` as an index? Not as a binary var but 1, 2, 3 etc so that (patient_id, visit_date, pseudo_visit_id) is a unique identifier. 
If it is very rare that (patient_id and visit_date) is not a unique identifier (which I would expect), we might want to collate the rows somehow!

iaindillingham:
I'm going to close this PR in favour of #1396. For the rationale, see #1358.

PR #1396, opened 2023-07-06
feat: Add the `ons_cis_raw` table

iaindillingham:
`ons_cis_raw` contains the columns requested by #1358, with the exception of the `pseudo_visit_id` column (see below), as well as the columns of the `ons_cis` table.

I've taken a conservative approach to mapping from the database table columns (`ONS_CIS_New`) to the ehrQL table (`ons_cis_raw`) columns: I've lower-cased the former to give the latter.

I'm adding `ons_cis_raw` because we need to do more exploration of the raw data before modifying `ons_cis` (e.g. checking the values of `visit_num` are drawn from the interval `[0-15]`), but it's hard to do more exploration without exposing the raw data through ehrQL.

`pseudo_visit_id` is not included, because `ONS_CIS_New.pseudo_visit_id` is of type `VARBINARY` and it's not clear how to expose that type through ehrQL.

### DIFF
--- a/ehrql/tables/beta/tpp.py
+++ b/ehrql/tables/beta/tpp.py
@@ -23,6 +23,7 @@ __all__ = [
     "emergency_care_attendances",
     "hospital_admissions",
     "appointments",
+    "ons_cis_raw",
     "ons_cis",
     "isaric_raw",
     "open_prompt",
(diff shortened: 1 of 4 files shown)


ehrql/tables/beta/tpp.py
```diff
@@ -237,6 +238,40 @@ class household_memberships_2020(PatientFrame):
     household_size = Series(int)
 
 
+@table
+class ons_cis_raw(EventFrame):
+    """
+    Data from the ONS Covid Infection Survey.
```
evansd, line 244: Is it worth clarifying how this table differs from `ons_cis` below? As it stands they have exactly the same description.

evansd (approved):

ehrql/tables/beta/tpp.py
iaindillingham: Yes, it is 🙂 I've clarified and force-pushed.

PR #1397, opened 2023-07-07
Better dummy data for imd_rounded

### LINKED ISSUE #1162: Generate more realistic dummy values for `addresses.imd_rounded`
At the moment we just pick values between 0 and 100 for integer columns:
https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/dummy_data/generator.py#L213-L215

This is confusing for the [`addresses.imd_rounded`](https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/tables/beta/tpp.py#L140) column because the [actual values](https://docs.opensafely.org/data-sources/systmone/#imd) lie between 0 and 32,800 but are always multiples of 100. See discussion here:
https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1679496923349789

## Add "range" constraint

We should add a new "range" constraint (`Constraint.Range`) to the list of [available constraints](https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/query_model/table_schema.py#L12-L42) and apply it to the [`addresses.imd_rounded`](https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/tables/beta/tpp.py#L140) column.

This can be modelled on the Python `range()` object which has a lower bound, upper bound and step size.

By defining an appropriate `description()` method details of this constraint should automatically appear in the documentation for the column.

Something like (this is just a rough sketch which I haven't spent long thinking about):
```py
class Range(BaseConstraint):
    minimum: int
    maximum: int
    step: int = 1

    @property
    def description(self):
        return f"Must be >= {self.minimum}, <= {self.maximum}, and a multiple of {self.step}"
```

## Use range constraint to generate dummy data

We can use this constraint to generate appropriate dummy data. For example, here's how we use `Constraint.Regex` when generating dummy data for string columns:
https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/dummy_data/generator.py#L219-L230

For integer columns the implementation could be something like (again, just a rough sketch):
```py
if range_constraint := column_info.get_constraint(Constraint.Range): 
    return self.rnd.randrange(
        range_constraint.minimum,
        range_constraint.maximum + 1,
        range_constraint.step,
    )
```

## Respect range constraint in "column specs"

For bonus points we can teach the `column_specs` module to respect this constraint as well. This would allow us to select the most appropriate integer type when outputting results in a binary format.

We already have the basic scaffolding in place for this, but we currently only define a range for the `Count` aggregation:
https://github.com/opensafely-core/databuilder/blob/a55f42aec2b8e978812d6030f3c7a1aec2386366/databuilder/query_model/column_specs.py#L102-L114

We could expand this to handle `SelectColumn` using something like (rough sketch alert):
```py
@get_range.register(SelectColumn)
def get_range_for_select_column(series):
    root = get_root_frame(series.source)
    range_constraint = root.schema.get_column_constraint_by_type(series.name, Constraint.Range)
    if range_constraint:
        return range_constraint.minimum, range_constraint.maximum
    return None, None
```

rebkwok:
Adds a range constraint, applies it to the imd_rounded columns, and uses it in dummy data generation.
Fixes #1162

### DIFF
--- a/ehrql/query_model/table_schema.py
+++ b/ehrql/query_model/table_schema.py
@@ -41,7 +41,7 @@ class Constraint:
         def description(self):
             return f"Matches regular expression: `{self.regex}`"
 
-    class Range(BaseConstraint):
+    class ClosedRange(BaseConstraint):
         minimum: int
         maximum: int
         step: int = 1
--- a/ehrql/query_model/table_schema.py
+++ b/ehrql/query_model/table_schema.py
@@ -41,6 +41,15 @@ class Constraint:
         def description(self):
             return f"Matches regular expression: `{self.regex}`"
 
+    class Range(BaseConstraint):
+        minimum: int
+        maximum: int
+        step: int = 1
+
+        @property
+        def description(self):
+            return f"Must be >= {self.minimum}, <= {self.maximum}, and a multiple of {self.step}"
+
 
 @dataclasses.dataclass(frozen=True)
 class Column:
--- a/ehrql/query_model/table_schema.py
+++ b/ehrql/query_model/table_schema.py
@@ -48,7 +48,9 @@ class Constraint:
 
         @property
         def description(self):
-            return f"Must be >= {self.minimum}, <= {self.maximum}, and a multiple of {self.step}"
+            if self.step == 1:
+                return f"Always >= {self.minimum} and <= {self.maximum}"
+            return f"Always >= {self.minimum}, <= {self.maximum}, and a multiple of {self.step}"
 
 
 @dataclasses.dataclass(frozen=True)
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -107,6 +107,7 @@ from ehrql.tables.beta.tpp import (
   <dd markdown="block">
 
 
+ * Must be >= 0, <= 32800, and a multiple of 100
   </dd>
 </div>
 
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -107,7 +107,7 @@ from ehrql.tables.beta.tpp import (
   <dd markdown="block">
 
 
- * Must be >= 0, <= 32800, and a multiple of 100
+ * Always >= 0, <= 32800, and a multiple of 100
   </dd>
 </div>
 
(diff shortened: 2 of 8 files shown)


iaindillingham (approved):
Here, `Range` is a closed interval; it includes its endpoints (e.g. `[0, 1]`). Python's `range` is a half-open interval; it includes one endpoint but not the other (e.g. `[0, 1)`).

The nature of the range is described, and the description will appear in the documentation, so a user shouldn't be surprised - they need never know about the `Range` class. However, a developer might, as the class is in one location and the implementation is in another location. To mitigate the potential surprise, how would you feel about calling the class `ClosedRange`? I don't feel strongly about it, but thought I'd ask the question.

ehrql/query_model/table_schema.py
```diff
@@ -41,6 +41,15 @@ def __post_init__(self):
...
             return f"Matches regular expression: `{self.regex}`"
 
+    class Range(BaseConstraint):
+        minimum: int
+        maximum: int
+        step: int = 1
+
+        @property
+        def description(self):
+            return f"Must be >= {self.minimum}, <= {self.maximum}, and a multiple of {self.step}"
```
evansd, line 51: I think the description moods have changed since I first wrote that ticket (see the other constraint descriptions) so for consistency this should use something like "Always" rather than "Must be".

I also wonder if we should get rid of the "multiple of" bit if `step` is 1 as that's going to look a bit weird.

If you do end up adding logic to the `description` property then you might find coverage moaning at you, in which case there are already some unit tests for the other constraints which you could add to:
https://github.com/opensafely-core/ehrql/blob/94fea127a67d824c438bcaecf7927834b40d8c4e/tests/unit/query_model/test_table_schema.py#L105-L116

evansd (approved):
:+1: Glad to see this getting fixed. Thanks.

A bit of bikeshedding over the description but nothing blocking.

rebkwok:
> Here, `Range` is a closed interval; it includes its endpoints (e.g. `[0, 1]`). Python's `range` is a half-open interval; it includes one endpoint but not the other (e.g. `[0, 1)`).
> 
> The nature of the range is described, and the description will appear in the documentation, so a user shouldn't be surprised - they need never know about the `Range` class. However, a developer might, as the class is in one location and the implementation is in another location. To mitigate the potential surprise, how would you feel about calling the class `ClosedRange`? I don't feel strongly about it, but thought I'd ask the question.

@evansd Do you have an opinion about this? I think I agree with Iain, it could be confusing that it's called Range, uses python's range for implementation, but implements a slightly different version of range than the python one.

evansd:
Thanks Iain. Yeah, I worried a bit about the difference between this and Python `range` when I was speccing it out, but I thought using a half-open interval would be too confusing for non-Python people. It hadn't occurred to me to just give it a different name; that's a good suggestion.

PR #1398, opened 2023-07-07
fix: Use sqlean package instead of system sqlite3

rebkwok:
This avoids the issue of some versions of sqlite missing required extensions and ensures all environments should be using the same version of sqlite bundled with sqlean.

Currently only the math extension is required and enabled.


ehrql/query_engines/sqlite_dialect.py
```diff
@@ -1,9 +1,16 @@
...
 
 
 class SQLiteDialect(SQLiteDialect_pysqlite):
     supports_statement_cache = True
 
+    @classmethod
+    def import_dbapi(cls):
+        # Use sqlean rather than the system version
+        sqlean.extensions.enable("math")
+        return sqlean
```
evansd, line 12: I don't pretend to know exactly what difference this makes, but the default SQLAlchemy dialect returns the `sqlite3.dbapi2` object, rather than `sqlite3`:
https://github.com/sqlalchemy/sqlalchemy/blob/400aa8a676ba1a0a1536ae52a20caa93726525dd/lib/sqlalchemy/dialects/sqlite/pysqlite.py#L503-L507

Obviously the tests all pass here so I guess what you've got works fine. But I wonder if it might avoid possible future weirdness just to copy example what SQLAlchemy does and use:
```
return sqlean.dbapi2
```

evansd (approved):
Great, thanks for doing this!

ehrql/query_engines/sqlite_dialect.py
rebkwok: :+1: that seems safest

PR #1399, opened 2023-07-07
Drive by docs tweaks

inglesp:


docs/tutorial/running-ehrql.md
```diff
@@ -324,15 +324,14 @@ to skip setting up the template,
 we create this file entirely by hand.
 
 The `project.yaml` file defines two actions: `generate_dataset` and `summarise_dataset`.
-Each of these actions defines an `output`,
-which has the potential data sensitivity indicated.
+Each of these actions specifies one or more `outputs`.
```
iaindillingham, line 327: If we're monospacing this because we're referring to the things that live in the directory, then the directory is called `output` in the [research-template](https://github.com/opensafely/research-template).

iaindillingham (approved):

Issue #1360, opened 2023-06-29
Avoid locking `sys.tables` when using `SELECT ... INTO`


iaindillingham:
This issue mirrors opensafely-core/cohort-extractor#965.

ehrQL writes results to a table in the temporary database (`OpenCORONATempTables`) using `SELECT ... INTO`

https://github.com/opensafely-core/ehrql/blob/7115972c780e79c57441e7e006a2e708c1770d6b/ehrql/query_engines/mssql.py#L230-L249

However, [`SELECT ... INTO` can lock `sys.tables` for long periods of time][1]; in effect, ehrQL jobs run in sequence.[^1]

To avoid locking `sys.tables` for long periods of time, we could first create the table in the temporary database and then, in a separate transaction, populate it. We would, however, need to be able to determine its schema.

@evansd and @Jongmassey discuss this issue in [this Slack thread][2].

[^1]: Indeed, any jobs that modify the database run in sequence.

[1]: https://www.sqlshack.com/sql-server-lock-issues-when-using-a-ddl-including-select-into-clause-in-long-running-transactions/
[2]: https://bennettoxford.slack.com/archives/C63UXGB8E/p1687962434219019

PR #1400, opened 2023-07-07
Avoid locking `sys.table` when creating temporary tables

evansd:
Addresses #1360, but I don't want to close this until we've confirmed this is working in prod and actually fixes the issue.

As well as addressing the above this PR fixes an absolute howler of a logic bug on my part, extends the tests to cover that bug, and modifies the Hypothesis tests to exercise the temporary table logic.

Removing the autoincrement behaviour from our test fixtures avoids an issue where, if the "results" query ends up selecting directly from an autoincrement column (an "identity column" in MSSQL-speak), then the temporary table inherits this "identity" property and MSSQL refuses to let you write to it.

It _is_ possible to work around this (and I include the details below for future reference) but it adds complexity and I don't believe that we will encounter this problem in the real data so I propose not to add the workaround at this point.

---

You can demonstrate the "identity column" issue by re-enabling autoincrement behaviour in `orm_utils` and then running:
```
pytest tests/integration/query_engines/test_mssql.py
```

This will fail with an error like:
```
Cannot insert explicit value for identity column in table '<table_name>' when IDENTITY_INSERT is set to OFF
```

The problem is that the `SELECT * INTO` construct we use to create the temporary table is causing the temporary table to inherit the identity column status of the table being queried, and MSSQL is then refusing to allow writes to it because it thinks it should be managing these values.

It's possible to [disable this behaviour](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-identity-insert-transact-sql) on a per-table basis using:
```
SET IDENTITY_INSERT <table_name> ON;
```

This will allow writing values to this column. But a further complication is that if we try to execute this statement against a database which _doesn't_ have an identity column then we get an error. And ehrQL doesn't know enough about how the database is configured to be able to know if a given temporary table will end up with an identity column or not.

To work around this we can wrap the statement in a TRY/CATCH so that we can always safely run it and it will be a no-op if it isn't needed.

As a final gotcha, this property can only be enabled against a single table at a time within a given session. So we need to re-enable identity insert protection after we've written to the table (again, using a TRY/CATCH in case the table didn't have an identity column in the first place).

Putting all this together, here is the relevant SQLAlchemy construct to live in `query_engines/mssql_dialect.py`:
```py
# https://learn.microsoft.com/en-us/sql/t-sql/statements/set-identity-insert-transact-sql
class SetIdentityInsert(Executable, ClauseElement):
    inherit_cache = True

    def __init__(self, table, enabled, ignore_errors=False):
        self.table = table
        self.enabled = enabled
        self.ignore_errors = ignore_errors

    def get_children(self):
        return (self.table,)


@compiles(SetIdentityInsert)
def visit_set_identity_insert(element, compiler, **kw):
    compiled = "SET IDENTITY_INSERT {} {}".format(
        compiler.process(element.table, asfrom=True, **kw),
        "ON" if element.enabled else "OFF",
    )
    if element.ignore_errors:
        return f"BEGIN TRY\n{compiled}\nEND TRY\nBEGIN CATCH END CATCH"
    else:
        return compiled
```

And this can be used by wrapping the insert query in `query_engines/mssql.py`:
```py
SetIdentityInsert(table, enabled=True, ignore_errors=True)
table.insert().from_select(columns, query),
SetIdentityInsert(table, enabled=False, ignore_errors=True)
```


ehrql/query_engines/mssql.py
```diff
@@ -233,17 +233,42 @@ def temporary_table_from_query(table_name, query, index_col=0, schema=None):
         sqlalchemy.Column(c.name, c.type, key=c.key) for c in query.selected_columns
     ]
     table = GeneratedTable(table_name, sqlalchemy.MetaData(), *columns, schema=schema)
-    table.setup_queries = [
-        # Use the MSSQL `SELECT * INTO ...` construct to create and populate this
-        # table
-        SelectStarInto(table, query.alias()),
+    # The "#" prefix indicates a session-scoped temporary table which won't persist if
+    # we open a new connection to the database
+    table.is_persistent = not table_name.startswith("#")
```
bloodearnest, line 238: How this even work before?!

inglesp (approved):
Well done.

evansd:
We can confirm that this works by running two copies of a minimal dataset definition simultaneously.

Here is the dataset definition:
```py
from ehrql import Dataset
from ehrql.tables.beta.core import ons_deaths, patients

dataset = Dataset()
dataset.define_population(patients.exists_for_patient())
dataset.has_death_certificate = ons_deaths.exists_for_patient()
```

Here is the SQL it generates:
```sql
-- Setup query 001 / 005
SELECT * INTO [#tmp_1] FROM (SELECT DISTINCT [ONS_Deaths].[Patient_ID] AS patient_id 
FROM [ONS_Deaths]) AS anon_1;

-- Setup query 002 / 005
CREATE CLUSTERED INDEX [ix_#tmp_1_patient_id] ON [#tmp_1] (patient_id);

-- Setup query 003 / 005
SELECT * INTO [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2 FROM (
SELECT patients.patient_id AS patient_id, CASE WHEN ([#tmp_1].patient_id IS NOT NULL) THEN 1 WHEN ([#tmp_1].patient_id IS NULL) THEN 0 END AS has_death_certificate 
FROM (
            SELECT
                Patient_ID as patient_id,
                DateOfBirth as date_of_birth,
                CASE
                    WHEN Sex = 'M' THEN 'male'
                    WHEN Sex = 'F' THEN 'female'
                    WHEN Sex = 'I' THEN 'intersex'
                    ELSE 'unknown'
                END AS sex,
                CASE
                    WHEN DateOfDeath != '99991231' THEN DateOfDeath
                END As date_of_death
            FROM Patient
        ) AS patients LEFT OUTER JOIN [#tmp_1] ON [#tmp_1].patient_id = patients.patient_id 
WHERE patients.patient_id IS NOT NULL) AS anon_1 WHERE 0=1;

-- Setup query 004 / 005
INSERT INTO [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2 (patient_id, has_death_certificate)
SELECT patients.patient_id AS patient_id, CASE WHEN ([#tmp_1].patient_id IS NOT NULL) THEN 1 WHEN ([#tmp_1].patient_id IS NULL) THEN 0 END AS has_death_certificate 
FROM (
            SELECT
                Patient_ID as patient_id,
                DateOfBirth as date_of_birth,
                CASE
                    WHEN Sex = 'M' THEN 'male'
                    WHEN Sex = 'F' THEN 'female'
                    WHEN Sex = 'I' THEN 'intersex'
                    ELSE 'unknown'
                END AS sex,
                CASE
                    WHEN DateOfDeath != '99991231' THEN DateOfDeath
                END As date_of_death
            FROM Patient
        ) AS patients LEFT OUTER JOIN [#tmp_1] ON [#tmp_1].patient_id = patients.patient_id 
WHERE patients.patient_id IS NOT NULL;

-- Setup query 005 / 005
CREATE CLUSTERED INDEX [ix_OPENCoronaTempTables_dbo_results_20230707_1311_d8834f26d4e2_patient_id] ON [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2 (patient_id);

-- Results query
SELECT [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2.patient_id, [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2.has_death_certificate 
FROM [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2;

-- Cleanup query 001 / 002
DROP TABLE IF EXISTS [OPENCoronaTempTables].dbo.results_20230707_1311_d8834f26d4e2;

-- Cleanup query 002 / 002
DROP TABLE IF EXISTS [#tmp_1];
```

And here are the logs of the two queries:
Screenshot from 2023-07-07 14-08-45
Screenshot from 2023-07-07 14-09-45

We can see that query 3, which creates the temporary table using the `WHERE 0=1` trick, runs in under a second which shows that the query planner is able to spot that it doesn't actually need to fetch any data in order to execute the query.

Query 4, which fetches the results and populates the temporary table, runs in ~1 minute each time and we can see from the timestamps that the two queries are running simultaneously.

PR #1410, opened 2023-07-10
Fix `open_prompt`

iaindillingham:
This PR bundles a fix, two features, and an update to the documentation that all stem from a chore: fetching and building the TPP schema. As a consequence, I bundled them into a single PR.

We start with 2102d94, which was a consequence of running

```
python tests/lib/update_tpp_schema.py fetch
python tests/lib/update_tpp_schema.py build
```

following a recent release of the [OpenSAFELY-TPP database schema][1] report.

At our request, TPP have changed the schema of the `OpenPROMPT` table, so 914bb0a, ea91391, and, ce510b6 incorporate these changes into ehrQL. Finally, b166ccc updates the documentation.

[1]: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -2140,7 +2140,7 @@ GitHub.
     <code>CTV3 (Read v3) code</code>
   </dt>
   <dd markdown="block">
-The response to the question, as a CTV3 code. Alternatively, if the question admits a number as the response, then the question, as a CTV3 code.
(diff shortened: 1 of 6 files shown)


evansd (approved):
An admirably clear PR (especially given the complexity of the data model you're having to deal with)!

I now finally understand what the `CodedEvent_ID` column was doing in this table, which is a bonus.

Incidentally, the `fetch` and `build` steps are wrapped together in a Just command `just update-tpp-schema` if you find yourself needing to run this often.

PR #1413, opened 2023-07-10
feat: Expose `Appointment.Status`

### LINKED ISSUE #1308: Expose `Appointment.Status`
It looks as though [_Project #152: Digital access to primary care for older people during COVID_][1] would benefit from access to `Appointment.Status` through ehrQL. Because of #1307, not all values of this column will be present.

Right, how do I do that, then?

Add a property to the `EventFrame`. Don't forget to add constraints; these are used by the dummy-data generator and by the documentation.

https://github.com/opensafely-core/ehrql/blob/b453a7ed1691ba12c31107aa8f7fbeb177b3515a/ehrql/tables/beta/tpp.py#L202-L205

Update the query. Don't forget to join to [the `DataDictionary` table][2] to get human-readable status descriptions rather than status codes. (These are the values of `Constraint.Categorical`, above.)

https://github.com/opensafely-core/ehrql/blob/b453a7ed1691ba12c31107aa8f7fbeb177b3515a/ehrql/backends/tpp.py#L208-L220

Run the tests. Fix the tests. Remember [Conventional Commits][3].

[1]: https://www.opensafely.org/approved-projects/#project-152
[2]: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#DataDictionary
[3]: https://www.conventionalcommits.org/en/v1.0.0/

milanwiedemann:
This PR exposes the `Appointment.Status` and joins to the DataDictionary to map the numeric values to more descriptive strings.

Closes #1308

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -220,8 +220,17 @@ class TPPBackend(BaseBackend):
                 Appointment_ID AS appointment_id,
                 Patient_ID AS patient_id,
                 CAST(BookedDate AS date) AS booked_date,
-                CAST(StartDate AS date) AS start_date
-            FROM Appointment
+                CAST(StartDate AS date) AS start_date,
+                Description AS status
+            FROM Appointment AS appt
+            LEFT JOIN (
+                SELECT
+                    Code,
+                    Description
+                FROM DataDictionary
+                WHERE [Table] = 'Appointment'
+            ) AS dict
+            ON appt.Status = dict.Code
         """
     )
 
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -317,6 +317,19 @@ The date the appointment was due to start
   </dd>
 </div>
 
+<div markdown="block">
+  <dt id="appointments.status">
+    <strong>status</strong>
+    <a class="headerlink" href="#appointments.status" title="Permanent link">🔗</a>
+    <code>string</code>
+  </dt>
+  <dd markdown="block">
+The status of the appointment
+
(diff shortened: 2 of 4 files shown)


ehrql/backends/tpp.py
```diff
@@ -218,10 +218,18 @@ def medications(self):
         """
             SELECT DISTINCT
                 Appointment_ID AS appointment_id,
+                CAST (BookedDate AS date) AS booked_date,
+                CAST(StartDate AS date) AS start_date,
```
iaindillingham, line 222: By moving these lines, we've introduced noise into the diff. We don't need to move them.

ehrql/backends/tpp.py
```diff
@@ -218,10 +218,18 @@ def medications(self):
...
-                CAST(StartDate AS date) AS start_date
-            FROM Appointment
+                Description AS status
+            FROM Appointment AS appt
+            LEFT JOIN (
+                SELECT
+                    Code,
+                    Description
+                FROM DataDictionary
+                    WHERE [Table] = 'Appointment') AS dict
```
iaindillingham, line 231: I'd suggest indenting `WHERE` to the same level as `FROM` and terminating the sub-query on a new line

```suggestion
                WHERE [Table] = 'Appointment'
            ) AS dict
```

iaindillingham (approved):
I've left a couple of nice-to-have but not essential suggestions. If you choose to make them, then you can `git commit --amend --no-edit` to avoid an interactive rebase.

One more: think about the title and body of the commit message.

I have a template to remind me of [the seven rules of a great Git commit message][1]. In this case, wrap the body to 72 characters; and use both title and body to explain what and why rather than describe how. You've done the latter to some extent, but there's some room for improvement. How about...

```
feat: Add appointments.status

This exposes Appointment.Status (a database table) as
appointments.status (an ehrQL table), mapping unfriendly status codes to
friendly status descriptions from DataDictionary. (We do something
similar in cohort-extractor, but we hard-code the status descriptions.)

```

[1]: https://cbea.ms/git-commit/

PR #1415, opened 2023-07-11
Add dump-example-data command

### LINKED ISSUE #1414: Bake sandbox data into ehrQL
Instead of making users download sandbox data separately, some data should be baked into the `ehrql` package, so that `opensafely exec ehrql:v0 sandbox` just works.

inglesp:
This slightly simplifies the tutorial setup process, and means we can retire https://github.com/opensafely-core/ehrql-example-data.

I considered instead changing the sandbox, so that running it without arguments would use data that's part of the ehrQL package, but discounted that because users wouldn't be able to inspect or modify the example data, and it would mean that we couldn't use the same dummy data throughout the tutorial.

Fixes #1414.


docs/tutorial/installation-and-setup.md
```diff
@@ -125,7 +125,7 @@ and we can help you get started.
 ## Downloading some dummy data
 
 Before running queries against real data in an OpenSAFELY backend,
-you can test your queries against dummy data on your own computer.
+you should test your queries against dummy data on your own computer.
 
 In the rest of this documentation,
 we will use the same small set of dummy data.
```
StevenMaude, line 131: suggestion: Maybe add here:

> we will use the same small set of dummy data.
> The example dummy data for this tutorial is provided for you by ehrQL.

Or similar.

StevenMaude (approved):

evansd (approved):
:ok_hand:

PR #1418, opened 2023-07-12
Expose `ONS_CIS_New.pseudo_visit_id`

### LINKED ISSUE #1416: Expose `ONS_CIS_New.pseudo_visit_id`
This was requested by @LindaNab in #1358, but I hesitated because I needed to think about how to expose a `varbinary` DB column as an ehrQL column. With Dave's help, I've figured it out.

iaindillingham:
Closes #1416

The `ONS_CIS_New.pseudo_visit_id` column is a `varbinary`. Investigation
on L2 tells us the values are 32 bytes (256 bits) in length, so we
assume the values are SHA-256 hashes.

When we expose this column, we want the values to be intelligible
strings. Consequently, when we convert from a `varbinary` to a
`varchar`, we also convert each binary character to two hex characters
by passing `2` as the third positional argument to `CONVERT`. The T-SQL
documentation for [binary styles][1] makes this clear, if you read it
enough:

> 1, 2 A data_type of type character requires a binary expression.
> Each binary character is converted into two hexadecimal characters.
> The characters 0x aren't added to the left of the converted result
> for style 2.

Because of this conversion, we expose this column as `ons_cis.visit_id`
rather than `ons_cis_raw.visit_id`.

[1]: https://learn.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-2017#binary-styles

### DIFF
--- a/tests/integration/backends/test_tpp.py
+++ b/tests/integration/backends/test_tpp.py
@@ -1,3 +1,4 @@
+import hashlib
 from datetime import date
 
 import pytest
@@ -634,6 +635,7 @@ def test_ons_cis(select_all):
         ONS_CIS_New(
             Patient_ID=1,
             visit_date=date(2021, 10, 20),
+            pseudo_visit_id=hashlib.sha256(b"1").digest(),
             visit_num=1,
             last_linkage_dt=date(2022, 8, 15),
             nhs_data_share=1,
@@ -646,6 +648,7 @@ def test_ons_cis(select_all):
         {
             "patient_id": 1,
             "visit_date": date(2021, 10, 20),
+            "visit_id": hashlib.sha256(b"1").digest().hex().upper(),
             "visit_num": 1,
             "is_opted_out_of_nhs_data_share": True,
             "last_linkage_dt": date(2022, 8, 15),
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -1670,6 +1670,18 @@ Data from the ONS Covid Infection Survey.
 </div>
 
 <div markdown="block">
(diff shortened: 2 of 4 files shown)


tests/integration/backends/test_tpp.py
```diff
@@ -645,6 +648,7 @@ def test_ons_cis(select_all):
         {
             "patient_id": 1,
             "visit_date": date(2021, 10, 20),
+            "visit_id": base64.b16encode(hashlib.sha256(b"1").digest()).decode(),
```
evansd, line 651: I think you can just do:
```suggestion
            "visit_id": hashlib.sha256(b"1").digest().hex(),
```

In fact, you could use the `.hexdigest()` method, but that's specific to `hashlib` whereas `.hex()` is a method on all bytestrings so I think it's preferable here.

evansd (approved):
Looks good.

I keep meaning to expand `MappedTable` to support applying casts to columns so we don't always have to switch to `QueryTable` everytime. I'll get to it one day!

tests/integration/backends/test_tpp.py
iaindillingham: Thanks, Dave. Both `.hexdigest()` and `.hex()` seem to use characters `[a-z]`, whilst `b16encode.decode()` and T-SQL seem to use characters `[A-Z]`. I've force pushed a change that calls `.upper()` on the resulting hex.

https://github.com/opensafely-core/ehrql/blob/0bfd64f0be5cb15e6b4390d2e8058f0060973a7c/tests/integration/backends/test_tpp.py#L651

evansd: Ah, makes sense. Thanks.

PR #1419, opened 2023-07-13
Add experimental command for assuring dataset definitions

inglesp:


iaindillingham (approved):
Thanks for talking this through, @inglesp. You mentioned that you're going to add a comment about the structure of `patient_data`, which will be useful.

PR #1432, opened 2023-07-17
Fix ONS Deaths code example in docs

Jongmassey:
The [code example in the current docs for ONS Deaths](https://docs.opensafely.org/ehrql/how-to/examples/#finding-each-patients-date-place-and-cause-of-death-from-ons-records) fails with the following error because ONS Deaths table is not one row per patient.

```
Compiling dataset definition from analysis/dataset_definition.py [ehrql.main] 
     Failed to import 'analysis/dataset_definition.py':

       File "/workspace/analysis/dataset_definition.py", line 35, in <module>
         dataset.date_of_death = ons_deaths.date
         ^^^^^^^^^^^^^^^^^^^^^
     TypeError: Invalid variable 'date_of_death'. Dataset variables must return one row per patient
```

This change provides a working code example that aggregates it to 1RPP and explains why in the preceding text.

### DIFF
--- a/docs/how-to/examples.md
+++ b/docs/how-to/examples.md
@@ -126,7 +126,8 @@ dataset.date_of_death = patients.date_of_death
 :notepad_spiral: This value comes from the patient's EHR record.
 
 There is generally a lag between the death being recorded in ONS data and appearing in the primary care record,
-but the date itself is usually reliable when it appears.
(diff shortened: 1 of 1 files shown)


iaindillingham (approved):
Good spot! Thanks for updating the docs.

PR #1436, opened 2023-07-19
Generate dummy patients even when no groups supplied

evansd:
The heuristic for calculating the population size is obviously terrible, but it's not yet obvious exactly how and why it's terrible so we defer trying to improve it beyond making sure we don't accidentally multiply by zero.


iaindillingham (approved):
Thanks for separating the refactoring and the fix; doing so made it easier to review.

Issue #1242, opened 2023-05-02
Overnight generative tests fail with MSSQL operational error


evansd:
The scheduled, overnight generative tests now _almost_ always fail with an error like:
```py
pymssql._pymssql.OperationalError(
    701,
    (
        b"There is insufficient system memory in resource pool 'default' to run this query."
        b"DB-Lib error message 20018, severity 17:"
        b"\nGeneral SQL Server error: Check messages from the SQL Server\n"
    ),
)
```

Followed by multiple instances of:
```py
pymssql._pymssql.OperationalError(
    701,
    b'DB-Lib error message 20017, severity 9:\n'
    b'Unexpected EOF from the server (localhost)\n'
    b'DB-Lib error message 20002, severity 9:\n'
    b'Adaptive Server connection failed (localhost)\n'
)
```

I say "almost always" because we have had one successful run in the last [25 days](https://github.com/opensafely-core/ehrql/actions/workflows/generative-tests.yml).

The first time we saw this was against this commit: https://github.com/opensafely-core/ehrql/commit/69db82aa09281cb056413c2c0050c573a68f9c59

However, this was the second test run against this commit and the first time was successful so there's not necessarily anything interesting about this commit itself.

Furthermore, we'd had recently merged [this fix](https://github.com/opensafely-core/ehrql/pull/1182) for a separate generative tests issue involving SQLite. It's quite possible that this SQLite issue was serving to mask the MSSQL issue by preventing the tests from running long enough to trigger it.

I can't reproduce this by running the tests locally, however I _can_ reproduce it if I artificially constraint the memory available to the container by editing this line:
https://github.com/opensafely-core/ehrql/blob/9bdcd1a31cdb98dfeaa2d8907c17215d4a6e2fe9/tests/lib/databases.py#L171

And adding the `mem_limit` argument:
```py
    command="/opt/mssql/bin/sqlservr",
    mem_limit="1024m",
```

Unfortunately, we don't seem to get any meaningful debugging information out of the MSSQL container.

evansd:
Hopefully closed by #1618

Issue #796, opened 2022-10-20
Add a Trino query engine


evansd:
This is needed for EMIS support.

This might be useful:
https://github.com/trinodb/trino-python-client

Also, apparently the PyHive dialect we already use can speak Trino/Presto:
https://github.com/dropbox/PyHive

Note, Trino has quite a different execution model from a traditional SQL database, giving it quite different performance characteristics. It might be worth using CTEs (as we do with SQLite) so we can provide the entire query upfront and give Trino a chance to optimise the number of passes it needs to do over the data.

evansd:
Trino provides a unified SQL layer over a wide variety of underlying data sources (accessed using what it calls "connectors").

There's an in-memory connector which might be useful for testing:
https://trino.io/docs/current/connector/memory.html

PR #1437, opened 2023-07-19
Add a trino query engine

rebkwok:
See #796 

Adds a trino query engine and the setup to run tests with it.

The trino container used for tests uses the official [trino image](https://hub.docker.com/r/trinodb/trino) with a [memory connector](https://trino.io/docs/current/connector/memory.html).

The generative tests are taking a lot longer to run now, and with 200 examples they were failing (after about 1 hour) with an [mssql memory error](https://github.com/opensafely-core/ehrql/actions/runs/5881025163/job/15948622443) which is _probably_ unrelated. I've removed the `GENTESTS_COMPREHENSIVE` flag and reduced the default number of examples to 100 in the `just test-all` command that runs in CI, which gets this branch passing in approximately the same time as before, without hitting the mssql memory error, and I'll defer dealing with that for the existing issue #1242.


ehrql/query_engines/base_sql.py
```diff
@@ -414,13 +417,21 @@ def to_first_of_month(self, date):
...
+        return self.get_aggregate_subquery(aggregate_function, *args).label("greatest")
 
     @get_sql.register(Function.MinimumOf)
     def get_sql_minimum_of(self, node):
+        aggregate_function = self.aggregate_functions()["minimum_of"]
         args = self.get_nary_function_args(node)
-        return self.get_aggregate_subquery(sqlalchemy.func.min, *args).label("least")
+        return self.get_aggregate_subquery(aggregate_function, *args).label("least")
+
+    def aggregate_functions(self):
```
Jongmassey, line 430: I'd be inclined to make the 'base' functions the canonical ANSI SQL functions for this which are `greatest` and `least`

ehrql/query_engines/base_sql.py
```diff
@@ -719,3 +727,33 @@ def get_sort_conditions(frame):
...
+    #   ["b", None, "", "a"],
+    #   [None, "", "a", "b"]
+    #   ["", "a", "b", None]
+    # ]
+    # Calling coalesce on each of these returns ["a", "b", "a", "a"] to be used in the
+    # aggregate function
+    len_cols = len(columns)
+    # if there is only one column in the list, just return it. Coalesce requires at least two
+    # arguments
+    if len_cols == 1:
```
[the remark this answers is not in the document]
rebkwok: I don't think so - I think it would error before that, but will check

ehrql/query_engines/base_sql.py
```diff
@@ -256,9 +256,12 @@ def get_sql_floordivide(self, node):
         float_result = self.get_sql_truedivide(node)
         return sqlalchemy.cast(SQLFunction("FLOOR", float_result), sqlalchemy.Integer)
 
+    def cast_to_int(self, value):
+        return sqlalchemy.cast(value, sqlalchemy.Integer)
+
```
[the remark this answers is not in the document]
rebkwok: We're not consistent with this across various methods, and I dithered about it.  Mostly I went with the one that minimised duplicated code (e.g. two out of three engines used the same implementation, I left it on the base engine - partly that was because I couldn't decide how/if we should share methods between engines)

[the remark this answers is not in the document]
iaindillingham: That's a great strategy. Let's stick with it.

ehrql/query_engines/base_sql.py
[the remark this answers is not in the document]
rebkwok: No, this would only happen if you tried to write e.g. `minimum_of()` or `maximum_of()` (with no arguments) in a dataset definition. That does produce a pretty nasty error message, but it happens earlier than this stage, so maybe something to deal with, but not necessarily here.  I made #1508 for it.

ehrql/query_engines/base_sql.py
```diff
@@ -634,6 +634,14 @@ def get_results(self, variable_definitions):
         results_query = self.get_query(variable_definitions)
         setup_queries, cleanup_queries = get_setup_and_cleanup_queries(results_query)
         with self.engine.connect() as connection:
+
+            def do_cleanup():
+                for i, cleanup_query in enumerate(cleanup_queries, start=1):
+                    log.info(
+                        f"Running cleanup query {i:03} / {len(cleanup_queries):03}"
+                    )
+                    connection.execute(cleanup_query)
```
evansd, line 652: I can understand exactly why you made this change, but I'm not convinced it's the right approach here. There are any number of ways we could still fail to clean up e.g. failing during one of the setup queries, or a failure which leaves the connection dead. So we're going to need some out-of-band mechanism to clean these things up in any case. I think it would leave our code easier to reason about if we just exited on error and didn't attempt to clean up.

tests/integration/utils/test_sqlalchemy_query_utils.py
```diff
@@ -32,5 +32,7 @@ def test_insert_many(engine):
         connection.execute(insert_many)
         response = connection.execute(sqlalchemy.select(table))
         results = list(response)
+        # Explicitly drop the table as it persists in the Trino engine
```
evansd, line 35: Oh yeah, I guess it must persist in the MSSQL engine too (it's just a vanilla table, not a temporary one) but for whatever reason it doesn't cause us problems there. Cleaning up like this seems sensible in any case.

ehrql/query_engines/base_sql.py
evansd: As to what goes in the base engine and what goes in the child engines, I don't think we've really got a consistent story yet (or certainly not one that's consistently applied).

I've had a vague idea of trying to do it along the lines Jon suggests ("standard" stuff goes in the base engine) but I don't think I've really followed that in practice.

I'm fine with just deciding on pragmatic grounds for now and we can always rework this stuff later without too much difficulty if we want to.

Justfile
```diff
@@ -207,9 +207,9 @@ test-generative *ARGS: devenv
     #!/usr/bin/env bash
     set -euo pipefail
 
-    examples=${GENTEST_EXAMPLES:-200}
+    examples=${GENTEST_EXAMPLES:-100}
     [[ -v CI ]] && echo "::group::Run tests (click to view)" || echo "Run tests"
-    GENTEST_EXAMPLES=$examples GENTEST_COMPREHENSIVE=t $BIN/python -m pytest \
+    GENTEST_EXAMPLES=$examples $BIN/python -m pytest \
```
evansd, line 212: I think we are going to need something that does the job of the comprehensive check, otherwise it's too easy to forget to add new operations to the generative tests. But I've had an idea of how to do this which might dramatically speed things up, and let us get away with far fewer examples in CI. I need to do some experimenting first.

evansd (approved):
This was far less horrific than I feared it might be. But I think what that actually means is that it was exactly as horrific as I imagined, but all the horror has been nicely rebased away into a beautiful series of commits which makes it look like you took a stroll through the park, effortlessly making things work at each stage :)

Less facetiously, this was a really nice PR to review and thanks for all the work that obviously went into it.

I think the only thing that does need addressing is making those inline tables globally unique. But hopefully that's a fairly simple change.

ehrql/query_engines/base_sql.py
rebkwok: OK - I think the global unique ID for the table might help with this. One of the issues was that if the generative tests use an inline table and encounter one of the ignored errors, the inline table didn't get cleaned up and other gen tests would fail. Although if it's in the same test run, it might still be a problem. I'll try it.

Justfile
rebkwok: yes, turning it off here was just throwing it over the fence to #1242

tests/generative/test_query_model.py
```diff
@@ -252,13 +252,18 @@ def run_error_test(query_engines, data, variable):
...
+    # main culprit
     (
         sqlalchemy.exc.DBAPIError,
         re.compile(
             ".+TrinoQueryError.+the query may have too many or too complex expressions.+"
         ),
     ),
+    (
+        sqlalchemy.exc.ProgrammingError,
+        re.compile(".+TrinoUserError.+QUERY_TEXT_TOO_LARGE.+"),
```
evansd, line 265: I wonder how much this is to with the fact the we're using CTEs as our "reify query" mechanism (as that's the default in BaseSQL). It means we end up with one enormous query at the end, rather than a series of smaller queries creating temporary tables.

I think this is the right thing to do for now. There's going to be a whole piece of work in figuring out how all this performs in production (which is likely to be _very_ different from how the MSSQL database performs) and optimising it. But there's no point spending time on that until we've got something solid to work with.

evansd (approved):

PR #1438, opened 2023-07-20
feat: Support bare `when` expressions

inglesp:
This simplifies `case` expressions that have a single `when` clause.

We can now replace:

    category = case(
        when(size < 15).then("small"),
        default="large",
    )

with:

    category = when(size < 15).then("small").otherwise("large")


evansd (approved):
It's pleasing how simple this turned out, isn't it?

There are a couple of follow up things I'd like to do at some point, but definitely not now. One is to at least consider changing `default` to `otherwise` for consistency. The other is to add error handling so that passing `when().then()` without `otherwise` into other functions (or assigning to a dataset) gives a helpful error message.

Are you happy for me to make tickets for those?

inglesp:
> It's pleasing how simple this turned out, isn't it?

Yes, I almost feel diddled.

> Are you happy for me to make tickets for those?

Go for it!

PR #1441, opened 2023-07-20
Move `refomat_docstring` to `utils` and use it more widely

evansd:
The `reformat_docstring()` function makes it much more convenient to write embedded Markdown in triple-quoted strings by stripping off the nested indentation. We want to be able to use for things other than just docstrings, so we move it out of the `docs` module and into `string_utils`.

### DIFF
--- a/ehrql/utils/string_utils.py
+++ b/ehrql/utils/string_utils.py
@@ -1,0 +1,13 @@
+import textwrap
+
+
+def strip_indent(s):
+    """
+    Remove indentation from a multiline string
+
+    This is especially useful for taking docstrings and displaying them as markdown.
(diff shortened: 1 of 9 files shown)


ehrql/utils/string_utils.py
```diff
@@ -0,0 +1,13 @@
+import textwrap
+
+
+def strip_indent(s):
+    """
+    Remove indentation from a multiline string
+
+    This is especially useful for taking docstrings and displaying them as markdown.
+    Note that before de-indenting we string leading newlines but not leading whitespace
```
inglesp, line 9:
```suggestion
    Note that before de-indenting we strip leading newlines but not leading whitespace
```

inglesp (approved):

PR #1442, opened 2023-07-20
CLI reference docs

evansd:
This is still WIP.


docs/includes/generated_docs/cli.md
```diff
@@ -0,0 +1,593 @@
...
+
+</div>
+
+<div class="attr-heading" id="ehrql.help">
+  <tt>-h, --help</tt>
+  <a class="headerlink" href="#ehrql.help" title="Permanent link">🔗</a>
+</div>
+<div markdown="block" class="indent">
+show this help message and exit
+
```
StevenMaude, line 78: nit: Again, not a big problem, but I think there's an extra empty line being added before the closing `div`.

StevenMaude (approved):
This looks really great :bookmark_tabs: :+1:

ehrql/__main__.py
```diff
@@ -60,236 +73,367 @@ def main(args, environ=None):
...
+            """
+            Generate dummy data for a dataset and write out tables as CSV.
+
+            This generates the same dummy data that `generate-dataset` would, but
+            instead of using this to produce a dataset it writes the underlying data
+            tables out as CSV (one file per table).
+
+            The directory containing these CSV files can then be used as the
+            [`--dummy-tables`](#generate-dataset.dummy-tables) argument to
+            `generate-dataset` to produce the dataset.
```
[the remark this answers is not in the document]
evansd: Oh yes, I meant to mention this in the PR body. You're right that in an ideal world we'd filter the markdown syntax (or at least, the non-useful parts of it) out of the console help. But I think there'd be quite a bit of work involved in that and I'm not sure it's worth it right now. I think it makes sense to prioritise making the docs useful in HTML given that's where I expect most of our users to encounter them. And it's OK if they're imperfect in the console as long as they're not unusable.

docs/includes/generated_docs/cli.md
evansd: It turns out that this is because there's a slot in the argument template which is only used for subcommands, and when that's empty we get an empty line. Given that the Markdown compiles to exactly the same HTML either way I'm inclined not to worry about it.

ehrql/__main__.py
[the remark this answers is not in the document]
StevenMaude: Yes, completely agree.

I think the benefit of having them available in the documentation far outweighs this nitpick.

docs/includes/generated_docs/cli.md
StevenMaude: Yep, also all fine, it was more a "in case you hadn't seen". But it's not a big problem.

PR #1443, opened 2023-07-21
fix: Set `environ` when generating measures

iaindillingham:
We did this when generating a dataset, but not when generating measures. Consequently, environment variables such as `TEMP_DATABASE_NAME` were available to subclasses of `BaseBackend` when generating datasets but, not when generating measures.

Users, like @alschaffer, who included the `medications` table in their measures definitions would have seen references to `PLACEHOLDER_FOR_TEMP_DATABASE_NAME` in the logs because `TEMP_DATABASE_NAME` was not passed to `BaseBackend.config`. Now, they won't.

### DIFF
--- a/ehrql/__main__.py
+++ b/ehrql/__main__.py
@@ -183,6 +183,7 @@ def add_create_dummy_tables(subparsers, environ, user_args):
 def add_generate_measures(subparsers, environ, user_args):
     parser = subparsers.add_parser("generate-measures", help="Generate measures")
     parser.set_defaults(function=generate_measures)
+    parser.set_defaults(environ=environ)
     parser.set_defaults(user_args=user_args)
     parser.add_argument(
         "--output",


iaindillingham:
I don't think it's currently possible -- and possibly even desirable -- to test this. We don't, for example, test `ehrql.__main__.add_generate_dataset`.

evansd (approved):
Argh! I can't believe I missed this!

This must've been a nightmare to debug. Thanks for fixing this!

iaindillingham:
It wasn't that bad, @evansd, because in the PR that introduced the custom medication dictionary you suggested using a descriptive default

https://github.com/opensafely-core/ehrql/blob/d6ec2c9573a0a9fc6bdf0955df7e391053eb4f1f/ehrql/backends/tpp.py#L107-L109

So, debugging was a case of working back from `config` to `os.environ` and noticing where the missing link was located. It would have been a lot harder without the descriptive default 🙂

PR #1453, opened 2023-07-24
Ignore warnings so vs code pytest integration works

Jongmassey:
e7cc523 introduced a warning which causes Pytest to error when VS Code Python Testing extension runs `python  -m pytest --collect-only --rootdir tests -s --cache-clear .` to discover the tests in the project. This causes the extension to refuse to load the discovered tests, the change in this PR tells Pytest not to raise an Error on Warnings and thus makes the extension work again

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,7 +80,7 @@ add_ignore = [
 [tool.pytest.ini_options]
 addopts = "--tb=native --strict-markers --dist=loadgroup"
 testpaths = ["tests"]
-filterwarnings = ["ignore::UserWarning"]
+filterwarnings = ['ignore: PYTHONHASHSEED environment variable not set to 0']
 
 [tool.ruff]
 line-length = 88
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ add_ignore = [
 [tool.pytest.ini_options]
 addopts = "--tb=native --strict-markers --dist=loadgroup"
 testpaths = ["tests"]
+filterwarnings = ["ignore::UserWarning"]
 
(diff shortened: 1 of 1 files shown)


inglesp:
Raising an error on warning can be very useful.  Better would be to address this particular warning.

One way to do this would be to tell pytest to set the `PYTHONHASHSEED` variable.  You can do this with [pytest-env](https://pypi.org/project/pytest-env/).

(A better way to do it might be to set the environment variable when your virtualenv is activated, but since everyone's setup is different, there's no easy way of applying this to all developers.)

Jongmassey:
good point, how about only ignoring that specific warning?

iaindillingham (approved):

StevenMaude:
PR with the pytest-env approach: https://github.com/opensafely-core/ehrql/pull/1459

It does fix this problem for me.

evansd:
_Copying this from elsewhere as a comment on a closed PR is likely to get lost._

Unfortunately (and non-obviously), the fix in #1459 won't quite work: it will silence the warning but not address the underlying issue. The problem is that changes to the PYTHONHASHSEED value made after the process starts [don't have any effect](https://stackoverflow.com/questions/30585108/disable-hash-randomization-from-within-python-program) – though the official documentation fails to mention this.

I think  we might be able to fix this for VSCode users by adding a `.env` file with:
```
PYTHONHASHSEED=0
```

It looks like these will be [loaded by default](https://code.visualstudio.com/docs/python/environments#_environment-variables) under the standard configuration.

For developers using different tooling we'll just have to document that variables specified in the `.env` file should be set somehow.

PR #1459, opened 2023-07-26
Specify `PYTHONHASHSEED=0` in pytest configuration

StevenMaude:
#1453 describes the rationale for this problem; warnings can appear if you run `pytest` directly, because this environment variable may not be set.

This uses the `pytest-env` package to do so, which is maintained by the pytest developers.

@inglesp suggested this approach over suppressing the pytest warning that results if you don't have `PYTHONHASHSEED=0` set.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,9 @@ add_ignore = [
 [tool.pytest.ini_options]
 addopts = "--tb=native --strict-markers --dist=loadgroup"
 testpaths = ["tests"]
+env = [
+  "PYTHONHASHSEED=0",
+]
 
 [tool.ruff]
 line-length = 88
--- a/requirements.dev.in
+++ b/requirements.dev.in
@@ -11,6 +11,7 @@ pip-tools
 pre-commit
 pytest
 pytest-cov
+pytest-env
 pytest-mock
 pytest-xdist
 ruff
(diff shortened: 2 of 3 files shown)


Jongmassey:
works in vs code for me

Jongmassey (approved):
assuming CI completes successfully, this seems great!

evansd:
Unfortunately (and non-obviously), this fix won't quite work: it will silence the warning but not address the underlying issue. The problem is that changes to the PYTHONHASHSEED value made after the process starts [don't have any effect](https://stackoverflow.com/questions/30585108/disable-hash-randomization-from-within-python-program) – though the official documentation fails to mention this.

I think  we might be able to fix this for VSCode users by adding a `.env` file with:
```
PYTHONHASHSEED=0
```

It looks like these will be [loaded by default](https://code.visualstudio.com/docs/python/environments#_environment-variables) under the standard configuration.

For developers using different tooling we'll just have to document that variables specified in the `.env` file should be set somehow.

StevenMaude:
Oh, that is an unfortunate gotcha.

Jongmassey:
>For developers using different tooling we'll just have to document that variables specified in the .env file should be set somehow.

a just action to activate .venv and load everything in .env?

evansd:
Anything using `just` has this set already. It's things which aren't using `just` that we need to worry about:
https://github.com/opensafely-core/ehrql/blob/4770ec6f1d387c9aead150f30dfdba9d8c94e208/Justfile#L10-L13

PR #1463, opened 2023-07-27
Remove ONS CIS data

iaindillingham:
Remove ONS CIS data


benbc (commented):
I'd like someone from the Data team to review this too, please.

How much effort would it be to remove references from `tests/lib/tpp_schema.*` as well?

iaindillingham:
> How much effort would it be to remove references from tests/lib/tpp_schema.* as well?

That's harder, as the CSV file is generated by the _OpenSAFELY-TPP database schema_ report and the Python file is generated from the CSV file. We could...

* Ask for the tables to be removed from the schema or (better) from the database
* Change *tests/lib/update_tpp_schema.py* to remove the references from the Python file (but they'd still be present in the CSV file)
* Remove references from the CSV file and the Python file, but they'd be added back (on a branch) when the `update-tpp-schema` GH action next runs
* Change the report, regenerate, check, release, publish

Jongmassey (approved):
LGTM

PR #1467, opened 2023-07-28
PoC for allowing hinting for tables that restrict population

benbc:


Jongmassey:
I'm not sure I prefer the complexity of this rather than inferring that if there is no primary population table then the inline patient table is the population table. 

I can see a possible advantage of this approach to allow population-restriction with an dataset definition that draws from both the `patients` table AND the inline table AND has a `inline_table.exists_for_patient()` i.e. an inner join is required and would be more performant than a LOJ and only applying the `exists_for_patient()` further down the query stack. 

I might have just talked myself into the hinting approach for answering the question of "how do we infer the correct join type for a table" as opposed to drilling through the query model to look for a filtering function that implies an inner join.

PR #1470, opened 2023-07-31
feat: Revert changes to avoid DB lock

iaindillingham:


ghickman (approved):

evansd:
There's just one commit in the original PR that wants reverting (the "avoid lock" one). The rest are bug/testing fixes and should stay.

iaindillingham:
Thanks for your comments, @evansd and @benbc. I should have been more targeted with this PR: I simply chose to revert everything in #1400. If I've understood your comments correctly, then I should revert only 0e535f5ef4a3e002e7bc10b0bcc78e41a94f9a34.

That said, I'm unclear why I should also revert 98cbddce2d0f6e7d7c90ae9cef70ae2bdbe13246, @benbc. Should I do so because we're trying, as far as we can, to take the system back to early July? Something else?

benbc:
Oh, I didn't see Dave's comment. Yes -- I just meant that you were right to do it because it's a bit more straightforward to back out the whole PR. Totally fine to just revert the one commit that Dave pointed to.

iaindillingham:
Thanks again @evansd and @benbc. We now revert only 0e535f5ef4a3e002e7bc10b0bcc78e41a94f9a34.

benbc (approved):

PR #1473, opened 2023-08-02
Fix `open_prompt.ctv3_code`

iaindillingham:
aaee849 adds `OpenPROMPT.CTV3Code` to TPP schema as a temporary measure, until the "OpenSAFELY-TPP database schema" report is next released. We don't want to release that report now, because we don't want to block the OpenPROMPT team.

9600870 is the feature commit, mapping `open_prompt.ctv3_code` to `OpenPROMPT.CTV3Code`.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -378,9 +378,7 @@ class TPPBackend(BaseBackend):
             CASE
                 WHEN CodeSystemId = 0 THEN ConceptId
             END AS snomedct_code,
-            CASE
-                WHEN CodeSystemId = 2 THEN ConceptId
-            END ctv3_code,
+            CTV3Code AS ctv3_code,
             CAST(CreationDate AS date) AS creation_date,
             CAST(ConsultationDate AS date) AS consultation_date,
             Consultation_ID AS consultation_id,
--- a/tests/lib/update_tpp_schema.py
+++ b/tests/lib/update_tpp_schema.py
@@ -141,6 +141,12 @@ def apply_schema_modifications(by_table):
         {"ColumnName": "CodingSystem", "ColumnType": "int"},
     ]
 
+    # TODO: Remove this when the "OpenSAFELY-TPP database schema" report is next
+    # released.
+    by_table["OpenPROMPT"] += [
+        {"ColumnName": "CTV3Code", "ColumnType": "varchar", "MaxLength": 50},
+    ]
+
     # We don't get column collation information but we know this matters in some cases
     # because you can't compare columns across tables unless the collations are
     # compatible. We add collations here for the two critical columns whose collations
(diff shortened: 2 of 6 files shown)


tests/lib/update_tpp_schema.py
```diff
@@ -141,6 +141,12 @@ def apply_schema_modifications(by_table):
         {"ColumnName": "CodingSystem", "ColumnType": "int"},
     ]
 
+    # TODO: Remove this when the "OpenSAFELY-TPP database schema" report is next
+    # released.
```
StevenMaude, line 145: suggestion: It's potentially worth opening an issue and adding it to the backlog, so it doesn't get forgotten about.

ehrql/backends/tpp.py
```diff
@@ -378,9 +378,7 @@ def medications(self):
             CASE
                 WHEN CodeSystemId = 0 THEN ConceptId
             END AS snomedct_code,
-            CASE
-                WHEN CodeSystemId = 2 THEN ConceptId
-            END ctv3_code,
+            CTV3Code AS ctv3_code,
```
StevenMaude, line 381: question: Can you explain a little more of the context of what's happening here, and what problem it fixes, please? I know very little about OpenPROMPT.

StevenMaude: Is it something like "the existing data in the column is not what is expected, and we're having to pull that from another temporary column"?

StevenMaude (approved):
The code changes look fine.

I'd personally appreciate a bit more explanation of why the changes are there, only to aid my understanding.

That could also be useful in the commit messages for future reference/historical digging, although it looks like this is a temporary fix anyway.

StevenMaude:
This might also be useful to explain to someone — in other words, me — with no context:

> …until the "OpenSAFELY-TPP database schema" report is next released. We don't want to release that report now, because we don't want to block the OpenPROMPT team.

* How will the new report address the problem?
* Why would releasing the report block the team?

ehrql/backends/tpp.py
iaindillingham: CTV3 codes were originally in `OpenPROMPT.CodedEvent_ID`. This was an error, so they were removed and this column became a FK to `CodedEvent.CodedEvent_ID`. It was assumed that the codes in `OpenPROMPT.ConceptId` -- which are a mix of CTV3 "Y" codes and SNOMED CT codes -- would be sufficient (`OpenPROMPT.CodeSystemId` is the flag that tells us the coding system). However, they were _not_ sufficient: there is an imperfect mapping between SNOMED CT codes and "regular" (non-"Y") CTV3 codes. So, all CTV3 codes were added to `OpenPROMPT.CTV3Code`. The above diff overcomes the imperfect mapping.

iaindillingham:
Thanks, @StevenMaude. Sorry the comments weren't as clear as they could have been. Had quite a lot on this week.

ehrql/backends/tpp.py
StevenMaude: Got it; thanks!

StevenMaude:
Thanks for the clarification!

PR #1480, opened 2023-08-07
fix: Restrict intermediate queries to population using JOIN

evansd:
Previously we ran every intermediate query against all the patients that happened to be in the table being queried and then only filtered the results down to the defined population when combining the results at the end.

This has the advantage of simplicity but it does mean the intermediate results tables are larger than they need to be and, potentially, the intermediate queries are slower.

This PR changes the logic so we first build a table containing just the patient IDs we want in our population and then use this as the left-most table in any subsequent joins.

There will be contexts in which this results in faster queries and contexts in which it results in slower ones. Our current hypothesis is that it will be a net performance benefit and that the worst case slowdowns will not be pathalogical. But until we test this on some realistic dataset definitions we won't know.

Note that the current implementation is not particularly lovely as it makes use of instance state to avoid passing the population table around, and thus necessitates some cleanup at the end. If the approach turns out to be a win then we can consider a cleaner factoring.

For interest, here is the original year-old Slack thread in which this approach was discussed:
https://bennettoxford.slack.com/archives/C03FB777L1M/p1657899866106729?thread_ts=1657897715.091099&cid=C03FB777L1M

### DIFF
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -52,6 +52,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
 
     global_unique_id: str
     counter = 0
+    population_table = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -77,18 +78,31 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         `get_setup_and_cleanup_queries` on the query object.
         """
         variable_definitions = apply_transforms(variable_definitions)
+
+        population_expression = self.get_predicate(variable_definitions["population"])
+        population_query = self.select_patient_id_for_population(population_expression)
+        population_query = population_query.where(population_expression)
+        population_query = apply_patient_joins(population_query)
+        population_table = self.reify_query(population_query)
+        # Store a reference to the population table so that all subsequent queries can
+        # make use of it
+        self.population_table = population_table
+
         variable_expressions = {
             name: self.get_expr(definition)
             for name, definition in variable_definitions.items()
             if name != "population"
         }
-        population_expression = self.get_predicate(variable_definitions["population"])
-        query = self.select_patient_id_for_population(population_expression)
+        query = sqlalchemy.select(population_table.c[0])
         query = query.add_columns(
             *[expr.label(name) for name, expr in variable_expressions.items()]
         )
-        query = query.where(population_expression)
         query = apply_patient_joins(query)
+
+        # Clear all shared state (yuck)
+        self.population_table = None
+        self.get_sql.cache_clear()
+        self.get_table.cache_clear()
 
         return query
 
@@ -116,8 +130,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
             ]
             # Create a table which contains the union of all these IDs. (Note UNION
             # rather than UNION ALL so we don't get duplicates.)
-            population_table = self.reify_query(sqlalchemy.union(*id_selects))
-            return sqlalchemy.select(population_table.c.patient_id)
+            all_ids_table = self.reify_query(sqlalchemy.union(*id_selects))
+            return sqlalchemy.select(all_ids_table.c.patient_id)
         elif len(tables) == 1:
             # If there's only one table then use the IDs from that
             return sqlalchemy.select(tables[0].c.patient_id.label("patient_id"))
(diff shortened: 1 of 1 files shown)


Jongmassey:
How would this behave with a population definition that references a computed column (e.g. _only include patients who have more than 5 prescriptions of this medication in the past year_ )? 

As I read it, the computed column that supports the population would be included in the `WHERE` clause of the `population_query` but not the `SELECT` list and would then be computed again later in a per-variable temp-table (assuming it was itself included as a variable).

evansd:
Queries that form part of the population definition won't benefit at all from this, they'll just get compiled exactly as they would before. But there shouldn't be any duplication: any temp tables created as part of the population query will get re-used if they're referenced again later.

Jongmassey:
>Queries that form part of the population definition won't benefit at all from this, 

Yes I guess that's perfectly logical.

> any temp tables created as part of the population query will get re-used if they're referenced again later

that's what I was worried about, brill

evansd:
I just tried re-running [this job](https://jobs.opensafely.org/the-healthcare-utilisation-of-people-with-long-covid-openprompt/healthcare_utilisation_openprompt/19635/k3tlmdgtryog65j7/) using the code on this branch. This is the kind of job for which we'd expect the most improvement because it has a tiny, easily defined population so it's in no sense representative but it's good to see what the best case looks like.

It's _quite_ good: total runtime dropped from 2 hours to 6 minutes.

evansd:
For the less happy path I chose [this job](https://jobs.opensafely.org/impact-of-covid-19-on-diabetes-care-in-england/covid-diabetes-care/19560/sbweasdv44dbmtt6/) which has a population definition of:
```py
dataset.define_population(
    (dataset.age < 110) & current_registration.exists_for_patient()
)
```

Resulting in a population of just under 26 million, which is generally as large as we get.

Previously this took 34 minutes to get to the point of downloading results; using the join on the population table it took 4 hours 30 minutes.

In particular, a couple of queries jumped from taking 3.5 minutes to taking 1 hour 15 minutes (query 005/011 and query 007/013 in the logs below). Both of these were queries on the `CodedEvent` table which had changed from (effectively):
```sql
SELECT ... FROM CodedEvent
```
To
```sql
SELECT ... FROM population_table
JOIN CodedEvent ON population_table.patient_id = CodedEvent.patient_id
```

Logs:
 * [query-timing-log-original.png](https://github.com/opensafely-core/ehrql/assets/19630/2940e526-bb63-4344-a254-b72bc2209cb7)
 * [query-timing-log-with-join.png](https://github.com/opensafely-core/ehrql/assets/19630/f8027d05-214e-4b3f-9c68-af6075b9e69c)
 * [query-original.sql.txt](https://github.com/opensafely-core/ehrql/files/12333690/query-original.sql.txt)
 * [query-with-join.sql.txt](https://github.com/opensafely-core/ehrql/files/12333694/query-with-join.sql.txt)

evansd:
Disappointed by the above, I tried a new approach which was, instead of a join, to add a WHERE clause like `patient_id IN (SELECT patient_id from #population_table)` to the relevant queries:

```diff
     def get_select_query_for_node_domain(self, node):
         """
         Given a many-rows-per-patient node, return the SELECT query corresponding to
         domain of that node
         """
         frame = get_domain(node).get_node()
         select_table, conditions = get_table_and_filter_conditions(frame)
         table = self.get_table(select_table)
         where_clauses = [self.get_predicate(condition) for condition in conditions]
+        if self.population_table is not None:
+            where_clauses.append(
+                table.c.patient_id.in_(
+                    sqlalchemy.select(self.population_table.c.patient_id)
+                )
+            )
         if where_clauses:
             query = query.where(sqlalchemy.and_(*where_clauses))
         return query
```

This is much more promising:

Job | Population | Original Runtime | With `JOIN` | With `IN (SELECT)`
-- | -- | -- | -- | --
[OpenPrompt](https://jobs.opensafely.org/the-healthcare-utilisation-of-people-with-long-covid-openprompt/healthcare_utilisation_openprompt/19635/k3tlmdgtryog65j7/) | 1,200 | 2 hours | 6 minutes | 8 minutes
[Diabetes](https://jobs.opensafely.org/impact-of-covid-19-on-diabetes-care-in-england/covid-diabetes-care/19560/sbweasdv44dbmtt6/) | 26 million | 30 minutes | 4.5 hours | 43 minutes

I wouldn't pay too much attention to the precise timings here as the server was at different levels of busyness and the OpenPrompt population increased between runs.

But it looks like this approach to filtering might give us some very significant speed-ups in the best case while not costing us too much in the worst case.

Jongmassey:
How about an `exists` rather than an `in` ? (sometimes can be better)

evansd:
Because Someone On the Internet had a post showing identical query plans for the `IN` and `EXISTS` versions and the `IN` was syntactically simpler. I can well believe they might have different plans under some circumstances though.

evansd:
Closing in favour of:
 * https://github.com/opensafely-core/ehrql/pull/1490

PR #1481, opened 2023-08-08
Improve performance of `medications` queries

evansd:
TPP provide a `MedicationDictionary` table which maps internal drug IDs to standard dm+d codes. Where this dictionary is incomplete we want to be able to supply custom entries. The facility to do so was added in:

 * https://github.com/opensafely-core/ehrql/pull/1348

Unfortunately the combination of the two joins and the COALESCE and NULLIF expressions caused abysmal query performance.

We can test this using a simple query template:
```sql
SELECT COUNT(*) FROM (
  /* table definition */
) AS t
WHERE t.date >= '20220101' AND t.date <= '20221231' AND t.dmd_code IN (
  /* prednisolone codelist */
)
```

Using the old, single-join table definition given here the query takes 48 seconds:
https://github.com/opensafely-core/ehrql/blob/488a9933425e7f4f169192c8563b2ae9f7954a58/ehrql/backends/tpp.py#L106-L112

Using the two-joins-and-coalesce approach the same query takes 22 minutes:
https://github.com/opensafely-core/ehrql/blob/a17cbeab3eecec2455a974e07dbfc6fa36df4094/ehrql/backends/tpp.py#L114-L122

The approach we take in this PR is to dynamically construct a single, combined dictionary table with NULL entries removed and then do a single join against that. It's not obvious that should perform better but it turns out that it does: the same query using this approach takes just over 1 minute.

Prednisolone codelist, for reference:
https://github.com/opensafely-core/ehrql/blob/a17cbeab/tests/acceptance/external_studies/openprompt-long-covid-vaccines/codelists/opensafely-asthma-oral-prednisolone-medication.csv

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -110,16 +110,40 @@ class TPPBackend(BaseBackend):
         temp_database_name = self.config.get(
             "TEMP_DATABASE_NAME", "PLACEHOLDER_FOR_TEMP_DATABASE_NAME"
         )
+
+        # We construct a medication dictionary table which combines the table provided
+        # by TPP (removing blank entries) with a custom dictionary we supply, taking
+        # care to remove any entries in our custom dictionary which are already defined
+        # in the TPP dictionary. If we didn't do this then duplicate entries would
+        # result in duplicate MedicationIssue rows when we do the join later.
+        #
+        # This query looks a bit gnarly, but MSSQL is sensible enough to just execute it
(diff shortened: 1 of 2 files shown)


inglesp (approved):

evansd:
@Jongmassey I don't think I would ever have considered this approach had you not suggested it, so a fair share of the credit/blame belongs on your plate!

Jongmassey:
I've thought some more about this last night and have an idea for maybe a cleaner query that could perform the same

PR #1482, opened 2023-08-08
Remove join between `Appointment` and `DataDictionary`

evansd:
This join (added in #1413) looks innocuous but it caused a catastrophic performance regression: queries which previously took 15 minutes were taking 11 hours.

The problem is compounded by the presence of the DISTINCT clause (which we hope is only temporary, see #1053) which seems to increase query times by 4-5x. However this is dwarfed by the join which increases query times by another 40-50x on top of that.

Replacing the join with a hardcoded case expression is only marginally slower than not including the `status` field at all so we do that instead.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -224,16 +224,24 @@ class TPPBackend(BaseBackend):
                 Patient_ID AS patient_id,
                 CAST(BookedDate AS date) AS booked_date,
                 CAST(StartDate AS date) AS start_date,
-                Description AS status
-            FROM Appointment AS appt
-            LEFT JOIN (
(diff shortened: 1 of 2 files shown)


inglesp (approved):
What are the risks that the hardcoded lookup gets out of date?

evansd:
I don't think they're that great. It's possible that TPP might add new status types in future but I can't see them reassigning codes.

evansd:
Note for future archaeologists: the reason the join performed so badly is that `Appointment.Status` is an int and `DataDictionary.Code` is a string. MSSQL performed the implicit type casting necessary to make this work as expected, but performance was abysmal.

PR #1483, opened 2023-08-09
fix: Make behaviour of `duration.ending_on` more intuitive

inglesp:
Specifically, `duration.ending_on(date)` now returns a list of intervals where the last interval ends on the given date.  Before this change, the last interval began on the given date.

ending_on is not currently used, so this change is safe to make:

https://github.com/search?q=org%253Aopensafely+ending_on


evansd:
Thanks for doing this. At the risk of being annoying, I'm not quite sure the implementation is correct though because I don't think it gives the right answer for e.g. `years(-3).starting_on("2020-01-01")` for which I think the old implementation _did_ give the right answer i.e. `starting_on` shouldn't give you intervals ending on that date just because you give it negative numbers.

evansd (approved):
On seconds thoughts, a simpler solution would just be to disallow negative arguments to `starting_on`/`ending_on`. If that turns out to be a thing people want we can work on supporting it. I just want to make sure we don't silently have weird behaviour.

For example:
```diff
@@ -777,7 +777,7 @@ class Duration:
 
         Useful for generating the `intervals` arguments to [`Measures`](#Measures).
         """
-        return self._generate_intervals(date, self.value, "starting_on")
+        return self._generate_intervals(date, self.value, 1, "starting_on")
 
     def ending_on(self, date):
         """
@@ -797,10 +797,10 @@ class Duration:
 
         Useful for generating the `intervals` arguments to [`Measures`](#Measures).
         """
-        return self._generate_intervals(date, -self.value, "ending_on")
+        return self._generate_intervals(date, self.value, -1, "ending_on")
 
     @classmethod
-    def _generate_intervals(cls, date, value, method_name):
+    def _generate_intervals(cls, date, value, sign, method_name):
         date = parse_date_if_str(date)
         if not isinstance(date, datetime.date):
             raise TypeError(
@@ -812,7 +812,11 @@ class Duration:
                 f"{cls.__name__}.{method_name}() can only be used with a literal "
                 f"integer value, not an integer series"
             )
-        return date_utils.generate_intervals(cls._date_add_static, date, value)
+        if value < 0:
+            raise ValueError(
+                f"{cls.__name__}.{method_name}() can only be used with positive numbers"
+            )
+        return date_utils.generate_intervals(cls._date_add_static, date, value * sign)
```

While we're modifying this it might be worth changing the docstring wording to make it clearer that this is only really useful for the measures framework and that confused Rose earlier and will no doubt confuse others.

evansd:
@inglesp Would you like me to finish this off on the lines suggested? I don't want this to languish too long.

inglesp:
Yes please, sorry to leave this dangling.

Issue #1444, opened 2023-07-21
Capture MSSQL query IO statistics


Jongmassey:
Following the changes to [cohort-extractor](https://github.com/opensafely-core/cohort-extractor/commit/057dcda640a633a8da8efe21c2a151b010ad62bb) and [ehrql ](https://github.com/opensafely-core/ehrql/commit/0e535f5ef4a3e002e7bc10b0bcc78e41a94f9a34) which removed the locking of the mssql system tables during the final results query select statements, this newfound level of concurrency resulted in an increase in storage requirements for the `OpenCORONATempTables` transaction log and [errors ](https://docs.google.com/document/d/1_0lAkLxMl4Hgf1W4MsMg1reIgOX5VSdnLxiZtcbpRjc/edit)caused by this file becoming full. 

We have permissions to query the size of this file, but do not have permissions to query how full it is. It is therefore challenging to know when a "transaction log full" error is forthcoming, or what is the proper level of concurrency for database jobs to reduce the risk of this happening.

We may be able to better estimate the most appropriate concurrency level by recording the IO statistics for completed queries of this type.

The first set of statistics which may help are enabled by the `SET STATISTICS IO` [option](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-statistics-io-transact-sql?view=sql-server-ver16), which doesn't tell us directly the transaction log usage for the statement, but the number of logical and physical reads made in the processing of the query which may be a reasonable proxy.

The second set are enabled by the `SET STATISTICS XML` [option](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-statistics-xml-transact-sql?view=sql-server-ver16). This option requires the `VIEWPLAN` permission, which we do not have on the `OpenCORONA` database but do have on the `OpenCORONATempTables` and `TempDB` table. We would need to subsequently parse this XML to extract the required information (e.g. number of rows, row size) which isn't directly available in the IO stats. Ad-hoc visualisation can be done with Brent Ozar's [Paste The Plan](https://www.brentozar.com/pastetheplan/).

both of these statistics come as "messages" alongside the result set and[ require a message handler to be registered](https://pymssql.readthedocs.io/en/stable/ref/_mssql.html?highlight=messages#_mssql.MSSQLConnection.set_msghandler) on the connection in `pymssql`

Jongmassey:
[...]

PR #1490, opened 2023-08-14
Restrict intermediate queries to population

evansd:
Previously we ran every intermediate query against all the patients that happened to be in the table being queried and then only filtered the results down to the defined population when combining the results at the end.

This has the advantage of simplicity but it does mean the intermediate results tables are larger than they need to be and, potentially, the intermediate queries are slower. It also means that the database has to evaluate the population inclusion logic at the same time as doing a very wide join across all the intermediate tables which creates a big burst of CPU pressure right at the final query.

This PR changes the logic so we first build a table containing just the patient IDs we want in our population and then use this to restrict the matching patients in any subsequent queries. The final wide join thus requires no WHERE clause.

Where the query population is a small fraction of the total population in the database this can result in some dramatic speedups. As the fraction of the population increases this improvement diminishes and where the population includes all (or almost all) possible patients including this condition will make things slower. Initial testing seems to show that the speedups are so significant, and the slowdowns rare and mild enough, that this is still a net benefit but we'll need to keep this under review.

For interest, here is the original year-old Slack thread in which this approach was discussed:
https://bennettoxford.slack.com/archives/C03FB777L1M/p1657899866106729?thread_ts=1657897715.091099&cid=C03FB777L1M

### DIFF
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -116,8 +116,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
             ]
             # Create a table which contains the union of all these IDs. (Note UNION
             # rather than UNION ALL so we don't get duplicates.)
-            population_table = self.reify_query(sqlalchemy.union(*id_selects))
-            return sqlalchemy.select(population_table.c.patient_id)
+            all_ids_table = self.reify_query(sqlalchemy.union(*id_selects))
+            return sqlalchemy.select(all_ids_table.c.patient_id)
         elif len(tables) == 1:
             # If there's only one table then use the IDs from that
             return sqlalchemy.select(tables[0].c.patient_id.label("patient_id"))
@@ -631,8 +631,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         domain of that node
         """
         frame = get_domain(node).get_node()
-        select_table, conditions = get_table_and_filter_conditions(frame)
-        table = self.get_table(select_table)
+        table_node, conditions = get_table_and_filter_conditions(frame)
+        table = self.get_table(table_node)
         where_clauses = [self.get_predicate(condition) for condition in conditions]
         query = sqlalchemy.select(table.c.patient_id.label("patient_id"))
         if where_clauses:
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -52,6 +52,7 @@ class BaseSQLQueryEngine(BaseQueryEngine):
 
     global_unique_id: str
     counter = 0
+    population_table = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -77,18 +78,38 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         `get_setup_and_cleanup_queries` on the query object.
         """
         variable_definitions = apply_transforms(variable_definitions)
+
+        # Generate a table containing the IDs all of patients matching the population
+        # definition
+        population_expression = self.get_predicate(variable_definitions["population"])
+        select_patient_id = self.select_patient_id_for_population(population_expression)
+        population_query = select_patient_id.where(population_expression)
+        population_query = apply_patient_joins(population_query)
+        population_table = self.reify_query(population_query)
+        # Store a reference to the population table so that we can use it while
+        # generating the variable expressions below
+        self.population_table = population_table
+
         variable_expressions = {
             name: self.get_expr(definition)
             for name, definition in variable_definitions.items()
             if name != "population"
         }
-        population_expression = self.get_predicate(variable_definitions["population"])
-        query = self.select_patient_id_for_population(population_expression)
+        query = sqlalchemy.select(population_table.c.patient_id)
         query = query.add_columns(
             *[expr.label(name) for name, expr in variable_expressions.items()]
         )
-        query = query.where(population_expression)
         query = apply_patient_joins(query)
+
+        # We use an instance variable to store the population table in order to avoid
+        # having to thread it through all our `get_sql`/`get_table` method calls. But
+        # this means that we can't safely re-use cached values across different calls to
+        # `get_query()` because the same query node may now generate different SQL
+        # depending on the population of the query to which it belongs. So we have to
+        # reset the caches and the population table reference.
+        self.population_table = None
+        self.get_sql.cache_clear()
+        self.get_table.cache_clear()
 
         return query
 
@@ -635,6 +656,23 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         table = self.get_table(table_node)
         where_clauses = [self.get_predicate(condition) for condition in conditions]
         query = sqlalchemy.select(table.c.patient_id.label("patient_id"))
+        # If we've already defined the population table (which we will have, other than
+        # when we're still in the middle of compiling the population query) then we can
+        # add an extra condition to our query which restricts it to just those
+        # patient_ids present in the population. Where the query population is a small
+        # fraction of the total population in the database this can result in some
+        # dramatic speedups. As the fraction of the population increases this
+        # improvement diminishes and where the population includes all (or almost all)
+        # possible patients including this condition will make things slower. Initial
+        # testing seems to show that the speedups are so significant, and the slowdowns
+        # rare and mild enough, that this is still a net benefit but we'll need to keep
+        # this under review.
+        if self.population_table is not None:
+            where_clauses.append(
+                table.c.patient_id.in_(
+                    sqlalchemy.select(self.population_table.c.patient_id)
+                )
+            )
         if where_clauses:
             query = query.where(sqlalchemy.and_(*where_clauses))
         return query
--- a/tests/integration/measures/test_calculate.py
+++ b/tests/integration/measures/test_calculate.py
@@ -10,6 +10,11 @@ from ehrql.tables import EventFrame, PatientFrame, Series, table
 @table
 class patients(PatientFrame):
     sex = Series(str)
+
+
+@table
+class addresses(EventFrame):
+    date = Series(date)
     region = Series(str)
 
 
@@ -26,6 +31,7 @@ def test_get_measure_results(engine):
     foo_event_count = events_in_interval.where(events.code == "foo").count_for_patient()
     had_event = events_in_interval.exists_for_patient()
     event_value = events_in_interval.value.sum_for_patient()
+    region = addresses.sort_by(addresses.date).last_for_patient().region
 
     intervals = years(3).starting_on("2020-01-01")
     measures = Measures()
@@ -41,7 +47,7 @@ def test_get_measure_results(engine):
         "foo_events_by_region",
         numerator=foo_event_count,
         denominator=event_count,
-        group_by=dict(region=patients.region),
+        group_by=dict(region=region),
         intervals=intervals,
     )
     measures.define_measure(
@@ -55,7 +61,7 @@ def test_get_measure_results(engine):
         "event_value_by_region",
         numerator=event_value,
         denominator=patients.exists_for_patient(),
-        group_by=dict(region=patients.region),
+        group_by=dict(region=region),
         intervals=intervals,
     )
     measures.define_measure(
@@ -64,7 +70,7 @@ def test_get_measure_results(engine):
         denominator=patients.exists_for_patient(),
         group_by=dict(
             sex=patients.sex,
-            region=patients.region,
+            region=region,
         ),
         intervals=intervals,
     )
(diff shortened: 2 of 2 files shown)


evansd:
Here are some initial performance comparisons:

Job | Population | Original Runtime | New Runtime
-- | -- | -- | --
[OpenPrompt 1](https://jobs.opensafely.org/the-healthcare-utilisation-of-people-with-long-covid-openprompt/healthcare_utilisation_openprompt/19635/k3tlmdgtryog65j7/) | 1,200 | 2 hours | 8 minutes
[OpenPrompt 2](https://jobs.opensafely.org/the-healthcare-utilisation-of-people-with-long-covid-openprompt/healthcare_utilisation_openprompt/19687/lrtplflb2dx3mick/) | 6,900 | 44 hours | 36 minutes
[Diabetes](https://jobs.opensafely.org/impact-of-covid-19-on-diabetes-care-in-england/covid-diabetes-care/19560/sbweasdv44dbmtt6/) | 26 million | 44 minutes | 57 minutes

As you can see, for jobs with small populations there is a very significant speed up.

For jobs with large populations – in this case `(age < 110) & current_registration.exists_for_patient()` – there is some slowdown, but bear in mind these are wall-clock times and subject to noise according to what else the database is having to do at the time. We should [improve our logging of query stats](https://github.com/opensafely-core/ehrql/issues/1444) so that we can make some direct comparisons of CPU cost.

It would also be good to parse population sizes out of our historical job logs so we can determine the distribution of population sizes which would help us estimate the impact (whether positive or negative) of this change.

A further option here is to be selective about when we enable this behaviour so that we only use it where we expect a positive impact. There are a few possibilities:

 * **Build the population table; count its rows; use that count to determine the best behaviour**
   This is the most correct approach, but adds significant architectural complexity. We'd no longer have separate compile and execute phases and, as determining the thresholds would be backend-specific, the query engine would need to involve the backend somehow in this logic.

 * **Use some heuristic for identifying queries we expect to benefit**
   The one that springs immediately to mind is "does the population definition restrict the population to an inline table?". This should be relatively easy to determine, and it's hard to see circumstances under which this obtains and the new query approach isn't a massive benefit. Of course, there may be other kinds of query which would also benefit and we'll miss those. But maybe that's OK for now? If we come across other queries which would benefit we can try to identify the patterns they identify and extent our heuristic accordingly.

 * **Allow the user to explicitly opt in to this behaviour somehow**
   This is in some ways the simplest, but I worry deeply about going down this road.

evansd:
I suppose I should have qualified that as "use some _purely syntactic_ heuristic" i.e. one that doesn't depend on what any of the tables actually contain. Both DBAs and query planners use table stats to guide them which this approach wouldn't be able to do. But it may well be good enough for our purposes in any case.

sebbacon:
Yes, that's how I understood it. I could have been clearer that I was commenting with respect to this:

> there may be other kinds of query which would also benefit and we'll miss those. But maybe that's OK for now? If we come across other queries which would benefit we can try to identify the patterns they identify and extent our heuristic accordingly.

I mean, that's all fine and normal, right?!

evansd:
Ah yes, got you – and yes, it's definitely normal but right now I don't feel like the word "fine" applies to anything to with databases :)

Jongmassey:
I'm not sure you've got enough actual CPU (not wall) time data to develop such a heuristic

evansd:
Well my first candidate was "if the population is restricted to a inline patient table then use the new approach". I think the improvements in that case are _so_ far above the noise floor that we don't need to worry about doing anything more accurate there.

But yes, if we want to start investigating other opportunities we're definitely going to need more and better data.

Of course, we don't even have enough data yet to know if we need to be selective about applying this approach. It could be, for all we know, that it's fine to use it all the time and it doesn't appreciably slow down even queries with large populations.

evansd:
Now we've got CPU time logging I was able to do a head to head comparison of the above `covid-diabetes-care` job using both the code on main and the code on this branch.

In wall-clock terms, the code on this branch was faster: 1hr 7m vs 1hr 40m.

In terms of aggregate CPU time, the code on this branch was _also_ faster: 10070.121s vs 11735.910s.

So, tentatively, I think that means all this discussion of when best to apply to approach is moot because we should just apply it all the time.

I'd like to do a more detailed comparison of the logs though to see how the performance of individual queries compares.

evansd:
I've managed to find some jobs with unusually large populations where this approach is slightly slower than previously.

In both cases, the population filter doesn't include any kind of practice registration requirement: one is defined as `(age <= 100) & (age >= 18) & sex.contains("male")` and the other simply as `age > 0`. This is quite unusual in my experience as generally registration history is key for knowing over what period a patient's records can be relied upon.

The comparative stats are:

Dataset | Pop. Size | Tables | Total time (before) | Total time (after) | Query time (before) | Query time (after) | CPU time (before) | CPU time (after)
-- | -- | -- | -- | -- | -- | -- | -- | --
[shielding](https://github.com/opensafely/shielding-evaluation/blob/c6e099f8ac2f4b0788f0343a72d6e5289e18deec/analysis/datasets.py) | 27 million | 55 | 1h 7m | 1hr 9m | 35m 55s | 39m 51s | 23,498s | 29,676s
[openprompt](https://github.com/opensafely/openprompt-hrqol/blob/a7499e62ee8096745048f723b58662caa794988e/analysis/define_openprompt_baseline.py) | 34 million | 10 | 15m | 16m | 1m 58s | 4m 21s | 708s | 637s

The difference between "total time" and "query time" is the time spent downloading results which in both cases dominates the total runtime meaning there's little overall difference from the user's perspective.

Wall-clock query times are slower in both cases under the new approach, but the CPU time picture is mixed: slightly faster in one case, significantly slower in the other.

On this basis, I'm minded to say that we should merge this PR as is and not try to get clever about when we do and don't apply it. There will be some cases like the above where we underperform, but I think these will be rare, the underperformance will be relatively mild, and I've got no confidence we could accurately identify them in advance.

ehrql/query_engines/base_sql.py
```diff
@@ -77,19 +78,39 @@ def get_query(self, variable_definitions):
...
+        # Generate a table containing the IDs all of patients matching the population
+        # definition
+        population_expression = self.get_predicate(variable_definitions["population"])
+        select_patient_id = self.select_patient_id_for_population(population_expression)
+        population_query = select_patient_id.where(population_expression)
+        population_query = apply_patient_joins(population_query)
+        population_table = self.reify_query(population_query)
+        # Store a reference to the population table so that we can use it while
+        # generating the variable expressions below
+        self.population_table = population_table
```
inglesp, line 91: I understand why you've done this, but making this class non-trivially stateful makes me a little bit uneasy.  This code's already quite demanding to grok, but at least it's more-or-less functional.  That said, I can't think of an improvement on your approach that doesn't involve lots of tedious plumbing, which would likely make the code harder to read.

ehrql/query_engines/base_sql.py
```diff
@@ -77,19 +78,39 @@ def get_query(self, variable_definitions):
...
 
+        # We use an instance variable to store the population table in order to avoid
+        # having to thread it through all our `get_sql`/`get_table` method calls. But
+        # this means that we can't safely re-use cached values across different calls to
+        # `get_query()` because the same query node may now generate different SQL
+        # depending on the population of the query to which it belongs. So we have to
+        # reset the caches and the population table reference.
+        self.population_table = None
+        self.get_sql.cache_clear()
+        self.get_table.cache_clear()
```
inglesp, line 112: None of the tests exercise the clearing of the caches, so if we add further cached single dispatch methods, there's a chance we'll miss this.  Can we try to hit this in a test somehow?  And can we add comments to `get_sql` and `get_table` about when the caches need to be cleared, as a clue to future us.

Or what about extracting most of the methods on `BaseSQLQueryEngine` into a `BaseSQLCompiler` class?  Then in `BaseSQLQueryEngine.get_query`, we instantiate a new compiler, and throw it away at the end of the method.

inglesp (approved):
> I'm minded to say that we should merge this PR as is and not try to get clever about when we do and don't apply it. 

Agreed, from your research this looks like a sensible approach.

Non-rhetorical question: do we exercise the population logic enough in tests?  It's obviously tested in every integration test, but in a very limited way.  And the generative tests always use a population of all patients in all tables.  Am not sure.

evansd:
Thanks @inglesp, all good questions as usual!

> making this class non-trivially stateful makes me a little bit uneasy

Yes, absolutely. Me too. I had a comment about this in the original PR description, which I managed to lose in the edits, that once we're confident this is a helpful change we should refactor it. But I was reluctant to put the work in to do that while there was still a chance we might want to back it all out.

> what about extracting most of the methods on `BaseSQLQueryEngine` into a `BaseSQLCompiler` class?

The trouble with this is that it makes the subclassing more awkward as now you've potentially got two classes you need to subclass and wire-up correctly.

I think the way to do this which best encapsulates the complexity without leaking it everywhere is to use the query model rewriting machinery to insert special new population filter operations into the query. Then the compilation methods go back to being pure functions of their query model inputs and everything becomes easier to reason about again.

I'm pretty sure that this can be made to work without _too_ much graph-theoretic gymnastics, but I haven't quite got the details straight in my head yet.

I should write a separate ticket for this so we don't lose track.

---

> None of the tests exercise the clearing of the caches

This actually isn't true, although it is true that there's no explicit tests for this. But the reason we need to worry about running multiple queries with a single engine instance is because that's what we do for measures. And if you remove either the cache clearing or the `population_table` NULLing lines then the measures tests fail.

So, arguably, while it's not great that the coverage is implicit here, it is testing the thing we actually care about for the reasons we actually care about it. And, as above, we'd like to factor the cache clearing away altogether in any case.

But I'm happy to add any extra tests you think would be worthwhile.

---

> Non-rhetorical question: do we exercise the population logic enough in tests? It's obviously tested in every integration test, but in a very limited way. And the generative tests always use a population of all patients in all tables. Am not sure.

No! And this is something I've been worrying about too. In fact, I did some exploratory work to look at what it would take to extend the generative tests in this way. It's not trivial for a few reasons. One is that not every BoolPatientSeries is a valid population definition (e.g. `~ons_deaths.exists_for_patient()` is not valid) so we need to make sure that Hypothesis can generate a high enough proportion of valid ones to not get stuck. The other is that it's very easy to generate population definitions which don't match any of the generated patients meaning that you're not actually testing the variable definition at all.

I think these are all surmountable. It will take some work but is absolutely worth doing. I'll write up a ticket for this too.

As for the correctness of this particular change, as part of the performance testing I've run it against real data using a diversity of dataset definitions and checked that the results matched the original results. So that, plus the fact that the change is relatively easy to reason about, gives me a tolerable level of confidence.

I'd definitely want Hypothesis testing in place before we start getting fancy with query model transforms though.

inglesp:
Thanks, that all sounds reasonable.

One thing:

> And if you remove either the cache clearing or the `population_table` NULLing lines then the measures tests fail.

They pass for me!

```
$ git diff
diff --git a/ehrql/query_engines/base_sql.py b/ehrql/query_engines/base_sql.py
index 37d9ec3d..38908feb 100644
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -108,8 +108,8 @@ class BaseSQLQueryEngine(BaseQueryEngine):
         # depending on the population of the query to which it belongs. So we have to
         # reset the caches and the population table reference.
         self.population_table = None
-        self.get_sql.cache_clear()
-        self.get_table.cache_clear()
+        # self.get_sql.cache_clear()
+        # self.get_table.cache_clear()
 
         return query
 
$ PYTHONHASHSEED=0 pytest -k measures
========================================== test session starts ==========================================
platform linux -- Python 3.11.2, pytest-7.4.0, pluggy-1.0.0
rootdir: /home/inglesp/work/ebmdatalab/ehrql
configfile: pyproject.toml
testpaths: tests
plugins: env-0.8.2, xdist-3.3.1, cov-4.1.0, hypothesis-6.82.4, mock-3.11.1
collected 2234 items / 2201 deselected / 33 selected                                                    

tests/docker/test_cli.py .                                                                        [  3%]
tests/integration/test_main.py ....                                                               [ 15%]
tests/integration/measures/test_calculate.py ....                                                 [ 27%]
tests/unit/test___main__.py .                                                                     [ 30%]
tests/unit/test_main.py ...                                                                       [ 39%]
tests/unit/measures/test_dummy_data.py ..                                                         [ 45%]
tests/unit/measures/test_measures.py ..................                                           [100%]

================================== 33 passed, 2201 deselected in 9.89s ==================================
```

evansd:
> They pass for me!

Gah, you're quite right! I could have sworn I tried commenting out the cache and `population_table` lines separately but maybe I didn't hit save at the right point before re-running the tests or something.

In any case, the issue is that the measures test isn't quite extensive enough. I've expanded it slightly and now it does fail if any one of those three lines is commented out.

evansd:
Have now run the newly extended generative tests against this branch locally using 10000 examples and haven't found any issues so I think we're good to go.

PR #1492, opened 2023-08-15
chore: Add build date and commit to Docker image

### LINKED ISSUE #1484: Docker images should be tagged with the git commit
This is useful for debugging. We do this when building images for other services e.g.
https://github.com/opensafely-core/job-server/blob/62bb8ab679bc19731291fe8c10f513c8dfa3ea99/docker/Dockerfile#L179-L184
```Dockerfile
# finally, tag with build information. These will change regularly, therefore
# we do them as the last action.
ARG BUILD_DATE=unknown
LABEL org.opencontainers.image.created=$BUILD_DATE
ARG GITREF=unknown
LABEL org.opencontainers.image.revision=$GITREF
```

It's not _essential_ as the images are tagged with version, for which there is a corresponding git tag. But it would be nice to have an immutable identifier here and nice to have things consistent across our images.

StevenMaude:
Fixes #1484.

These details are visible with `docker inspect`.

Based on:

opensafely-core/job-server@fbb5d1747263db429c367b4f8533b3307e8ac096

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -130,3 +130,11 @@ COPY ehrql /app/ehrql
 RUN python -m compileall /app/ehrql
 COPY databuilder /app/databuilder
 RUN python -m compileall /app/databuilder
+
+# The following build details will change.
+# These are the last step to make better use of Docker's build cache,
+# avoiding rebuilding image layers unnecessarily.
+ARG BUILD_DATE=unknown
+LABEL org.opencontainers.image.created=$BUILD_DATE
+ARG GITREF=unknown
+LABEL org.opencontainers.image.revision=$GITREF
--- a/Justfile
+++ b/Justfile
@@ -129,8 +129,11 @@ build-ehrql:
     #!/usr/bin/env bash
     set -euo pipefail
 
+    export BUILD_DATE=$(date -u +'%y-%m-%dT%H:%M:%SZ')
+    export GITREF=$(git rev-parse --short HEAD)
+
     [[ -v CI ]] && echo "::group::Build ehrql (click to view)" || echo "Build ehrql"
-    DOCKER_BUILDKIT=1 docker build . -t ehrql-dev
+    DOCKER_BUILDKIT=1 docker build --build-arg BUILD_DATE="$BUILD_DATE" --build-arg GITREF="$GITREF" . -t ehrql-dev
     [[ -v CI ]] && echo "::endgroup::" || echo ""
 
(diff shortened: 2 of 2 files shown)


StevenMaude:
```sh
$ docker inspect ghcr.io/opensafely-core/ehrql:dev
```
```json
            "Labels": {
                "org.opencontainers.image.authors": "tech@opensafely.org",
                "org.opencontainers.image.created": "23-08-15T10:25:11Z",
                "org.opencontainers.image.description": "ehrQL action for opensafely.org",
                "org.opencontainers.image.ref.name": "ubuntu",
                "org.opencontainers.image.revision": "04711b33",
                "org.opencontainers.image.source": "https://github.com/opensafely-core/ehrql",
                "org.opencontainers.image.title": "ehrql",
                "org.opencontainers.image.url": "opensafely.org",
                "org.opencontainers.image.vendor": "OpenSAFELY",
                "org.opencontainers.image.version": "20.04",
                "org.opensafely.action": "ehrql",
                "org.opensafely.base.build-date": "23-08-14T04:22:40.291Z",
                "org.opensafely.base.vcs-ref": "5742c38"
            }
```

iaindillingham (approved):

evansd:
I don't think this tags the prod image?
https://github.com/opensafely-core/ehrql/blob/fd51ded8aa332dd4bfdef2e2f275df749bbd86c9/.github/workflows/build-and-deploy.yml#L104-L111

iaindillingham:
Oh, that's annoying: I was worried about not tagging the prod image but checked only the Justfile; I didn't think of checking the GH actions. @StevenMaude, I think Dave is correct.

StevenMaude:
I think there might be confusion over "tag" versus "label".

As far as I can tell, the example in #1484 only adds the commit ID as a label only, not as an image tag. But also refers to "tag" in the comment of that example's Dockerfile, even though it's an image label.

So, there's no image tag added here, but we could do so.

evansd:
> I think there might be confusion over "tag" versus "label"

Sorry, yes, you're quite right I was confusing those two and we just want a label here, not a tag.

But I don't think the label's being applied either – unless I'm missing something – because the workflow which builds the prod image doesn't pass the relevant arguments to the `docker build`.

StevenMaude:
Right, now I understand!

That's a consequence of us not using the Justfile in GitHub Actions, and my assumption we did.

PR #1498, opened 2023-08-15
Clarify behaviour of omitting output option

StevenMaude:
This was discussed in Slack:

https://bennettoxford.slack.com/archives/C02HJTL065A/p1692101347621419?thread_ts=1692095561.720749&cid=C02HJTL065A

### DIFF
--- a/docs/tutorial/running-ehrql.md
+++ b/docs/tutorial/running-ehrql.md
@@ -242,6 +242,9 @@ to run the dataset definition.
     * :notepad_spiral: If the `--dummy-tables` option is omitted,
       randomly generated data will be used instead.
 * `--output output/dataset.csv` specifies the path to the output CSV file.
+    * :notepad_spiral: If the `--output` option is omitted,
+      the output will be not be saved to a file,
+      but displayed on screen.
 
 #### What you should see when you run the command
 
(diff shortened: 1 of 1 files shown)


sebbacon:
I wonder if it belongs [here](https://docs.opensafely.org/ehrql/explanation/output-formats/) too?

sebbacon (approved):
Definitely an improvement.

As a random aside, I'm not a fan of semantic linebreaks (presumably that's what I can see here?). Personal taste I guess but I find it harder to read, not easier.

StevenMaude:
> semantic linebreaks (presumably that's what I can see here?)

Yes, that's what this is.

I didn't like it that much at first, and now I do more, but I'm not really tied to it.

StevenMaude:
There's a bit of a mix of semantic and non-semantic line breaks throughout the ehrQL documentation.

I think the actual benefit is actual more for reading diffs, than when adding new lines; like this [good example](https://rhodesmill.org/brandon/2012/one-sentence-per-line/).

StevenMaude:
I'll merge for now anyway, because of this mixture we have.

StevenMaude:
> I wonder if it belongs [here](https://docs.opensafely.org/ehrql/explanation/output-formats/) too?

I didn't see that :man_facepalming: — but it's a very good suggestion, so I'll add in another PR.

PR #1500, opened 2023-08-15
Automatically apply T1OO exclusions

evansd:
To facilitate deployment we temporarily set the default configuration to `True`. A subsequent PR reverses this:
 * https://github.com/opensafely-core/ehrql/pull/1588

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -1,8 +1,11 @@
+from urllib import parse
+
 import ehrql.tables.beta.core
 import ehrql.tables.beta.smoketest
 import ehrql.tables.beta.tpp
 from ehrql.backends.base import MappedTable, QueryTable, SQLBackend
 from ehrql.query_engines.mssql import MSSQLQueryEngine
+from ehrql.query_model import nodes as qm
 
 
 class TPPBackend(SQLBackend):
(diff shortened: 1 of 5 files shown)


inglesp (approved):
Very neat.  I can't think of a way that this could be circumvented by a user, but for something so important, the single test feels a bit light.  Should we be more thorough in testing here?

evansd:
> Should we be more thorough in testing here?

What would that look like?

inglesp:
> > Should we be more thorough in testing here?
> 
> What would that look like?

I don't know!  Something something Hypothesis?  I can't see what value it would give though, beyond extra belts to our braces.

ehrql/backends/tpp.py
```diff
@@ -22,6 +25,62 @@ class TPPBackend(BaseBackend):
...
+        parts = parse.urlparse(dsn)
+        params = parse.parse_qsl(parts.query, keep_blank_values=True)
+
+        include_t1oo_values = []
+        other_params = []
+        for key, value in params:
+            if key == "opensafely_include_t1oo":
+                include_t1oo_values.append(value)
+            else:
+                other_params.append((key, value))
```
rebkwok, line 44: `parse_qs` would give us the list of `opensafely_include_t1oo` values without having to loop through `params`

```
In [1]: url = "mssql://foo/db?opensafely_include_t1oo=false&opensafely_include_t1oo=false&param1=1&param1=3&param2="
In [2]: parts = urlparse(url)
In [3]: params = parse_qs(parts.query, keep_blank_values=True)
In [4]: params
{'opensafely_include_t1oo': ['false', 'false'],
 'param1': ['1', '3'],
 'param2': ['']}
include_t1oo_values = params.pop("opensafely_include_t1oo", [])
```

Of course, then we'd still need to loop to get tuple pairs for `other_params`, so maybe it doesn't make much difference. It could be a one-liner, but I'm never sure nested list comprehensions are worth it for readability.  It's a simple one though, so :shrug: 
```
other_params =  [(k, val) for k, vals in params.items() for val in vals]
```

ehrql/backends/tpp.py
```diff
@@ -22,6 +25,62 @@ class TPPBackend(BaseBackend):
...
+
+        if len(include_t1oo_values) == 1:
+            self.include_t1oo = include_t1oo_values[0].lower() == "true"
+        elif len(include_t1oo_values) != 0:
+            raise ValueError(
+                "`opensafely_include_t1oo` parameter must not be supplied more than once"
```
rebkwok, line 50: Do we care if it's supplied more than once if the value is the same each time?

rebkwok (approved):

ehrql/backends/tpp.py
evansd: Interestingly, I originally used a `set` here for exactly this behaviour. But then I thought we're in control of this URL and there's no good reason we should be setting it twice and if we do the resulting URL is likely to be confusing, so it would be better to be strict.

ehrql/backends/tpp.py
evansd: I think you can actually pass the results from `parse_qs` directly into `urlencode` so we wouldn't even need to unpack. My silly reason for not doing that was that it wouldn't preserve parameter order for repeated params and I was obsessing over returning the URL as close to how it was supplied as possible.

But now I think about it that really _is_ a silly reason. I'll change it to your approach.

rebkwok (approved):

PR #1507, opened 2023-08-18
Log MSSQL queries along with timing and IO stats

### LINKED ISSUE #1451: Production logging of SQL and performance
Cohortextractor genereates detailed logs for every query in production, including:

- Formatted SQL (truncated if needed)
- Execution time

This is fairly essential to diagnosing issues in production, as we have *zero* visibility into the DB itself.  We need similar logging output for ehrql.

Possibly related to #1444.

evansd:
This logs each setup/cleanup query executed by the MSSQL engine, along with total query time and stats gathered using `SET STATISTICS TIME/IO ON`.

Note that while we log the query we choose not to log the parameters at this stage for a few reasons:
 * there's a risk of leaking patient data into the logs;
 * the parameters can be _very_ verbose;
 * the structure of the query is generally the most interesting thing;
 * the full query is always recoverable afterwards using `dump-dataset-sql` locally.

We can revisit this if necessary later.

I have tried to make the logs as easy as possible to parse visually (for the stressed support person looking for smoking guns) while making sure they're reasonably easy to parse programmatically if we want to gather aggregate timing data or something like that.

Example below (though the timings are all zero because its from an empty local db):

Screenshot from 2023-08-18 10-57-00

Closes #1451 #1444

### DIFF
--- a/ehrql/query_engines/mssql.py
+++ b/ehrql/query_engines/mssql.py
@@ -6,6 +6,7 @@ from sqlalchemy.sql.functions import Function as SQLFunction
 
 from ehrql.query_engines.base_sql import BaseSQLQueryEngine, apply_patient_joins
 from ehrql.query_engines.mssql_dialect import MSSQLDialect, SelectStarInto
+from ehrql.utils.mssql_log_utils import execute_with_log
 from ehrql.utils.sqlalchemy_exec_utils import (
     ReconnectableConnection,
     execute_with_retry_factory,
@@ -205,7 +206,7 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
         with ReconnectableConnection(autocommit_engine) as connection:
             for i, setup_query in enumerate(setup_queries, start=1):
                 log.info(f"Running setup query {i:03} / {len(setup_queries):03}")
-                connection.execute(setup_query)
+                execute_with_log(connection, setup_query, log.info)
 
             # Re-establishing the database connection after an error allows us to
             # recover from a wider range of failure modes. But we can only do this if
@@ -236,7 +237,7 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
 
             for i, cleanup_query in enumerate(cleanup_queries, start=1):
                 log.info(f"Running cleanup query {i:03} / {len(cleanup_queries):03}")
-                connection.execute(cleanup_query)
+                execute_with_log(connection, cleanup_query, log.info)
 
     # implement the "table value constructor trick"
     # https://stackoverflow.com/questions/71022/sql-max-of-multiple-columns/6871572#6871572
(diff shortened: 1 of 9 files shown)


bloodearnest:
I appreciate the human readable formatting, but it does make it a bit difficult to grep.

With chortextractor, we had a query id to link multiple log output lines together, so you could do things like grep "query=n" logs.txt.

Any chance you could add similar to each stats line? like "setup query n/m"? Then a grep might catch them all together? Also, if we want to grep and sort the slowest querys, we can do that and then list the "query id" to go looking manually in the logs at the sql.


Also, any chance we could add a cpu_ms/elapsed_ms value to the timings output? grep can't do maths, but I think queries where that number is high would be good to find easily.

evansd:
> Any chance you could add similar to each stats line? like "setup query n/m"? 

Good idea. Done and screenshot updated.

> Also, any chance we could add a cpu_ms/elapsed_ms value to the timings output?

Do you mean the sum of parse and exec? If so I'm less sure about this as I'm not sure that number is terribly interesting: parse times are so utterly dwarfed by exec times in production that I can't think when we'd care about them.

bloodearnest:
> > Also, any chance we could add a cpu_ms/elapsed_ms value to the timings output?
> 
> Do you mean the sum of parse and exec? If so I'm less sure about this as I'm not sure that number is terribly interesting: parse times are so utterly dwarfed by exec times in production that I can't think when we'd care about them.


No, I mean `cpu time / elapsed time`. It gives an indication of high cpu usage queries if high, or of high cpu contention with other if lower and long elapsed.

In cohortextractor logs, you can see ratios or ~30:1 for some queries.

It's the only potential indication we've got of high cpu usage for a particular query (*after its completed, anyway).

evansd:
image

What am I missing? Or what are you missing? :)

(Ignoring the fact that these are all zero in the example.)

evansd:
Ohhhh, ha! I've just realised you mean literally one divided by the other. I'm so used to read "a / b time" as a shorthand for "a time" or "b time".

Jongmassey:
Wow, this pretty heroic - I hadn't expected it to be quite this far-reaching! I'm slightly struggling to see where in the compilation of `InsertMany` we prevent the `VALUES` bit of the SQL being logged?

evansd:
@bloodearnest Another good suggestion (once I actually understood what you were saying!). Have pushed and updated the screenshot.

evansd:
> I'm slightly struggling to see where in the compilation of `InsertMany` we prevent the `VALUES` bit of the SQL being logged?

It's not so much that we _prevent_ it being logged as we don't go out of our way to do something special to get it logged. We end up going down the first branch here, and just compiling a plain insert statement. The gnarly bit is what we have to do if we're trying to produce fully compiled SQL strings, for which there's no directly analogue to `InsertMany`.
https://github.com/opensafely-core/ehrql/blob/df11e81dea373c8ba12822a6b1d50261b5a09cd3/ehrql/utils/sqlalchemy_query_utils.py#L220-L241

bloodearnest (approved):

bloodearnest:
approve from a "what it outputs" perspective!

Jongmassey (approved):

PR #1511, opened 2023-08-21
Update the TPP schema

iaindillingham:
7782930 and b82af45 remove modifications, which have been incorporated in the [OpenSAFELY-TPP database schema][1] report. c882f5b calls `just update-tpp-schema`. Finally, aac4616 fixes a bug in the TPP backend which emerged as a consequence of the previous commits. It implies that the `clinical_events` ehrQL table has not been used to date, or that the change occurred recently, because in both cases, a query against this table would have errored.

[1]: https://reports.opensafely.org/reports/opensafely-tpp-database-schema/


evansd:
>  in both cases, a query against this table would have errored.

MSSQL is case-insensitive by default (putting the Microsoft in SQL there) so these queries will have worked OK, but better to have the case matching for the sake of sanity.

iaindillingham:
Well, who'd have thought! Thanks for mentioning it, @evansd. It seemed odd that this column hadn't been accessed before now.

PR #1512, opened 2023-08-22
Generative test improvements

### LINKED ISSUE #1458: Tests are running slower than they used to
On GitHub, these now take about 20 minutes, when they used to take about 6 minutes.

If this is known and acceptable, then that's fine and this issue can be closed.

I've not tested anything much yet. But from skimming the [CI run logs](https://github.com/opensafely-core/ehrql/actions/workflows/main.yml), it looks like it might be the merge of #1236 that has led to this increase.

Edit: it's the additional generative tests introduced in 7e1fb3652c38587280088a1975c836d7e574494f that are the source.

evansd:
This PR separates out checking that we can generate examples for every query model operation from actually running tests using those examples. This means that the comprehensiveness check is now fast enough (2 seconds) to run locally by default and, as a bonus, it now shows up as a normal test failure rather than a weird error in the teardown phase.

We also reduce the default maximum depth of the query tree in generated examples as this makes a huge difference to how long test generation takes. The scheduled workflow, where the serious generative testing happens, restores the depth to its original value.

We also make derandomisation the default setting, which gives better parity between CI and local tests, and then explicitly opt in to randomisation in `just generative-tests` and in the scheduled workflow.

(As an aside, I think we might want to look at lowering the max depth value somewhat even in the scheduled tests. The kinds of high-depth examples Hypothesis generates tend not to be very interesting (e.g. massive stacks of nested date operations). What we really want it to do is generate as many different compositions of types of operation as possible; and it may be that a larger number of lower-depth examples would achieve that better.)

Closes #1458

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -204,7 +204,7 @@ test-generative *ARGS: devenv
 
     examples=${GENTEST_EXAMPLES:-200}
     [[ -v CI ]] && echo "::group::Run tests (click to view)" || echo "Run tests"
-    GENTEST_EXAMPLES=$examples GENTEST_COMPREHENSIVE=t $BIN/python -m pytest \
+    GENTEST_EXAMPLES=$examples $BIN/python -m pytest \
         --cov=ehrql \
         --cov=tests \
         --cov-report=html \
--- a/Justfile
+++ b/Justfile
@@ -197,6 +197,7 @@ test-unit *ARGS: devenv
 # Set GENTEST_MAX_DEPTH to change the depth of generated query trees.
 test-generative *ARGS: devenv
     GENTEST_EXAMPLES=${GENTEST_EXAMPLES:-200} \
+    GENTEST_RANDOMIZE=${GENTEST_RANDOMIZE:t} \
       $BIN/python -m pytest tests/generative {{ ARGS }}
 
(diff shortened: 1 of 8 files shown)


Justfile
```diff
@@ -189,27 +189,30 @@ test-unit *ARGS: devenv
...
+# Run the generative tests only, configured to use more than the tiny default
+# number of examples. Optional args are passed to pytest.
 #
 # Set GENTEST_DEBUG env var to see stats.
 # Set GENTEST_EXAMPLES to change the number of examples generated.
+# Set GENTEST_MAX_DEPTH to change the depth of generated query trees.
 test-generative *ARGS: devenv
-    $BIN/python -m pytest tests/generative {{ ARGS }}
+    GENTEST_EXAMPLES=${GENTEST_EXAMPLES:-200} \
+    GENTEST_RANDOMIZE=${GENTEST_RANDOMIZE:t} \
```
rebkwok, line 200: This one isn't documented (there's a mention of randomizing with `just test-generative`, and it's unlikely to be something we want to turn on and off locally, but since it can be specified as an env var, I think it'd be worth adding it to the developer docs

rebkwok (approved):
This is much nicer, thanks! Just a comment on documentation.

PR #1513, opened 2023-08-22
fix: Fix formatting of tables with large numbers

evansd:
I didn't anticipate quite how high some of these IO stats would get in production!


tests/unit/utils/test_mssql_log_utils.py
```diff
@@ -118,7 +118,9 @@ def test_format_table_io():
             "scans": 1,
         },
         "Worktable": {
-            "lob_logical": 0,
+            # Check we correctly handle the case when the figure is wider than the
+            # header (which does actually happen in production!)
+            "lob_logical": 1_00_000_000_000,
```
inglesp, line 123: centinanonit: this looks odd to me -- I wouldn't have expected the initial underscore.

inglesp (approved):

tests/unit/utils/test_mssql_log_utils.py
evansd: Oh yes, quite right.

PR #1517, opened 2023-08-23
docs: Remove construction warnings

iaindillingham:
Anecdotally, we know that construction warnings worry users: "What does 'construction' mean?", "Is everything going to break?" Furthermore, they fall into two categories:

* TODOs ("This will be improved in the future")
* TPP-specific examples

Arguably, we see little benefit from either category. The second is often redundant, as adjacent headings signal that examples are TPP-specific.

### DIFF
--- a/docs/introduction/getting-help.md
+++ b/docs/introduction/getting-help.md
@@ -1,6 +1,3 @@
-:construction: We're still working on ehrQL and its documentation:
-**hearing what you find tricky to do or understand will help us improve ehrQL**.
-
 If you are confused or stuck, muddled, addled, befuddled or bemused,
 and cannot easily find what you are looking for in this documentation, then:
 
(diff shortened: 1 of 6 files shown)


docs/introduction/getting-help.md
```diff
@@ -1,6 +1,3 @@
-:construction: We're still working on ehrQL and its documentation:
-**hearing what you find tricky to do or understand will help us improve ehrQL**.
-
```
inglesp, line 3: I think this comment still provides value, in that it contextualises the rest of this page.

inglesp (approved):
Possibly worth keeping the one example I've mentioned, but up to you.

docs/introduction/getting-help.md
iaindillingham: I disagree. The context of this page is how the user can get help from us; this comment is about how we can get help from the user.

PR #1518, opened 2023-08-23
fix: Increase number of examples used in generative tests comprehensive check

rebkwok:
The previous number of examples was mostly allowing this comprehensive check to pass, but it was quite close to the minimum needed, so would fail on the occasional local run. Increasing it to 250 adds around 1.5s to the time the tests take to run, but should make the test more robust.


evansd:
In one sense we don't need this now we've got the fix in #1533. We obviously are quite close the line of not generating enough examples though so I can well imagine that next time we change something in the generative tests we'll have to increase the example count. Probably worth waiting until that happens though, and then we can work out how much we actually need to increase it by.

PR #1520, opened 2023-08-23
docs: Explain each of Diátaxis modes

iaindillingham:
Diátaxis has four modes of documentation. Here, we give each mode its own landing page, which we link to from ehrQL's landing page. In each case, we explain the mode of documentation. We are careful to distinguish "studying" from "working with"; and "practical steps" from "theoretical knowledge".

Notice that the navigation panel has become slightly more user-friendly: clicking "Tutorial" now takes the user to the tutorial's landing page, for example. The landing pages don't, at present, do more than mirror the navigation panel. This is intentional and will change in time.

We don't touch the introduction. Again, this is intentional.

### DIFF
--- a/docs/explanation/index.md
+++ b/docs/explanation/index.md
@@ -1,0 +1,5 @@
+The [explanation](explanation/index.md) provides theoretical knowledge for studying ehrQL.
+
+* [ehrQL backend tables](explanation/backend-tables.md)
+* [ehrQL output formats](explanation/output-formats.md)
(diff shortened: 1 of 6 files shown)


iaindillingham:
I guess I've taken it as a given that we accept the Diátaxis approach. It might be that the practical steps/theoretical knowledge axis isn't useful to us; if so, then we can remove it (and let Diátaxis know!).

PR #1522, opened 2023-08-23
Document clinical events table

### LINKED ISSUE #1373: Document clinical_events table

rebkwok:
fixes #1373

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -89,10 +89,7 @@ class TPPBackend(BaseBackend):
         """
             SELECT
                 Patient_ID AS patient_id,
-                CASE
-                    WHEN ConsultationDate = '9999-12-31T00:00:00' THEN NULL
-                    ELSE CAST(ConsultationDate AS date)
-                END AS date,
+                CAST(NULLIF(ConsultationDate, '9999-12-31T00:00:00') AS date) AS date,
                 NULL AS snomedct_code,
                 CTV3Code AS ctv3_code,
                 NumericValue AS numeric_value
@@ -100,10 +97,7 @@ class TPPBackend(BaseBackend):
             UNION ALL
             SELECT
                 Patient_ID AS patient_id,
-                CASE
-                    WHEN ConsultationDate = '9999-12-31T00:00:00' THEN NULL
-                    ELSE CAST(ConsultationDate AS date)
-                END AS date,
+                CAST(NULLIF(ConsultationDate, '9999-12-31T00:00:00') AS date) AS date,
                 ConceptId AS snomedct_code,
                 NULL AS ctv3_code,
                 NumericValue AS numeric_value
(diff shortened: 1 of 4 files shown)


ehrql/backends/tpp.py
```diff
@@ -89,15 +89,21 @@ class TPPBackend(BaseBackend):
         """
             SELECT
                 Patient_ID AS patient_id,
-                CAST(ConsultationDate AS date) AS date,
+                CASE
+                    WHEN ConsultationDate = '9999-12-31T00:00:00' THEN NULL
+                    ELSE CAST(ConsultationDate AS date)
+                END AS date,
```
evansd, line 95: Since the events of the last few weeks, I'm now super nervous about changes to the query structure of large, widely-used tables like this. I'm not saying this one in particular is likely to cause problems. But I'd like us to do at least some basic performance testing on this query to make sure it performs OK when filtering by date.

There are also a few equivalent ways of writing this which it would be good to compare the performance of e.g.
```
CAST(NULLIF(ConsultationDate, '9999-12-31T00:00:00')) AS date
```

Doing this is painful in the sense that it requires manually constructing SQL queries in the console and using `SET STATISTICS TIME ON` to compare CPU times (the only metrics we have access to). But I don't see an alternative at the moment.

rebkwok: @evansd Based on investigation in [#1523](https://github.com/opensafely-core/ehrql/issues/1523#issuecomment-1706844168) it looks like NULLIF may be marginally more efficienct, so I've updated it to use that. And also as discussed there, it doesn't seem like this change is going to make things worse

ehrql/tables/beta/tpp.py
```diff
@@ -33,6 +33,20 @@
 
 @table
 class clinical_events(EventFrame):
+    """
+    Each record corresponds to a single clinical or consultation event for a patient.
+
+    Each record contains either a SNOMED-CT code or a CTv3 code (never both). Note that
+    date may be null.
```
evansd, line 40: This is true, but I'm not sure it quite captures the weird way this table is structured.

I might say something like this:

> Each event is recorded twice: once with a CTv3 code, and again with the equivalent SNOMED-CT code. Each record will have only one of the `ctv3_code` or `snomedct_code` columns set and the other will be null. This allows you to query the table using either a CTv3 codelist or SNOMED-CT codelist and all records using the other coding system will be effectively ignored.

That's maybe not the clearest wording, but hopefully you see what I'm getting at.

evansd (approved):

ehrql/tables/beta/tpp.py
rebkwok: Ah TIL that I didn't understand the underlying tables anyway.

rebkwok: I spent a while trying to figure out how to make the wording clearer, but I couldn't come up with anything better, so just going with your suggested wording :)

PR #1529, opened 2023-08-24
Disable writing results to temporary database

evansd:
The only reason we were doing this was to give us robustness against connection errors (the dreaded "DBPROCESS is dead or not enabled"). Having the results in a persistent table allowed us to close the connection, reconnect, and continue downloading results in a way which wouldn't be possible with session-scoped temporary tables.

However, while we still see this error frequently with Cohort Extractor, we have never once seen it with ehrQL. I initially thought this might be luck, but ehrQL has seen enough use at this point that that seems extremely unlikely and instead that _something_ in the ehrQL's connection stack fixes the underlying issue. (We can speculate on what that might be in the comments.)

Apart from being unnecessary, writing results to the temporary database causes other problems: it acquires a global lock which means only one query at a time can compile its results; it bloats the transaction log on the temporary database such that if we work around the locking problem we end up filling the transaction log and blowing up.

While fixing this issue in ehrQL doesn't solve the problem for Cohort Extractor it still has benefits:
 * ehrQL jobs no longer compete with each other or with Cohort Extractor jobs for a shared lock;
 * writing results to a session-scoped table doesn't create the same transaction log pressure as writing to the temporary database and should generally place less load on the database.

Obviously making changes of this kind is likely to cause a certain nervousness. But this change doesn't involve doing a new sort of query – it just involves doing more of what we're doing already. And we've been happily writing massive amounts of data to session-scoped tables in parallel for years without issue.

Nevertheless, out of an abundance of caution/institutional scar tissue from when we tried to fix the locking problem, I'm only disabling the behaviour here and leaving all the supporting code and tests in place. The idea being that it will be easy to revert this change cleanly if we suddenly realise there's a problem.

A [follow-up PR][PR] which actually removes all the related code will be left in draft until we're confident things are working.

[PR]: https://github.com/opensafely-core/ehrql/pull/1524

### DIFF
--- a/ehrql/query_engines/mssql.py
+++ b/ehrql/query_engines/mssql.py
@@ -153,7 +153,12 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
         # Write results to a temporary table and select them from there. This allows us
         # to use more efficient/robust mechanisms to retrieve the results.
         table_name, schema = self.get_results_table_name_and_schema(
+            # This is a minimally invasive way to disable the persistent results table
+            # in case we want to re-enable it in a hurry. We need the toggle so that
+            # tests can use it to retain coverage.
             self.config.get("TEMP_DATABASE_NAME")
+            if self.config.get("PERSIST_RESULTS_TABLE")
+            else None
         )
         results_table = temporary_table_from_query(
             table_name, results_query, index_col="patient_id", schema=schema
--- a/tests/integration/query_engines/test_mssql.py
+++ b/tests/integration/query_engines/test_mssql.py
@@ -42,7 +42,10 @@ def test_get_results_using_temporary_database(mssql_engine, temp_database_name):
 
         results = mssql_engine.extract_qm(
             variable_definitions,
-            config=dict(TEMP_DATABASE_NAME=temp_database_name),
+            config=dict(
+                TEMP_DATABASE_NAME=temp_database_name,
+                PERSIST_RESULTS_TABLE="t",
+            ),
         )
 
         assert results == [


sebbacon:
> We can speculate on what that might be in the comments

Go on, then!

sebbacon:
I approve of this approach. Is it worth writing a little more detail about a plan for monitoring it?

* What sorts of errors might we be looking out for?
* Are there performance smoke tests we can run (or even select a couple of in-flight projects to re-run the extraction for before/after)?
* How can we communicate this to tek-support?
* How long should we run the experiment for, before deciding to remove the code?

evansd:
As to why ehrQL might not hit the same problem Cohort Extractor does, I still don't know.

The relevant differences between the two are (as far as I can make out):

 * Python version: `3.8.10` vs `3.11.4`

 * pymssql version: `2.2.3` vs `2.2.7`

 * version of FreeTDS library that comes bundled in the pymssql wheel: `1.3.4` vs `1.3.13`
    ```
   python -c 'from pymssql import _mssql; print(_mssql.get_dbversion())'
   ```

 * ehrQL uses SQLAlchemy to manage the connection whereas Cohort Extractor uses the DBAPI object directly

There's no screamingly obvious smoking gun in the changes for either pymssql or FreeTDS:
https://github.com/pymssql/pymssql/compare/v2.2.3...v2.2.7
https://github.com/FreeTDS/freetds/compare/v1.3.4...v1.3.13

Quite possibly it's the Python version but I wouldn't know where to start looking for that.

It strikes me now that it might be worth looking at upgrading the Python version (and pymssql version) for Cohort Extractor. If it makes the connection issue go away that it avoids the database locking problem entirely.

evansd:
Yes, I very much agree. The issue is more where and when to make the announcement so that we can be confident everyone has seen it and will remember. I'd suggest doing it at the point we decide to deploy. The announcement itself can include the instructions (revert the commit), I don't think there's enough to warrant a separate guide.

evansd:
For reasons which we're yet to fully understand, in ehrQL the final join query is _not_ one of the long-running or heavy CPU queries. So I don't think the same concerns apply here as for Cohort Extractor.

Not that CPU usage monitoring would not be a very good thing.

inglesp (approved):

PR #1530, opened 2023-08-24
Move and rework remaining introductory content

iaindillingham:
Setting aside d894284, which I'm sneaking in because I don't want to open another PR, this PR moves the remaining introductory content into the landing page (aee62a9) and the explanation section (376a9ec). As with previous documentation PRs, I've tried to reduce the word count whilst retaining the information; this is good for our users (less to read) and good for us (less to maintain).


docs/tutorial/index.md
```diff
@@ -1,5 +1,5 @@
 The tutorial provides practical steps for studying ehrQL.
```
milanwiedemann, line 1: We have more than one tutorial, so I would use _tutorials_ here

mkdocs.yml
```diff
@@ -12,9 +12,6 @@ docs_dir: docs
 
 nav:
   - ehrQL: index.md
-  - Introduction:
-    - Guidance for existing cohort-extractor users: introduction/guidance-for-existing-cohort-extractor-users.md
-    - Introduction to ehrQL concepts: introduction/introduction-to-concepts.md
   - Tutorial:
     - Tutorial: tutorial/index.md
```
milanwiedemann, line 16: I would use _tutorials_ here because we have more than one. We also use plural in the 'How-to guides' subsection

mkdocs.yml
```diff
@@ -36,6 +33,7 @@ nav:
     - ehrQL backend tables: explanation/backend-tables.md
     - ehrQL output formats: explanation/output-formats.md
     - Using ehrQL in OpenSAFELY projects: explanation/using-ehrql-in-opensafely-projects.md
+    - Guidance for existing cohort-extractor users: explanation/guidance-for-existing-cohort-extractor-users.md
```
milanwiedemann, line 36: Can we delete 'existing' here? Not sure it adds anything?

milanwiedemann (approved):
I think this is already so much clearer than the version before! Some suggestions in the comments, feel free to ignore!

Also, I couldn't comment on code that you didnt change but had one more thought about the landing page `docs/index.md`. I think we could make it clearer that the tutorial is required to understand the next steps (i.e., dont start working with ehrQL before working through the tutorials because you'll need to understand these concepts)?

Something like this maybe:

1. The [tutorial***s***](tutorial/index.md) provides practical steps for studying ehrQL ***and will teach you the core concepts needed to work with ehrQL***.
2. The [how-to guides](how-to/index.md) provide practical steps for working with ehrQL in your project.

Or we could make it clearer somewhere else that it makes sense to fully understand the tutoral before starting to work with ehrQL?

PR #1534, opened 2023-08-25
Improve the structure of the tutorial

iaindillingham:
I've made a couple of small changes to the structure of the tutorial. First, I've separated the section about installing ehrQL from the section about generating dummy data. Second, I've named the sections consistently; _Installing_, _Generating_, _Writing_ etc. Because the documentation hasn't been around that long, I've renamed the files as well as renamed the sections.

Why change the structure of the tutorial? Anecdotally, we know that the content in the four main sections is (seen as) similar. I'd argue that by separating installing ehrQL and generating dummy data, and by naming the sections consistently, we separate the tutorial from, say, the how-to guides.


docs/tutorial/generating-dummy-data.md
milanwiedemann, line 1: I think the title '**Generating dummy data**' might be confusing here. Here we only describe dumping the 10 example patients that are only really needed to use the sandbox. So far, I think our users expect dummy data to be one file, i.e the output from an ehrQL action, and not 6 seperate files that represent the TPP database. 

I think we should clarify somewhere WHAT we mean by dummy data. Do we mean (1) multiple csv files that represent the database or (2) the output from an ehrQL action?

I think we should say "dummy tables" if we talk about 1 and use "dummy data" when we talk about 2, the output from an ehrQL action (similar to how we talk about it in cohortextractor https://docs.opensafely.org/study-def-expectations/#providing-your-own-dummy-data)

docs/tutorial/writing-dataset-definition.md
milanwiedemann, line 1: This is mainly about using the sandbox to understand how ehrQL works, but not directly about how to develop your `dataset_definition.py` file. Only a short section at the beginning and end of this page talk about dataset definitions.

I feel like the "Core types of object" subsection breaks the flow of the dataset definition example and would introduce these concepts earlier.

I think we should break this page into two, first something about "Core data types in ehrQL" and then something like "Writing a dataset definition".

mkdocs.yml
```diff
@@ -14,9 +14,10 @@ nav:
   - ehrQL: index.md
   - Tutorial:
     - Tutorial: tutorial/index.md
-    - Installation and setup of ehrQL: tutorial/installation-and-setup.md
+    - Installing ehrQL: tutorial/installing-ehrQL.md
+    - Generating dummy data: tutorial/generating-dummy-data.md
```
milanwiedemann, line 18: Just adding my ideas for heading here
```suggestion
    - Downloading dummy tables: tutorial/generating-dummy-data.md
```

mkdocs.yml
```diff
@@ -14,9 +14,10 @@ nav:
...
+    - Writing a dataset definition: tutorial/writing-dataset-definition.md
```
milanwiedemann, line 19:
```suggestion
    - Core data types in ehrQL: tutorial/ehrql-data-types.md
    - Writing a dataset definition: tutorial/writing-dataset-definition.md
```

milanwiedemann (commented):
I think we're making it much eaiser to follow the tutorial with these changes! 

I got confused when I saw the "Generating dummy data" section and think we should try to be cleaer what we mean by dummy data (vs dummy tables). I also think that the current page "Writing a dataset definition" includes a lot of information that breaks the flow. It's important but could get it's own page and be introduced earlier? I think would also fit quite well after the dummy tables page, because it helps the users understand the transformations that are needed to get to one row per patient?

iaindillingham:
I'm not going to touch the "Writing a dataset definition" page. I think the heading captures the intent; but you're right that the content needs improvement. I'd planned to remove the information about frames and series, because it's in the reference. On reflection, however, I think the intent of this page is to write a dataset definition and interact with it in the sandbox.

docs/tutorial/writing-dataset-definition.md
iaindillingham: I agree that the content needs improvement. However, I don't intend to improve it in this PR 🙂

docs/tutorial/generating-dummy-data.md
iaindillingham: The CLI docs use _dummy data_ in the context of:

* patient-level and event-level tables, i.e. dummy input data
* a dataset table, i.e. dummy output data

So, I think the distinction is between _input_ and _output_. I'd suggest calling this section _Generating dummy input data_, for consistency with the CLI docs.

A reason for not making the distinction between _dummy tables_ and _dummy data_ is that there may be a single dummy table.

docs/tutorial/writing-dataset-definition.md
iaindillingham: As discussed, we now have two pages: *Working with frames and series* and *Writing a dataset definition*.

milanwiedemann (approved):
Fantastic, I like and approve the changes!

PR #1546, opened 2023-08-30
docs: Remove cohort-extractor deprecation

iaindillingham:
We're not quite ready to say "ehrQL replaces cohort-extractor", so we remove the cohort-extractor deprecation notice. We discussed doing so during this morning's data team meeting.


inglesp (approved):
Why are we deleting the guidance for existing users?

iaindillingham:
> Why are we deleting the guidance for existing users?

Because it contains statements like

> ehrQL is a replacement for the existing OpenSAFELY cohort-extractor tool for new studies

> cohort-extractor is now in maintenance mode

> In principle, provided that ehrQL supports all the features that you need, then you can replace any cohort-extractor steps in an existing OpenSAFELY project with ehrQL instead.

and I don't think we're quite ready for that level of commitment. Given the choice between rewording this section and removing it, I decided to remove it. It will return, in a more concrete form, in the future.

Issue #1538, opened 2023-08-25
Clarify `dump-example-data` and `create-dummy-tables`


iaindillingham:
ehrQL distinguishes between output data and input data:

* _Output data_ are a dataset.
* _Input data_ are patient-level and event-level frames and series.

To generate dummy output data, a researcher can call `generate-dataset` outside an L3 environment. It's clear what this sub-command does: _generate_ is defined as [_to produce or create something_][1] and a _dataset_ is defined as output data, so this command produces or creates output data.

To generate dummy input data, a researcher can call `dump-example-data` or `create-dummy-tables`. Anecdotally, it's not clear what these sub-commands do.[^1] We could clarify them by:

* naming them `generate-example-input-data` and `generate-input-data`;
* replacing them with a single `generate-input-data` sub-command, whose arguments distinguish between the former and the latter;
* something else.

[1]: https://www.oxfordlearnersdictionaries.com/definition/english/generate

[^1]: @milanwiedemann and I were discussing the ehrQL tutorial. Milan thought that `create-dummy-tables` generated output data, because his dataset definition referenced only the `patients` table.

iaindillingham:
@inglesp and I discuss the underlying issue in [this Slack thread][1]: Essentially, should we use

* **dummy data** to mean dummy *input* data;
* **dummy dataset** to mean dummy *output* data.

If so, then we should update _Generating dummy output data_ and _Command line interface_ in [the ehrQL section of the OpenSAFELY documentation][2], as well as any other references.

[1]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1693918523779469
[2]: https://docs.opensafely.org/ehrql/

PR #1551, opened 2023-09-01
Improve the content of the tutorial

iaindillingham:
I've attempted to improve the content of the tutorial. There are three types of commit.

The **first** type of commit, which is easiest to review, moves content from a page to an index file and creates a similarly named directory to the page. Doing so means we can store images adjacent to the content; we can avoid, in other words, having to think too hard about what to name similar images in different sections.

The **second** type of commit, which is the most substantive, attempts to improve the content. When @inglesp asked me how I intended to make the tutorial more tutorial-like, I cited two guidelines from the [Diátaxis][1] approach:

* describe concrete steps
* offer minimum explanation

I've attempted to adhere to these guidelines. I've also incorporated Peter's feedback in [this Slack thread][2]; most importantly, a reader should install VS Code to complete the tutorial.

The **third** type of commit moves existing content and adds new content. I don't think a discussion of `sandbox` vs `exec` vs `run` belongs in the tutorial, so I've moved this page to the explanation. Nevertheless, the tutorial should be _meaningful_ and _usefully complete_ (Diátaxis again). I think that writing dummy output data to the terminal meets these criteria.

Are you sitting comfortably? Then we'll begin. <https://iaindillingham-tutorial.databuilder.pages.dev/tutorial/>

[1]: https://diataxis.fr/tutorials/
[2]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1692983111046249


sebbacon:
From a quick skim, it seems very clear.

One question: reading the "Installing ehrQL" bit, I found myself wondering how Codespaces might fit it. I realise that's presumably for a future PR. But I'm aware we have NHS folks very interested in ehrQL and they can't install Docker. So I'm wondering if a short note ("can't install Docker? Get in touch in Slack and we can help you do it in the cloud" or something) might be helpful, as a placeholder until we develop a Codespaces workflow and documentation. What do you think?

docs/tutorial/generating-dummy-output-data/index.md
```diff
@@ -0,0 +1,41 @@
...
+Scroll up to see the column headers;
+notice that two column headers correspond to the two patient series
+— `asthma_med_date` and `asthma_med_code` —
+that you added to the dataset definition.
+
+Also notice the lines that start with `[info]`.
+These show that the dummy data generator generated dummy data for 5,000 patients;
+found 500 that matched the dataset definition;
+and wrote them to
+— in this case, displayed them in —
```
sebbacon, line 33: How about "...and displayed them in the terminal"? I don't think we need bother the reader with TTY technicalities?

iaindillingham: Fair point! I'm not paid by the word, so I'll remove these 🙂

iaindillingham:
I'm reluctant to add caveats to the tutorial, @sebbacon. The tutorial aims to do one thing, and do it well. If we're confident that Codespaces are ready to use, then we should also be ready to document their use.

docs/tutorial/working-with-frames-and-series/index.md
```diff
@@ -0,0 +1,213 @@
...
+7 | 8 | 2021-01-27
+9 | 9 | 2015-03-14
+```
+
+## Transform input data into output data
+
+**Tables contain input data**.
+As a researcher, your task is to transform tables into a dataset.
+A dataset, like a patient frame, has one row per patient.
+Unlike a patient frame, however, **a dataset contains output data**.
```
sebbacon, line 121: I found this section hard to parse. Thinking it through aloud:

> **Tables contain input data**.

Me: Hm, why "input" data? Isn't it just data? Is it because we care about where it came from? 

> As a researcher, your task is to transform tables into a dataset.

Me: what is a dataset? Is it like a table? How is it different?

> A dataset, like a patient frame, has one row per patient.

Me: what's a patient frame again? [scrolls back] Oh, it's a table. Hang on, why aren't we calling it a table?

> Unlike a patient frame, however, **a dataset contains output data**.

Me: Doesn't a patient frame contain data?! Is that _input_ data? How are they different? I thought it was _tables_ that contain input data. 


I've gotta dash, sorry, but happy to try to come up with some alternative suggestions later if it helps!

iaindillingham:
I'd argue that the explanation of the Django development server is unnecessary. It's called a _development_ server! What is it for, if not development? Why should it matter to me that it's lightweight? (What does _lightweight_ mean, in this context?) Why should it matter to me that it's written in Python? The effect of me running a command called `runserver` is to start the Django development server. Might it not be better to call the command `rundevserver`, so the effect of me running it is self-explanatory?

I'd suggest that the answer to the question "Why is Django asking me to start the development server?" is "Because doing so will save me time".

Anyhow, I'm unconvinced by your hypothesized reaction, @inglesp 🙂. I feel that in many cases, you'd accept that each _this_ was a piece of a jigsaw puzzle, and that this-by-this explanations were unnecessary. (I'm also reminded of Rose's feedback that the documentation was very "wordy". More explanations; more words.) There are, however, clearly some instances of _this_ that would benefit from an explanation: I'd like to learn which, like you, from some users. So, I will:

* Walk through the tutorial with some users
* Note which parts would benefit from an explanation
* Add the explanations

I think that the distinction between dummy input data and dummy output data is very important, so I'd like to talk more about your comment. I used this distinction to help Milan understand the difference between `generate-dataset`, `dump-example-data`, and `create-dummy-tables` (#1538). I think that users could make better use of dummy input data to overcome some of the limitations of the dummy data generator; by extension, we could help them by making this distinction clear.

docs/tutorial/working-with-frames-and-series/index.md
iaindillingham: Thanks, @sebbacon. I'm unconvinced by this section; but I couldn't remove _everything_ from the existing tutorial 🙂 I think it was originally here to motivate the transformations that follow. We now have them, and many more, in [*Using ehrQL to answer specific questions*](https://docs.opensafely.org/ehrql/how-to/examples/). 

It doesn't help that "table" is overloaded. Here, we mean specifically ehrQL tables (not, for example, tabular data structures). I'll clarify.

We care very much about where the data came from: it came from L2. But in this case, it didn't, because it's dummy input data: it came from a proxy L2. As a researcher, you should know that data come from somewhere and go to somewhere. You should know that ehrQL sits in the middle.

I feel that the "What's a dataset?" question is answered: one row per patient; output data. The reference to a patient frame isn't helpful; I'll remove.

iaindillingham: I think I'd also like to add that to a certain extent, the names are given: Whilst we may decide that we could tighten up "table", "frame", "input data", and "output data", I've tried to update the tutorial to reflect what we have now, rather than what we might have in the future.

PR #1559, opened 2023-09-04
Fix hash randomisation test

evansd:
The fix applied in #1459 turns out to not actually work. This PR fixes the test which made it _look_ like it worked, and attempts to fix the underlying DevEx issue in another way.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,7 @@
+# This file defines environment variables we want to be set in development
+# environments.  Commands run via Just should have these set already but we
+# want to support other invocation mechanisms. We use a `.env` file as VSCode's
+# Python extension (but not the terminal) should pick these up by default, see:
+# https://code.visualstudio.com/docs/python/environments#_environment-variables
+
+PYTHONHASHSEED=0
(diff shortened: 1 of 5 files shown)


.env
```diff
@@ -0,0 +1,7 @@
+# This file defines environment variables we want to be set in development
+# environments.  Commands run via Just should have these set already but we
+# want to support other invocation mechanisms. We use a `.env` file as VSCode
+# should pick these up by default, see:
```
iaindillingham, line 4: I can see myself using the terminal in VS Code and wondering why it isn't picking up the environment variables defined within `.env`. Of course, that probably wouldn't happen for _this_ environment variable, because I'd use `just test`. But I can see `.env` growing.

```suggestion
# want to support other invocation mechanisms. We use a `.env` file as VSCode's
# Python extension (but not the terminal) should pick these up by default, see:
```

iaindillingham (approved):

.env
evansd: Oh thanks! I hadn't realised this so it's a useful clarification.

PR #1564, opened 2023-09-05
Use Codespaces for tutorial

iaindillingham:
We replace the sections of the tutorial that set up OpenSAFELY locally with a Codespaces approach. Why? Because the Codespaces approach involves less setup, is more reliable, and is free for personal accounts; everyone should be able to follow it.

As this is a tutorial, there's no expectation that a user will use the dataset definition they write in their own project. Consequently, there's no expectation that they will move the repo from their personal account to the opensafely organization. Consequentially, a user need not use opensafely/research-template, so we need not describe, for example, what *project.yaml* does. (We already do that, anyway, in *The project pipeline*.)

That said, we do need to set up a codespace somehow. Enter [opensafely/ehrql-tutorial](https://github.com/opensafely/ehrql-tutorial), which is a template. (It's not called opensafely/ehrql-tutorial-template, so a user can't inadvertently create it in the opensafely organization.) The _devcontainer.json_ in this repo aims to take away some of the less useful/interesting tasks that were in the tutorial (executing `dump-example-data`), so isn't an exact copy of the _devcontainer.json_ within opensafely/research-template.


StevenMaude (approved):
This looks fine; I do have a question.

StevenMaude:
Question: Where does the guidance go for someone setting up ehrQL who doesn't want to use Codespaces? This information is now removed entirely.

For the tutorial, it's fine, and it's good to have a single route. Effectively, this will eventually replace the main Getting Started tutorial that uses cohort-extractor. It's also reasonable that a user wants to run everything on their computer too, and this isn't well-explained that I can find anywhere.

StevenMaude:
I'll add that I had [some information on Codespaces](https://docs.opensafely.org/github-codespaces/) already; this may overlap very slightly with what's here, or be useful to point to from the ehrQL section.

iaindillingham:
> Question: Where does the guidance go for someone setting up ehrQL who doesn't want to use Codespaces? This information is now removed entirely.

If they're only setting up ehrQL, then all they need to is run `opensafely pull ehrql`. And if they didn't do that, then the first time they used ehrQL, `opensafely` would do it for them. To put that another way, there isn't a page on setting up cohort-extractor; why should there be a page for ehrQL?

Thanks for highlighting *Running OpenSAFELY in GitHub Codespaces*. User feedback suggested that too much linking was distracting -- it was one of the reasons for preferring videos to pages -- so I'll not link to it directly from the tutorial.

PR #1565, opened 2023-09-05
Apply strict equality to frozensets

### LINKED ISSUE #1562: Python's `0 == False` behaviour causes us to generate incorrect SQL
In Python, `frozenset({False}) == frozenset({0})`. This means that where we have two `is_in` queries using these values, the values will be treated as equivalent by the caching mechanism and we'll return the same SQL for both. This works OK for databases which happen to treat `0` and `False` equivalently (or just use `0` as their falsey value) but for others like Trino which are more strict it doesn't work.

For example, this test here passes on all other query engines.

```py
def test_is_in_type_confusion(spec_test):
    table_data = {
        p: """
              | i1 | b1
            --+----+----
            1 | 0  | T
            2 | 1  | T
            3 | 1  | F
            """,
    }
    spec_test(
        table_data,
        p.i1.is_in([0]) | p.b1.is_in([False]),
        {
            1: True,
            2: False,
            3: True,
        },
    )
```

But for Trino it results in:
```
trino.exceptions.TrinoUserError: TrinoUserError(type=USER_ERROR, name=TYPE_MISMATCH, message="line 1:111: IN value and list items must be the same type or coercible to a common type. Cannot find common type between integer and boolean, all types (without duplicates): [integer, boolean]", query_id=20230904_162155_03259_8c7pc)
```

Note that we have the same problem with floats.

This was discovered by Hypothesis but buried in an absolutely monstrous query which had to be simplified by hand.

This is related to the issue fixed in https://github.com/opensafely-core/ehrql/pull/578 but can't be quite so straightforwardly fixed as we don't control equality for `frozenset`. Possibly we'll have to start using our own type here which might have other benefits (e.g. we could make it retain insertion order and then we don't have to muck about with PYTHONHASHSEED to get stable orderings).

evansd:
This extends the fix applied in #578 to frozensets to address #1562. We also get `Value` to reject any container types for which we haven't defined an equality method to prevent this issue reoccurring when we add new types.

Closes #1562

### DIFF
--- a/tests/unit/query_model/test_nodes.py
+++ b/tests/unit/query_model/test_nodes.py
@@ -457,8 +457,14 @@ def test_any_type_acts_as_an_escape_hatch():
     "lhs,cmp,rhs",
     [
         (10, "==", 10),
+        (10, "!=", 11),
         (10, "!=", 10.0),
+        (True, "==", True),
+        (True, "!=", False),
         (1, "!=", True),
+        (1.0, "!=", True),
+        (0, "!=", False),
+        (0.0, "!=", False),
     ],
 )
 def test_comparisons_between_value_nodes_are_strict(lhs, cmp, rhs):
--- a/tests/unit/query_model/test_nodes.py
+++ b/tests/unit/query_model/test_nodes.py
@@ -465,6 +465,21 @@ def test_any_type_acts_as_an_escape_hatch():
         (1.0, "!=", True),
         (0, "!=", False),
         (0.0, "!=", False),
+        # Test that we're strict about booleans embedded in frozensets
+        (frozenset({True}), "==", frozenset({True})),
+        (frozenset({True}), "!=", frozenset({False})),
+        (frozenset({1}), "!=", frozenset({True})),
+        (frozenset({1.0}), "!=", frozenset({True})),
+        (frozenset({0}), "!=", frozenset({False})),
+        (frozenset({0.0}), "!=", frozenset({False})),
+        (frozenset({True, False}), "==", frozenset({False, True})),
+        # Test the fast path for identical frozensets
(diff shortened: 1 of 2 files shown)


tests/unit/query_model/test_nodes.py
```diff
@@ -453,13 +453,55 @@ class SomeInternalOperation(Series):
...
+        (example := frozenset({1, 2, 3}), "==", example),
+        # Test the fast path for different sized frozensets
+        (frozenset({1, 2, 3}), "!=", frozenset({1, 2})),
+        # Test the normal path for frozenset equality
+        (frozenset({1, 2, 3}), "==", frozenset({3, 2, 1})),
+        (frozenset({1, 2, 3}), "!=", frozenset({4, 5, 6})),
+    ],
+)
+def test_comparisons_between_value_nodes_are_strict(lhs, cmp, rhs):
+    if cmp == "==":
```
bloodearnest, line 486: :heart:

iaindillingham: Could we not use `operator` here? We do when testing the query language. For example:

https://github.com/opensafely-core/ehrql/blob/ea07908488e07c7f84b14a120aa428eab3da8d98/tests/unit/test_query_language.py#L387-L403

iaindillingham (approved):

tests/unit/query_model/test_nodes.py
evansd: Yeah we could. Those QL tests are part of a whole series of similar tests which use a whole bunch of different operators so that definitely seems like the right approach there. For just `==` and `!=` this way felt a bit more obvious and simple to me, but I wouldn't object to them being done either way really.

PR #1572, opened 2023-09-06
docs: Fix appointments docstring

iaindillingham:
As @alschaffer found, the workspace has moved. Whilst OpenSAFELY Jobs redirected, the information about roles was confusing.


inglesp (approved):
Fine.  Do we have any plans to make this public?

iaindillingham:
>  Do we have any plans to make this public?

Issue #1574, opened 2023-09-07
Add a function for instantiating objects of class `Dataset`


iaindillingham:
When watching Rose work through the ehrQL tutorial recently, I realized just how confusing

```py
dataset = Dataset()
```

must be. As an exercise, try to describe what's happening above, without using the words _instantiate_, _object_, or _class_ (Alternatively, try to describe this issue's title.)

We know, anecdotally, that ehrQL seems to be more Python-like than cohort-extractor. Perhaps we could make it less so with

```py
dataset = create_dataset()
```

I think @inglesp raised this issue on Slack awhile back. I also think that there was a brief discussion about whether the verb should be "create", "make", or whatever. I don't think it matters; but the ehrQL tutorial instructs a user to [_Create the dataset_](https://docs.opensafely.org/ehrql/tutorial/writing-a-dataset-definition/#create-the-dataset).

evansd:
I'm in favour of this. Here's the [original thread](https://bennettoxford.slack.com/archives/C03FB777L1M/p1682605407408419) where a few options were discussed including:
 * `new_dataset()`
 * `make_new_dataset()`
 * `make_empty_dataset()`
 * `create_empty_dataset()`

I think `create_dataset()` is my preference so far though.

evansd:
For parity we should also define a `create_measures()` function.

inglesp:
> I think `create_dataset()` is my preference so far though.

Mine too.  I like the verb + noun, and I like the succinctness.

iaindillingham:
Note to self: don't forget to update the ehrQL tutorial.

PR #1575, opened 2023-09-08
Improve the tutorial in response to user feedback

iaindillingham:
I'm sure there will be several similar PRs. However, [the user feedback][1] that this PR incorporates came from @rose-higgins. Thanks, Rose!

Three commits describe concepts in more depth (b225736, 53cbcf6, and cbc4bea). Hopefully their commit messages provide sufficient justification for their inclusion.

Two commits replace the terms _frame_ and _series_, with the terms _table_ and _column_ (e133076, a86e220). Whilst I hope their commit messages provide sufficient justification for their inclusion, I'd also like to pick out some quotes from the user feedback:

> “I don’t know what ‘the patients table is a patient frame’ means”
> “I’m a bit confused about the difference between a table and a frame. I’m not sure what a frame is.”
> “I’m not sure what the difference between a table and a dataset is”
> “Code event series throws me off; code event series seems like a new classification type. Maybe it was previously called a date of birth event series?”

I feel we can really help users by _not_ introducing frames and series during the tutorial. This may seem strange, if you're a software developer ("Surely you'd want to know that the type was a patient frame?"). However, I'd argue that pedagogically, the result is faster, more accurate learning.

Related to "strange, if you're a software developer" see #1574.

[1]: https://docs.google.com/document/d/1fMOvr7k8Hl-16mOvPpghl16TD3-8vi2S-u8d1H6fkaA/edit#heading=h.898xxbxilnop


sebbacon:
The observation that they don't need to know _anything_ about frames etc is one of those delightfully simple "oh yeh that's obvious now you mention it" things. I think it's a great improvement

docs/tutorial/setting-up/index.md
```diff
@@ -48,6 +48,19 @@ In this section, you will run an OpenSAFELY command to test that the codespace i
...
+    such as ehrQL,
+    that have command line interfaces.
+    Whilst *terminal* and *shell* are often used synonymously,
+    a shell runs within a terminal.
+
+    You may have used a shell before:
+    Git Bash, for example, bundles Git with the Bash shell for the Windows operating system.
+    The codespace's default shell is also the Bash shell,
+    so if you've used Git Bash before,
+    then you'll find the codespace's default shell familiar.
```
inglesp, line 62: Applying Dillingham's razor: do we really need this?

docs/tutorial/working-with-frames-and-series/index.md
```diff
@@ -115,7 +115,13 @@ Similarly, the `medications.date` column has many rows per patient.
...
 Your task, as a researcher, is to transform tables
 — such as `patients` and `medications` —
-into a dataset.
+into a dataset that is suitable for analysis.
+
+??? tip "Tables and datasets"
+    In ehrQL, **tables and datasets perform different functions**.
+    Whilst both can be represented as rows and columns,
+    a *dataset* is a group of patients with common statistical characteristics.
+    In other words, a *dataset* is a *cohort*.
```
inglesp, line 127: Is it?  The reason we're not using "cohort" is because it was understood to mean "the set of patients in a study".

docs/tutorial/working-with-frames-and-series/index.md
```diff
@@ -96,6 +98,7 @@ patient_id        | row_id            | date              | dmd_code
 ```
 
 Similarly, the `medications.date` column has many rows per patient.
+Notice that the column is indexed by `patient_id` and `row_id`.
```
inglesp, line 101: No: it's only indexed by `row_id`.

docs/tutorial/working-with-frames-and-series/index.md
```diff
@@ -68,15 +67,16 @@ The `patients.date_of_birth` column is a patient series; it has one row per pati
 
 ## Work with event data
 
-### Import the `medications` table
+Working with event data means working with many different tables.
```
inglesp, line 71: Not sure about this.  You can work with event data through working with a single table.  I read this as implying that you need to use many different tables to work with event data.

inglesp (approved):
Some good changes here, but I'm on the fence about dropping frames and series.

You say:

> When working through the tutorial, a user repeatedly questioned what frames and series were

To what extent is that because we don't actually define these terms in the tutorial?

I'd want to avoid a repeat of the first day of A Level chemistry: "right students, everything you learnt for your GCSEs was wrong...".  If we make the change here, do you want us to make the change everywhere in the docs?  If so, could we have a discussion with the team?

Some background: the reason we chose to differentiate between frames/tables and series/columns was to try to make clear that nothing actually happens until you submit your query for execution.  Of course, this is now blurred by the sandbox.

iaindillingham:
> To what extent is that because we don't actually define these terms in the tutorial?

In the tutorial, we say "The `patients` table is a patient frame; it has one row per patient." and "The `medications` table is an event frame; it has many rows per patient." So, I'd argue that we _do_ define these terms in the tutorial.

> I'd want to avoid a repeat of the first day of A Level chemistry: "right students, everything you learnt for your GCSEs was wrong...".

That isn't the case, here. Some tables _are_ frames; some tables _are_ series. Within *ehrql/tables/beta/core.py* and *ehrql/tables/beta/tpp.py*, you'll find classes that inherit from `PatientFrame` and `EventFrame` as well as `Series`. If you learn during A Level ehrQL that tables can be further divided into these types, then you didn't learn anything wrong during GCSE ehrQL.

> If we make the change here, do you want us to make the change everywhere in the docs? If so, could we have a discussion with the team?

No, I don't want us to do make the change everywhere in the docs.

> Some background: the reason we chose to differentiate between frames/tables and series/columns was to try to make clear that nothing actually happens until you submit your query for execution. Of course, this is now blurred by the sandbox.

It's blurred in the [_Language reference_](https://docs.opensafely.org/ehrql/reference/language/), too. For example:

> You can think of a Frame as being like a table in a database, in that it contains multiple rows and multiple columns.

inglesp:
> > To what extent is that because we don't actually define these terms in the tutorial?
> 
> In the tutorial, we say "The `patients` table is a patient frame; it has one row per patient." and "The `medications` table is an event frame; it has many rows per patient." So, I'd argue that we _do_ define these terms in the tutorial.

I don't think these are really definitions -- or at least, they're not framed as such.  It's like saying "Zebedee is a zebra; it has stripes".  Should I interpret that to mean that stripes define a zebra?

> Some tables _are_ frames, some columns  _are_ series

Yes, but some are not.  In this PR weu're occasionally using "column" to mean "series".  It almost certainly doesn't matter in this case, but I'd like to avoid introducing sloppiness if we can help it, even if it's very mild and pedagogically reasonable.

~~~

However... the more I think about it, the more I'd be willing to consider dropping "frame" and "series" completely in favour of "table" and "column".  But I'd like to discuss this with the team.

So maybe go ahead and merge this now -- it's an improvement on what came before -- and then we could discuss terminology when you're next in.

iaindillingham:
> I don't think these are really definitions -- or at least, they're not framed as such. It's like saying "Zebedee is a zebra; it has stripes". Should I interpret that to mean that stripes define a zebra?

Whilst these aren't framed as definitions, I think they fulfil that function. Let's see what it takes to frame Zebedee as a definition.

> "Zebedee is a zebra; it has stripes."
> "Zebedee is a zebra; Zebedee has stripes."
> "Zebedee is a zebra; zebras have stripes."
> "Zebedee is a zebra; zebras are striped."
> "Zebedee is a zebra; zebras are striped wild horses."

I think the "wild horses" really moves us from definition-like to definition. The problem with frames and series, though, is that it's hard to find an equivalent to "striped wild horses" that doesn't require further definition.

> The `patients` table is a patient frame. _Patient frames_ are containers for patient-level data (they have one row per patient). Sometimes, this data will come directly from a backend. However, this data may also be the result of transforming data from a backend.

* What's a container?
* What's a backend?
* What does _transform_ mean?

Perhaps we should go the other way and remove the word `table` and move the code to a `frames` module?

> `patients` is a patient frame. ...

---

> Some tables _are_ frames; some columns _are_ series
>> Yes, but some are not.

I think you mean some frames are not tables; and some series are not columns. This is correct: `medications` is a frame and a table; `medications.sort_by(medications.date).first_for_patient()` is a frame, but not a table. However, I don't think that from a user's perspective, this matters.

---

> I'd like to avoid introducing sloppiness

I'd very gently, but very clearly, like to object to the word _sloppiness_. We're trying to find the level of abstraction that facilitates effective learning. If mapping programming language to human languages was lossless, then we wouldn't need programming languages.

docs/tutorial/working-with-frames-and-series/index.md
iaindillingham: The primary key is `row_id`. The foreign key is `patient_id`. But if we want to look up the medications for a patient, and we don't want to describe why our data structure of one series plus an index is actually a data structure of two series plus an index, it helps to think of an _index_ as a look-up table.

docs/tutorial/working-with-frames-and-series/index.md
iaindillingham: You'll have to explain further: why is a dataset not a set of patients in a study? Because there's more than one dataset? If so, isn't that an implementation detail?

docs/tutorial/setting-up/index.md
iaindillingham: Rose noticed that the shell was Bash and not Git Bash. Whilst `n=1` is a terrible sample size, our `n` is never going to be a large number. I think we need this, if only because Bash behaves slightly differently to Git Bash. For example, in some Git Bashes I've used, tab-completing a command adds an extension to the command: `column` becomes `column.exe`.

docs/tutorial/working-with-frames-and-series/index.md
iaindillingham: Fair point. I've updated this to read:

> Unlike patient data, there are many sources of event data: the `medications` table, for example. To work with the `medications` table, first import it into the sandbox.

iaindillingham:
Thanks for your constructive critique, @inglesp 🙂 I'm going to do as you say, and merge this now.

PR #1582, opened 2023-09-11
feat: Expose OPA table

### LINKED ISSUE #1460: Expose the `OPA` database table
In cohort-extractor, we query the `OPA` database table with [the `TPPBackend.patients_outpatient_appointment_date` method][1]. In ehrQL, we expose the `OPA_Cost`, `OPA_Diag`, and `OPA_Proc` database tables as the `opa_cost`, `opa_diag`, and `opa_proc` ehrQL tables. We also join `OPA` to each of these database tables, to expose `OPA.Appointment_Date` and `OPA.Referral_Request_Received_Date`. However, we don't expose `OPA` directly and a user can't, for example, filter by `OPA.Treatment_Function_Code`. For parity with cohort-extractor, we should.

The method can filter `OPA` on _attendance_ and _first attendance_, where values of `OPA.Attendance_Status` that encode _attendance_ and _first attendance_ are given by the HES data dictionary. We should consider how best to replicate this behaviour in ehrQL to avoid, say, a user sorting by `OPA.Appointment_Date` and returning the first record for each patient.

Recently, @markdrussell asked about exposing the `OPA` database table, as he's translating the cohort-extractor study definition into an ehrQL dataset definition for the [early-inflammatory-arthritis][2] study.[^1] However, the lack of the database table in ehrQL isn't blocking the study.

[^1]: https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1690384405516059

[1]: https://github.com/opensafely-core/cohort-extractor/blob/v1.86.2/cohortextractor/tpp_backend.py#L2856-L2966
[2]: https://jobs.opensafely.org/early-inflammatory-arthritis-during-covid-19/early-inflammatory-arthritis/

inglesp:
Only columns that are queryable in cohort-extractor have been exposed.

Fixes #1460.

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -20,6 +20,7 @@ from ehrql.tables.beta.tpp import (
     medications,
     occupation_on_covid_vaccine_record,
     ons_deaths,
+    opa,
     opa_cost,
     opa_diag,
     opa_proc,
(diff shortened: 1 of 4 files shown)


evansd (approved):
The fact that we can so easily add a new table and have confidence that it will do what we expect feels like a real payoff from all that work we did automatically importing the TPP schema and creating the backend integration tests.

PR #1587, opened 2023-09-13
feat: Make size of dummy data population configurable

inglesp:


inglesp:
After our discussion yesterday, I'm sharing this early to get feedback on the approach.

Questions:

* The current API would be to do `dataset.dummy_data_config.population_size = 1000`.  Does this expose too much of the implementation?  Should I change it to `dataset.set_dummy_data_config(population_size=1000)`?
* I've bunged `DummyDataConfig` into `ehrql/query_language.py` to avoid circular imports -- ideally it would be in `ehrql/dummy_data/config.py`.  Does anyone have any tips for resolving circular import problems?  Or is this a wart we can live with?
* How should I document this?

evansd:
> The current API would be to do `dataset.dummy_data_config.population_size = 1000`.  Does this expose too much of the implementation?  Should I change it to `dataset.set_dummy_data_config(population_size=1000)`?

I feel like calling a method rather than assigning an attribute would be more consistent with what we do elsewhere. Currently we use attribute assignment only with names that the user makes up, and we don't have any "magic" attribute names (we have reserved attribute names like `patient_id` but I'd argue that's slightly different). So my preference would be to use a method here.

As for the method name, I have a memory that "set" as a verb – while feeling very natural to us – is somehow confusing to users. Maybe `configure_dummy_data` would be clearer?

> I've bunged `DummyDataConfig` into `ehrql/query_language.py` to avoid circular imports -- ideally it would be in `ehrql/dummy_data/config.py`.  Does anyone have any tips for resolving circular import problems?  Or is this a wart we can live with?

Oh, that's a pain. I mean, it doesn't feel terrible to me to have it in `query_language` but it's annoying that we've managed to tie ourselves in a circle like that. I'd say leave it there for now and we can look at maybe picking things apart a bit better later.
 what would cause the circularity if it was in `dummy_data.config`? 

> How should I document this?

Another advantage of a method call over assigning to a config object is that we naturally get reference documentation for this. Obviously that's not sufficient by itself, but it's better than nothing.  I guess we need some proper documentation on dummy data (how it works, the different ways you can generate it etc) and this would be described there.

It _could_ also be worth sticking a reference to it in the logs where we say how many dummy rows we're trying to generate:
https://github.com/opensafely-core/ehrql/blob/c3a3417a25cf96d66bbb2e7fb045d27bb79a8606/ehrql/dummy_data/generator.py#L55-L58

I don't know if anyone would actually look at that though.

ehrql/query_language.py
```diff
@@ -68,14 +74,24 @@ def define_population(self, population_condition):
...
+
     def __setattr__(self, name, value):
         if name == "population":
             raise AttributeError(
                 "Cannot set variable 'population'; use define_population() instead"
             )
         if name in self.variables:
             raise AttributeError(f"'{name}' is already set and cannot be reassigned")
-        if name in ("patient_id", "variables"):
+        if name in ("patient_id", "variables", "dummy_dataset_config"):
```
evansd, line 94: For the sake of completeness, might be worth adding this as a test case here:
https://github.com/opensafely-core/ehrql/blob/c3a3417a25cf96d66bbb2e7fb045d27bb79a8606/tests/unit/test_query_language.py#L121-L131

evansd (approved):
Excellent. I'm so glad we've finally got this in place.

One thing which we should think about as soon as we can (although I wouldn't want it to block merging this PR) is extending this behaviour to measures as well.

PR #1598, opened 2023-09-18
docs: Remove unnecessary dummy data logging

iaindillingham:
Follows #1589. I've not updated the associated screenshot (opensafely_exec.png). This is intentional: unless (until?) we generate screenshots automatically, I think a small amount of drift is acceptable.


evansd (approved):
Ah, thanks! It's so hard to keep on top of the ripple out effects of these small changes.

PR #1604, opened 2023-09-20
Improve the tutorial in response to user feedback

iaindillingham:
The user feedback that this PR incorporates came from @cewalters, @StevenMaude, and @wjchulme. Thanks, Caroline, Steve, and Will!

If you're interested in the rationale for each commit, then step through them one-by-one.

Preview: <https://iaindillingham-tutorial.databuilder.pages.dev/tutorial/>


docs/tutorial/setting-up/index.md
```diff
@@ -40,7 +40,8 @@ A *codespace* is a development environment that's hosted in the cloud.
 
 1. Click **Create codespace on main**
 
-The codespace will open in a new tab.
+It will take roughly two minutes to create the codespace,
+which will open in a new tab.
```
inglesp, line 44: Is this a one-time cost?

docs/tutorial/working-with-tables/index.md
```diff
@@ -1,4 +1,5 @@
 In this section, you will work with tables in the sandbox.
+The *sandbox* is an environment for experimenting with ehrQL.
```
inglesp, line 2: Mention dummy tables here?

docs/tutorial/working-with-tables/index.md
```diff
@@ -30,6 +30,20 @@ first import it into the sandbox.
...
+
+    ```pycon
+    >>> from ehrql.tables.beta.core import patients
+    ```
+
+    into the sandbox and pressing ++enter++ displays
+
+    ```pycon
+    >>>
+    ```
```
inglesp, line 45: In the interests of keeping things snappy, do we need the code sample here?

inglesp (approved):

docs/tutorial/setting-up/index.md
iaindillingham: Yes, in the once-per-codespace sense. In other words, if you do the tutorial twice, in different codespaces, then you pay twice; if you do the tutorial twice, in the same codespace, then you pay once.

I investigated [prebuilding][1], but

> If it currently takes more than 2 minutes to create a codespace for a repository, you are likely to benefit from using prebuilds.

So, it seems we are unlikely to benefit from using prebuilds. Even if we were, then I'm not sure that we could, because the prebuild would be associated with the template repo, not the repo the user created from the template repo.

[1]: https://docs.github.com/en/codespaces/prebuilding-your-codespaces

docs/tutorial/working-with-tables/index.md
iaindillingham: Catherine spent a couple of minutes clicking around after typing

```py
>>> from ehrql.tables.beta.core import patients
```

into the sandbox and pressing enter, and ended up opening *example-data/patients.csv*. Whilst I agree this snippet isn't snappy, I wanted to be extra extra sure that the user didn't expect any output. Consequently, I'd argue that we do need it.

docs/tutorial/working-with-tables/index.md
iaindillingham: Good point. Will do 👍🏻

docs/tutorial/working-with-tables/index.md
inglesp: Won't that just be served by saying

> Importing a table into the sandbox doesn't display any output.

?

Don't feel strongly though!

PR #1606, opened 2023-09-20
Extend generative tests to exercise population logic

### LINKED ISSUE #1593: Generative tests should exercise population logic
At present, the generative tests used a fixed population definition:
https://github.com/opensafely-core/ehrql/blob/d41539e88dae5cf2a06fcfb7979f1a13d854fb34/tests/generative/test_query_model.py#L169-L172

Which just includes every patient in every table in the test:
https://github.com/opensafely-core/ehrql/blob/d41539e88dae5cf2a06fcfb7979f1a13d854fb34/tests/generative/data_setup.py#L50-L52

This makes sense when the thing under test is `variable` because we want to evaluate it for as many patients as possible. But it also means that we never test more exotic population definitions and so bugs in the population handling logic won't be exposed.

This came up in discussion on this PR:
 * https://github.com/opensafely-core/ehrql/pull/1490

And I also suspect it would have caught several, possibly all, of the bugs identified here:
 * https://github.com/opensafely-core/ehrql/pull/1592

So I think we want to extend the generative tests to generate population definitions as well as variables. As a population definition is just a one-row-per-patient boolean series, most of the mechanics for doing this are already in place. But there are a few complication factors.

First, not every one-row-per-patient boolean series is a valid population definition (see the [`population_validation`](https://github.com/opensafely-core/ehrql/blob/d41539e88dae5cf2a06fcfb7979f1a13d854fb34/ehrql/query_model/population_validation.py#L23-L75) model for details). Hopefully we can use `hypothesis.assume()` to reject invalid examples and they won't be frequent enough to be problematic.

The second challenge is that the more exotic the population definition the less the chance that any of the generated data will actually match it and so we'll end up with an empty result set. And when this happens, we're not really testing the `variable` at all.

I wonder if a way round this would be to execute every example twice: once using the generated population definition and again with the "all patients" population definition.

evansd:
As a test of this I tried commenting out all the fixes added in #1592 and Hypothesis quite quickly found all the bugs, which is encouraging in a post-bolt-stable-door-closing kind of way.

Because the generative tests are now doing more work they take longer. Running the 100 examples required by CI used to take 10 seconds locally for me and now takes 16 seconds. I think this is acceptable given the extra coverage we're getting.

Closes #1593


tests/generative/test_query_model.py
```diff
@@ -169,56 +177,69 @@ def test_handle_date_errors(query_engines, operation, rhs):
...
-            for i, result in enumerate(first_results):
-                assert result == pytest.approx(
-                    second_results[i], rel=1e-5
-                ), f"Mismatch between {first_name} and {second_name}"
-        else:
-            assert (
-                first_results == second_results
-            ), f"Mismatch between {first_name} and {second_name}"
+    assert isinstance(dummy_data_generator.get_data(), list)
+    # Using a simplifed population definition which should always have matching patients
```
iaindillingham, line 234:
```suggestion
    # Using a simplified population definition which should always have matching patients
```

iaindillingham (approved):
It's been awhile since I looked at the generative tests, but I followed along. One tiny nitpick, but, 👍🏻.

PR #1611, opened 2023-09-22
fix: Use temporary tables instead of CTEs for Trino

evansd:
These aren't actually temporary tables in the usual sense because Trino doesn't support those: they're standard, persisted tables which we delete after the query has completed. This does mean that we're not guaranteed to clean them in case of an error, but we use a naming scheme which includes the date as a prefix and so is amenable to having a scheduled cleanup job to clear old tables.

This is roughly what Cohort Extractor used to do when run in EMIS.

This change is needed because, after getting the serious generative tests running, we were hitting errors like:

    TrinoUserError(
        type=USER_ERROR,
        name=QUERY_HAS_TOO_MANY_STAGES,
        message=(
            "Number of stages in the query (178) exceeds the allowed "
            "maximum (150). If the query contains multiple aggregates "
            "with DISTINCT over different columns, please set the "
            "'mark_distinct_strategy' session property to 'none'. If the "
            "query contains WITH clauses that are referenced more than "
            "once, please create temporary table(s) for the queries in "
            "those clauses.",
        ),
        query_id=20230922_115126_08994_ksi7v,
    )

### DIFF
--- a/ehrql/query_engines/trino.py
+++ b/ehrql/query_engines/trino.py
@@ -6,6 +6,7 @@ from ehrql.query_engines.base_sql import BaseSQLQueryEngine, get_cyclic_coalesce
 from ehrql.query_engines.trino_dialect import TrinoDialect
 from ehrql.query_model.nodes import Position
 from ehrql.utils.sqlalchemy_query_utils import (
+    CreateTableAs,
     GeneratedTable,
     InsertMany,
 )
@@ -131,5 +132,16 @@ class TrinoQueryEngine(BaseSQLQueryEngine):
         ]
         table.cleanup_queries = [
             sqlalchemy.schema.DropTable(table),
+        ]
+        return table
+
+    def reify_query(self, query):
+        table_name = f"ehrql_{self.global_unique_id}_tmp_{self.get_next_id()}"
+        table = GeneratedTable.from_query(table_name, query)
+        table.setup_queries = [
+            CreateTableAs(table, query),
+        ]
+        table.cleanup_queries = [
+            sqlalchemy.schema.DropTable(table, if_exists=True),
         ]
         return table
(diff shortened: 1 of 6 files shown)


ehrql/query_engines/trino.py
```diff
@@ -133,3 +133,17 @@ def create_inline_patient_table(self, columns, rows):
...
+        columns = [
+            sqlalchemy.Column(c.name, c.type, key=c.key) for c in query.selected_columns
+        ]
+        table = GeneratedTable(table_name, sqlalchemy.MetaData(), *columns)
+        table.setup_queries = [
+            CreateTableAs(table, query),
+        ]
+        table.cleanup_queries = [
+            sqlalchemy.schema.DropTable(table, if_exists=True),
+        ]
```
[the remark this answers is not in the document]
evansd: This is a good idea, and I have added such a test parameterised across all query engines.

tests/unit/utils/test_sqlalchemy_query_utils.py
```diff
@@ -192,6 +192,27 @@ def test_get_setup_and_cleanup_queries_with_insert_many():
...
+        "\tstring VARCHAR\n"
+        ")"
+    )
+
+
+def test_generated_table_from_query_with_metadata():
+    metadata = sqlalchemy.MetaData()
+    query = sqlalchemy.select(sqlalchemy.literal(1).label("number"))
+    table = GeneratedTable.from_query("some_table", query, metadata=metadata)
+    assert table.metadata is metadata
```
inglesp, line 215: Is this a test to appease coverage?  What would actually break if you couldn't pass metadata to the function?

inglesp (approved):

tests/unit/utils/test_sqlalchemy_query_utils.py
evansd: Right now, I don't think anything would break. But if we're creating novel SQLAlchemy elements I'd like them to be good SQLAlchemy citizens where that's a low effort thing to achieve.

inglesp: OK!

PR #1618, opened 2023-09-26
Work around memory leaks in long-running gentests

evansd:
To work around memory leak issues in our database containers (or possibly badly configured garbage collection, but the effect is the same) we run the tests in multiple smaller batches, recreating the containers between each batch.

This also allows us to set the amount of testing we do based on the time it takes, rather than trying to craft an appropriate number of examples that fits within the allotted runtime.

We now also run solely the actual generative test, rather than repeatedly running the surrounding meta tests.


inglesp (approved):
Neat

PR #1626, opened 2023-09-27
Drive-by enhancements to docs

iaindillingham:
Two drive-by enhancements to the docs. See the commit messages for their rationales 🙂 Both draw on the rather dense, but nevertheless useful [_Developing Quality Technical Information: A Handbook for Writers and Editors_](https://learning.oreilly.com/library/view/developing-quality-technical/9780133119046/).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,7 +6,8 @@ and execute them with the command line interface to generate **datasets with one
 ehrQL allows researchers to access data sources from primary and secondary care,
 as well as from organisations such as the Office for National Statistics (ONS).
 
-!!! info "Why did we create a new query language for healthcare research?  Read more in [this blog post](https://www.bennett.ox.ac.uk/blog/2023/09/why-ehrql/)."
+Why did we create a new query language?
+You can find out about ehrQL's [philosophy and design goals](https://www.bennett.ox.ac.uk/blog/2023/09/why-ehrql/) on the Bennett Institute blog.
 
 ## ehrQL's documentation
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,5 +1,5 @@
 *ehrQL* (rhymes with *circle*), the **Electronic Health Records Query Language**,
-is a query language and command line interface designed for the specific application domain of EHR data.
+is a query language and command line interface that we designed for the specific application domain of EHR data.
 Researchers write **dataset definitions** with the query language
 and execute them with the command line interface to generate **datasets with one row per patient**.
 


docs/index.md
```diff
@@ -6,7 +6,7 @@ and execute them with the command line interface to generate **datasets with one
 ehrQL allows researchers to access data sources from primary and secondary care,
 as well as from organisations such as the Office for National Statistics (ONS).
 
-!!! info "Why did we create a new query language for healthcare research?  Read more in [this blog post](https://www.bennett.ox.ac.uk/blog/2023/09/why-ehrql/)."
+You can find out [why researchers should use ehrQL](https://www.bennett.ox.ac.uk/blog/2023/09/why-ehrql/) on the Bennett Institute blog.
```
madwort, line 9: As well as being a good idea, it's also written in one of our organisation policy documents that we should always do this.

madwort (approved):

docs/index.md
iaindillingham:
```suggestion
You can find out about ehrQL's [philosophy and design goals](https://www.bennett.ox.ac.uk/blog/2023/09/why-ehrql/) on the Bennett Institute blog.
```

docs/index.md
```diff
@@ -1,12 +1,12 @@
 *ehrQL* (rhymes with *circle*), the **Electronic Health Records Query Language**,
-is a query language and command line interface designed for the specific application domain of EHR data.
+is a query language and command line interface that was designed for the specific application domain of EHR data.
```
evansd, line 2: Was the original really unclear? Obviously, excessively elliptical constructions can be problematic; but to me that modified sentence is clunkier but no easier to parse.

iaindillingham: I've been reading _Developing Quality Technical Information_ and also thinking more about teaching/learning, which includes thinking more about people who don't speak English as their first language. I don't think the original was unclear; I was, however, persuaded that the revision was clearer when I saw a similar example translated into French and German, and how the "that was" made the translations more accurate.

docs/index.md
iaindillingham: My reasoning is that _[Link to answer]_ leads to _Click here_ and _Click here_ is undesirable. However, avoiding _Click here_ is cumbersome, hence the rewrite.

Let's think about the question. What kind of reader would ask, "Why did you decide to create a new language?" What kind of reader would ask, "What are ehrQL's philosophy and design goals?" I'd argue that this section of the docs should be pitched to the second kind of reader, rather than the first.

docs/index.md
evansd: Right yes, this in an interesting point. I wonder if we might end up needing to explicitly distinguish different sorts of writing, with different sorts of goal and audience, within the Bennett Institute. There's definitely a place for maximising non-native readability (of which reference docs would be the obvious example) but I wouldn't  want to eviscerate all personality from our writing more generally.

evansd (approved):

docs/index.md
iaindillingham: Suggestion accepted 🙂

PR #1627, opened 2023-09-27
Restructure the language reference

### LINKED ISSUE #1623: Restructure the language reference
The structure of the language reference could better reflect how a user encounters the language in their dataset definitions:

- Dataset
    - create_dataset
    - Dataset
- Frames
- Series
- Date Arithmetic
- Codelists
    - codelist_from_csv
- Functions
    - case
    - maximum_of
    - minimum_of
- Measures

In the above, Dataset and create_dataset move from General to Dataset; codelist_from_csv moves from General to Codelists; and General is renamed Functions.

iaindillingham:
Closes #1623


evansd (approved):
Looks great!

Were you also going to rename "Date Arithmetic" to "Dates"? I think you mentioned that in the original ticket and it seemed a good idea to me.

As an aside, I'd find it easier to review if the source file changes were separate from the compiled docs changes, which I tend to do as a single commit at the end. However I'm aware that currently our pre-commit hook _prevents_ you from doing things like that. I wonder if that's an argument for removing it? We could have a separate "compile docs" check in CI which would fail quickly if you forgot to rebuild the docs. I guess it depends how often people forget to do it.

iaindillingham:
> Were you also going to rename "Date Arithmetic" to "Dates"?

I was, and then I read that section properly and realized that it _really is_ about date arithmetic. For some reason, I thought the content had changed, but the title hadn't. I guess there's an argument that the prose is about date arithmetic and the docstrings are about dates, but that argument seemed pedantic.

iaindillingham:
> I'd find it easier to review if the source file changes were separate from the compiled docs changes

Yup, I agree with you. I don't think I'd like to JDI, but will create an issue.

PR #1634, opened 2023-09-29
Remove `check-docs` pre-commit hook

### LINKED ISSUE #1632: Remove `check-docs` pre-commit hook
It's easier to review changes to the documentation when changes to the source files and changes to the compiled files are in separate commits (#1627). However, the `check-docs` pre-commit hook makes separating changes to source/compiled files hard: essentially, you have to bypass _all_ the pre-commit hooks with `git commit --no-verify` (`git commit -n`).

Bypassing the `check-docs` pre-commit hook isn't especially error-prone, because the similarly named GitHub action (_.github/workflows/check-docs.yml_) checks that the compiled files reflect the source files. Consequently, if we removed the pre-commit hook, then we could more easily separate changes to source/compiled files, and still catch cases where the compiled files did not reflect the source files.

iaindillingham:
Closes #1632

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,14 +13,6 @@ repos:
       types: [python]
       require_serial: true
 
-    - id: check-docs
-      name: check-docs
-      entry: just precommit-generate-docs
-      language: system
-      types: [python]
-      always_run: true
-      require_serial: true
-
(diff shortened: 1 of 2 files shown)


evansd (approved):
Worth removing the `just precommit-generate-docs` command as well?

PR #1639, opened 2023-10-02
fix: Replace negative IMD values with NULLs

### LINKED ISSUE #1548: Replace negative IMD values with NULLs
In TPP, -1 is used to indicate a missing IMD value.  We should replace this with NULL in the [table definition in the TPP backend](https://github.com/opensafely-core/ehrql/blob/99cbe958693eb3c7f1c63dae5d52647f450b84e0/ehrql/backends/tpp.py#L158).  As with #1523, we should check the performance impact before making any changes.

There's unlikely to be a problem with backwards compatibility, because if we write a dataset to arrow, we'll try to use an unsigned int for this column (because of [this constraint](https://github.com/opensafely-core/ehrql/blob/99cbe958693eb3c7f1c63dae5d52647f450b84e0/ehrql/tables/beta/tpp.py#L142)) and it'll crash.  But it's possible users are writing to CSV, so we should double check, through GH code search.

iaindillingham:
Although the issue mentions negative IMD values, the only negative IMD values are -1s. So, we can use NULLIF to replace -1s with NULLs.

I've checked the performance impact of this change by comparing the execution time of the new query (NULLIF) to the old query.

The first template query counted the number of rows where imd_rounded was GTE zero. In this case, there was an 88ms difference in elapsed time between new and old, favouring old.

The second template query counted the number of rows where imd_rounded was NULL (new query) or was -1 (old query). In this case, there was a 33ms difference in elapsed time between new and old, favouring new.

Fixes #1548

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -31,7 +31,7 @@ class TPPBackend(BaseBackend):
                 CAST(addr.EndDate AS date) AS end_date,
                 addr.AddressType AS address_type,
                 addr.RuralUrbanClassificationCode AS rural_urban_classification,
-                addr.ImdRankRounded AS imd_rounded,
+                NULLIF(addr.ImdRankRounded, -1) AS imd_rounded,
                 CASE
                     WHEN addr.MSOACode NOT IN ('NPC', '') THEN addr.MSOACode
                 END AS msoa_code,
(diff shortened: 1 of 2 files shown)


rebkwok (approved):
Are there any ehrQL studies using IMD that are expecting -1 values and need to be updated?

iaindillingham:
> Are there any ehrQL studies using IMD that are expecting -1 values and need to be updated?

Great question! Aside from the OpenPROMPT studies, three studies seem to extract rounded IMD without first grouping around quintiles (the examples I've seen would be unaffected). I've let researchers know,[^1] as I don't think we're checking every repo and opening PRs.[^2]

[^1]: https://bennettoxford.slack.com/archives/C02HJTL065A/p1696323936511239
[^2]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1696321019637789

PR #1640, opened 2023-10-04
Rename some SUS related items

StevenMaude:
See individual commit messages for the reasoning.

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -22,10 +22,10 @@ from ehrql.tables.beta.tpp import (
     medications,
     occupation_on_covid_vaccine_record,
     ons_deaths,
-    opa,
     opa_cost,
     opa_diag,
     opa_proc,
+    opa_raw,
     open_prompt,
     patients,
     practice_registrations,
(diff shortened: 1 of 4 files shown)


StevenMaude:
Questions:
* Do we need to rename all the SUS tables to `_raw` for consistency? It doesn't feel like we need to do this; it's only the `opa_raw` table that has values that we might remap in future. But perhaps we want to selectively add columns and move them from the `_raw` table to a finalised table. This would require updating a few studies that use these `opa*` tables.
* Should we rename `emergency_care_attendances` to `ec_attendances` or `ec_diagnoses` or similar? This would be for consistency with the other `ec` tables. It does not correspond to a table in TPP's schema: it's a join of the `EC` table and the `EC_Diagnosis` table.

StevenMaude:
thought: Two things about how we're naming things that are identifiers at the moment:

`addresses.address_id`
`apcs.apcs_ident`
`apcs_cost.apcs_ident`
`ec.ec_ident`
`emergency_care_admissions.id` (which this PR proposes to change)
`hospital_admissions.id`
`vaccinations.vaccination_id`

1. They are inconsistent.
2. Some have the repetition of the table source in the identifier name: `apcs_cost.ident` would be cleaner than `apcs_cost.apcs_ident`.

Maybe that's fine though because of:

* the original naming in the source data
* if they actually refer to different kinds of identifier
* if there are ever actually multiple kinds of identifiers in a table

StevenMaude:
I reviewed uses of these names in the `opensafely` organization and could not find any. Can recheck before merge.

StevenMaude:
Discussed with @inglesp and we're going to close this one. Some of the surrounding data needs reviewing anyway.

Issue #1325, opened 2023-06-15
Test all the ehrQL documentation code examples


StevenMaude:
#1317 replaces the documentation content with that from the wiki.

The existing test "framework" simply ran all the dataset definitions and checked the outputs hadn't changed.

For simplicity, this will be dropped from the content temporarily.

We will need to do something similar for these examples. However, it is a little more complicated than previously. There are the categories of examples that we should cover

* working dataset definitions (do they give the correct output, and run without error?)
* (deliberately) non-working dataset definition (do they fail as expected?)
* incomplete dataset definition; some examples use `codelist_from_csv(XXX)`, where `XXX` is the literal string `XXX` and not a valid argument (can we fix these up?)
* Python prompt examples (do they give the correct output, and run without error?)

StevenMaude:
This probably needs #1342 completing first.

StevenMaude:
I made this comment in #1342, but pulled out to this issue as more relevant here:

> There are MkDocs plugins that will pull code out for testing, but I don't think that's probably going to work in our case, because:
>
> * we don't just have Python code in any case; they could be Python console examples
> * we have multiple ways of running that code, that need specifying
> * we don't need to always include all of the code for a full example, in our documentation
> * we sometimes deliberately want to include failures

We probably also have the requirement that:

* the code is formatted and linted: #1346, which is trickier to do in code blocks (although not impossible)
* the outputs are persisted, and validated 

Let's just have a quick look at what's out there for MkDocs, to confirm this:

* https://github.com/koaning/mktestdocs — more for checking that code blocks run, though you can asserts as tests, and I think those asserts would just end up in the code blocks
* https://github.com/pawamoy/markdown-exec — more for generating content from code blocks 
* https://github.com/entangled/mkdocs-plugin — possibly the more flexible of all of these options, you can specify Makefiles to build included content; but I guess won't handle errors

The other problem is that all these solutions presumably run on documentation build, which would be a little slow to process several OpenSAFELY projects and dataset definitions.

StevenMaude:
We're now testing the complete dataset definitions following #1648, so the initial work for this is done.

It may be still valid to test the other categories of code examples mentioned here.

PR #1648, opened 2023-10-10
test: Check complete ehrQL dataset definitions in docs

StevenMaude:
# Summary

This is an alternative to #1475 which separated out code examples as a prelude to testing them.
Rather than separating out all the included code examples, we test:

* complete dataset definitions — included as `ehrql` code blocks in the ehrQL documentation Markdown source under the `docs/` directory
* dataset definitions in `.py` modules anywhere under the `docs/` directory

It partially addresses #1325 by testing the complete dataset definition examples.

As a result of this testing, most of the existing examples have been validated as working in this PR, and a few corrected in #1677.

The alternative would be to use an already existing package to do this. I [summarised several of these in #1475](https://github.com/opensafely-core/ehrql/pull/1475#issuecomment-1725419508). Unfortunately, most of them don't really do exactly what we want. So having our own approach, without adding extra dependencies, seems reasonable. **We could re-review these options in future and see if any do become more suitable.**

# What gets tested

## Code blocks labelled as `ehrql`

It tests ehrQL example code blocks that are marked with the `ehrql` syntax. This is a simple example:

``````
```ehrql
…
```
``````

These code blocks still retain Python syntax highlighting. The only difference is that they get extracted and tested by the test suite.

**This test code uses the extension that extracts these code blocks out. Therefore, if a code block displays correctly in the ehrQL documentation, it should also be tested.** Allowed code block syntax in our current documentation configuration goes beyond "a line starting with three backticks"; see the [SuperFences documentation](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/).

## Python source dataset definitions

Full dataset definitions in the `docs/` directory are tested.

:warning: It is currently assumed that any `*.py` file in the `docs/` directory is a working dataset definition for now. We could establish a filename convention or comment marker in the file to indicate it should be tested.

## What isn't tested

There are other code-related examples included in the ehrQL documentation that are not checked:

* Sections or snippets of dataset definitions that are incomplete
* The contents of datasets output from dataset definitions
* Dataset definitions that deliberately fail
* The tracebacks of failing dataset definitions 
* Python interactive console sessions
* OpenSAFELY projects
* Commands given at the terminal

## How search and testing of examples works

Checking is rudimentary and works as follows for inline dataset definition examples:

1. The test code extracts code blocks from any `*.md` files in `docs/`.
1. The ehrQL `generate_dataset` function is run against the code blocks. This uses random dummy data.
1. We check there's some valid output CSV data that we can iterate through, without checking what the data is.

Broadly, the same check takes place for the Python dataset definitions:

1. The test code searches the documentation directory for `*.py` files.
1. The ehrQL `generate_dataset` function is run against the code from each file. This uses random dummy data.
1. We check there's some valid output CSV data that we can iterate through, without checking what the data is.

## Workarounds that are used

* `codelist_from_csv()` gets monkeypatched so we just return some arbitrary codelist codes, and the validation of codes is patched out. This could be improved by establishing a convention of column names in the documentation examples that are use to detect which test CSV codes should be used: for example, `codelist_from_csv(filename="my_file", column="snomedct")`

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -71,7 +71,10 @@ markdown_extensions:
           format: !!python/name:pymdownx.superfences.fence_code_format
   - toc:
       permalink: "🔗"
-  - pymdownx.highlight
+  - pymdownx.highlight:
+      extend_pygments_lang:
+        - name: ehrql
+          lang: python
   - pymdownx.superfences
   - pymdownx.tabbed:
       alternate_style: true
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -75,7 +75,6 @@ markdown_extensions:
       extend_pygments_lang:
         - name: ehrql
           lang: python
-  - pymdownx.superfences
   - pymdownx.tabbed:
       alternate_style: true
   - pymdownx.keys
--- a/tests/docs/test_complete_examples.py
+++ b/tests/docs/test_complete_examples.py
@@ -1,3 +1,4 @@
+import copy
 import csv
 import inspect
 import unittest.mock
@@ -10,6 +11,21 @@ import mkdocs.config
 import pytest
 
 import ehrql.main
+
+
+def get_markdown_extension_configuration():
+    """Returns a dictionary representing the mkdocs.yml Markdown extension
+    configuration.
+
+    It should only be required to run this function once,
+    at the start of these tests."""
+    config_path = Path(__file__).parents[2] / "mkdocs.yml"
+    config = mkdocs.config.load_config(config_file_path=str(config_path))
+    assert "pymdownx.superfences" in config["markdown_extensions"]
+    return config["mdx_configs"]
+
+
+MARKDOWN_EXTENSION_CONFIGURATION = get_markdown_extension_configuration()
 
 
 @dataclass
@@ -60,18 +76,18 @@ class MarkdownFenceExtractor:
         """Retrieves the existing extensions settings from the mkdocs.yml
         configuration, replacing any custom SuperFences fences with a special
         test custom fence to extract all fences."""
-        config_path = Path(__file__).parents[2] / "mkdocs.yml"
-        config = mkdocs.config.load_config(config_file_path=str(config_path))
-        assert "pymdownx.superfences" in config["markdown_extensions"]
-        config["mdx_configs"]["pymdownx.superfences"]["custom_fences"] = [
+        config = copy.deepcopy(MARKDOWN_EXTENSION_CONFIGURATION)
+        config["pymdownx.superfences"]["custom_fences"] = [
             {
                 # "name" specifies fences to extract.
+                # "*" indicates fences unhandled by other custom fences;
+                # as we have no other custom fences, "*" processes all fences.
                 "name": "*",
                 "class": "test",
                 "format": self._fence_null_format,
             },
         ]
-        return config["mdx_configs"]
+        return config
 
     def _extract_fences(self, content):
         markdown.Markdown(
--- a/tests/docs/test_complete_examples.py
+++ b/tests/docs/test_complete_examples.py
@@ -1,9 +1,7 @@
 import csv
 import inspect
-import typing
 import unittest.mock
 import uuid
-from collections.abc import Generator, Iterator
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -28,23 +26,18 @@ class MarkdownFenceExtractor:
     See https://facelessuser.github.io/pymdown-extensions/extensions/superfences/
     """
 
-    def __init__(self, content: str) -> None:
-        self.fences: list[MarkdownFence] = []
-        self.extension_configs: typing.Mapping[
-            str, typing.Any
-        ] = self._configure_superfences()
+    def __init__(self, content):
+        self.fences = []
+        self.extension_configs = self._configure_superfences()
         self._extract_fences(content)
 
     def _fence_null_format(
-        # Argument types are taken from mkdocs-code-validator,
-        # which has an MIT license,
-        # but they will be removed before merge.
         self,
-        src: str,
-        language: str,
-        css_class: str | None,
-        options: typing.Mapping[str, typing.Any],
-        md: markdown.Markdown,
+        src,
+        language,
+        css_class,
+        options,
+        md,
         **kwargs,
     ) -> str:
         """Extract the fences in the same way
@@ -80,7 +73,7 @@ class MarkdownFenceExtractor:
         ]
         return config["mdx_configs"]
 
-    def _extract_fences(self, content: str) -> None:
+    def _extract_fences(self, content):
         markdown.Markdown(
             extensions=["pymdownx.superfences"],
             extension_configs=self.extension_configs,
@@ -101,22 +94,20 @@ class DatasetDefinitionExample:
     fence_number: int | None
     source: str
 
-    def relative_path(self) -> Path:
+    def relative_path(self):
         """Return the relative path of the dataset definition source file
         to the source code root."""
         source_code_path = Path(__file__).parents[2]
         return self.path.relative_to(source_code_path)
 
 
-def discover_paths(glob_string: str) -> Generator[Path, None, None]:
+def discover_paths(glob_string):
     """Generate a list of matching files for a glob in the documentation source path."""
     docs_path = Path(__file__).parents[2] / "docs"
     return docs_path.glob(glob_string)
 
 
-def find_complete_ehrql_examples_in_markdown(
-    file: typing.TextIO,
-) -> Iterator[DatasetDefinitionExample]:
+def find_complete_ehrql_examples_in_markdown(file):
     """Yields extracted code blocks labelled as ```ehrql from a Markdown file.
 
     Incomplete ehrQL dataset definitions should be labelled as ```python,
@@ -133,9 +124,7 @@ def find_complete_ehrql_examples_in_markdown(
             yield example
 
 
-def generate_complete_ehrql_examples() -> (
-    Generator[DatasetDefinitionExample, None, None]
-):
+def generate_complete_ehrql_examples():
     """Yields all complete ehrQL DatasetDefinitionExamples from the Markdown documentation."""
     markdown_paths = list(discover_paths("**/*.md"))
     assert len(markdown_paths) > 0, "No Markdown files found"
@@ -158,7 +147,7 @@ def generate_complete_ehrql_examples() -> (
         )
 
 
-def create_example_test_case_id(example: DatasetDefinitionExample) -> str:
+def create_example_test_case_id(example):
     """Returns a test case ID for pytest from a specific DatasetDefinitionExample."""
     test_id = f"{example.relative_path()}"
     if example.fence_number is not None:
@@ -166,7 +155,7 @@ def create_example_test_case_id(example: DatasetDefinitionExample) -> str:
     return test_id
 
 
-def validate_dataset_output(dataset_path: Path) -> None:
+def validate_dataset_output(dataset_path):
     """Validates that an output dataset file is a CSV."""
     with open(dataset_path) as f:
         csv_content = f.readlines()
@@ -189,9 +178,7 @@ class DatasetDefinitionTestError(Exception):
     generate_complete_ehrql_examples(),
     ids=create_example_test_case_id,
 )
-def test_ehrql_generate_dataset_example(
-    tmp_path: Path, example: DatasetDefinitionExample
-) -> None:
+def test_ehrql_generate_dataset_example(tmp_path, example):
     tmp_filename_base = str(uuid.uuid4())
 
     tmp_dataset_definition_path = tmp_path / (tmp_filename_base + ".py")
--- a/tests/docs/test_complete_examples.py
+++ b/tests/docs/test_complete_examples.py
@@ -1,0 +1,292 @@
+import csv
+import inspect
+import typing
+import unittest.mock
+import uuid
+from collections.abc import Generator, Iterator
+from dataclasses import dataclass
+from pathlib import Path
+
+import markdown
+import mkdocs.config
+import pytest
+
+import ehrql.main
+
+
+@dataclass
+class MarkdownFence:
+    """Represents a Markdown fence."""
+
+    source: str
+    language: str
+
+
+class MarkdownFenceExtractor:
+    """Extracts fences from a markdown.Markdown object using the SuperFences extension.
+
+    See https://facelessuser.github.io/pymdown-extensions/extensions/superfences/
+    """
+
+    def __init__(self, content: str) -> None:
+        self.fences: list[MarkdownFence] = []
+        self.extension_configs: typing.Mapping[
+            str, typing.Any
+        ] = self._configure_superfences()
+        self._extract_fences(content)
+
+    def _fence_null_format(
+        # Argument types are taken from mkdocs-code-validator,
+        # which has an MIT license,
+        # but they will be removed before merge.
+        self,
+        src: str,
+        language: str,
+        css_class: str | None,
+        options: typing.Mapping[str, typing.Any],
+        md: markdown.Markdown,
+        **kwargs,
+    ) -> str:
+        """Extract the fences in the same way
+        as the SuperFences extension does, and make them accessible to the test code.
+
+        Returns an empty string.
+
+        This null formatter exists only for this purpose.
+        See https://facelessuser.github.io/pymdown-extensions/extensions/superfences/#formatters
+
+        "All formatters should return a string as HTML."
+
+        We don't require the formatted text,
+        only that this method is run and we can access the source
+        and language."""
+        self.fences.append(MarkdownFence(source=src, language=language))
+        return ""
+
+    def _configure_superfences(self):
+        """Retrieves the existing extensions settings from the mkdocs.yml
+        configuration, replacing any custom SuperFences fences with a special
+        test custom fence to extract all fences."""
+        config_path = Path(__file__).parents[2] / "mkdocs.yml"
+        config = mkdocs.config.load_config(config_file_path=str(config_path))
+        assert "pymdownx.superfences" in config["markdown_extensions"]
+        config["mdx_configs"]["pymdownx.superfences"]["custom_fences"] = [
+            {
+                # "name" specifies fences to extract.
+                "name": "*",
+                "class": "test",
+                "format": self._fence_null_format,
+            },
+        ]
+        return config["mdx_configs"]
+
+    def _extract_fences(self, content: str) -> None:
+        markdown.Markdown(
+            extensions=["pymdownx.superfences"],
+            extension_configs=self.extension_configs,
+        ).convert(content)
+
+
+@dataclass
+class DatasetDefinitionExample:
+    """Stores details of a complete ehrQL dataset definition example.
+
+    The origin of such an example may be a Markdown fence,
+    or a standalone Python module."""
+
+    path: Path
+    # This fence number count includes all fences,
+    # not just the ehrQL fences.
+    # Standalone Python modules are not given a fence number.
+    fence_number: int | None
+    source: str
+
+    def relative_path(self) -> Path:
+        """Return the relative path of the dataset definition source file
+        to the source code root."""
+        source_code_path = Path(__file__).parents[2]
+        return self.path.relative_to(source_code_path)
+
+
+def discover_paths(glob_string: str) -> Generator[Path, None, None]:
+    """Generate a list of matching files for a glob in the documentation source path."""
+    docs_path = Path(__file__).parents[2] / "docs"
+    return docs_path.glob(glob_string)
+
+
+def find_complete_ehrql_examples_in_markdown(
+    file: typing.TextIO,
+) -> Iterator[DatasetDefinitionExample]:
+    """Yields extracted code blocks labelled as ```ehrql from a Markdown file.
+
+    Incomplete ehrQL dataset definitions should be labelled as ```python,
+    and not with ```ehrql."""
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: tests/docs/test_complete_examples.py [hunk 1]
```diff
@@ -0,0 +1,185 @@
...
+    markdown_paths = discover_markdown_paths()
+    assert len(markdown_paths) > 0, "No Markdown files found"
+
+    for p in markdown_paths:
+        with open(p) as f:
+            yield from find_complete_ehrql_examples_in_markdown(f)
+
+
+def create_example_test_case_id(example: DatasetDefinitionExample) -> str:
+    """Generate helpful test case IDs for pytest from a DatasetDefinitionExample."""
```
StevenMaude, line 130:
```suggestion
    """Create helpful test case IDs for pytest from a DatasetDefinitionExample."""
```

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def discover_markdown_paths() -> list[str]:
+    """Find Markdown files in the documentation source path."""
+    docs_path = Path(__file__).parents[2] / "docs"
+    # Note that we're currently not excluding any paths like the generated docs.
+    # Do we need to? I don't think so, because if examples aren't labelled as
+    # ```ehrql then they won't be checked.
+    markdown_paths = list(docs_path.glob("**/*.md"))
+    return markdown_paths
```
StevenMaude, line 35:
```suggestion
    return list(docs_path.glob("**/*.md"))
```

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,185 @@
...
+    #
+    # ```ehrql
+    # …
+    # EOF
+    in_ehrql_block = False
+    example = None
+
+    for line_number, line in enumerate(file, start=1):
+        # Check if the line marks the start of a ehrQL code block?
+        if line.startswith("```ehrql") or line.startswith("``` ehrql"):
```
StevenMaude, line 70: thought: I'd started with just trying to get the content of interest out. We could instead pull out *all* code blocks using the backtick formatting, and then filter those for the syntax(es) we want at the end. That might be `ehrql-dataset` and `ehrql-measure`, potentially.

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,185 @@
...
+            if in_ehrql_block:
+                # TODO: Should this fail completely?
+                # NB: This drops the incomplete block entirely.
+                warnings.warn(
+                    f"In {example.source_filename}, a new ehrQL code block starts on line {line_number} before the previous one has been closed."
+                )
```
StevenMaude, line 76: thought: This is only checking this condition for `ehrql` code blocks. It's not going to warn if a different kind of code block is not closed:

``````
```python
some code
```ehrql
more code
```
``````

But the important context here is that this means that an example we might expect to check will not be checked, because it's incomplete, so maybe that's not too important.

File: tests/docs/test_complete_examples.py [hunk 5]
```diff
@@ -0,0 +1,185 @@
...
+
+
+@pytest.mark.parametrize(
+    "example", generate_complete_ehrql_examples(), ids=create_example_test_case_id
+)
+# TODO: This code only handles examples that run with generate_dataset.
+# Do we need to handle measures?
+def test_ehrql_dataset_example(example: DatasetDefinitionExample) -> None:
+    """Check that an ehrQL generate dataset example works."""
+    # We need to have actual file paths to pass to generate_dataset.
```
StevenMaude, line 141: context: This is because the dataset definition loading uses [`importlib.util.spec_from_file_location`](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location), which takes a file path and then loads it, so we can't pass in a `StringIO` or similar.

Maybe there's a better way?

tests/docs/test_find_docs_examples.py
```diff
@@ -0,0 +1,82 @@
...
+            test_complete_examples.find_complete_ehrql_examples_in_markdown(example)
+        )
+
+    assert len(result) == 1
+    assert result[0] == expected
+
+
+def test_find_docs_examples_warning_end_of_file():
+    example = StringIO("```ehrql")
+    example.name = "test_find_docs_examples_warning_open_block"
```
StevenMaude, line 53:
```suggestion
    example.name = "test_find_docs_examples_warning_end_of_file"
```

tests/docs/test_find_docs_examples.py
```diff
@@ -0,0 +1,82 @@
...
+    with pytest.warns():
+        result = list(
+            test_complete_examples.find_complete_ehrql_examples_in_markdown(example)
+        )
+    assert len(result) == 0
+
+
+def test_find_docs_examples_warning_end_of_file_with_code():
+    example = StringIO("```ehrql\nsome code\n```\n```ehrql")
+    example.name = "test_find_docs_examples_warning_open_block"
```
StevenMaude, line 65:
```suggestion
    example.name = "test_find_docs_examples_warning_end_of_file_with_code"
```

tests/docs/test_find_docs_examples.py
```diff
@@ -0,0 +1,82 @@
...
+
+import pytest
+
+from . import test_complete_examples
+
+
+def test_find_docs_examples_working():
+    example = StringIO("```ehrql\nsome code\n```")
+    # Unlike file objects, StringIO objects do not have a name.
+    # We use this in the code being tested, to save somewhat redundantly passing the name.
```
StevenMaude, line 11:
```suggestion
    # We access the file's name in the relevant code being tested, to save somewhat redundantly passing the name.
```

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,185 @@
...
+import ehrql.main
+
+
+@dataclass
+class DatasetDefinitionExample:
+    """Store details of a dataset definition example."""
+
+    source_filename: str
+    start: int = None
+    end: int = None
```
StevenMaude, line 21:
```suggestion
    start_line: int = None
    end_line: int = None
```

StevenMaude:
question: I was going to check the newly added dataset definition under "writing-a-more-complex-dataset-definition". However, on looking, I found [this is actually a separate snippet-included file](https://github.com/opensafely-core/ehrql/blob/236f9d72a3ed95640b205370ea6d442cc924b9ec/docs/tutorial/writing-a-more-complex-dataset-definition/index.md#L8) and *not* written inline.

So we'd need an extra test here to also run against any Python modules that are dataset definitions, and some convention (filename?) of specifying unambiguously that a Python file is a dataset definition to be tested.

What are the actual requirements here? Do we need to handle:

* only inline written dataset definitions in code blocks, as this PR does?
* only file-included dataset definitions as in #1475?
* both?

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,186 @@
...
+        temp_dataset_definition_path = temp_directory_path / (
+            temp_filename_base + ".py"
+        )
+        with open(temp_dataset_definition_path, "w") as f:
+            f.write(example.read())
+
+        temp_dataset_path = temp_directory_path / (temp_filename_base + ".csv")
+
+        # TODO: can we monkeypatch codelist_from_csv so that we can test documentation examples
+        # that use a codelist?
```
StevenMaude, line 169: thought: This might work, I think, for the simple examples.

tests/docs/__init__.py
StevenMaude, line 1: question: Is this a sensible path for these tests to live?

Or some other path?

There is a `tests/unit/docs`, but these don't feel like unit tests, they are end-to-end tests of the examples.

tests/docs/test_complete_examples.py
StevenMaude: I'm having a go at this anyway, because it might not be too much work.

tests/docs/__init__.py
StevenMaude: And should there be a `just` command to specifically check the examples?

StevenMaude:
Almost all the dataset definitions in the examples page are now tested. This is via some slightly dubious hacks in the test code.

iaindillingham (commented):
It's great to see that you've improved the examples in response to the test. I've left several comments, but in general I'd strongly suggest that for each `TODO`, you make and document a decision, and then remove the `TODO`. If you don't, then who will?

I'm uneasy about monkeypatching `codelist_from_csv`. Would it be possible to point each call to a codelist? (We'd only need one, and it need not be a real one.)

As an aside, I stepped through the PR commit by commit, but found the narrative hard to follow, as some later commits refactor some earlier commits. Rebasing can be a massive pain, but it makes for an easier review. (Dave is a master of the narrative PR.)

inglesp (commented):
It's really good to see all the errors this has found!  I'd love to see those commits pulled out of this PR and committed as soon as possible.

A few things I'd like to discuss:

* Can we simplify the markdown parsing code?  For instance, you'll be able to pull out all `ehrql` blocks with a regex.  It might not catch all the errors that you're checking for, but we'd expect mkdocs to do that.  Can we even hook into mkdoc's markdown parser?
* Should datasets always define a population?  It would simplify our code, and wouldn't add too much to the examples.
* Can we avoid mocking by providing real CSV files?  I'm a bit allergic to mocking, perhaps irrationally, but a comment on a commit makes me think that there's a risk they're going to hide errors here.
* There's quite a lot of TODOs in the code.  I've tried to give my answers to some of them.  I'd like them to be resolved before this PR's merged though.

(My review has crossed with Iain's and I've not had the chance to review his review, so we might be saying the same thing twice -- sorry about that.)

StevenMaude:
Thanks both for the review; the suggestions are good improvements.

StevenMaude:
@iaindillingham @inglesp:

* TODOs; yes, I'll make documented decisions on these and remove them.
* > Can we avoid mocking… (of `codelist_from_csv`)?
   * I'll have another go at this to see what's possible. I'm not sure if there are issues with mismatches of the types of expected codes that the function returns, and that the dataset definitions use. Likewise with the `column` argument; if different dataset definitions could be using different column names, I'm not quite sure how that will work.

@inglesp:

* > Should datasets always define a population?
  * Yes, Iain [suggested that the examples should.](https://github.com/opensafely-core/ehrql/pull/1648#discussion_r1368896038) — which takes some of the current "magic" out from the test.
* > Can we simplify the markdown parsing code? (to use either a regex or Python Markdown) … It might not catch all the errors that you're checking for, but we'd expect mkdocs to do that. 
  * I'll have another look at this. I don't believe MkDocs catches incomplete code blocks. At least, it doesn't even warn about them. But maybe Python Markdown — which MkDocs uses — can?

StevenMaude:
The fixes to examples are in #1677, so I'll rebasing this PR to remove those commits.

StevenMaude:
I looked into offloading the code block extraction to the same Markdown tooling that the documentation uses, and [it's only possible if you drop the line numbers](https://github.com/facelessuser/pymdown-extensions/discussions/2214).

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,255 @@
...
+    markdown_paths = discover_paths("**/*.md")
+    # TODO: remove assert?
+    assert len(markdown_paths) > 0, "No Markdown files found"
+
+    for p in markdown_paths:
+        with open(p) as f:
+            yield from find_complete_ehrql_examples_in_markdown(f)
+
+    # TODO: establish a naming convention to specify dataset definition should be tested.
+    dataset_definition_source_paths = discover_paths("**/*.py")
```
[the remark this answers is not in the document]
StevenMaude: Actually, I think it's fine to leave it as is.

StevenMaude:
I've also added `define_population()` lines to the examples that are being tested and removed the workaround that patched this in for dataset definitions that didn't have it.

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,220 @@
...
+            wraps=wrapped_codelist_from_csv,
+        ),
+        # There is no codelist code that satisfies constraints of all code systems,
+        # so patch out the validity check and just pass in a fake codelist.
+        # If we adopted a standard for example column names, we could use valid codes.
+        unittest.mock.patch(
+            "ehrql.codes.BaseCode.__post_init__",
+            return_value=None,
+        ),
+    ):
```
StevenMaude, line 206: @inglesp: I've had another go at running `codelist_from_csv()` with less mocking, so the real `codelist_with_csv()` function is now at least called, albeit with patched arguments.

(I can do more to validate the filename is a correct argument.)

We always need to patch out the filename to point a temporary CSV. If we decided on a convention for these in examples: like `column="snomedct"`, we could actually pass through the `column` and `category_column`.

This would then allow actual codes to be used: we could create the temporary CSV could have these columns.

I'm not sure if this is getting too complex to be worthwhile though.

StevenMaude: (I fixed the checks here to use `inspect.signature`.)

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def find_complete_ehrql_examples_in_markdown(
+    file: typing.TextIO,
+) -> Iterator[DatasetDefinitionExample]:
+    """Yields extracted code blocks labelled as ```ehrql from a Markdown file.
+
+    Incomplete ehrQL dataset definitions should be labelled as ```python,
+    and not with ```ehrql."""
+    # There are three kinds of possible ehrql code blocks that we might see.
```
StevenMaude, line 47: [This isn't true](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/#nested-fence-format), I'll have a go at correcting this.

StevenMaude:
Because it was bugging me, I've had a go at trying out different routes to extracting the code blocks:

* The existing very naive approach which is simple, but could miss blocks:
  ````
  ```ehrql
  …
  ```
  ````
* Using the SuperFences regexes to match more flexibly as to what code blocks actually allowed in the documentation, which is more tricky to get right without introducing potential bugs or inconsistencies.
* Using the custom fences feature of SuperFences to extract the blocks.
  * This does have the advantage of not writing our own code to extract the blocks, and, because we use the same code, we know that we will extract the same blocks that the documentation "sees" when being built.
  * The downsides are: no line numbers — only the number of the block in the page, so we'd have to print any failing code instead to help identify it — and without direct warning of missing the closing blocks. (But indirectly running the code should still catch failures, because code in an unclosed example instead would get amalgamated with the next example and intervening text. Except for examples at the end of the page, but these would be complete, just unclosed, and would just be a formatting issue, not a code issue.)
  * So I'm leaning towards this right now.

Can tidy and write up a bit more next week.

StevenMaude:
On my computer, this used to take about 10-12 seconds to run these tests, and now takes over 60 seconds following the sandbox PR #1653.

It also breaks my monkeypatching of `codelist_from_csv()` so a workaround is needed.

evansd:
@StevenMaude I'd suggest not using `main.generate_dataset` to test these but rather using `loaders.load_dataset_definition_unsafe()`:
https://github.com/opensafely-core/ehrql/blob/eae0ece1e197bd089e46a51e45e0e40b20f94409/ehrql/loaders.py#L229

This will be much faster and allow your monkeypatching to work.

StevenMaude:
@evansd: I'd hope to get rid of the monkeypatching and maybe use real CSVs in future (#1694); but I'll bypass it for now; thanks.

StevenMaude:
# What's changed from when this PR was last reviewed:

* We now use the [SuperFences extension](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/) to extract the Markdown fences. This is the extension we're actually using when building the documentation to do so. This approach has several advantages, and a couple of disadvantages over the previous very crude regex —  matching ` ```ehrql ` at the start of a line — approach 
  * :heavy_plus_sign: Internally, it's still using regular expressions for matching, but they are abstracted away. We're also using an [approach that the pymdownx developer approves of](https://github.com/facelessuser/pymdown-extensions/discussions/1291).
  * :heavy_plus_sign: It is clear when previewing the documentation what will be tested: anything with an `ehrql` syntax that is displayed as a code block.
  * :heavy_plus_sign: Further, there should be no risk of including code blocks that have a valid appearance but are not tested.
  * :heavy_plus_sign: Though not implemented, we could in future repurpose the [attribute features of fences](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/#injecting-classes-ids-and-attributes) for more. (These are also in plain Python Markdown code fences, but our simple regex approach didn't pull these out.)   
  * :heavy_minus_sign: We lose the warnings for incomplete code blocks that were previously present. This is mitigated by:
    * incomplete fences at the end of a Markdown source file not being formatted as a code block, so this should eventually get spotted
    * fences that are accidentally left open but closed by a subsequent fence will include something like: two dataset definitions and intervening text, and most likely fail to run 
  * :heavy_minus_sign: We lose the direct line number referencing for Markdown fences. It is [not possible to extract these](https://github.com/facelessuser/pymdown-extensions/discussions/2214#discussioncomment-7394383). In place, I've extracted the ordinal position as "fence number", and, on error, the whole dataset definition is displayed.
 
# What I'm happy to defer

* #1694
* #1697 

These are self-contained improvements that could be made on top of this PR. This PR has been open long enough that it's probably preferable to merge it soon, to get the benefit of continuously testing some of the examples. Both of the issues describe potential self-contained improvements.

StevenMaude:
Failure is due to Cloudflare's current API glitches as of 2023-11-02:

> `Failed to get Pages project`

tests/docs/test_run_generate_dataset_example.py
```diff
@@ -0,0 +1,92 @@
...
+
+    with pytest.raises(ehrql.loaders.DefinitionError) as e:
+        test_complete_examples.test_ehrql_generate_dataset_example(tmp_path, example)
+        # This should only fail if the test fails, so skip coverage.
+        assert (e.__context__) == TypeError(  # pragma: no cover
+            "codelist_from_csv() given incorrect arguments"
+        )
+
+
+def test_run_generate_dataset_example_gives_unreadable_csv(tmp_path):
```
StevenMaude, line 70: thought: There are potentially more tests we could add here; for example, checking that ` ```{.ehrql} ` works, but as we're not using that feature yet — maybe we would if wanting to add extra inline data — I haven't.

mkdocs.yml
```diff
@@ -71,8 +71,10 @@ markdown_extensions:
           format: !!python/name:pymdownx.superfences.fence_code_format
   - toc:
       permalink: "🔗"
-  - pymdownx.highlight
-  - pymdownx.superfences
```
StevenMaude, line 75: note: We already included `pymdownx.superfences` in the lines above (that are unchanged, and not displayed by default).

tests/docs/test_run_generate_dataset_example.py
```diff
@@ -0,0 +1,91 @@
...
+            from ehrql import create_dataset
+
+            dataset = create_dataset()
+            dataset.define_population(not_a_function())
+            """
+        ),
+    )
+    with pytest.raises(test_complete_examples.DatasetDefinitionTestError) as exc_info:
+        test_complete_examples.test_ehrql_generate_dataset_example(tmp_path, example)
+    assert type(exc_info.value) is test_complete_examples.DatasetDefinitionTestError
```
StevenMaude, line 41: comment: not currently in the Pytest docs at time of writing, but [mentioned in this PR](https://github.com/pytest-dev/pytest/pull/11578)

tests/docs/test_run_generate_dataset_example.py
```diff
@@ -0,0 +1,91 @@
...
+
+            dataset = create_dataset()
+            dataset.define_population(patients.exists_for_patient())
+            """
+        ),
+    )
+    with unittest.mock.patch("ehrql.main.generate_dataset", return_value=None):
+        with pytest.raises(
+            test_complete_examples.DatasetDefinitionTestError,
+            match=r"Check of output dataset CSV failed for example",
```
StevenMaude, line 86: question: I actually wanted to use [exception notes as added to Python 3.11](https://docs.python.org/3/tutorial/errors.html) in place of a custom exception to add these messages as notes to the original exception.

pytest [apparently supports these](https://github.com/pytest-dev/pytest/issues/11223) including using `match` to check against the notes. But I was finding that pytest was only seeing the first line of a `FileNotFoundError` exception which I was trying to raise.

```
AssertionError: Regex pattern did not match.                                                                                                                                                                 
 Regex: 'Check of output dataset CSV failed for example'                                                                                                                                                     
 Input: "[Errno 2] No such file or directory: /tmp/… 
```

I don't know if I'm doing something wrong in how I was trying to raise the exception. In [this code](https://github.com/opensafely-core/ehrql/pull/1648/commits/e4b718eb4a683b31b2007329e86c3c4e18a4beb3), I was doing:

```python
try:
    validate_dataset_output(tmp_dataset_path)
except Exception as e:
    e.add_note(f"Check of output dataset CSV failed for example: \n{formatted_example}")
    raise
```

If anyone think exception notes are a better approach and knows what I'm doing wrong, I'd love to know!

tests/docs/test_complete_examples.py
```diff
@@ -0,0 +1,279 @@
...
+        assert "pymdownx.superfences" in config["markdown_extensions"]
+        config["mdx_configs"]["pymdownx.superfences"]["custom_fences"] = [
+            {
+                # "name" specifies fences to extract.
+                "name": "*",
+                "class": "test",
+                "format": self._fence_null_format,
+            },
+        ]
+        return config["mdx_configs"]
```
StevenMaude, line 74: improvement: Was tidying things up and moved this into the class. But it should only happen once for the entire test, instead of for each Markdown file.

inglesp (approved):
This is good to see.  I'm sure there are some more tweaks you could make, but I'd be happy for this to be merged now.

Issue #1523, opened 2023-08-23
Replace ceiling dates with NULL in TPP tables


rebkwok:
In some TPP tables, missing or unknown dates are represented by a ceiling date (9999-12-31), which can cause confusion or coding errors (see e.g. https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1692786760750299)

We already replace ceiling end dates in the [practice_registrations table](https://github.com/opensafely-core/ehrql/blob/da8fede4dad0ea16979e88a7e8215369d73fa749/ehrql/backends/tpp.py#L66), and consultation date in the clinical_events table (as of #1522).  For any other date fields in TPP tables we should consider whether ceiling dates should be replaced with NULL.

For each date/datetime field, we should check:
- is it nullable? If not, does the database contains ceiling dates?
- could a ceiling date for this field represent something meaningful or useful?
- will any existing ehrQL studies be affected by changing ceiling dates to NULL?
- what are the performance implications of replacing ceiling dates with NULL?

Date/datetime fields in TPP (based on database schema report):

### OPENPROMPT - Non-nullable; no ceiling dates in data
- OpenPROMPT.ConsultationDate
- OpenPROMPT.CreationDate

### ICNARC - All nullable, no ceiling dates
- ICNARC.DateOfDeath
- ICNARC.HospitalAdmissionDate
- ICNARC.HospitalDischargeDate
- ICNARC.IcuAdmissionDateTime
- ICNARC.IcuDischargeDateTime
- ICNARC.OriginalHospitalAdmissionDate
- ICNARC.OriginalIcuAdmissionDate
- ICNARC.UltimateHospitalDischargeDate
- ICNARC.UltimateIcuDischargeDate

### ISARIC_Patient_Data_Topline (all varchar, all nullable - the following are converted to date in cohort-extractor):
(Only used in cohort-extractor, ehrql uses ISARIC_New)
 - assess_or_admit_date
 - cestdat 
 - daily_dsstdat
 - daily_lbdat
 - dlvrdtc_rptestcd
 - dsstdat 
 - dsstdtc
 - hostdat 
 - hostdat_transfer
 - hosttim (cohort extractor converts to date, but the field is a timestring in format 00:00:00)
 - influenza_2021_vaccined
 - symp_or_assess_date

### ISARIC_New
All varchar, all nullable, date-like fields can contain "NA"; currently in exploration stage in ehrQL 
 
### APCS
[...]

PR #1651, opened 2023-10-12
Add `WL_Clockstops` and `WL_OpenPathways`

iaindillingham:
This adds the database table `WL_Clockstops` as the ehrQL tables `wl_clockstops_raw` and `wl_clockstops`, and the database table `WL_OpenPathways` as the ehrQL tables `wl_openpathways_raw` and `wl_openpathways`.

The raw ehrQL tables should help researchers understand the structure and semantics of the database tables (i.e. undertake _data development_). I've added their corresponding backend tables as instances of `QueryTable` rather than `MappedTable`, because the database tables contain binary columns that must be converted to string columns using custom SQL.

If we treat `wl_clockstops` as a view, then filtering each of the date columns in turn for rows that fall within a range takes from roughly 40ms to roughly 500ms. Adding another filter, such as also filtering by `priority_type_code`, takes from roughly 100ms to 1500ms. If we treat `wl_openpathways` as a view, then similar queries take about 500ms longer, possibly because many values of `wl_openpathways.current_pathway_period_start_date` are `NULL`.

I've added the constraints using the information in National Waiting List Dataset Submission v12.[^1] The categorical constraints on `priority_type_code` and `waiting_list_type` are especially useful, as they ensure that the dummy data are more realistic. As @alschaffer notes, however, in the real tables around 10% of values in each of these columns do not match the categorical constraints. Consequently, these values will be missing (`NULL`) in a real dataset.

You can find the rendered documentation at <https://iaindillingham-add-nhse-wl.databuilder.pages.dev/reference/schemas/beta.tpp/#wl_clockstops>. Thanks for your help documenting the ehrQL tables, @alschaffer.

---

~I used the National Waiting List Dataset Submission v12[^1] to help document the columns. However, I'm still unclear what Clock Stops and Open Pathways are. Could you help me describe the ehrQL tables, @alschaffer? Are there URLs to these datasets that I could add to the documentation?~

~I haven't been able to test the performance characteristics of the cooked ehrQL tables, because I cannot connect to the VPN.[^2] However, when I can, I will. Similarly, I haven't been able to check that the constraints reflect the data, rather than the documentation.~

[^1]: https://bennettoxford.slack.com/archives/C02HJTL065A/p1697015782102079?thread_ts=1694531849.195469&cid=C02HJTL065A
[^2]: https://bennettoxford.slack.com/archives/C010SJ89SA3/p1696602442369059

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -535,3 +535,35 @@ class TPPBackend(SQLBackend):
             ON vax.VaccinationName_ID = ref.VaccinationName_ID
         """
     )
+
+    wl_clockstops_raw = QueryTable(
+        # columns passed to CONVERT are likely to contain sha256 hashes
+        """
+            SELECT
+                Patient_ID AS patient_id,
+                ACTIVITY_TREATMENT_FUNCTION_CODE AS activity_treatment_function_code,
+                PRIORITY_TYPE_CODE AS priority_type_code,
+                CONVERT(
+                    varchar(max),
+                    PSEUDO_ORGANISATION_CODE_PATIENT_PATHWAY_IDENTIFIER_ISSUER,
+                    2
+                ) AS pseudo_organisation_code_patient_pathway_identifier_issuer,
+                CONVERT(
+                    varchar(max),
+                    PSEUDO_PATIENT_PATHWAY_IDENTIFIER,
+                    2
+                ) AS pseudo_patient_pathway_identifier,
+                CONVERT(
+                    varchar(max),
+                    Pseudo_Referral_Identifier,
+                    2
+                ) AS pseudo_referral_identifier,
+                Referral_Request_Received_Date AS referral_request_received_date,
+                REFERRAL_TO_TREATMENT_PERIOD_END_DATE AS referral_to_treatment_period_end_date,
+                REFERRAL_TO_TREATMENT_PERIOD_START_DATE AS referral_to_treatment_period_start_date,
+                SOURCE_OF_REFERRAL_FOR_OUTPATIENTS AS source_of_referral_for_outpatients,
+                Waiting_List_Type AS waiting_list_type,
+                Week_Ending_Date AS week_ending_date
+            FROM WL_ClockStops
+        """
+    )
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -635,3 +635,43 @@ class TPPBackend(SQLBackend):
             FROM WL_OpenPathways
         """
     )
+
+    wl_openpathways = QueryTable(
+        """
+            SELECT
+                Patient_ID AS patient_id,
+                ACTIVITY_TREATMENT_FUNCTION_CODE AS activity_treatment_function_code,
+                CONVERT(DATE, Current_Pathway_Period_Start_Date, 23) AS current_pathway_period_start_date,
+                CASE PRIORITY_TYPE_CODE
+                    WHEN '1' THEN 'routine'
+                    WHEN '2' THEN 'urgent'
+                    WHEN '3' THEN 'two week wait'
+                END AS priority_type_code,
+                CONVERT(
+                    varchar(max),
+                    PSEUDO_ORGANISATION_CODE_PATIENT_PATHWAY_IDENTIFIER_ISSUER,
+                    2
+                ) AS pseudo_organisation_code_patient_pathway_identifier_issuer,
+                CONVERT(
+                    varchar(max),
+                    PSEUDO_PATIENT_PATHWAY_IDENTIFIER,
+                    2
+                ) AS pseudo_patient_pathway_identifier,
+                CONVERT(
+                    varchar(max),
+                    Pseudo_Referral_Identifier,
+                    2
+                ) AS pseudo_referral_identifier,
+                CONVERT(DATE, REFERRAL_REQUEST_RECEIVED_DATE, 23) AS referral_request_received_date,
+                CONVERT(
+                    DATE,
+                    NULLIF(REFERRAL_TO_TREATMENT_PERIOD_END_DATE, '9999-12-31'),
+                    23
+                ) AS referral_to_treatment_period_end_date,
+                CONVERT(DATE, REFERRAL_TO_TREATMENT_PERIOD_START_DATE, 23) AS referral_to_treatment_period_start_date,
+                SOURCE_OF_REFERRAL AS source_of_referral,
+                NULLIF(Waiting_List_Type, 'NULL') AS waiting_list_type,
+                CONVERT(DATE, Week_Ending_Date, 23) AS week_ending_date
+            FROM WL_OpenPathways
+        """
+    )
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -567,3 +567,38 @@ class TPPBackend(SQLBackend):
             FROM WL_ClockStops
         """
     )
+
+    wl_clockstops = QueryTable(
+        """
+            SELECT
+                Patient_ID AS patient_id,
+                ACTIVITY_TREATMENT_FUNCTION_CODE AS activity_treatment_function_code,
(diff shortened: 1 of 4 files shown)


alschaffer:
Also, for those variables where you provided the possible values from the data dictionary (priority_type_code, waiting_list_type), I note that a not insignificant proportion (~10%) have other values. I don't know if that is something you usually note in the schema.

iaindillingham:
> Also, for those variables where you provided the possible values from the data dictionary (priority_type_code, waiting_list_type), I note that a not insignificant proportion (~10%) have other values. I don't know if that is something you usually note in the schema.

Elsewhere, we have linked to short data reports from the schema. Have you produced something similar for your data development work, @alschaffer? As it stands, the proportion with other values will appear as missing (`NULL`). Have I missed any categories, or do the other values represent data entry (or similar) errors?

alschaffer:
Here is where I've provided a summary of the data holdings and quality: https://docs.google.com/document/d/1Es1ODHeyUdIj6MXgFonZtQ2ncPC0_oaSfN8piGQc0ks/edit 

And this folder has some more granular data exploration, including frequency distributions for many variables:
[G:\Shared drives\OpenSAFELY\Data\Waiting List Data\Notebook html output](https://drive.google.com/drive/folders/10FGmVbSL1n4Ib4OVYsVecnpVK1dZNiU5?usp=drive_link)

I suspect those other categories are a combination of old codes, local codes, and typos (e.g. reversing the order of letters, inputting numbers instead of letters). If it's not too much trouble, it would be useful to include the other categories, just for the _waiting_list_type_ variable. While they may not end up being important, I'd prefer not to exclude them right off the bat. 

For ClockStops, the categories in the data are (in descending order of frequency: ORTT, IRTT, PTLO, PTLI, ONON, RTTO, INON, RTTI.

For OpenPathways, it is: ORTT, ONON, IRTT, INON, PTLO, PTLI, OTHR, PTLN, PLTO, RTTO, PLTI, PTI, PTL0, RTTI, PTL, PTO, PTL1.

(FYI this is from the notebook pdf output in the folder above.) You could use the same categories for both datasets, since it's the same variable.

iaindillingham:
Thanks, @alschaffer. I think there's a slight -- but understandable -- misunderstanding.

I've applied categorical constraints to `priority_type_code` and `waiting_list_type` in both `wl_clockstops` and `wl_openpathways` (ehrQL tables). These categorical constraints apply only when we generate dummy data.

I've mapped values of `PRIORITY_TYPE_CODE` in `WL_ClockStops` and `WL_OpenPathways` (database tables). These mappings apply when we extract real data. The value that's extracted is the value in the mapping, or `NULL`.

I've not mapped values of `Waiting_List_Type` in either `WL_ClockStops` or `WL_OpenPathways`. The value that's extracted is the value in the column.

sebbacon:
> The categorical constraints on priority_type_code and waiting_list_type are especially useful, as they ensure that the dummy data are more realistic. As @alschaffer notes, however, in the real tables around 10% of values in each of these columns do not match the categorical constraints. Consequently, these values will be missing (NULL) in a real dataset.

Just checking my understanding: does this mean we're effectively dropping 10% of non-valid data for these fields in the non-raw tables?

iaindillingham:
> Just checking my understanding: does this mean we're effectively dropping 10% of non-valid data for these fields in the non-raw tables?

We're not dropping anything. Values of `priority_type_code` are mapped, because most values of `PRIORITY_TYPE_CODE` either obviously conform to the data dictionary or don't. Values of `waiting_list_type` are not mapped, because it's less clear whether they conform or don't.

alschaffer:
Thanks Iain, I get it now, I forgot that the values in `PRIORITY_TYPE_CODE` needed to be mapped. And most of the "non-conforming" values in that variable are not interpretable anyway.

ehrql/backends/tpp.py
```diff
@@ -535,3 +535,139 @@ def medications(self):
...
+                    PSEUDO_PATIENT_PATHWAY_IDENTIFIER,
+                    2
+                ) AS pseudo_patient_pathway_identifier,
+                CONVERT(
+                    varchar(max),
+                    Pseudo_Referral_Identifier,
+                    2
+                ) AS pseudo_referral_identifier,
+                CONVERT(DATE, REFERRAL_REQUEST_RECEIVED_DATE, 23) AS referral_request_received_date,
+                CONVERT(DATE, REFERRAL_TO_TREATMENT_PERIOD_END_DATE, 23) AS referral_to_treatment_period_end_date,
```
rebkwok, line 666: Some of the WL table date fields contain ceiling values (see #1523); this is one of them.  Do we want to replace ceiling dates with Null? (This may be too early to do it, since it's still in the exploratory stage)

ehrql/tables/beta/tpp.py
```diff
@@ -863,3 +867,205 @@ class vaccinations(EventFrame):
...
+        ),
+        constraints=[Constraint.Regex(r"[a-zA-Z0-9]{2}")],
+    )
+    waiting_list_type = Series(
+        str,
+        constraints=[Constraint.Categorical(["ORTT", "IRTT", "ONON", "INON"])],
+    )
+    week_ending_date = Series(
+        datetime.date,
+        description="The Sunday of the week that the pathway relates to",
```
rebkwok, line 1070: I haven't looked at the tables in a while, but when I did previously, there were instances of this date being empty, or in the future. It should never be in the future afaik, since it should be the date the data was uploaded.  Should there be some sort of note to the user in the docs?

Just checked again; there are no empty week_ending_dates now, but there are 23,554 future ones in the wel_openpathways table, and 1816 in the wl_clockstops table ("future" being after this Sunday, in case the date entered is the date of the next Sunday rather than the one just gone).

tests/integration/backends/test_tpp.py
```diff
@@ -1262,6 +1265,174 @@ def test_vaccinations(select_all):
...
+            PRIORITY_TYPE_CODE="1",
+            PSEUDO_ORGANISATION_CODE_PATIENT_PATHWAY_IDENTIFIER_ISSUER=sha256_digest(1),
+            PSEUDO_PATIENT_PATHWAY_IDENTIFIER=sha256_digest(1),
+            Pseudo_Referral_Identifier=sha256_digest(1),
+            Referral_Request_Received_Date="01-02-2023",
+            REFERRAL_TO_TREATMENT_PERIOD_END_DATE="03-04-2025",
+            REFERRAL_TO_TREATMENT_PERIOD_START_DATE="02-03-2024",
+            SOURCE_OF_REFERRAL_FOR_OUTPATIENTS="",
+            Waiting_List_Type="ORTT",
+            Week_Ending_Date="03-03-2024",
```
rebkwok, line 1292: The data in the db is in YYYY-MM-DD format; is it DD-MM-YYYY here deliberately, to check that the raw table just produces whatever string is there? If so, could there maybe be a comment to say so?

rebkwok (approved):

ehrql/backends/tpp.py
iaindillingham: I think that we do, as any exploratory work can be done with the `raw_` tables. I'll replace ceiling dates with NULLs.

iaindillingham: Thanks for your comment.

tests/integration/backends/test_tpp.py
iaindillingham: This is an error on my part. I didn't have DB access when I wrote the tests, so I relied on the codebook. When I regained DB access, I checked the tables and updated the tests, but missed this test. Thanks for spotting it!

ehrql/tables/beta/tpp.py
iaindillingham: > Should there be some sort of note to the user in the docs?

On balance, I think we shouldn't put this kind of information in the docs, as it may become outdated without us knowing; empty values for the `Week_Ending_Date` column is a good example. I know @alschaffer has investigated these tables. Is there a plan to publish those investigations, Andrea?

PR #1654, opened 2023-10-14
fix: Always validate categorical data in user-provided arrow dummy dataset

inglesp:
When a user provides their own dummy dataset, we read it in and then write it out again.  If a dataset variable is categorical, and if the dataset is written out to an arrow file, an error will be thrown by pyarrow if the corresponding column contains unexpected values.

As such, when we read in a user's dummy dataset from an arrow file, we need to validate that the values in a categorical column are in the set of expected values, even if pyarrow does not think that the column is categorical.

### DIFF
--- a/ehrql/file_formats/arrow.py
+++ b/ehrql/file_formats/arrow.py
@@ -204,8 +204,13 @@ class ArrowDatasetReader(BaseDatasetReader):
                 f"Type mismatch in column '{name}': "
                 f"expected {spec.type}, got {column.type}"
             )
-        if pyarrow.types.is_dictionary(column.type) and spec.categories is not None:
-            column_categories = column.dictionary.to_pylist()
+        if spec.categories is not None:
+            if pyarrow.types.is_dictionary(column.type):
+                column_categories = column.dictionary.to_pylist()
+            else:
+                column_categories = sorted(
+                    {v for v in column.to_pylist() if v is not None}
+                )
             unexpected_categories = set(column_categories) - set(spec.categories)
             if unexpected_categories:
                 return (
--- a/tests/integration/test_file_formats.py
+++ b/tests/integration/test_file_formats.py
@@ -142,6 +142,28 @@ def test_read_dataset_validates_categories(test_file):
         read_dataset(test_file, column_specs)
 
 
+def test_read_dataset_validates_categories_on_non_categorical_column(test_file):
+    # This tests that categories are validated even if an arrow column is not
+    # categorical. This is relevant if a user provides their own dummy dataset without
+    # making the columns categorical.
+
+    if test_file.suffix != ".arrow":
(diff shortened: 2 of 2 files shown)


ehrql/file_formats/arrow.py
```diff
@@ -204,8 +204,11 @@ def _validate_column(self, name, column, spec):
                 f"Type mismatch in column '{name}': "
                 f"expected {spec.type}, got {column.type}"
             )
-        if pyarrow.types.is_dictionary(column.type) and spec.categories is not None:
-            column_categories = column.dictionary.to_pylist()
+        if spec.categories is not None:
+            if pyarrow.types.is_dictionary(column.type):
+                column_categories = column.dictionary.to_pylist()
+            else:
+                column_categories = [v for v in column.to_pylist() if v is not None]
```
iaindillingham, line 211: Won't `column_categories` contain duplicate values? Also, what would happen if the categories were in different batches?

iaindillingham (commented):
I'm not confident I understand `ArrowDatasetReader` enough to approve. I've left a comment.

ehrql/file_formats/arrow.py
inglesp: > Won't `column_categories` contain duplicate values?

It's setted on the next line.

> Also, what would happen if the categories were in different batches?

It doesn't matter -- it's only a problem if the column contains values that are not in the column spec, not vice versa.

iaindillingham: > It's setted on the next line.

Ah, I see. Thank you! But don't we use a list of `column_categories`, which may contain duplicate values, when returning the string? (I may have misunderstood what `column` contains.)

> It doesn't matter -- it's only a problem if the column contains values that are not in the column spec, not vice versa.

We validate only the first batch, don't we? So if a column in the second batch contained values that were not in the column spec, then we wouldn't know, would we?

evansd (approved):
Iain is exactly right about the behaviour here, _however_ I think this may be acceptable in any case. We have a parallel situation with CSV where we can't fully validate the categories without reading the entire file, so we just read the first N rows (for some small N) and validate those. The idea is that you hopefully get immediate feedback if you've done something very silly, but full validation is only going to come when the file is read in its entirety.

So (apart from the issue of duplicates in the error message) I think this may be good to go with just a comment explaining that we're not attempting full validation here.

I expect that dummy data files will probably only consist of a single batch in any case so in practice this won't be much of a problem.

One thing that might improve the situation though would be catching the KeyError in the arrow writer and raising something more helpful.

inglesp:
@evansd what's the rationale for not validating the whole dataset?  We use `read_dataset` to read in a user's own dummy dataset, and in the implementation of `table_from_file`.  In either case, do we expect huge unvalidatable files?

Issue #1633, opened 2023-09-29
Draw the owl


iaindillingham:
As @inglesp has pointed out, the ehrQL tutorial is similar to _How to draw an owl_.

![How to draw an owl](https://github.com/opensafely-core/ehrql/assets/477263/c8409908-812f-493d-9314-3d7ba629d239)

Upon conclusion of the ehrQL tutorial, the reader has created a repo, created (and deleted) a codespace, interacted with the sandbox, created a minimal dataset definition, and generated a dummy dataset that is displayed in the terminal (i.e. it is not written to a file).

To become a competent user of ehrQL, however, the reader should also:

- Expand the dataset definition
- Write a dummy dataset to a file
- Commit the dataset definition to `main`

## Expand the dataset definition

I'd like to check with a couple of researchers about what "expand" most usefully means,[^2] but based on [this dataset definition][1], which @alschaffer said was written by her pilots without her help,[^1] I think "expand" probably means:

- Combining Boolean series to define the population (e.g. was born on or before a date and is alive and is either male or female; was registered with a practice on a date; was registered with a practice for a minimum of `k` days)
- Adding some simple demographic variables, such as age and sex
- Adding a complex demographic variable, such as ethnicity (`codelist_from_csv`)
- Adding a complex socioeconomic variable, such as IMD quintile (`case`)
- Deriving a variable, such as counting the number of medications within the last 30 days (`.is_in`, `.is_on_or_between`, `days`, `.count_for_patient`)

## Write a dummy dataset to a file

The reader should add an associated action to *project.yaml*, which they will run with `opensafely run [action]`. They should compare and contrast `run` with `exec`, noticing that `exec` is good for eyeballing the data but `run` is good for developing downstream actions, especially when the dummy dataset isn't written to a CSV file.

## Commit the dataset definition to `main`

Upon conclusion of the ehrQL tutorial, the reader will be at "Initial commit" and be ready to run the associated action on OpenSAFELY Jobs. (Crating a project and workspace, and using OpenSAFELY Jobs is out of scope.) Also, they will have created an artefact inside the codespace that persists outside the codespace.

The reader shouldn't commit the dataset definition to a feature branch and open a pull request, because different projects and different organizations have different guidelines about feature branches and pull requests.

[1]: https://github.com/opensafely/end-of-life-carequality/blob/main/analysis/ehrql/dataset_definition_ehrql_example.py

[...]

PR #1655, opened 2023-10-14
Add "Writing a more complex dataset definition"

iaindillingham:
In this section, the reader works through the dataset definition in the example study (1a500f7). This section is a "rough cut". It needs to be improved, but it also needs to be read by readers. Please consider this when reviewing 🙂 Specifically, I know

* this section breaks the flow of the tutorial
* it's unclear whether the reader should type the code in the sandbox, in the dataset definition, or in both
* there's no output, even when it looks like there should be output
* there's no workflow, as such

I'm asking to merge this section now, because we are asking users to use ehrQL now, which is causing some anxiety.[^1][^2] We know that the tutorial needs to be improved (#1633), but I don't feel I have time to make all the improvements (see above). Nevertheless, I feel this section is a step in the right direction.

Thanks for your help writing the example study, @alschaffer.

[^1]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1697207949530699
[^2]: https://bennettoxford.slack.com/archives/C02HJTL065A/p1697193101523839

### DIFF
--- a/docs/tutorial/example-study/codelists/codelists.json
+++ b/docs/tutorial/example-study/codelists/codelists.json
@@ -1,0 +1,16 @@
+{
+  "files": {
+    "opensafely-asthma-inhaler-salbutamol-medication.csv": {
+      "id": "opensafely/asthma-inhaler-salbutamol-medication/2020-04-15",
+      "url": "https://codelists.opensafely.org/codelist/opensafely/asthma-inhaler-salbutamol-medication/2020-04-15/",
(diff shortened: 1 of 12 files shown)


inglesp (approved):
Happy for this to go in now.  I'm not sure about having separate paths through the tutorial for first- and second-time readers.  And in terms of sequencing, I'd want to have this after we introduce dummy datasets.  But that can wait.

iaindillingham:
I agree, @inglesp 🙂

PR #1660, opened 2023-10-16
chore: Drop mkdocs-table-reader-plugin

iaindillingham:
The [mkdocs-table-reader-plugin][1] plugin adds a Markdown element for inserting tabular data (e.g. CSVs) into a document. We don't use the element: `rg read_csv docs/` doesn't match it, and the docs build without the plugin. Let's drop it.

[1]: https://github.com/timvink/mkdocs-table-reader-plugin

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -52,11 +52,6 @@ extra_css:
 watch:
   - docs
 
-plugins:
-  - table-reader:
-      base_path: "docs_dir"
-
-
 # WARNING!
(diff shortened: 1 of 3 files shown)


inglesp (approved):
84 reasons to approve this!

image

evansd (approved):
Delighted that this lets us drop both Pandas and numpy!

iaindillingham:
> Delighted that this lets us drop both Pandas and numpy!

Me too! I was surprised that such an innocuous-looking plugin depended on these heavyweights.

PR #1669, opened 2023-10-20
Refactor backends to support adding hooks

evansd:
This pulls some of the refactoring commits out of #1500 because we want to make use of them before merging that PR.


evansd:
Yeah, the punchline is really the final commit of #1500. But I'm doing other work which fiddles around with backends and I'll end up in rebase hell if I don't merge this now.

PR #1696, opened 2023-11-02
fix: Upgrade dependencies

inglesp:
Dependabot still isn't doing this for us (see #964).


inglesp:
The deployment to Cloudflare Pages is unrelated -- Cloudflare are having issues at the moment.

rebkwok (approved):

PR #1705, opened 2023-11-07
Move raw tables into separate module

### LINKED ISSUE #1665: Move raw tables into separate module
We have ehrQL tables that transform their corresponding database tables to some degree.[^2] Examples include `patients`, `ons_deaths`, and `wl_clockstops`. We also have raw ehrQL tables that give more-or-less direct access to the database tables. At present, `isaric_raw`, `wl_clockstops_raw`, and `wl_openpathways_raw` are the only raw tables.

Tables and raw tables live side-by-side in the same modules, which in practice means side-by-side in `ehrql.tables.beta.tpp`. To distinguish between tables and raw tables, raw tables have the prefix `raw_`. However, as @sebbacon pointed out,[^1] this gives raw tables the same status as tables. This is undesirable: we'd prefer researchers to use tables rather than raw tables. To encourage researchers, we should lower the status of raw tables.

Moving raw tables to `ehrql.tables.beta.raw.tpp` would lower the status of raw tables. It would also give a clear location in [the table schemas documentation][1] for describing them. As only [opensafely/isaric-exploration][] references a raw table, we can move the raw tables without a deprecation period, as long as we open a PR on the study so as not to confuse Millie and Will.

To make this issue more tractable, I've chosen to avoid the question of what "beta" means. And of what makes a table a table, rather than a raw table. (Some tables don't involve transforming the database tables; some raw tables do.) And of whether "raw" is the right adjective.

[^1]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1697533816749729
[^2]: @inglesp has called these "cooked" ehrQL tables 👨🏻‍🍳 

[1]: https://docs.opensafely.org/ehrql/reference/schemas/
[opensafely/isaric-exploration]: https://github.com/opensafely/isaric-exploration

iaindillingham:
After approval, but before merge, I'll open a PR on [opensafely/isaric-exploration](https://github.com/opensafely/isaric-exploration). (It's the only study to make use of a raw table.)

Closes #1665


evansd (approved):
Thanks for doing the hard work to make this easy to review! All looks great.

PR #1706, opened 2023-11-07
Replace ceiling dates in PatientAddress with NULL

### LINKED ISSUE #1668: Replace ceiling start_date and end_date in PatientAddress table with null
The start_date and end_date fields in the PatientAddress table are non-nullable date fields that can take ceiling values (9999-12-31).  These should be replaced with NULL to avoid unexpected behaviour if a user was to specify start/end > YYYYMMDD.

Performance issues checked in #1523, see comment: https://github.com/opensafely-core/ehrql/issues/1523#issuecomment-1768948693

milanwiedemann:
Replace ceiling `start_date` and `end_date` in PatientAddress table with `NULL`.

As described in [this](https://github.com/opensafely-core/ehrql/issues/1668#issuecomment-1795200793) comment, this change should not impact the results of existing studies

Closes #1668

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -79,8 +79,8 @@ class TPPBackend(SQLBackend):
             SELECT
                 addr.Patient_ID AS patient_id,
                 addr.PatientAddress_ID AS address_id,
-                CAST(addr.StartDate AS date) AS start_date,
-                CAST(addr.EndDate AS date) AS end_date,
+                CAST(NULLIF(addr.StartDate, '9999-12-31T00:00:00') AS date) AS start_date,
+                CAST(NULLIF(addr.EndDate, '9999-12-31T00:00:00') AS date) AS end_date,
                 addr.AddressType AS address_type,
                 addr.RuralUrbanClassificationCode AS rural_urban_classification,
                 NULLIF(addr.ImdRankRounded, -1) AS imd_rounded,
--- a/tests/integration/backends/test_tpp.py
+++ b/tests/integration/backends/test_tpp.py
@@ -131,6 +131,16 @@ def test_addresses(select_all):
             LocationRequiresNursing="Y",
             LocationDoesNotRequireNursing="N",
         ),
+        PatientAddress(
+            Patient_ID=1,
+            PatientAddress_ID=5,
+            StartDate="9999-12-31T00:00:00",
+            EndDate="9999-12-31T00:00:00",
+            AddressType=3,
+            RuralUrbanClassificationCode=4,
+            ImdRankRounded=1000,
+            MSOACode="NPC",
+        ),
     )
     assert results == [
         {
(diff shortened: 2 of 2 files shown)


iaindillingham (approved):
Before merging, could you [estimate the performance impact][1] of these changes and add the timings as a comment on this PR? Thank you 🙂 

[1]: https://github.com/opensafely-core/ehrql/wiki/Estimating-the-performance-impact-of-changes-to-%22query-table%22-SQL-in-TPP

inglesp:
> Before merging, could you [estimate the performance impact](https://github.com/opensafely-core/ehrql/wiki/Estimating-the-performance-impact-of-changes-to-%22query-table%22-SQL-in-TPP) of these changes and add the timings as a comment on this PR? Thank you 🙂

Hasn't this already been done?  See @rebkwok's notes on #1668:

> Performance issues checked in #1523, see comment: [#1523 (comment)](https://github.com/opensafely-core/ehrql/issues/1523#issuecomment-1768948693)

ehrql/backends/tpp.py
```diff
@@ -79,8 +79,8 @@ def modify_query_variables(self, variables):
             SELECT
                 addr.Patient_ID AS patient_id,
                 addr.PatientAddress_ID AS address_id,
-                CAST(addr.StartDate AS date) AS start_date,
-                CAST(addr.EndDate AS date) AS end_date,
+                CAST(NULLIF(addr.StartDate, '9999-12-31') AS date) AS start_date,
+                CAST(NULLIF(addr.EndDate, '9999-12-31') AS date) AS end_date,
```
evansd, line 83: I think that for a couple of subtle, non-obvious reasons this might not be quite right.

One is that, before we do the casting, `StartDate` and `EndDate` are actually datetimes not dates and so we should really be using a datetime literal to compare them with like we do here:
https://github.com/opensafely-core/ehrql/blob/aeec978ee040f57cded5fc9bb804f5020473cafc/ehrql/backends/tpp.py#L173

In practice, MSSQL will do an implicit cast of the date to a datetime so the comparison won't throw an error. But it would be better not to rely on that behaviour.

The second issue is that `YYYY-MM-DD` isn't a reliable date format in MSSQL because, depending on the localisation settings, it can actually interpret it as `YYYY-DD-MM`. The reliable date format for MSSQL is the non-standard `YYYYMMDD` format (i.e. `99991231`). It's kind of incredible that this is the case, and it's not the sort of thing I'd expect many people to know. Of course, this goes away if we switch to using datetimes (the reliable format for datetimes _does_ have dashes in).

```suggestion
                CAST(NULLIF(addr.StartDate, '9999-12-31T00:00:00') AS date) AS start_date,
                CAST(NULLIF(addr.EndDate, '9999-12-31T00:00:00') AS date) AS end_date,
```

iaindillingham: 😱 Thanks for highlighting this, Dave. (And also for your description of the issue in #92.)

milanwiedemann: Thanks Dave! And it looks like we did do this before but I missed it, e.g., https://github.com/opensafely-core/ehrql/blob/aeec978ee040f57cded5fc9bb804f5020473cafc/ehrql/backends/tpp.py#L173

But looking through the code I found another datetime variable (EndDate from [RegistrationHistory](https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#RegistrationHistory)) where we currently just check for the date in YYYY-MM-DD format:

https://github.com/opensafely-core/ehrql/blob/aeec978ee040f57cded5fc9bb804f5020473cafc/ehrql/backends/tpp.py#L549

Is this something where we should switch to `9999-12-31T00:00:00` as well"?

evansd: Already done! :grinning: 
https://github.com/opensafely-core/ehrql/pull/1708

milanwiedemann: hahahahahah, so fast!

PR #1708, opened 2023-11-08
fix: Use standard idiom for ceiling date handling

evansd:
I happened to notice these when looking at something else. In one case we were comparing a datetime with a date and using the "dangerous" date literal format (see #92 for the horrific details).


iaindillingham (approved):
Thanks for clarifying the dangerous date literal.

Issue #512, opened 2022-05-27
More efficient implementation of codelists and categories


evansd:
Currently we implement codelists using IN queries, e.g.
```sql
WHERE code IN ("big", "list", "of", "codes")
```
And we map codes to categories using CASE expressions e.g.
```sql
CASE
  WHEN code = "abc" THEN "category_1"
  WHEN code = "def" THEN "category_2"
  ELSE "category_3"
END
```

There are potential scaling issues with the `IN` queries as certain databases have limits on how many values we can include and some of the codelists get quite large.

There are potential scaling _and_ performance issues with the `CASE` expressions.

In both cases the answer is to do what we did in cohortextractor and create temporary tables with the codes in and then use JOINs.

This will involve an additional [`apply_transforms()`](https://github.com/opensafely-core/databuilder/blob/4a49fc7c72fdf1d06ae75ad01ec91455ad22749c/databuilder/query_model_transforms.py#L26) step to find places in the query which could benefit from this optimisation and modify the query graph to use it.

The patterns we're looking for are:
 * `IN` queries with a list of values over a certain length;
 * `CASE` expressions when the conditions all compare a single input expression with a list of fixed values, and where the output expressions are all fixed values.

I expect there will be a bit of fiddling around with collations. In the TPP database, different coding systems use different text collations and the one used in the codelist table needs to match whatever the target column in the database is.

evansd:
The last time we implemented this we used [multi-valued](https://docs.sqlalchemy.org/en/14/core/dml.html#sqlalchemy.sql.expression.Insert.values.params.*args) insert statements to populate the codelist tables.

However, as the linked docs say, this isn't actually the best way of bulk inserting rows:

> It is essential to note that **passing multiple values is NOT the same as using traditional executemany() form**. The above syntax is a **special** syntax not typically used. To emit an INSERT statement against multiple rows, the normal method is to pass a multiple values list to the `Connection.execute()` method, which is supported by all database backends and is generally more efficient for a very large number of parameters.

This approach seems to have also [caused problems](https://github.com/opensafely-core/databuilder/pull/490/commits/3efde664f7373d6c4412cf56493bd5c3bcb88e0d) for Graphnet when attempting to run using an Azure Synapse database.

The correct approach is described [here](https://docs.sqlalchemy.org/en/14/tutorial/dbapi_transactions.html#tutorial-multiple-parameters) and involves creating a single insert statement which is then passed to `execute()` along with a list of parameters.

[...]

PR #1710, opened 2023-11-08
Support large codelists

evansd:
This addresses the easier half of #512 which involves restructuring large `x IN(a, b, c, ...)` queries to use a temporary table containing the parameters. (The more difficult half is rewriting large `.to_category()` queries to use a JOIN rather than a CASE expression.)

The core of the implementation is relatively simple; most of the complexity is in handling column collations for MSSQL.

### DIFF
--- a/tests/integration/backends/test_tpp.py
+++ b/tests/integration/backends/test_tpp.py
@@ -17,6 +17,7 @@ from tests.lib.tpp_schema import (
     APCS_Cost,
     APCS_Der,
     Appointment,
+    Base,
     CodedEvent,
     CodedEvent_SNOMED,
     CustomMedicationDictionary,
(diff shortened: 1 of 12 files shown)


tests/integration/backends/test_tpp.py
```diff
@@ -1484,6 +1485,123 @@ def test_registered_tests_are_exhaustive():
...
+        {"patient_id": patient_id, "matches": value in matching_values}
+        for patient_id, value in patient_values
+    ]
+
+
+def test_nonstandard_collation_test_is_exhaustive():
+    # Walk over every ORM class we use in this test module and collect all the columns
+    # on those classes which use a nonstandard collation. We don't inspect every class
+    # in the schema because that contains tables we don't currently use. But the moment
+    # we start using a table we want it to be included.
```
inglesp, line 1575: This is a bit weird, and only works (I think?) because we have `test_registered_tests_are_exhaustive`.  Could you add a comment explaining this?

inglesp (approved):
I share your mild architectural discomfort, but I don't think it's too bad.  It introduces a little bit more indirection, but really not very much new code.  (Imagine trying this in a library with no indirection like cohort-extractor!)

> The more difficult half is rewriting large `.to_category()` queries to use a JOIN rather than a CASE expression.

If this is going to be much harder, do we think this is worth doing now?

evansd:
> I share your mild architectural discomfort, but I don't think it's too bad.

I suppose another mitigating factor here is that it isn't _infectious_ complexity, in the sense that when we eventually come to unpick it I don't think we'll have to make sweeping changes to large parts of the codebase – more a case of shifting a few things around here and there so the code has a more coherent story and is easier to follow.

> If this is going to be much harder, do we think this is worth doing now?

I think it would make sense to create a new ticket for it and prioritise it separately. It would be nice to know that we can handle very large categorised codelists but I'm not sure we've seen any in the wild so far, and so it probably makes sense to tackle other issues first.

If I had a very clear plan in my head of how it would all work then it might make sense to ride the wave and do it now. But I don't, as yet.

tests/integration/backends/test_tpp.py
evansd: Yes, that's right (and yes it is more than a bit weird). I'll try to clarify in the comments.

The alternative approach I thought of was to walk the full schema and have an explicit list of columns we don't use. But my worry was that nothing keeps that up-to-date if we start using one of those columns. Piggy backing off the imports (in combination with `test_registered_tests_are_exhaustive`) was the only way I could come with of capturing this automatically.

PR #1712, opened 2023-11-08
Include collations in TPP schema

evansd:
These are now included in the schema CSV following:
 * https://github.com/opensafely/tpp-database-schema/pull/99


inglesp (approved):
Automation's paying off!

PR #1714, opened 2023-11-09
Support configuration of dummy data population size for measures

### LINKED ISSUE #1689: Support `measures.configure_dummy_data()`
We need to be able to set the dummy population size for measures in the same way as for datasets.

rebkwok:
Fixes #1689


evansd (approved):
Looks good, thanks for taking this on. There's just one bit of unnecessary but harmless code I've commented on.

PR #1722, opened 2023-11-13
feat: Expose historical APCS data for further exploration in ehrQL

inglesp:


evansd (approved):

iaindillingham:
These would be good candidates for raw tables.

iaindillingham:
Oh, they are!

ehrql/tables/beta/raw/tpp.py
```diff
@@ -318,3 +318,53 @@ class wl_openpathways_raw(EventFrame):
 
 
 wl_openpathways = table(wl_openpathways_raw)
+
+
+@table
+class apcs_historical(EventFrame):
```
iaindillingham, line 324: I was expecting this to be undecorated and called `apcs_historical_raw`, with a similarly named property on the backend. We don't have to, though. If we chose to do so in the future, then we could without affecting users.

PR #1727, opened 2023-11-14
fix: Fx bug in MSSQL horizontal aggregation workaround

### LINKED ISSUE #1724: Inline table error in generative tests (MSSQL)
Generative tests are failing with an MSSQL error similar to:
```
Invalid object name '#inline_data_3'
```

E.g. https://github.com/opensafely-core/ehrql/actions/runs/6831028936

I can reduce this to the following variable and still get the error:
```
variable = Function.MaximumOf(
    (
        Case(
            {
                Function.In(
                    Value("aaa"), Value(frozenset({"", "aba", "cb", "aaa", "a"}))
                ): Value(3)
            },
            default=Value(9),
        ),
    )
)
```
If I remove the outermost `Function.MaximumOf()`, the test passes.

Adding a breakpoint in mssql `get_results()` to inspect the setup queries:
```
ipdb> for query in setup_queries: print(str(query))
SELECT * INTO "#tmp_1" FROM (SELECT DISTINCT p1.patient_id AS patient_id 
FROM p1) AS anon_1
CREATE INDEX "ix_#tmp_1_patient_id" ON "#tmp_1" (patient_id)
SELECT * INTO "#tmp_2" FROM (SELECT "#tmp_1".patient_id AS patient_id 
FROM "#tmp_1" 
WHERE "#tmp_1".patient_id IS NOT NULL) AS anon_1
CREATE INDEX "ix_#tmp_2_patient_id" ON "#tmp_2" (patient_id)
SELECT * INTO "#results" FROM (SELECT "#tmp_2".patient_id AS patient_id, (SELECT max(aggregate_values.aggregate) AS max_1 
FROM (VALUES (CASE WHEN (:param_1 IN (SELECT "#inline_data_3".value 
FROM "#inline_data_3")) THEN :param_2 ELSE :param_3 END)) AS aggregate_values (aggregate)) AS v 
FROM "#tmp_2") AS anon_1
CREATE INDEX "ix_#results_patient_id" ON "#results" (patient_id)
```

Appears to be missing a setup query to create #inline_data_3, which is should be the table representing the `Value(frozenset({"", "aba", "cb", "aaa", "a"}))`.  I put a breakpoint in `create_inline_table()` to verify that the table is indeed created, with appropriate setup queries, but it's not making its way into the final query.
```
ipdb> table.setup_queries
[<sqlalchemy.sql.ddl.CreateTable object at 0x7f96fa6ffc90>, <ehrql.utils.sqlalchemy_query_utils.InsertMany object at 0x7f96fa6fffd0>, <sqlalchemy.sql.ddl.CreateIndex object at 0x7f96fa6fc1d0>]
ipdb> for query in table.setup_queries: print(str(query))
CREATE TABLE "#inline_data_3" (
	value VARCHAR(3)
)
<ehrql.utils.sqlalchemy_query_utils.InsertMany object at 0x7f96fa6fffd0>
CREATE INDEX "ix_#inline_data_3_value" ON "#inline_data_3" (value)
```

evansd:
We work around the absence of `GREATEST/LEAST` support in our version of MSSQL using a table value constructor and a scalar select. This required a hacky workaround to get SQLAlchemy's ScalarSelect object to "remember" whch tables it was selected from. However this hack turned out to be insufficient: it handled the `_from_objects` property but not the `get_children()` method.

Adding support for large codelists in #1710 exposed the problem here as it became possible to construct aggregations which would "lose" the reference to the codelist table and therefore never create it.

Fortunately, Hypothesis found this issue for us before it affected any users. Nice one.

The workaround we now use (making our own ScalarSelect subclass with the desired behaviour) is still a bit of a hack, but slightly less so; and it solves the problem in question.

Closes #1724


rebkwok (approved):
I'm not sure I feel very qualified to review this, but I think it all makes sense to me, and thank you for the detailed comments :)

iaindillingham (approved):
Thank you for the clear comments!

PR #1728, opened 2023-11-14
Add ethnictity_from_sus table

### LINKED ISSUE #937: Add `ethnicity_from_sus` table
This is to replicate the [`with_ethnicity_from_sus`](https://github.com/opensafely-core/cohort-extractor/blob/2287eae86184f90451524befd4608b62b89236bc/cohortextractor/tpp_backend.py#L2716) function in Cohort Extractor which is quite widely used.

The logic involved here (taking values from various columns in various tables, UNIONing them together, and finding the most frequent) is not something we can currently support in ehrQL. So we'll have to implement this as a `QueryTable` where the logic is done in SQL and the results presented as a "view" in ehrQL.

The Cohort Extractor function also handles mapping the alphabetic character codes to numeric ones. We're deliberately not doing that here for two reasons:
 * ehrQL makes it very easy to do this kind of thing using `.map_values({"A": 1, "B": 2, ...})`;
 * these numeric codes are something of a [legacy artefact](https://bennettoxford.slack.com/archives/C03FB777L1M/p1673603741625489) that we'd like to move away from.

The ehrQL side of things could look something like this ([`beta/tpp.py`](https://github.com/opensafely-core/databuilder/blob/b2b9388f7af5da4c5b7f871e1f66187667e148f5/databuilder/tables/beta/tpp.py)):
```py
@table
class ethnicity_from_sus(PatientFrame):
    """
    Finds the most frequently used ethnicity code for each patient from the various SUS tables.

    Specifically it uses:

        APCS.Ethnic_group
        EC.Ethnic_Category
        OPA.Ethnic_Category

    Codes are as defined by "Ethnic Category Code 2001" — the 16+1 ethnic data
    categories used in the 2001 census:
    https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html

    Codes beginning Z ("Not stated") and 99 ("Not known") are ignored.

    Where there is a tie for the most common code the lexically greatest code is used.
    """
    code = Series(
        str,
        constraints=[
            Constraint.Categorical(
                list("ABCDEFGHJKLMNPRS")
            ),
        ],
        description=(
            "First character of recorded ethncity code:\n"
            "https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html"
        )
    )
    local_code = Series(
        str,
        constraints=[
            Constraint.Regex(
                list("[ABCDEFGHJKLMNPRS][0-9A-Z]")
            ),
        ],
        description=(
            "Two character code with both national prefix and optional local suffix:\n"
            "https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html"
        ),

    )
```

The backend table will need to be implemented as a `QueryTable` in [`backends/tpp.py`](https://github.com/opensafely-core/databuilder/blob/b2b9388f7af5da4c5b7f871e1f66187667e148f5/databuilder/backends/tpp.py) (SQL adapted from Cohort Extractor and untested):
```py
    ethnicity_from_sus = QueryTable(
        """
            SELECT
              Patient_ID AS patient_id,
              SUBSTRING(local_code, 1, 1) AS code,
              local_code
            FROM (
              SELECT
                Patient_ID,
                local_code,
                ROW_NUMBER() OVER (
                    PARTITION BY Patient_ID
                    ORDER BY COUNT(local_code), local_code DESC
                ) AS row_num
              FROM (
                SELECT
                  Patient_ID,
                  Ethnic_group AS local_code
                FROM
                  APCS
                UNION ALL
                SELECT
                  Patient_ID,
                  Ethnic_Category AS local_code
                FROM
                  EC
                UNION ALL
                SELECT
                  Patient_ID,
                  Ethnic_Category AS local_code
                FROM
                  OPA
              ) t
              WHERE
                local_code IS NOT NULL
                AND local_code != ''
                AND local_code != '99'
                AND local_code NOT LIKE 'Z%'
              GROUP BY Patient_ID, local_code
            ) t
            WHERE row_num = 1
        """
    )
```

We'll also need to add a test for this table in [`tests/integration/backends/test_tpp.py`](https://github.com/opensafely-core/databuilder/blob/b2b9388f7af5da4c5b7f871e1f66187667e148f5/tests/integration/backends/test_tpp.py) (the exhaustiveness check will remind us if we forget). As the query here is more complex than the average it will probably require a bit more thorough a test than some of the other tables.

rebkwok:
Fixes #937 

ethnicity_from_sus involves collecting all [ethnicity codes](https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html) from the EC, APCS and OPA tables and finding the most frequent one.  APCS and OPA can have a 2 character code, both the national and optional local codes. EC can only take a single character, the national code. Only the national code is relevant for the frequency counts, so we extract it from the APCS and OPA tables before counting/sorting.

see also [slack discussion](https://bennettoxford.slack.com/archives/C03FB777L1M/p1699982433341039)

### DIFF
--- a/docs/includes/generated_docs/schemas/beta.tpp.md
+++ b/docs/includes/generated_docs/schemas/beta.tpp.md
@@ -16,6 +16,7 @@ from ehrql.tables.beta.tpp import (
     ec,
     ec_cost,
     emergency_care_attendances,
+    ethnicity_from_sus,
     hospital_admissions,
     household_memberships_2020,
     medications,
@@ -995,6 +996,46 @@ TODO
   <dd markdown="block">
 
 
+  </dd>
+</div>
+
+  </dl>
+</div>
+
+
+<p class="dimension-indicator"><code>one row per patient</code></p>
+## ethnicity_from_sus
+
+This finds the most frequently used national ethnicity code for each patient from
+the various SUS (Secondary Uses Service) tables.
+
+Specifically it uses ethnicity codes from the following tables:
+
+    APCS (In-patient hospital admissions)
+    EC (A&E attendances)
+    OPA (Out-patient hospital appointments)
+
+Codes are as defined by "Ethnic Category Code 2001" — the 16+1 ethnic data
+categories used in the 2001 census:
+https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html
+
+Codes beginning Z ("Not stated") and 99 ("Not known") are excluded.
+
+Where there is a tie for the most common code the lexically greatest code is used.
+<div markdown="block" class="definition-list-wrapper">
+  <div class="title">Columns</div>
+  <dl markdown="block">
+<div markdown="block">
+  <dt id="ethnicity_from_sus.code">
+    <strong>code</strong>
+    <a class="headerlink" href="#ethnicity_from_sus.code" title="Permanent link">🔗</a>
+    <code>string</code>
+  </dt>
+  <dd markdown="block">
+First character of recorded ethncity code (national code):
+https://www.datadictionary.nhs.uk/data_elements/ethnic_category.html
+
+ * Possible values: `A`, `B`, `C`, `D`, `E`, `F`, `G`, `H`, `J`, `K`, `L`, `M`, `N`, `P`, `R`, `S`
   </dd>
 </div>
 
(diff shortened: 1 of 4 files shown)


evansd (approved):
:+1: Thanks for all the attention to detail that went into this.

evansd:
Oh, I suppose it needs pedantic prefix to generate a new release.

Issue #1450, opened 2023-07-24
Consolidate documentation which patients we have data for


inglesp:
We need a page in the documentation that explains:

* which patients we have data for
* what happens to a patient's data when they move between practices
* how to censor populations in ehrQL to include only patients continuously registered at one practice (epi input required here)

Some of this already exists, eg [this page on SystmOne](https://docs.opensafely.org/data-sources/systmone/) or [in the variable reference](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.with_complete_gp_consultation_history_between)  We should consolidate this, and link to it from the relevant parts of the ehrQL docs (eg the practice_registrations docs).

inglesp:
See also [this discussion](https://bennettoxford.slack.com/archives/C02HJTL065A/p1689327065988359)

inglesp:
See also links in #1630.

StevenMaude:
To start with, I'm looking at the disparate bits of information. So reviewing the existing [TPP data sources page](https://docs.opensafely.org/data-sources/systmone/) first.

In relation to this specific issue of describing which patients are covered by the data, it feels like the only relevant paragraph is the following:

> SystmOne is a primary care clinical information system run by TPP (Leeds), used by roughly one third of GP practices in England, with records for approximately 44% of the English population. Its primary purpose is patient management, though the data may be used for research.

It could be moved into the ehrQL [TPP backends details](https://docs.opensafely.org/ehrql/reference/backends/#tpp), verbatim or in some edited form, I think. This paragraph is talking about the data available in the backend, as opposed to explaining the schema.

***

Incidentally, I think the remainder of the TPP data sources page could be:

* moved into the ehrQL backend or schema page
* possibly moved into the TPP database report
  * example:
    * > These are available in the OpenSAFELY-TPP database in the `CodedEvent` table.
* removed as it's redundant and the existing ehrQL schema documentation explains it;
  * example:
    *  > clinical… events are coded… using… Read codes… and… SNOMED CT codes"

The EMIS data sources page is [incredibly sparse](https://docs.opensafely.org/data-sources/emis/) and could also be moved into the appropriate ehrQL tables, when added.

It might be worth opening a separate documentation issue for this, if anyone agrees.

StevenMaude:
Some of the "Defining patient populations with censoring" linked from [#1496](https://github.com/opensafely-core/ehrql/issues/1496#issue-1851365764) feels like it is relevant here too.

[...]

PR #1729, opened 2023-11-14
Document which patients are covered

StevenMaude:
WIP for #1450.


docs/explanation/selecting-populations-for-study.md
StevenMaude, line 1: comment: I'm not convinced this is the best place for this page.

This page could equally be in the main documentation somewhere: it applies to the underlying data provided by the backends.

However, if we do add an ehrQL example here, then perhaps it fits somewhere within the ehrQL documentation (and if we intend everyone to use ehrQL in future).

docs/explanation/selecting-populations-for-study.md
```diff
@@ -0,0 +1,29 @@
...
+
+* From the time a patient leaves an NHS data set,
+  we cannot be confident of any facts about the patient relating to that data set,
+  unless they are completely transferred to another data set
+  available in the same backend.
+* A patient's total period of registration at one or more practices
+  comprises the entire period over which we can expect their data
+  to appear in other NHS data sets.
+  Researchers may wish to select patients with a continuous registration
+  over the time period of interest.
```
StevenMaude, line 25: comment: This is largely a selected quotation of words taken from a "Contracts domain modelling" document.

docs/explanation/selecting-populations-for-study.md
```diff
@@ -0,0 +1,29 @@
+This page is aimed at researchers working with NHS England patient record data
+as provided by the OpenSAFELY backends.
+
+## Continuity of patient data
+
+Within the NHS in England, it is usually assumed that
+a patient's current primary care record is complete:
+when a patient moves practice, their record moves with them.
+
+However, a given NHS data set may only cover a subset of patients in England:
```
StevenMaude, line 10: thought: The use of "data set" here has the potential for confusion with "dataset" as used in ehrQL, but I'm not sure if we have some other term for this.

docs/explanation/selecting-populations-for-study.md
```diff
@@ -0,0 +1,29 @@
...
+  available in the same backend.
+* A patient's total period of registration at one or more practices
+  comprises the entire period over which we can expect their data
+  to appear in other NHS data sets.
+  Researchers may wish to select patients with a continuous registration
+  over the time period of interest.
+
+## Selecting patients with a continuous registration with ehrQL
+
+…
```
StevenMaude, line 29: todo: This needs an example adding, if we have a suitable one.

I had a quick look at how `practice_registrations` was used with ehrQL. Is the following sufficient? https://github.com/opensafely/comparative-booster-ehrql-poc/blob/adcd0853aa637f49808469ae755cdf41ff3fc024/analysis/variables_lib.py#L26-L31

Or are there more robust examples to suggest?

StevenMaude (commented):
I'm not sure this is exactly what's wanted for #1450. The original issue talks about consolidating the information on a single page, but I'm not convinced that a single page is the place to put everything.

Suggestions are welcome.

docs/explanation/selecting-populations-for-study.md
iaindillingham: This isn't the worst place. And by locating this page within the ehrQL repo, it's clear who is responsible for maintaining it. For the moment, and possibly until we agree who is responsible for maintaining the documentation, I'd say that here was fine.

iaindillingham: The content also seems to be understanding-oriented/theoretical knowledge, so works as an explanation from the Diátaxis perspective.

docs/explanation/selecting-populations-for-study.md
iaindillingham: Possibly! Are you worried about taking credit for others' work? If so, think about [creating a commit with multiple authors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors).

docs/explanation/selecting-populations-for-study.md
iaindillingham: The docs talk about data sources.

docs/explanation/selecting-populations-for-study.md
StevenMaude: :+1:

docs/explanation/selecting-populations-for-study.md
StevenMaude: This was a little bit of not wanting to say "I wrote this", even though I committed it.

Unfortunately, online word processors don't give you the equivalent of `git blame` but just show revision history, which makes ascribing authorship a pain. But I've done the necessary searching and it was Seb that wrote most/all of that section.

ehrql/tables/beta/tpp.py
```diff
@@ -889,6 +883,16 @@ def for_patient_on(self, date):
...
+    def has_a_continuous_practice_registration_spanning(self, start_date, end_date):
+        """
+        Return whether a patient has a continuous practice registration
+        between the inclusive start and end dates.
+        """
+        return self.where(
+            self.start_date.is_on_or_before(start_date)
+            & (self.end_date.is_after(end_date) | self.end_date.is_null())
+        ).exists_for_patient()
+
```
StevenMaude, line 895: thought: This was taken from the [comparative booster study](https://github.com/opensafely/comparative-booster-ehrql-poc/commit/dccc16616371757be99ec1e1e9b023518e4dde95). But it would be good for a researcher to look over this before we make this available for users.

evansd: We've developed a convention (though I'm not sure if this is properly documented anywhere) that all aggregation methods (i.e. ones which produce a patient-series from an event-series) end with `_for_patient()`. I'm not sure if there's a neat way of rewording this method name to accommodate that. But I wonder in any case whether it would be better to split that logic out so you have a method like:
```python
    def spanning(self, start_date, end_date):
        """
        Filter registrations to just those spanning the entire period between
        `start_date` and `end_date` inclusive
        """
        return self.where(
            self.start_date.is_on_or_before(start_date)
            & (self.end_date.is_after(end_date) | self.end_date.is_null())
        )
```
And then the user can chain `.exists_for_patient()` on the end if they just want the boolean.

docs/explanation/selecting-populations-for-study.md
```diff
@@ -0,0 +1,28 @@
...
+## Continuity of patient data
+
+Within the NHS in England, it is usually assumed that
+a patient's current primary care record is complete:
+when a patient moves practice, their record moves with them.
+
+However, a given NHS data source may only cover a subset of patients in England:
+for example, only those included in a particular EHR vendor's database.
+Therefore, where a patient has moved to a practice outside the database,
+that patient would usually be censored from that date onwards.
```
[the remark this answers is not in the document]
StevenMaude: You're correct that it would be better to be more precise.

It's not my original wording: instead taken from a design document.

My best guess is I _think_ that it's (a), which also probably warrants (b) as mentioned in the following paragraph.

So maybe:

> Therefore, when a patient moves to a practice outside the database,
> that patient's data would usually be missing in that database from that date onwards.

It would be good for someone else to confirm (or improve!) this wording.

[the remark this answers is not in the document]
inglesp: Have asked in #research.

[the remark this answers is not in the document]
StevenMaude: @LindaNab responded, so going to use that wording.

PR #1742, opened 2023-11-16
docs: Import only from `tpp` in example

StevenMaude:
This simplifies the example dataset definition, and is the way we currently advocate writing a dataset definition.


evansd (approved):

docs/tutorial/pushing-to-github/the_staged_dataset_definition.png
iaindillingham, line 1: Sorry, I disagree with your rationale for cropping the screenshot. The tutorial doesn't crop them, to orient the reader. The message is "Here is the screen you can see, this is where in that screen I'd like to direct your attention." With a cropped version, the reader has to undertake the cognitively demanding task of matching what we give them, to the screen they can see.

Not cropping the screenshot also makes it easier to automatically generate it in the future. I've thought about how we might accomplish this, but it's not important enough to work on now.

If you're (quite understandably!) not up for regenerating the screenshot -- or indeed, all the screenshots, as that's not a bad thing -- then I'd be happy to.

StevenMaude: @iaindillingham: If you'd prefer to recreate it, that's fine. It's better you do that, because I won't be able to make it look the same as the others.

StevenMaude: I've instead hand-edited the existing screenshot to reflect the dataset definition changes. It was possible in this case because of the nature of the changes.

I still think it might be preferable to have screenshots not tied into content where possible, because updating them is another manual step. Automation might be better if that's possible.

PR #1745, opened 2023-11-17
Document `addresses` table

### LINKED ISSUE #1371: Document addresses table

milanwiedemann:
Improve documentation of `addresses` table in ehrQL. Most information comes from existing cohort-extractor docs ([address_as_of()](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.address_as_of), [Geographic and administrative groupings](https://docs.opensafely.org/data-sources/systmone/#geographic-and-administrative-groupings)).

Closes #1371

### DIFF
--- a/ehrql/tables/beta/tpp.py
+++ b/ehrql/tables/beta/tpp.py
@@ -42,23 +42,98 @@ __all__ = [
 
 @table
 class addresses(EventFrame):
-    address_id = Series(int)
-    start_date = Series(datetime.date)
-    end_date = Series(datetime.date)
-    address_type = Series(int)
-    rural_urban_classification = Series(int)
-    imd_rounded = Series(int, constraints=[Constraint.ClosedRange(0, 32_800, 100)])
+    """
+    Geographic characteristics of the home address a patient registers with a practice.
+    Each row in this table is one registration period per patient.
+    Occasionally, a patient has multiple active registrations on a given date.
+    The postcode from the address is mapped to an Output Area,
+    from which other larger geographic representations can be derived
+    (see various [ONS publications][addresses_ukgeographies] for more detail).
+
+    !!! tip
+        To group rounded IMD ranks by quintile:
+
+        ```py
+        imd = addresses.for_patient_on("2023-01-01").imd_rounded
+        dataset.imd_quintile = case(
+            when((imd >=0) & (imd < int(32844 * 1 / 5))).then("1 (most deprived)"),
+            when(imd < int(32844 * 2 / 5)).then("2"),
+            when(imd < int(32844 * 3 / 5)).then("3"),
+            when(imd < int(32844 * 4 / 5)).then("4"),
+            when(imd < int(32844 * 5 / 5)).then("5 (least deprived)"),
+            default="unknown"
+        )
+        ```
+
+    [addresses_ukgeographies]: https://www.ons.gov.uk/methodology/geography/ukgeographies
+    """
+
+    address_id = Series(
+        int,
+        description="Unique address identifier.",
+    )
+    start_date = Series(
+        datetime.date,
+        description="Date patient moved to address.",
+    )
+    end_date = Series(
+        datetime.date,
+        description="Date patient moved out of address.",
+    )
+    address_type = Series(
+        int,
+        description="Type of address.",
+    )
+    rural_urban_classification = Series(
+        int,
+        description="""
+            Rural urban classification:
+
+            * 1 - Urban major conurbation
+            * 2 - Urban minor conurbation
+            * 3 - Urban city and town
+            * 4 - Urban city and town in a sparse setting
+            * 5 - Rural town and fringe
+            * 6 - Rural town and fringe in a sparse setting
+            * 7 - Rural village and dispersed
+            * 8 - Rural village and dispersed in a sparse setting
+        """,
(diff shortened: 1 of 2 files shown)


ehrql/tables/beta/tpp.py
```diff
@@ -42,23 +42,104 @@
...
+    Addresses
+
+    Geographic characteristics of the home address a patient registers with a practice.
+    Each row in this table is one registration period per patient.
+    Occasionally, a patient has multiple active registrations on a given date.
+    The postcode from the address is mapped to an Output Area,
+    from which other larger geographic representations can be derived
+    (see various [ONS publications][addresses_ukgeographies] for more detail).
+
+    !!! tip
```
milanwiedemann, line 55: This is a `!!! warning` in the current cohort-extractor docs but I thought it's more of a `!!! tip`? This code is also in our [How-to guides - Calculating each patient's IMD quintile](https://docs.opensafely.org/ehrql/how-to/examples/#finding-each-patients-imd-rank) but I thought it'd be good to highlight here too?

ehrql/tables/beta/tpp.py
```diff
@@ -42,23 +42,104 @@
...
+            * 1 - Urban major conurbation
+            * 2 - Urban minor conurbation
+            * 3 - Urban city and town
+            * 4 - Urban city and town in a sparse setting
+            * 5 - Rural town and fringe
+            * 6 - Rural town and fringe in a sparse setting
+            * 7 - Rural village and dispersed
+            * 8 - Rural village and dispersed in a sparse setting
+        """,
+        constraints=[Constraint.ClosedRange(1, 8, 1)],
```
milanwiedemann, line 107: I added this `Constraint.ClosedRange(1, 8, 1)` because it seems logical from the description we use in our current docs, but I didn't check what the data looks like.

ehrql/tables/beta/tpp.py
```diff
@@ -42,23 +42,104 @@
...
+        description="Start date of registration period.",
+    )
+    end_date = Series(
+        datetime.date,
+        description="End date of registration period.",
+    )
+    address_type = Series(
+        int,
+        description="Type of address.",
+    )
```
milanwiedemann, line 92: I couldn't find any more specific information about the `address_type` column, does anyone else know more?

ehrql/tables/beta/tpp.py
inglesp: We don't need this warning in ehrQL!  We can write zeros just fine, as shown by this test:

```
def test_zeros(tmp_path):
    filename = tmp_path / "file.arrow"
    column_specs = {
        "patient_id": ColumnSpec(int),
        "imd": ColumnSpec(int, min_value=0, max_value=32_800),
    }
    results = [
        (123, 0),
        (456, None),
        (789, 100),
    ]
    write_dataset(filename, results, column_specs)
    table = pyarrow.feather.read_table(filename)
    output_rows = [tuple(d.values()) for d in table.to_pylist()]
    assert output_rows == results
```

ehrql/tables/beta/tpp.py
```diff
@@ -42,23 +42,104 @@
...
+    [addresses_ukgeographies]: https://www.ons.gov.uk/methodology/geography/ukgeographies
+    """
+
+    address_id = Series(
+        int,
+        description="Unique address identifier.",
+    )
+    start_date = Series(
+        datetime.date,
+        description="Start date of registration period.",
```
inglesp, line 83: I don't think it's correct to say "registration period" here.  What about "Date patient moved to address"?  And same for `end_date` below.

ehrql/tables/beta/tpp.py
inglesp: It's unused in cohort-extractor, so fine to leave as-is for now.

ehrql/tables/beta/tpp.py
inglesp: micro-nit: You could do `Constraint.ClosedRange(1, 8)`, because the `steps` parameter [defaults to 1](https://github.com/opensafely-core/ehrql/blob/4fdb6452aa1333310e1f0530ff66ee5788426706/ehrql/query_model/table_schema.py#L47).

ehrql/tables/beta/tpp.py
```diff
@@ -42,23 +42,104 @@
...
-    has_postcode = Series(bool)
+    has_postcode = Series(
+        bool,
+        description="Indicating whether a valid postcode is recorded for the patient.",
+    )
     # Is the address potentially a match for a care home? (Using TPP's algorithm)
-    care_home_is_potential_match = Series(bool)
+    care_home_is_potential_match = Series(
+        bool,
+        description="Indicating whether the patient's address matched with a care home.",
```
inglesp, line 131:
```suggestion
        description="Indicating whether the patient's address matched with a care home, using TPP's algorithm.",
```

inglesp (approved):
Two small changes please, but otherwise :shipit:

ehrql/tables/beta/tpp.py
milanwiedemann: oh thanks, I missed this! I deleted the warning but have kept the IMD tip here

ehrql/tables/beta/tpp.py
milanwiedemann: yes, I think that's better

PR #1746, opened 2023-11-17
Add disclosure control for measures

### LINKED ISSUE #1666: Support small number suppression and rounding of measures output
Measures generation with cohort extractor supported [optional small number suppression](https://github.com/opensafely-core/cohort-extractor/blob/1c281ea6308998514c569990d632b3cc5c793c5d/cohortextractor/measure.py#L34C1-L39C40).

Support for this in ehrQL will allow more consistent disclosure control. Since the implementation in cohort extractor, we have [updated our disclosure control guidance](https://www.opensafely.org/updated-output-checking-processes/), so this should be done in collaboration with an output checker to make sure the approach is consistent with the updated guidelines.

iaindillingham:
This adds disclosure control -- suppressing small numbers (less than or equal to seven) and then rounding numbers (to the nearest five) -- to the measures framework. The OpenSAFELY approach to disclosure control is documented in "[Updated disclosure control guidance][1]".

Disclosure control is enabled by default. It is disabled by calling:

```
measures.configure_disclosure_control(enabled=False)
```

[The OpenSAFELY docs][2] suggest suppressing small numbers by replacing them with `"[REDACTED]"` (i.e. a string). However, doing so adds complexity to the implementation, because some output formats have typed columns. I considered replacing small numbers with `None`, but this does the same: The complexity shifts from the output format to the calculation of the ratio.

Replacing small numbers with zeros doesn't add complexity to the implementation. As well as being of the same type, the calculation of the ratio before disclosure control is the same as the calculation of the ratio after disclosure control.

Consequently, small numbers are replaced with zeros.

[1]: https://www.opensafely.org/updated-output-checking-processes/
[2]: https://docs.opensafely.org/releasing-files/#redacting-counts-less-than-or-equal-to-7

Closes #1666.

### DIFF
--- a/ehrql/measures/measures.py
+++ b/ehrql/measures/measures.py
@@ -62,6 +62,11 @@ class Measure:
     intervals: tuple[tuple[datetime.date, datetime.date]]
 
 
+@dataclasses.dataclass
+class DisclosureControlConfig:
+    enabled: bool = True
+
+
 # This provides an interface for construction a list of `Measure` instances (as above)
 # and consists almost entirely of validation logic
 class Measures:
@@ -84,6 +89,7 @@ class Measures:
         self._measures = {}
         self._defaults = {}
         self.dummy_data_config = DummyDataConfig(population_size=None)
+        self.disclosure_control_config = DisclosureControlConfig()
 
     def define_measure(
         self,
@@ -299,6 +305,26 @@ class Measures:
         ```
         """
         self.dummy_data_config.population_size = population_size
+
+    def configure_disclosure_control(self, *, enabled=True):
+        """
+        Configure disclosure control.
+
+        By default, numerators and denominators are subject to disclosure control.
+        First, values less than or equal to seven are replaced with zero (suppressed);
+        then, values are rounded to the nearest five.
+
+        To disable disclosure control:
+
+        ```py
+        measures.configure_disclosure_control(enabled=False)
+        ```
+
+        For more information about disclosure control in OpenSAFELY, please see the
+        "[Updated disclosure control
+        guidance](https://www.opensafely.org/updated-output-checking-processes/)" page.
+        """
+        self.disclosure_control_config.enabled = enabled
 
     def __iter__(self):
         return iter(self._measures.values())
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -1,6 +1,8 @@
 import textwrap
 from datetime import date
 
+import pytest
+
 from ehrql.main import generate_measures
 from ehrql.query_engines.sqlite import SQLiteQueryEngine
 from ehrql.tables.beta.core import patients
@@ -23,21 +25,36 @@ measures.define_measure(
 )
 """
 
+DISABLE_DISCLOSURE_CONTROL = "\n\nmeasures.configure_disclosure_control(enabled=False)"
 
-def test_generate_measures(in_memory_sqlite_database, tmp_path):
+
+@pytest.mark.parametrize("disclosure_control_enabled", [False, True])
+def test_generate_measures(
+    in_memory_sqlite_database, tmp_path, disclosure_control_enabled
+):
     in_memory_sqlite_database.setup(
         make_orm_models(
             {
                 patients: [
                     dict(patient_id=1, date_of_birth=date(2020, 6, 1), sex="male"),
                     dict(patient_id=2, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=3, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=4, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=5, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=6, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=7, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=8, date_of_birth=date(2021, 6, 1), sex="female"),
+                    dict(patient_id=9, date_of_birth=date(2021, 6, 1), sex="female"),
                 ]
             }
         )
     )
 
     measure_definitions = tmp_path / "measures.py"
-    measure_definitions.write_text(MEASURE_DEFINITIONS)
+    if disclosure_control_enabled:
+        measure_definitions.write_text(MEASURE_DEFINITIONS)
+    else:
+        measure_definitions.write_text(MEASURE_DEFINITIONS + DISABLE_DISCLOSURE_CONTROL)
     output_file = tmp_path / "output.csv"
 
     generate_measures(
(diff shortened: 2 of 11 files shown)


inglesp (commented):
> The OpenSAFELY docs [2] suggest suppressing small numbers by replacing
them with `"[REDACTED]"`. However, doing so adds complexity to the
implementation.

How much complexity?

ehrql/measures/measures.py
```diff
@@ -300,6 +306,31 @@ def configure_dummy_data(self, *, population_size):
         """
         self.dummy_data_config.population_size = population_size
 
+    def configure_disclosure_control(self, *, enabled=True):
+        """
+        Configure disclosure control.
+
+        By default, the numerators and denominators of measures generated from real and
+        dummy tables, and by the dummy data generator, are subject to disclosure
+        control. First, values less than or equal to `7` are replaced with `0`
```
inglesp, line 315: Took me a couple of goes to parse this.  How about:

> By default, numerators and denominators are subject to disclosure control (unless the user has provided their own dummy data file).

ehrql/measures/measures.py
```diff
@@ -300,6 +306,31 @@ def configure_dummy_data(self, *, population_size):
...
+        (suppressed); then, values are rounded to the nearest `5`.
```
inglesp, line 316: Do we need to wrap numbers in backticks?

tests/integration/test_main.py
```diff
@@ -52,20 +63,36 @@ def test_generate_measures(in_memory_sqlite_database, tmp_path):
...
+    if disclosure_control_enabled:
+        assert output_file.read_text() == textwrap.dedent(
+            """\
+            measure,interval_start,interval_end,ratio,numerator,denominator,sex
+            births,2020-01-01,2020-12-31,,0,0,male
+            births,2020-01-01,2020-12-31,,0,0,female
+            births,2021-01-01,2021-12-31,,0,0,male
+            births,2021-01-01,2021-12-31,,0,0,female
+            """
+        )
```
inglesp, line 75: We discussed having richer data here, so that numerators/denominators weren't all replaced with zero.  That's a nice-to-have -- how much work would it be?

tests/integration/test_main.py
inglesp, line 1: We're missing a test about what happens when no configuration is set.  Addressing this would also smooth out the weirdness of having to escape curly braces in `MEASURE_DEFINITIONS` (why are we trying to put a dict in a set??  oh...).  Specifically, we could leave `MEASURE_DEFINITIONS` as-is, and then append to it as required in each test.

tests/integration/test_main.py
```diff
@@ -143,12 +177,23 @@ def test_generate_measures_dummy_tables(tmp_path):
...
+    else:
+        assert output_file.read_text() == textwrap.dedent(
+            """\
+            measure,interval_start,interval_end,ratio,numerator,denominator,sex
+            births,2020-01-01,2020-12-31,1.0,1,1,male
+            births,2020-01-01,2020-12-31,0.0,0,1,female
+            births,2021-01-01,2021-12-31,0.0,0,1,male
+            births,2021-01-01,2021-12-31,1.0,1,1,female
+            """
+        )
```
inglesp, line 199: As these multiline strings are re-used, would it make sense to have them as module-level constants?

inglesp (approved):

ehrql/measures/measures.py
evansd: I think Peter's rephrasing is clearer. I wonder though if we should just apply SDC to everything, including user-supplied dummy data. It's a better match for what's going to happen in production, and it gives us a simpler story here.

ehrql/measures/measures.py
```diff
@@ -300,6 +306,31 @@ def configure_dummy_data(self, *, population_size):
...
+
+        Measures provided by a dummy data file are not subject to disclosure control.
+
+        For more information about disclosure control in OpenSAFELY, please see the
+        "[Updated disclosure control
+        guidance](https://www.opensafely.org/updated-output-checking-processes/)" page.
+        """
+        # The following is covered by tests/integration/test_main.py, but coverage can't
+        # detect that.
+        self.disclosure_control_config.enabled = enabled  # pragma: no cover
```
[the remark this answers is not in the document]
evansd: It will be because the definition file is evaluated in a different process (for sandboxing purposes).

I think it might make sense to have a unit test for this anyway in `tests/unit/measures/test_measures.py`. It's obviously pretty trivial, but if we start configuring more stuff here and needing some kind of validation logic then it would be good to have a pre-existing test to expand on. And I feel like having the test is simpler than explaining the presence of the pragma.

ehrql/utils/sdc_utils.py
```diff
@@ -0,0 +1,17 @@
+"""Statistical Disclosure Control (SDC) utilities.
```
evansd, line 1: This is more of an aesthetic preference than anything particularly concrete, but I'm not sure I'd break this out into a utils module like this. Utils _generally_ speaking (although we're definitely not fully consistent here) generally contain code which isn't "core business logic" and either needs to be called from multiple places, or is just too long-winded and faffy to include in the module which needs it.

I wonder if a better structure would be to pull the code out of the `calculate` module and have a `ehrql.measures.disclosure_control` module which contains that code plus the stuff here in utils.

evansd (approved):

ehrql/measures/measures.py
iaindillingham: I'm going to make the reasoning explicit.

The question: Should the measures framework always respect `measures.disclosure_control_config.enabled`?

* If the answer is "no", then the story is more complex: one route for these kinds of data, another route for this kind of data.
* If the answer is "yes", then the story is simpler. What are the drawbacks?

If a user intentionally supplied an uncontrolled dummy data file, but didn't call `measures.configure_disclosure_control(enabled=False)`, then the measures framework would apply disclosure control to their dummy data. The drawback is that doing so might confuse the user. But as you say, a local run would match a production run: We would want to prompt the user either to change their dummy data or to configure disclosure control appropriately.

That's reasonable: I will update to always respect `measures.disclosure_control_config.enabled`.

ehrql/measures/measures.py
iaindillingham: We don't. No idea where those came from.

tests/integration/test_main.py
iaindillingham: I've retained `MEASURE_DEFINITIONS`, as you suggest, and created `DISABLE_DISCLOSURE_CONTROL`. I've used these like this

```py
if disclosure_control_enabled:
    measure_definitions.write_text(MEASURE_DEFINITIONS)
else:
    measure_definitions.write_text(MEASURE_DEFINITIONS + DISABLE_DISCLOSURE_CONTROL)
```

rather than creating a new variable and then calling `write_text` once because, I couldn't think of a name for the new variable.

ehrql/measures/measures.py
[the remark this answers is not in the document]
iaindillingham: That's correct.

I've added `test_configure_disclosure_control`. I've also updated `test_define_measures` and `test_define_measures_with_default_group_by` to test for the default value.

ehrql/utils/sdc_utils.py
iaindillingham: That makes sense, Dave. Thanks for describing your rationale. I have added `ehrql.measures.disclosure_control`.

I originally wanted to keep `apply_sdc_to_measure_results` next to `get_measure_results` to make it clear that the former returns a tuple that matches the structure of the latter (we're using the tuple as a record, not just as an immutable list, in other words). However, a separate module keeps everything related to disclosure control together, which is clearer.

I guess that this might suggest we need a `namedtuple` to make the record's structure explicit. I'm going to leave that question for another day.

tests/integration/test_main.py
iaindillingham: We did, and I neglected to include them. I have done so now, for this test.

tests/integration/test_main.py
iaindillingham: They're not, now, because `test_generate_measures` has richer data.

evansd (approved):
This is all remarkably neat and non-invasive! I think applying it to user-supplied dummy data was definitely the right call here, now I see how it simplifies the implementation.

PR #1748, opened 2023-11-17
docs: Document medications table

### LINKED ISSUE #1376: Document medications table

rebkwok:
Fixes #1376


brianmackenna:
Thank for tagging - I don't have time for a full review. @richiecroker can you ensure someone from your team does a full review please and is available for input for @rebkwok 

Some initial thoughts

richiecroker:
@brianmackenna Vicky will take a look from our team, but I can't seem to tag her

alschaffer:
A few comments...

> Burn-in periods refer to the process of excluding any data relating to a particular event for a period of time following that event. Burn-out periods refer to the process of excluding any data elating to a particular event for a period of time preceding that event.

I find this description confusing. I'm not sure what you mean by "event", as to my mind there are two relevant types of events here, the medication prescribing event, and the practice transfer event. Plus, there is a typo ("elating").

Second, example code is useful, but research is highly variable and so it's likely people may end up using different approaches to deal with this issue. I can envision cases where I might use something like the supplied code, but for certain research questions my first instinct would be to restrict my population to people with at least 3 months practice registration at the start of the study period/at least 3 months remaining at the end, within `define_population` instead. So the important thing is that there is enough information for a researcher to make an informed decision. For instance, when you say the data are unreliable, does that mean that the data recorded are potentially inaccurate (i.e., the medication was not actually prescribed despite being recorded as such), or simply that the prescribing patterns don't really reflect typical care?

It would also be helpful to state why a 12-wk burn in/out period was used. Is this the period during which we feel comfortable that medication prescribing will have stabilised? Is this what we recommend everyone uses?

speed-vm:
Thank you Becky. I have made some comments, but for my understanding I'd quite like to chat through available column options 

**Medications**
Ideally available columns would include:
Date of issue
dm+d - drug
Quantity - _where the formulation is tablet, capsule, ampoule, vial etc the quantity will be the number of tablets, capsules, ampoules, vials etc_
Intent - acute or repeat - _Was the prescription issued as a one off for example a course of antibiotics, or has it been issued as a repeat, for example a medication for blood pressure._ 
Formulation perhaps - tablets, caps, liquid

An objective 'discontinuation' column could be helpful if available for those doing work around adverse drug reactions / discontinuation in older adults for example.

rebkwok:
@brianmackenna @alschaffer @speed-vm Thanks for all the comments.  At the moment, we are not adding additional columns to the medications table. This PR is just about documenting what is there already.  I'll move comments about additional desired columns to another issue.

Burn-in/burn-out periods have been [discussed extensively](https://github.com/opensafely-core/ehrql/issues/1376#issuecomment-1684080652) around medications, which led me to think it was an important thing to document. @speed-vm Can you confirm what epi terms should be used as well/instead, and what they refer to? I'm used to terms like "wash out" periods in clinical trials, but I'm not sure if that's something also used in epi research.

> 'More events happening prior to a patient leaving a practice' Has this been described anywhere formally, might be nice to link to it?
Unfortunately no, as per Seb, the source for most of these observations is Chris from TPP :shrug:

alschaffer:
"Wash-out" period is frequently used in observational (non-clinical trial) research, but the term is used quite loosely (at least in my experience). I would say the way it's used is more similar to what you've described as "burn-in" period (excluding observations occurring within X days of an index event or start of study period). As far as I know what you've described as the "burn-out" period doesn't have an equivalent term in epidemiology.

speed-vm:
1) Here is a [doc](https://docs.google.com/document/d/1GDuvBBovS4z1Hq40UczkWjHIkkzLTprGwMkqJOtCIlY/edit) with edit for medicationIssue table. @brianmackenna have I got the right end of the stick. Without other columns exposed/available, it doesn't feel right to go too far beyond the previous docs. 

2. Caught up with @alschaffer . In agreement, with changes made. We could also think of no other common 'OS specific medication scenarios' at this stage that might warrant mentioning in that section.

evansd (approved):

PR #1754, opened 2023-11-20
Pure python

sebbacon:


inglesp:
This is some really interesting work, but I'm going to close the PR as it's not something we plan on merging at the moment.

PR #1759, opened 2023-11-20
Add some explanatory docs on measures

### LINKED ISSUE #1357: Document measures support
Added in #1276.

evansd:
Closes #1357

### DIFF
--- a/docs/explanation/measures.md
+++ b/docs/explanation/measures.md
@@ -1,0 +1,299 @@
+# Using the measures framework
+
+## Introduction
+
+The measures framework is used to calculate **quotients** (i.e. a numerator divided by a denominator) and to see how these **vary over time** and when broken down by different **groupings**.
+
+The numerators, denominators and groups are all defined using ehrQL queries, just like we would use with a dataset definition. But the way those definitions are used is a bit different.
+
+To explain the concepts involved we'll start with a basic but complete example and walk through each of the elements.
+
+
+## Basic example
+
+Suppose we want to know what proportion of the patients prescribed atorvastatin tablets in a given month were prescribed 80mg tablets, with the results broken down by sex and calculated for each of the first six months of 2022. Here's how we'd do that using the measures framework:
+
+### Code
+
+```python
+from ehrql import INTERVAL, case, create_measures, months, when
+from ehrql.tables.beta.core import medications, patients
+
+# Every measure definitions file must include this line
+measures = create_measures()
+
+# Small codelist for demonstration purposes; the real list would be longer
+atorvastatin_tablets = [
+    "39733211000001101",
+    "39695411000001103",
+    "39733011000001106",
+    "39733111000001107",
+]
+
+atorvastatin_80mg_tablets = [
+    "39733211000001101",
+]
+
+# The use of the special INTERVAL placeholder below is the key part of
+# any measure definition as it allows the definition to be evaluated
+# over a range of different intervals, rather than a fixed pair of dates
+rx_in_interval = medications.where(
+    medications.date.is_during(INTERVAL)
+)
+atorvastatin_rx = rx_in_interval.where(
+    medications.dmd_code.is_in(atorvastatin_tablets)
+)
+atorvastatin_80_rx = rx_in_interval.where(
+    medications.dmd_code.is_in(atorvastatin_80mg_tablets)
+)
+
+has_recorded_sex = patients.sex.is_in(["male", "female"])
+
+measures.define_measure(
+    name="atorva_80",
+    numerator=atorvastatin_80_rx.exists_for_patient(),
+    denominator=atorvastatin_rx.exists_for_patient() & has_recorded_sex,
+    group_by={
+        "sex": patients.sex
+    },
+    intervals=months(6).starting_on("2022-01-01"),
+)
+```
+
+### Running the example
+
+You can save this file as `measure_definition.py` and then run the [`generate-measures`](../../reference/cli/#generate-measures) command on it:
+```
+opensafely exec ehrql:v0 generate-measures measure_definition.py --output measures.csv
+```
+
+### Results
+
+This should produce a file called `measures.csv` whose contents look something like this:
+
+measure | interval_start | interval_end | ratio | numerator | denominator | sex
+-- | -- | -- | -- | -- | -- | --
+atorva_80 | 2022-01-01 | 2022-01-31 | 0 | 0 | 5 | female
+atorva_80 | 2022-01-01 | 2022-01-31 | 0.2 | 2 | 10 | male
+atorva_80 | 2022-02-01 | 2022-02-28 | 0 | 0 | 3 | female
+atorva_80 | 2022-02-01 | 2022-02-28 | 0.2 | 1 | 5 | male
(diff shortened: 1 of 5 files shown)


docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
+# Using the measures framework
+
+## Introduction
+
+The measures framework is used to calculate **quotients** (i.e. a numerator divided by a denominator) and to see how these **vary over time** and when broken down by different **groupings**.
+
+The numerators, denominators and groups are all defined using ehrQL queries, just like you would use with a dataset definition. But the way those definitions are used is a bit different.
```
iaindillingham, line 7: I can see that you've used "you" and "we" deliberately throughout this document. However, I think it's friendlier to maintain the illusion that we will be writing my code together.

```suggestion
The numerators, denominators and groups are all defined using ehrQL queries, just like we would use with a dataset definition. But the way those definitions are used is a bit different.
```

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+)
+
+has_recorded_sex = patients.sex.is_in(["male", "female"])
+
+measures.define_measure(
+    name="atorva_80",
+    denominator=atorvastatin_rx.exists_for_patient() & has_recorded_sex,
+    numerator=atorvastatin_80_rx.exists_for_patient(),
+    group_by={
+        "sex": patients.sex
```
iaindillingham, line 57: Why isn't this `has_recorded_sex`?

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+atorvastatin_80_rx = rx_in_interval.where(
+    medications.dmd_code.is_in(atorvastatin_80mg_tablets)
+)
+
+has_recorded_sex = patients.sex.is_in(["male", "female"])
+
+measures.define_measure(
+    name="atorva_80",
+    denominator=atorvastatin_rx.exists_for_patient() & has_recorded_sex,
+    numerator=atorvastatin_80_rx.exists_for_patient(),
```
iaindillingham, line 55: Could you reverse the order of `denominator` and `numerator`? Writing the thing on the bottom of the fraction above the thing on the top of the fraction is adding a small, but noticeable amount of cognitive effort. (I appreciate that you explain the denominator before you explain the numerator, but I don't think writing the arguments in explanation order is that important.)

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+series](../../reference/language/#BoolPatientSeries). (It is also possible
+to supply an integer here, but we'll cover this later.)
+
+The **`numerator`** arguments defines the _additional_ condition that
+patients must also meet to be included in the numerator. (Again, this
+can also be an integer as we'll discuss later.)
+
+The **`group_by`** argument defines how we would like our results broken
+down. It's optional – leaving it out means that we'll get a single pair
+of numerator/denominator numbers for each time interval – but most
```
iaindillingham, line 138: As we're working in tables, could we say "a single row for each time interval"?

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+date pairs but, as typing these all out by hand would be laborious, we
+provide several convenience functions for generating such lists:
+[`years`](../../reference/language/#years.starting_on),
+[`months`](../../reference/language/#months.starting_on) and
+[`weeks`](../../reference/language/#weeks.starting_on).
+
+
+### The `INTERVAL` placeholder
+
+The key difference between the ehrQL you'd write for a dataset definition and the ehrQL you write for a measure comes in this line here:
```
iaindillingham, line 158: Hey, I thought we were doing this together!

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+
+## Additional notes
+
+### Multiple measures
+
+We're not restricted to just one measure per file. If we wanted to add
+another measure using the same numerator and denominator but this time
+broken down by age band we would start by defining an age band variable
```
iaindillingham, line 195:
```suggestion
broken down by age band, then we would start by defining an age band variable
```

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+
+As we defined our measure to cover monthly intervals over a six month
+period, this means that this single measure will produce 60 rows – ten
+for each month.
+
+
+### Dummy data
+
+When run outside of the secure environment ehrQL will generate dummy
+data for measures just as it does with datasets. However note that
```
iaindillingham, line 289:
```suggestion
data for measures just as it does with datasets. However, note that
```

iaindillingham (approved):
Writing prose for ehrQL is harder than writing code for ehrQL. At least with the code, there's a linter, formatter, and test suite. I've made several comments; non are deal-breakers; all are more opinionated than I'd care to admit.

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+To explain the concepts involved we'll start with a basic but complete example and walk through each of the elements.
+
+
+## Basic example
+
+Suppose we want to know what proportion of the patients prescribed atorvastatin tablets in a given month were prescribed 80mg tablets, with the results broken down by sex and calculated for each of the first six months of 2022. Here's how we'd do that using the measures framework:
+
+### Code
+
+```python
```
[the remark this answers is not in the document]
evansd: This doesn't work unfortunately as this isn't a dataset definition.

docs/explanation/measures.md
```diff
@@ -0,0 +1,299 @@
...
+The **`group_by`** argument defines how we would like our results broken
+down. It's optional – leaving it out means that we'll get a single pair
+of numerator/denominator numbers for each time interval – but most
+measures define at least one set of groups. It is supplied as a
+dictionary mapping group names to group definitions.
+
+As we saw in the example above, each group name defined here ends up as
+column in the results. Each group definition should be a "categorical"
+patient series, that is: a patient series which takes only a fixed set
+of values.
```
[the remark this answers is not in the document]
evansd: In practice you can actually use anything here because we don't enforce constraints. But it only makes sense to group by series which have this property.

docs/explanation/measures.md
evansd: Because that's a boolean, not the thing we want to group by. It's a bit annoying we need to define that extra `has_recorded_sex` condition, but if we don't then the dummy data system will generate roughly equal quantities of `male`, `female`, `intersex` and `unknown` patients and we end up with 24 rows of output rather than 12.

Ideally we'd adjust the dummy data system to use a distribution that roughly reflects the population in the database, but I don't want to tackle that right now.

docs/explanation/measures.md
evansd: Yeah, I wondered about that too and didn't have a clear sense about which was least worst. Happy to swap to the order here.

docs/explanation/measures.md
evansd: I think I just used whatever felt most natural at the time I happened to be writing so I'm not sure it's particularly consistent. Thanks for picking up on these.

docs/explanation/measures.md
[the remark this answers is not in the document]
iaindillingham: That's a shame. Thanks for letting me know.

PR #1762, opened 2023-11-21
Support passing a series to is_in()

rebkwok:


rebkwok:
Currently the spec test I've added for doing `is_in()` with a series works for the in_memory engine, but for the sqlite engine it  returns True for all patients. If there is one matching patient, it returns True for all of them. If there are no matching patients it returns False for all of them (not None for the null patient 4).  It is doing _something_ that looks at patient ID, I think, because if there's a value in e.i1 which overlaps with p.i1 but for a different patient, it doesn't return that as True.

rebkwok:
This is now marginally better with [commit](https://github.com/opensafely-core/ehrql/pull/1762/commits/c4f80869eeab7470f5a50e4bd155ec8b5ad52f3c), in that the sqlite and trino engines now return the right result as long as there is data for the patient, but don't deal with None properly (i.e. in the spec test, patients 1, 2, and 3 are correct, but 4 returns False rather than None).

However, mssql is complaining about a syntax error, which seems like it might be related to [this issue](https://github.com/sqlalchemy/sqlalchemy/issues/3212), but I can't figure out how to apply the suggested solution to how we're using it.

rebkwok:
Passing the spec test now, using case/in_ instead of exists.

tests/spec/series_ops/test_containment.py
```diff
@@ -12,6 +12,16 @@
...
         """,
+    e: """
+          |  i1
+        --+-----
+        1 | 101
+        2 | 201
+        2 | 203
+        3 | 333
+        3 | 334
+        4 |
```
evansd, line 23: I'm afraid I don't see how the approach you've got at the moment is going to work because it's not doing any kind of join or filter on patient_id. I think the only reason it appears to work is because you have distinct values in `e` for each patient. If you to the bottom of `e` a line like:
```
2 | 301
```
Then I think you'll find all the SQL tests will fail (though the in-memory one should be fine).

What makes this problem difficult is that at the point where we're constructing the expression we don't have access to the relevant patient_id column.

It's tempting to try to do something like `lhs.table.c.patient_id` but that won't work in every case because we cant' always guarantee that `lhs` comes from a single table – it might be the result of some case expression over several tables.

What we're really want to say is: use the `patient_id` column of the primary table _whatever that ends up being_ in the final query; but I'm not sure how to implement that in SQLAlchemy.

In very old versions of ehrQL I used this horrible trick of having a "placeholder" ID column which got created here:
https://github.com/opensafely-core/ehrql/blob/58ebdbb97e79c96c33a20b669a1087769b2d88d4/databuilder/query_engines/sqlite.py#L59-L79

And then used the SQLAlchemy `replacement_traverse()` function to replace references to it with a real column later:
https://github.com/opensafely-core/ehrql/blob/58ebdbb97e79c96c33a20b669a1087769b2d88d4/databuilder/query_engines/sqlite.py#L295-L337

That's all pretty nasty though.

rebkwok: So in a previous [commit](https://github.com/opensafely-core/ehrql/pull/1762/commits/4f25e0c2f2659aa52101cdcdbdbb975b5e8ab9c8) I used patient_id from `self.population_table`, because the `get_select_query_for_node_domain` [already uses it](https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/base_sql.py#L756) - which also sort-of worked, but returned False for the patient that we expect to get None for.

rebkwok: Right, yes, and now I remember why I was trying not to use `exists` - because mssql doesn't like it

rebkwok:
So, with the last commit I was able to use the population table patient_id, with exists, and get mssql to allow it, and the spec test (with the extra non-unique value) is working - except for the None values. i.e. patient 4 has None for `i1` in both tables and we expect the result of the is_in to be None, but it's False because my exists query can't tell the difference

rebkwok:
@evansd This still fails in various ways but:

tests/spec/series_ops/test_containment.py - includes the test you suggested, making the values non-unique across patients.  All not-in-memory query engines do the same thing - they return the right result for patients with values, but don't detect null values.

Using the `self.population_table` seems to work, as long as it already exists, which it doesn't if the is_in query is part of the population definition query. I attempted to add a placeholder patient, and I added a temporary test in tests/integration/test_query_engines.py where all engines fail because they can't find the placeholder table (temporary in the sense that it should probably be a unit test in a different place, but this was the simplest place I found to make a test that failed as I expected it to).

However, I'm now failing to figure out how to do the replacement of the placeholder ID column reference. I think it only needs to be done on the population query, but I don't think I understand how to inspect the parts of the query in `replacement_traverse` properly. Any idea if it's going in the right direction?

evansd:
Oh great, that does sound like it's going in the right direction!

I think relying on the population_table might be a bit of a red-herring though, and might only work with the simplistic queries in the spec tests. The reason I say that is because I _think_ the patient_id column needs to come from whatever is the left-most table in the join. For the final results query that _will_ always be the population table. But for intermediate queries I don't think it will.

So if you can get something like the placeholder trick working then I'd just use that everywhere and forget about the population table.

My thought was that if you can construct a placeholder patient_id column you could modify `apply_patient_joins()`:
https://github.com/opensafely-core/ehrql/blob/8772e07c1729a0c027fc1326bb99bdd400ac6990/ehrql/query_engines/base_sql.py#L804-L826

And do _something_ like:
```py

query = replacement_traverse( 
    query, {}, replace=lambda obj: join_key if obj is PLACEHOLDER_PATIENT_ID else None 
)
```
Before the `implicit_joins = ...` line.

I've no idea if that would actual work though.

`replacement_traverse` is pretty weird. I haven't used it much but there are some docs on it here:
https://docs.sqlalchemy.org/en/21/core/visitors.html#sqlalchemy.sql.visitors.replacement_traverse

rebkwok:
Another attempt at the placeholder patient id column, with new confusions:

1) The `replacement_traverse` is working, in the sense that it's finding the placeholder patient ids and replacing them 

2) If I put the `replacement_traverse` before the `implicit_joins = ... ` line, the sqlite engine complains that it's confused about the joins because there are multiple possible froms that could be used. This happens even if I use the previous version of the code that only did the placeholder table if the population table wasn't available, where I know that there are no placeholder tables to replace.  In that case, a `str(query)` before and after the  `replacement_traverse` is identical, but it's obviously changed something even though there were no replacements.  If I put the `replacement_traverse` after the `implicit_joins = ... ` line, the sqlite engine is OK, and the spec test fails in the same way as the other engines (because it's not dealing with nulls yet)

3) I modified the test in `integrations/test_query_engine.py` to try and simplify a more realistic query so I could see what was going on (it's a weird nonsense query, but I think should be possible).  I can see that it replaces the patient placeholder column, but it's not replacing it with the right patient table 

The test does this:
```
dataset = Dataset()
dataset.define_population(clinical_events.exists_for_patient())
dataset.n = patients.date_of_birth.is_in(clinical_events.date)
```
`str(query)` before `replacement_traverse`:
```
WITH cte_1 AS 
(SELECT DISTINCT clinical_events.patient_id AS patient_id 
FROM clinical_events), 
cte_2 AS 
(SELECT cte_1.patient_id AS patient_id 
FROM cte_1 
WHERE cte_1.patient_id IS NOT NULL), 
cte_3 AS 
(SELECT clinical_events.patient_id AS patient_id, clinical_events.date AS date 
FROM clinical_events 
WHERE clinical_events.patient_id IN (SELECT cte_2.patient_id 
FROM cte_2))
 SELECT cte_2.patient_id, EXISTS (SELECT NULL AS anon_1 
FROM cte_3, patients_a9ea2dae7670, patients 
WHERE cte_3.patient_id = patients_a9ea2dae7670.patient_id AND cte_3.date = patients.date_of_birth) AS n 
FROM cte_2
```

After:
```
WITH cte_1 AS 
(SELECT DISTINCT clinical_events.patient_id AS patient_id 
FROM clinical_events), 
cte_2 AS 
(SELECT cte_1.patient_id AS patient_id 
FROM cte_1 
WHERE cte_1.patient_id IS NOT NULL), 
cte_3 AS 
(SELECT clinical_events.patient_id AS patient_id, clinical_events.date AS date 
FROM clinical_events 
WHERE clinical_events.patient_id IN (SELECT cte_2.patient_id 
FROM cte_2))
 SELECT cte_2.patient_id, EXISTS (SELECT NULL AS anon_1 
FROM cte_3, patients 
WHERE cte_3.patient_id = cte_2.patient_id AND cte_3.date = patients.date_of_birth) AS n 
FROM cte_2
```
I think the placeholder table needs to somehow be replaced with a reference to the patients table, not to the clinical events table. `patients.date_of_birth` represents the lhs of the `is_in` query, so the placeholder reference should be using the same, shouldn't it?

evansd:
@rebkwok Sorry for taking so long to get to this. The only way I could try to work out what was going on was to take your PR and hack about on it, which I've done so here:
 * https://github.com/opensafely-core/ehrql/pull/1795

I've got something that passes the spec tests but it definitely isn't right because the gentests find bugs in it if I run them for long enough.

Maybe the best next step is to have a call where we look at the differences in our approaches here and discuss?

inglesp:
Superseded by #1795

PR #1763, opened 2023-11-22
Add `count_episodes_for_patient()`

### LINKED ISSUE #678: Add a `count_episodes_for_patient` aggregation to date series
The exact name of this aggregation is TBC but the behaviour we need is that described by the `returning="number_of_episodes"` and `episode_defined_as` arguments as used in cohortextractor. That is, we want to count events but consider events not more than N days apart as being the same event. This N day interval is commonly referred to as a "washout period".

For example, given these dates and a washout period of 3 days, the episode count would be 2 (events A-C get merged into one episode, and events D-E into another): 

Event ID | Date
-- | --
A | 2020-01-01
B | 2020-01-04
C | 2020-01-06
D | 2020-01-20
E | 2020-01-22

Implicit in this example is the fact that the washout clock restarts with each new event. That is, even though events A and C are separated by more than 3 days they don't count as distinct episodes because event B is not more than 3 days from either. This seems to be the behaviour researchers expect and is also the easiest to implement in SQL.

We implement this in cohortextractor using the `LAG` window function:
https://github.com/opensafely-core/cohort-extractor/blob/76eb09fd43d1bd42133dd537524d2120a0d5ac66/cohortextractor/tpp_backend.py#L1461-L1487

I expect the query model to look something like this:
```
class AggregateByPatient:
    ...
    class CountEpisodes(Series[int]):
        source: Series[date]
        washout_period_days: Series[int]
```

evansd:
I've taken the decision to only accept a fixed "maximum gap" duration here i.e. you can't have a gap size which varies dynamically for each patient. I don't think it would actually be that difficult to support this if we had to, but it would require some thought and I don't feel like it would be worth it at this stage. 

I've let Hypothesis have a good run at this locally and, apart from a silly bug in the query rewriter which I'd forgotten to update, it didn't find issues.

Closes #678

### DIFF
--- a/docs/includes/generated_docs/language__series.md
+++ b/docs/includes/generated_docs/language__series.md
@@ -2001,6 +2001,39 @@ Return the maximum value in the series for each patient (or NULL if the patient
 has no values).
 </div>
 
+<div class="attr-heading" id="DateEventSeries.count_episodes_for_patient">
+  <tt><strong>count_episodes_for_patient</strong>(<em>maximum_gap</em>)</tt>
+  <a class="headerlink" href="#DateEventSeries.count_episodes_for_patient" title="Permanent link">🔗</a>
+</div>
+<div markdown="block" class="indent">
+Counts the number of "episodes" for each patient where dates which are no more
+than `maximum_gap` apart (specified in [`days()`](#days) or [`weeks()`](#weeks))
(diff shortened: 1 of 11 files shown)


inglesp (approved):
Pleasingly straightforward

PR #1780, opened 2023-11-27
fix: Only encode categorical strings as Arrow dictionaries

evansd:
ehrQL's notion of "categorical" covers variables of any type which take only an explicitly enumerated set of values, and Arrow supports creating dictionaries of any type. But non-string categoricals are unusual, not all that useful, and can cause problems with downstream software so we no longer create them.


inglesp (approved):

sebbacon:
Worth bearing in mind numeric categoricals are, IIRC, the default in Stata. But I also seem to recall @rebkwok's neat Stata stuff covering this off already.

PR #1783, opened 2023-11-27
fix: Population can only be defined once

### LINKED ISSUE #775: Calling `Dataset.set_population()` more than once should fail
@milanwiedemann noticed that you can do this, and that it's confusing. Only the last call defines the dataset.

We should warn or probably fail with an exception, if a dataset definition tries to do this.

milanwiedemann:
Currently `.define_population(...)` can be used multiple times in a dataset definition and only the last call defines the dataset. This raises an error if `.define_population(...)` is used more than once.

I didn't think we need to mention this in the docs, but we could add a short note to `define_population()`?

Fixes #775

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -63,6 +63,10 @@ class Dataset:
         dataset.define_population(patients.date_of_birth < "1990-01-01")
         ```
         """
+        if "population" in self.variables:
+            raise AttributeError(
+                "define_population() should be called no more than once"
+            )
         validate_population_definition(population_condition._qm_node)
         self.variables["population"] = population_condition
 
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -145,6 +145,13 @@ def test_dataset_rejects_invalid_variable_names(variable_name, error):
         setattr(Dataset(), variable_name, patients.i)
 
 
+def test_cannot_define_population_more_than_once():
+    dataset = Dataset()
+    dataset.define_population(patients.exists_for_patient())
+    with pytest.raises(AttributeError, match="no more than once"):
+        dataset.define_population(patients.exists_for_patient())
+
+
 def test_cannot_reassign_dataset_variable():
     dataset = Dataset()
     dataset.foo = patients.date_of_birth.year


evansd (approved):
Thanks for catching this Milan!

My view is that it ins't worth mentioning in the docs. It's just more words for people to read. If anyone tries to do it they'll get a clear enough error message and I think that's all we need.

ehrql/query_language.py
```diff
@@ -64,6 +64,10 @@ def define_population(self, population_condition):
         ```
         """
         validate_population_definition(population_condition._qm_node)
+        if "population" in self.variables:
+            raise AttributeError(
+                "Population is already defined and cannot be reassigned"
+            )
```
iaindillingham, line 70: I think this should move up one line (above `validate_population_definition`). If someone calls `dataset.define_population` twice, the first time correctly and the second time incorrectly, we want to tell them that they've called `dataset.define_population` twice, not that they called it incorrectly the second time. (And only then, when they've called it correctly, tell them that they've called it twice.)

I also think the error message could be clearer. We know that the population is stored within `dataset.variables["population"]`, but a user doesn't. The message we want to send to them is something like

```
define_population() should be called no more than once
```

Notice that we start with a lower case "d".

tests/unit/test_query_language.py
```diff
@@ -145,6 +145,14 @@ def test_dataset_rejects_invalid_variable_names(variable_name, error):
         setattr(Dataset(), variable_name, patients.i)
 
 
+def test_cannot_reassign_population():
+    dataset = Dataset()
+    dataset.define_population(patients.exists_for_patient())
+    dataset.year_of_birth = patients.date_of_birth.year
+    with pytest.raises(AttributeError, match="already defined"):
+        dataset.define_population(dataset.year_of_birth <= 2000)
```
iaindillingham, line 153: Why are we using a different population condition?

iaindillingham (commented):
I don't feel I can approve; but I do feel I can comment 🙂 Let me know whether you agree/disagree, and possibly seek a second opinion.

ehrql/query_language.py
milanwiedemann: thanks @iaindillingham, will fix now!

tests/unit/test_query_language.py
milanwiedemann: true, it should be simpler because the difference isn't relevant and we're not testing it.

tests/unit/test_query_language.py
```diff
@@ -145,6 +145,14 @@ def test_dataset_rejects_invalid_variable_names(variable_name, error):
         setattr(Dataset(), variable_name, patients.i)
 
 
+def test_cannot_reassign_population():
```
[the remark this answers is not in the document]
milanwiedemann: thx, renamed to `test_cannot_define_population_more_than_once`

iaindillingham (approved):

PR #1784, opened 2023-11-27
Fix long running gentests

evansd:
This adds a few follow up fixes to #1760:

 * Replace the `::test_query_model` test specifier which was accidentally removed.
 * Fix the bash conditional syntax (there was a missing space between `0` and `]`, but modern bash tends to favour the double bracket syntax in any case.)
 * Don't attempt database teardown on a broken connection (this was generating new failures which then prevented the test from exiting in the way we intended.)

### DIFF
--- a/.github/workflows/generative-tests.yml
+++ b/.github/workflows/generative-tests.yml
@@ -47,12 +47,12 @@ jobs:
 
             # Do the actual testing
             set +e
-            just test tests/generative/test_query_model.py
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
Looks good, sorry for the couple of silly mistakes that slipped through in teh first PR

evansd:
Blame the lazy reviewer :wink: 

TBH, I hadn't though at all about that `finally` clause until I saw the stacktrace.

PR #1787, opened 2023-11-29
feat: Bump version

iaindillingham:
071b90a included a conventional commit, but unfortunately this didn't trigger the tag-new-version job. (Possibly this was because it included the word "fix" twice?) It wasn't a transient error: I've rerun the workflow [1], and it didn't trigger the job the second time either.

I'd really, really like to tag a new version, and to build and release a new image, so I'm bumping the version with an empty commit.

[1]: https://github.com/opensafely-core/ehrql/actions/runs/7018779199


iaindillingham:
Backing out of this. It wasn’t the best way to achieve the desired result, and turned out not to be necessary because of subsequent merges to `main`.

PR #1788, opened 2023-11-29
fix: `UnboundLocalError` in `MeasureTimer`

milanwiedemann:
Andrea noticed an error[^1] that we didn't catch in the initial release of `MeasureTimer` in  #1757. Making `elapsed_time` an instance variable (`self.elapsed_time`) should fix the error I think? This also changes the name of the first argument of the `@classmethod` to the convention `cls`.

`UnboundLocalError: cannot access local variable 'elapsed_time' where it is not associated with a value`

[^1]: https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1701258939491439

### DIFF
--- a/ehrql/measures/calculate.py
+++ b/ehrql/measures/calculate.py
@@ -68,7 +68,7 @@ class MeasureTimer:
         self.counter = 0
 
     @classmethod
-    def from_grouped(csl, timeout, grouped):
+    def from_grouped(cls, timeout, grouped):
         num_iterations = sum(
             [len(intervals) for denominator, intervals in grouped.keys()]
         )
@@ -81,15 +81,15 @@ class MeasureTimer:
 
     def check_timeout(self, interval):
         if interval != self.previous_interval:
-            elapsed_time = time.time() - self.start_time
             self.counter += 1
             self.previous_interval = interval
-        if self.counter >= 12:
-            projected_time = elapsed_time / self.counter * self.num_iterations
-            if projected_time > self.timeout:
-                raise MeasuresTimeout(
-                    f"Generating measures exceeded {self.timeout}s time limit."
-                )
+            if self.counter >= 12:
+                self.elapsed_time = time.time() - self.start_time
+                projected_time = self.elapsed_time / self.counter * self.num_iterations
+                if projected_time > self.timeout:
+                    raise MeasuresTimeout(
+                        f"Generating measures exceeded {self.timeout}s time limit."
+                    )
 
 
 class MeasureCalculator:
--- a/tests/integration/measures/test_calculate.py
+++ b/tests/integration/measures/test_calculate.py
@@ -115,13 +115,15 @@ def test_get_measure_results_with_timeout(patched_time, in_memory_engine):
         numerator=foo_event_count,
         denominator=event_count,
         intervals=intervals,
+        group_by=dict(
+            sex=patients.sex,
+        ),
     )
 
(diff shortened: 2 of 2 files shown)


ehrql/measures/calculate.py
```diff
@@ -81,11 +81,11 @@ def start(self):
 
     def check_timeout(self, interval):
         if interval != self.previous_interval:
-            elapsed_time = time.time() - self.start_time
+            self.elapsed_time = time.time() - self.start_time
             self.counter += 1
             self.previous_interval = interval
         if self.counter >= 12:
-            projected_time = elapsed_time / self.counter * self.num_iterations
+            projected_time = self.elapsed_time / self.counter * self.num_iterations
```
evansd, line 88: I think this should fix it, but the alternative would be to only perform the `if self.counter >= 12` check when the interval changes. There's no point checking it every loop iteration because the values only change when we change interval.

You could either do that by indenting the whole block like:
```py
        if interval != self.previous_interval:
            self.previous_interval = interval
            self.counter += 1
            if self.counter >= 12:
                elapsed_time = time.time() - self.start_time
                projected_time = elapsed_time / self.counter * self.num_iterations
```

Or you could use a "guard clause" which is when you check a condition at the start of a function and then exit early if it's not met:
```py
        if interval == self.previous_interval:
            return
        self.previous_interval = interval
        self.counter += 1
        if self.counter >= 12:
            elapsed_time = time.time() - self.start_time
            projected_time = elapsed_time / self.counter * self.num_iterations
```

evansd (approved):
Sorry, I should have spotted this in review but obviously wasn't paying close enough attention.

The reason the test isn't catching this is that it doesn't cover the case where the loop gets to 12 iterations and _doesn't_ timeout. I think if you just change the mocked time calls to do 13 fast queries and then make the 14th slow it will expose the bug.

ehrql/measures/calculate.py
milanwiedemann: > The reason the test isn't catching this is that it doesn't cover the case where the loop gets to 12 iterations and doesn't timeout. I think if you just change the mocked time calls to do 13 fast queries and then make the 14th slow it will expose the bug.

Currently when I go through the loops, the 12th iteration doesn't timeout (1x slow and 11x fast) so `projected_time` (57500.0) get's calculated once and is below `self.timeout` (259200.0). The next iteration the very slow query pushes the `projected_time` to 461538.5 and times out. Also just doing 13 fast queries doesnt expose the bug 😕

evansd: Sorry, you're quite right. (I feel a bit like ChatGPT here: confidently asserting stuff and then apologising and correcting myself when it's pointed out I'm talking crap.)

The issue isn't that there aren't enough intervals. It's that there isn't more than _one row per interval_ because there aren't any groups. If you add this to the test measure definition (and change nothing else) then it hits the bug:
```py
        group_by=dict(
            sex=patients.sex,
        ),
```

I have actually checked that this is true this time – rather than just talking out of my arse.

milanwiedemann: thanks @evansd 😬 I think we're finally close to merging and fixing the bug. I just wanted to check whether we both got this recent change right.

You suggested moving the line `elapsed_time = time.time() - self.start_time` into the second if statement, which means that we're only taking the time from the 12th query onwards, but that's ok I think?

milanwiedemann: ok great! thanks for all the troubleshooting and suggestions :)

PR #1790, opened 2023-11-29
Clean up logical terminology

### LINKED ISSUE #1439: Consider changing `case(..., default=...)` to `case(..., otherwise=...)`
This would make it consistent with the "bare when clause" construction added in #1438 e.g
```py
category = when(size < 15).then("small").otherwise("large")
```

And I think it reads slightly better e.g.
```py
category = case(
    when(size < 15).then("small"),
    when(size < 25).then("medium"),
    otherwise="large",
)
```

We'd want to support both keywords for a time (possibly quite a long time) while we make sure that all the uses of the old style have been updated.

### LINKED ISSUE #1621: Consider renaming `if_null_then` to `when_null_then`
@evansd demonstrated the `case` statement during today's team talks. Because `if` is reserved in Python, we use `when` to indicate the condition in the `case` statement. Someone (@madwort?) suggested that for consistency, `if_null_then` should be `when_null_then`.

Someone else (@alschaffer?) also asked why it's `if_null_then` and not `if_null.then`, but I think that's a separate issue.

evansd:
This makes two changes to our logical functions:
 * Replaces `case(..., default=...)` with `case(..., otherwise=...)`
 * Replaces `if_null_then()` with `when_null_then()`

In both cases this gives us better consistency of terminology. The old functions/arguments are retained as aliases for now so there are no backwards compatibility concerns.

We plan to remove the aliases when we release `v1`.

Closes #1439
Closes #1621

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -211,7 +211,7 @@ class BaseSeries:
         """
         return case(
             when(self.is_not_null()).then(self),
-            default=self._cast(other),
+            otherwise=self._cast(other),
         )
 
     def is_in(self, other):
@@ -251,7 +251,7 @@ class BaseSeries:
                 when(self == from_value).then(to_value)
                 for from_value, to_value in mapping.items()
             ],
-            default=default,
+            otherwise=default,
         )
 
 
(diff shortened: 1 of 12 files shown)


ehrql/query_language.py
```diff
@@ -1390,14 +1390,20 @@ def case(*when_thens, default=None):
...
     """
+    if default is not None:
+        if otherwise is not None:
+            raise ValueError("Use `otherwise` instead of `default`")
+        otherwise = default
     cases = _DictArg((case._condition, case._value) for case in when_thens)
-    # If we don't want a default then we shouldn't supply an argument, or else it will
-    # get converted into `Value(None)` which is not what we want
-    if default is None:
+    # If we don't want a `otherwise` value then we shouldn't supply an argument, or else
```
inglesp, line 1409: piconit: s/a/an/

inglesp (approved):

Issue #797, opened 2022-10-20
Add an EMIS schema and backend


evansd:
This will involve adding a module at [`databuilder.tables.beta.emis`](https://github.com/opensafely-core/databuilder/tree/main/databuilder/tables/beta) which describes the data available in EMIS, and then a module at [`databuilder.backends.emis`](https://github.com/opensafely-core/databuilder/tree/main/databuilder/backends) which describes how to map the user-facing representation to the actual database tables.

Depends on:
 * #796

sebbacon:
Does this in turn depend on something like "define core contracts to allow querying of both EMIS and TPP"?

evansd:
No, I don't think it does — and by design. The plan is that (a) the use of the `tables.beta` namespace and (b) the layers of indirection that Data Builder gives us mean that we can get something out there for people to use and then, at our leisure, work towards a set of core contracts that both can implement.

That is, it's possible for `tables.beta.tpp` and `tables.tpp` to co-exist and look different but both end up mapped to the same underlying database tables. So I'm hoping that gives us enough flexibility to decouple "Have _some_ way of querying the data" from "Decide on the one true way that all EHR data should be represented" — which feels like a good thing.

PR #1793, opened 2023-11-30
Add patients and clinical events tables for EMIS backend

rebkwok:
The first part of #797 
- updates the patients and clinicial_events tables in the EMIS backend to use the correct source tables and columns
- adds tables.beta.emis with patients and clinical events tables
- sets up the test framework for emis tests


inglesp (approved):
Looking good, thanks.

What do you plan to do about the duplication between the TPP and EMIS backend tests?

rebkwok:
> What do you plan to do about the duplication between the TPP and EMIS backend tests?

Just doing the first bit in this PR to make sure it's along the right lines, I'll look at that in the next one

PR #1795, opened 2023-11-30
Support passing a series to `is_in()`

### LINKED ISSUE #518: Support passing a series to `is_in()`
We want to allow queries like this:
```
a_events = events.take(event.code == code_a)
b_events = events.take(event.code == code_b)
co_occurring_events = a_events.take(a_events.date.is_in(b_events.date))
```

On the ehrQL side, this involves the `is_in()` method spotting when it's been given a Series object, and applying the `CombineAsSet` aggregation to it before passing it to `Function.In`.

On the query engine side this involves ... I'm not quite sure yet. I got the basic approach working here:
https://github.com/opensafely-core/databuilder/commit/dfc8b5b1e106f0a5e113b7ab2127f1bbff60c6c0

The tricky bit is getting that `patient_id` reference on line 125. At the point where we're building this expression we don't yet know the `patient_id` expression of the whole query into which it's going to be embedded. Possibly this will require reviving something like the "placeholder" trick where we just stick _something_ in there and then use `replacement_traverse()` to rewrite the query later:
https://github.com/opensafely-core/databuilder/commit/58ebdbb97e79c96c33a20b669a1087769b2d88d4

evansd:
This means we can now support queries like:
```py
a_events = events.where(event.code.is_in(codelist_a))
b_events = events.where(event.code.is_in(codelist_b))
co_occurring_events = a_events.where(a_events.date.is_in(b_events.date))
```

Closes #518

### DIFF
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -230,12 +230,12 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     def get_sql_in_value(self, node):
         lhs = self.get_expr(node.lhs)
         if len(node.rhs.value) == 0:
-            # Special case handling for when LHS is NULL and RHS is an empty list: ehrQL
-            # evaluates this as NULL whereas SQL evaluates it as FALSE. (Of course, both
-            # agree than when RHS is empty and LHS is non-NULL then the result is
-            # FALSE.) So we want an expression which evaluates NULL when LHS is NULL,
-            # and FALSE otherwise. The expression `LHS != LHS` behaves as required.
-            return operators.ne(lhs, lhs)
+            # Special case handling for empty lists: ehrQL (like SQL) considers the
+            # expression `x IN (<empty list)` to be FALSE under all circumstances, even
+            # when `x` is NULL. Empty list expressions aren't valid in all databases and
+            # although SQLAlchemy works around this for us it's cleaner just to return
+            # constant FALSE here.
+            return sqlalchemy.literal(False)
         else:
             rhs = self.get_expr_for_multivalued_param(node.rhs)
             return lhs.in_(rhs)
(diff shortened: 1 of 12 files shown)


iaindillingham (commented):
Reading the spec tests, I found myself agreeing with SQL over ehrQL. [The Slack thread][1] was useful; I think I treat NULL as unknown.

[1]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1701765721146619

ehrql/query_engines/base_sql.py
```diff
@@ -216,8 +216,16 @@ def get_sql_is_null(self, node):
...
         lhs = self.get_expr(node.lhs)
-        rhs = self.get_expr_for_multivalued_param(node.rhs)
-        return lhs.in_(rhs)
+        if isinstance(node.rhs, Value) and len(node.rhs.value) == 0:
+            # Special case handling for when LHS is NULL and RHS is an empty list: ehrQL
+            # evaluates this as NULL whereas SQL evaluates it as FALSE. (Of course, both
+            # agree than when RHS is empty and LHS is non-NULL then the result is
+            # FALSE.) So we want an expression which evaluates NULL when LHS is NULL,
+            # and FALSE otherwise. The expression `LHS != LHS` behaves as required.
+            return operators.ne(lhs, lhs)
```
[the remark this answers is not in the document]
evansd: Ah, we don't! But the point is that _when_ it's NULL we want the result to be NULL, and otherwise we want it to be FALSE. And that's what `LHS != LHS` gives us.

iaindillingham (approved):
Thank you for breaking this into small, more digestible commits. It made it much easier to review. (I checked out the branch and stepped through each, to check my understanding.)

ehrql/query_engines/base_sql.py
```diff
@@ -728,7 +773,16 @@ def reify_query(self, query):
         e.g. using `.alias()` to make a sub-query, using `.cte()` to make a Common Table
         Expression, or writing the results of the query to a temporary table.
         """
-        return query.cte(name=f"cte_{self.get_next_id()}")
+        cte = query.cte(name=f"cte_{self.get_next_id()}")
```
rebkwok, line 776: The trino query_engine has its own `reify_query` method that overrides this one - does it need to handle the no_replacement_traverse as well?

rebkwok (approved):

ehrql/query_engines/base_sql.py
evansd: The problem is specifically with CTEs, not other kinds of table-like thing which SQLAlchemy generally considers immutable.

I did consider refactoring things so that `reify_query()` always applies this flag and then calls out to some other `actually_reify_query()` function which subclasses can override. But my thinking was that the test case will catch any future instances of this problem and then we can apply the flag as it's needed and avoid complicating the call stack here.

Issue #743, opened 2022-09-23
Decide how to handle `patient_id` values which aren't ints


evansd:
As gestured at here:
https://github.com/opensafely-core/databuilder/blob/68ca77dd132c7c61d9bb17add4cf7c67b97b0171/databuilder/column_specs.py#L31-L33

I think the solution is probably to default to using a `uint64` but to allow backends to specify a different type if necessary. The creates the slight awkwardness that when creating (or validating) dummy data we won't necessarily match the type that's going to appear in production. Given that patient IDs are largely opaque from the researcher's point of view I'm hoping this won't matter much in practice.

An alternative would be to always treat `patient_id` as an binary blob. But it would have to be a variable length blob as we don't know the size in advance, and these are less efficient to store and work with than fixed size types. And we'd need to use base64 or some equivalent to encode them in CSV. And overall I'd rather not pay this complexity tax everywhere for what's hopefully going to be a reasonably rare occurrence.

PR #1807, opened 2023-12-05
Add hashed_organisation column to every table created in the EMIS db

rebkwok:
See https://bennettoxford.slack.com/archives/C03FB777L1M/p1701708496981819

Every EMIS table contains a hashed_organisation column, which is required to be present in any table queried.  For any tables that we create during an ehrql query, we need to also include a hashed_organisation column, with a valid id (we have our own one, and we can retrieve that from the environment and use it for every row).


ehrql/backends/emis.py
```diff
@@ -30,6 +34,22 @@ class EMISBackend(SQLBackend):
         ehrql.tables.beta.smoketest,
     ]
 
+    EMIS_ORG_HASH = os.environ.get("EMIS_ORGANISATION_HASH", "emis_organisation_hash")
+    HASHED_ORG_COLUMN = sqlalchemy.literal(
+        EMIS_ORG_HASH, type_=sqlalchemy.VARCHAR
+    ).label("hashed_organisation")
```
evansd, line 40: `os.environ` is available as `self.config` in the Backend classes (and in the Query Engine classes too) so it's best to grab things from there e.g.
https://github.com/opensafely-core/ehrql/blob/896a357381d9ecc1aab57fa6a76d7d8ec8db3d7d/ehrql/backends/tpp.py#L502-L505

ehrql/backends/emis.py
```diff
@@ -30,6 +34,22 @@ class EMISBackend(SQLBackend):
...
+
+    def modify_inline_table_args(self, columns, rows):
+        columns.append(
+            sqlalchemy.Column(
+                self.HASHED_ORG_COLUMN.name,
+                self.HASHED_ORG_COLUMN.type,
+                key=self.HASHED_ORG_COLUMN.key,
+            )
+        )
+        rows = tuple([tuple([*row, self.EMIS_ORG_HASH]) for row in rows])
```
evansd, line 50: Because I can't resist code-golfing, you could write this slightly more simply as:
```oy
rows = tuple((*row, self.EMIS_ORG_HASH) for row in rows)
```

tests/integration/backends/test_emis.py
```diff
@@ -153,3 +157,67 @@ def test_patients(select_all_emis):
 
 def test_registered_tests_are_exhaustive():
     assert_tests_exhaustive(EMISBackend())
+
+
+def test_inline_table_includes_organisation_hash(trino_database):
```
evansd, line 162: I think there's probably a cleaner and simpler way of testing the thing you want here but I don't think that's worth worrying about this stage.

ehrql/query_engines/trino.py
```diff
@@ -138,9 +138,18 @@ def create_inline_table(self, columns, rows):
 
     def reify_query(self, query):
         table_name = f"ehrql_{self.global_unique_id}_tmp_{self.get_next_id()}"
+        # Doesn't work with compound select queries
+        # query = query.add_columns(self.backend.HASHED_ORG_COLUMN)
```
evansd, line 142: You can work around this by wrapping the query up as a subquery:
```py
query = sqlalchemy.select(*query.alias().columns)
```

I just tried this locally, and added the line:
```
query.add_columns(sqlalchemy.literal("fun_times").label("with_access_guard"))
```
And it passed the spec tests, which doesn't mean it _definitely_ works but is a good start.

tests/integration/backends/test_emis.py
evansd: I still think this, but on reflection I don't think we have to care about this right now. I'm satisfied the test _does_ test the thing it needs do. We can look at simplifying it later.

ehrql/backends/emis.py
```diff
@@ -25,6 +25,7 @@ class EMISBackend(SQLBackend):
     display_name = "EMIS"
     query_engine_class = TrinoQueryEngine
     patient_join_column = "registration_id"
+    patient_join_column_type = str
```
evansd, line 28: You're absolutely right that we need this and I'd completely forgotten about it; but it's going to be more complicated than this because we the code which works out what kind of Arrow schema to use for the results doesn't (and can't) know which particular backend is going to be in use when it runs. As this change stands, it will blow up when you try to write results to an Arrrow file (and possibly a CSV too, I can't remember how strict the CSV writer is).

There's an old ticket on this here which I'll update with my latest thinking:
https://github.com/opensafely-core/ehrql/issues/743

For now, I think we should pull this commit out (and I'll a reference to it from that ticket so we can re-use the work later). And then we can tackle it at some point as part of the "make EMIS actually work" project.

evansd (approved):
This is far smaller and more self-contained than I feared it might be, so that's great.

You'll see my main comment is about the `patient_id` stuff. Finessing the tests is something we can tackle another time.

tests/integration/backends/test_emis.py
rebkwok: I agree, but couldn't figure out the better way to do it (yet)

ehrql/backends/emis.py
```diff
@@ -30,6 +32,33 @@ class EMISBackend(SQLBackend):
         ehrql.tables.beta.smoketest,
     ]
 
+    def get_emis_org_column(self):
```
evansd, line 35: Oh I meant to say, this deserves some comments. At a minimum a link back to this PR so there's a breadcrumb trail to follow in working out what all this stuff is for. But I think a bit of inline explanation would be helpful as well.

PR #1808, opened 2023-12-05
WIP: docs: Add v1 release notes

### LINKED ISSUE #1764: Write release notes for ehrQL v1
These should describe the major changes for users moving from v0 to v1.

This probably should go in the ehrQL documentation.

Below are items that could be mentioned in these notes. Some of the items listed below are major changes that have happened in recent development for the ["deprecate cohort-extractor" milestone](https://github.com/opensafely-core/ehrql/milestone/2), and not just the breaking changes for v1.

It's debatable what might go into these notes, aside from breaking changes that will mark v1. It's trickier to write a more general changelog for v1, because that's been the entirety of development. This might be much simpler for an ehrQL v2. 

## General changes

- [x] the name change from Data Builder to ehrQL
- [x] direct import of `Dataset`; favour `create_dataset` (#1599)
- [x] disclosure control for measures (#1746)

## Breaking changes

- [x] deprecation of the `databuilder` Docker image; use `ehrql` (#1765)
  * do we need to suggest that the Data Builder image should be manually removal by users, via `docker rm` ? (it's not massive, but is possibly something like 0.5 GB of storage just sat there that will never get removed)
- [x] use of `ehrql:v1` in `project.yaml` instead of `ehrql:v0`
- [x] removal of `databuilder` imports; use `ehrql` imports (#1221)
- [x] use of raw tables (#1705); these are moved into the `raw` namespace
- [x] deprecation of `tpp.hospital_admissions.primary_diagnoses` (#1743); use `tpp.hospital_admissions.primary_diagnosis` (may be moot if `hospital_admissions` is consolidated with `apcs`)
- [x] possible consolidation of `apcs` and `hospital_admissions` (#1761)

(feel free to update)

StevenMaude:
Fixes #1764.


StevenMaude:
Do we want to include recent notable changes that are already included in current v0 versions of ehrQL, particularly those for this recent "deprecate cohort-extractor" milestone? For now, I've written about some of these, but we could remove them.

docs/reference/upgrading-ehrql-from-v0-to-v1.md
```diff
@@ -0,0 +1,123 @@
...
+
+  ```python
+  from ehrql.tables.beta.tpp import clinical_events
+  ```
+
+### Renaming of raw tables
+
+* The "raw" minimally transformed tables are now imported
+  with names starting `beta.raw`.
+  Some tables have also been renamed to remove `_raw` from their name.
```
iaindillingham, line 67: There shouldn't be any tables with this prefix. Although there are several classes with the prefix, they aren't decorated with the `table` decorator. When they are passed to this function, they are named without the prefix. That's a mouthful; an example is clearer. Compare `ehrql.tables.beta.raw.tpp.isaric_raw` with `ehrql.tables.beta.raw.tpp.isaric`.

StevenMaude (commented):
This looks fine :heavy_check_mark:

Only small suggestions I have:

* consider reordering the sections from something roughly like "how likely is this change to affect a user's work"; suggestion:
  * v0 to v1 upgrade
  * removal of `databuilder` imports — this is possibly the one exception in my suggested ordering: it may not be as common, but follows on naturally on from the v0 to v1 upgrade change
  * move of `tables.beta`
  * dummy dataset population size
  * renaming of language features
  * deprecation of `hospital_admissions`
  * querying `household_memberships_2020`
* consider linking to the relevant documentation for the particular feature being discussed, so, for example, the `when_null_then` in the language reference

I don't think these are required though: the document's short enough that this is fine as is.

(I can't approve it, because I opened it :upside_down_face:)

inglesp (approved):

PR #1812, opened 2023-12-06
Add EMIS vaccinations table

### LINKED ISSUE #797: Add an EMIS schema and backend

rebkwok:
Adds the EMIS vaccinations table. In cohort-extractor, the method to
query vaccinations also allowed querying by product code, but it did
so by inner joining the medications table on the immunisation table.
That gives us all medications for patients who also had a vaccination
procedure, which isn't the right thing to do here - a lot of those
medications will not be vaccination-related. In order to get the
same behaviour as cohort-extractor, users can query this vaccinations
table with a codelist for vaccination procedure codes, and the
medications table with a codelist for vaccination product codes.

Closes #797

### DIFF
--- a/docs/includes/generated_docs/schemas/emis.md
+++ b/docs/includes/generated_docs/schemas/emis.md
@@ -12,6 +12,7 @@ from ehrql.tables.emis import (
     medications,
     ons_deaths,
     patients,
+    vaccinations,
 )
 ```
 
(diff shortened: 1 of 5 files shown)


iaindillingham (commented):
✅ in principle, but not in practice, as it's going to be merged into the v1 branch.

PR #1814, opened 2023-12-06
feat: Deprecate hospital_admissions table

inglesp:
It's duplicated by the apcs table


inglesp:
Commits were cherry-picked into #1811.

PR #1815, opened 2023-12-07
Remove some MkDocs log noise

StevenMaude:
The MkDocs logs when serving or building the documentation are noisy, and are at least a screenful of information for me on my display configuration.

This hides information that might actually be useful to people when working on the documentation: new problems that might have been introduced when editing aren't obvious.

This PR reduces, but does not entirely remove, the noise by:

* fixing up some of the issues associated with the logs; *some links were actually broken in the OpenSAFELY documentation site*
* removing some of the false positives by configuring MkDocs options to exclude certain paths from checks

## Other notes

* The logs are now much cleaner when using `just docs-serve --clean` — this takes account of the `exclude_docs` option I've configured in `mkdocs.yml`. This excludes the `includes/` directory, which is a source of multiple false positives. **Perhaps we should make `--clean` the default in `just docs-serve`?** (Sometimes people might want to preview excluded documents, such as a `drafts/` directory, which is why this is not the default MkDocs behaviour.)
* There's a quirk where some URLs work in the ehrQL documentation preview, but not OpenSAFELY's documentation. The difference is that the ehrQL preview uses paths to pages like `/how-to/…`, while the OpenSAFELY site has paths like `/ehrql/how-to/…`. Because of this, some links — either from absolute paths or from relative paths that reach the URL root — work in the ehrQL preview, but are broken on the OpenSAFELY site. I'm not sure why the OpenSAFELY [link check doesn't catch these](https://github.com/opensafely/documentation/issues/1403). I opened [an issue](https://github.com/opensafely-core/ehrql/issues/1817) for restructuring the ehrQL preview.
* The OpenSAFELY documentation site has the same problem with including the ehrQL `includes/` as accessible pages. Because we still have the quirk of having to duplicate our configuration, we should [configure it there too](https://github.com/opensafely/documentation/issues/1402).
* The remaining issues are primarily around how relative paths are automatically generated in the Python code inside `ehrql/docs`. This might be a little bit trickier to get right, so I've [opened an issue to document the problem](https://github.com/opensafely-core/ehrql/issues/1818), and it can be a separate PR.

### DIFF
--- a/docs/how-to/assign-multiple-columns.md
+++ b/docs/how-to/assign-multiple-columns.md
@@ -7,7 +7,7 @@ This guide first shows you how to use `dataset.add_column` to assign one new col
 
 ## Assign one column to the dataset
 
-As mentioned in the tutorial on "[Writing a more complex dataset definition](/tutorial/writing-a-more-complex-dataset-definition/#assign-variables-to-the-dataset)",
+As mentioned in the tutorial on "[Writing a more complex dataset definition](../tutorial/writing-a-more-complex-dataset-definition/index.md#assign-variables-to-the-dataset)",
 we can assign a variable to the dataset by adding a dot and the name of the variable to the dataset,
 followed by an equals sign and the definition of the variable.
 In the following example, the name of the variable is `my_variable`
--- a/docs/includes/generated_docs/backends.md
+++ b/docs/includes/generated_docs/backends.md
@@ -30,15 +30,15 @@ time. For instance, students are often registered with a practice at home and a
 practice at university.
 
 !!! warning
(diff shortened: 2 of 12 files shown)


StevenMaude:
Checked the links by hand in the ehrQL preview, and a preview of the OpenSAFELY documentation site that incorporates this ehrQL branch.

Jongmassey (approved):
Looks good AFAICT

iaindillingham (approved):
🚀

StevenMaude:
I've left the `just docs-serve` command as is; `just docs-serve --clean` will give you cleaner logs without as many lines informing you of deliberate exclusions.

PR #1819, opened 2023-12-07
Expose TPP's CodedEventRange

### LINKED ISSUE #1491: Expose TPP's CodedEventRange table
`CodedEventRange` is a table which can be joined onto `CodedEvent` to gives extra information about numeric values -- specifically: lower bound, upper bound, and comparator.

It is exposed in cohort-extractor via [`patients.comparator_from`](https://docs.opensafely.org/study-def-variables/#cohortextractor.patients.comparator_from).

The simplest approach would be to update [the SQL used to build the `clinical_events` table](https://github.com/opensafely-core/ehrql/blob/4770ec6f1d387c9aead150f30dfdba9d8c94e208/ehrql/backends/tpp.py#L88-L106) to include a join on `CodedEventRange`, and then to add the new fields to [the table definition](https://github.com/opensafely-core/ehrql/blob/4770ec6f1d387c9aead150f30dfdba9d8c94e208/ehrql/tables/beta/tpp.py#L33-L38).  However, we would need to be satisfied that doing this wouldn't significantly affect performance of queries that don't use the joined table.

An alternative would be to add machinery to only include the join when necessary.

As such, the first task will be to measure the performance of typical queries with and without the join, to determine the impact of the join.

Jongmassey:
fixes #1491

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -257,7 +257,8 @@ class TPPBackend(SQLBackend):
                 CAST(NULLIF(ConsultationDate, '9999-12-31T00:00:00') AS date) AS date,
                 NULL AS snomedct_code,
                 CTV3Code AS ctv3_code,
-                NumericValue AS numeric_value
+                NumericValue AS numeric_value,
+                CodedEvent_ID
             FROM CodedEvent
             UNION ALL
             SELECT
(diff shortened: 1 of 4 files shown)


ehrql/tables/beta/tpp.py
```diff
@@ -364,6 +365,56 @@ class clinical_events(EventFrame):
...
+        "The lower bound of the reference range associated with an event's numeric_value",
+    )
+    upper_bound = Series(
+        float,
+        "The upper bound of the reference range associated with an event's numeric_value",
+    )
+    comparator = Series(
+        str,
+        description=(
+            "If an event's numeric_value is returned with a comparator,"
```
inglesp, line 402:
```suggestion
            "If an event's numeric_value is returned with a comparator, "
```

ehrql/backends/tpp.py
```diff
@@ -247,19 +247,41 @@ def modify_query_variables(self, variables):
...
+                cer.UpperBound AS upper_bound,
+                CASE cer.Comparator
+                    WHEN 3 THEN '~'
+                    WHEN 4 THEN '='
+                    WHEN 5 THEN '>='
+                    WHEN 6 THEN '>'
+                    WHEN 7 THEN '<'
+                    WHEN 8 THEN '<='
+                END COLLATE Latin1_General_CI_AS AS comparator
+            FROM ({clinical_events.query}) ce
```
inglesp, line 279: That's neat

tests/integration/backends/test_tpp.py
```diff
@@ -471,6 +472,163 @@ def test_clinical_events(select_all_tpp):
...
+            "numeric_value": 1.5,
+            "comparator": "<=",
+            "lower_bound": 6.0,
+            "upper_bound": 7.0,
+        },
+    ]
+    # cannot rely on input order equalling output order,
+    # dicts are neither hashable nor sortable
+    assert not [r for r in results if r not in expected]
+    assert not [e for e in expected if e not in results]
```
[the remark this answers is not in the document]
Jongmassey: It tests that all the comparator mappings in the CASE statement are correct (assuming they are correctly specified here), based on recent experience that feels like a good idea. Perhaps could be more concise.

The reason for the ordering is that the CTv3 codes come first because that table is first in the UNION. I guess I could just re-order the expected list to match that

PR #1821, opened 2023-12-07
Correct raw `ons_deaths` table names in documentation

StevenMaude:


docs/includes/generated_docs/schemas/beta.raw.emis.md
```diff
@@ -39,7 +39,7 @@ This table contains all registered deaths.
 The `ehrql.tables.beta.ons_deaths` table contains the earliest registered death.
 
 !!! tip
-    To return one row per patient from `ehrql.tables.beta.raw.ons_deaths`,
+    To return one row per patient from `ehrql.tables.beta.raw.core.ons_deaths`,
```
StevenMaude, line 42: comment: Again, this gets inherited from the `core` table.

StevenMaude:
Discussed with @inglesp and we agreed that this could be merged now. Even though the EMIS entry isn't strictly correct, it's currently completely incorrect, and _this PR improves the rest of the entries_.

I've opened an issue to discuss how we best override documentation for tables that are reused from `core`: #1837.

ehrql/tables/raw/core.py
```diff
@@ -37,10 +37,10 @@ class ons_deaths_raw(EventFrame):
 
     In the associated database table a small number of patients have multiple registered deaths.
     This table contains all registered deaths.
-    The `ehrql.tables.ons_deaths` table contains the earliest registered death.
+    The `ehrql.tables.raw.core.ons_deaths` table contains the earliest registered death.
```
iaindillingham, line 40: I think this should be...

```suggestion
    The `ehrql.tables.core.ons_deaths` table contains the earliest registered death.
```

iaindillingham (approved):
See comment, though 🙂

ehrql/tables/raw/core.py
StevenMaude: Yes, you're right, thanks!

PR #1854, opened 2023-12-20
Improve setting up `test_data` for `assure`

### LINKED ISSUE #1846: Stricter checks for specifying test data in assurance tests
Currently we dont check how many rows are added to a test table defined in assurance tests. For example, it is possible to add multiple rows for a patient to the `patients` table (only the last entry will get used), a one-row-per-patient-table. We should implement a check for this.

To implement this, we have to ask whether the `orm_class` object is for a one-row- or many-row-per-patient table, see: https://github.com/opensafely-core/ehrql/blob/f5a91b484768d5e2bf218f4c1c3b8885d44ddbd4/ehrql/assurance.py#L53

The quick-and-dirty way of asking whether the object is for a one-row- or many-row-per-patient table is to check (using `hasattr`) whether the object has an attribute called `row_id`. This gets set in [orm_utils.py](https://github.com/opensafely-core/ehrql/blob/f5a91b484768d5e2bf218f4c1c3b8885d44ddbd4/ehrql/utils/orm_utils.py#L16-L34), and eventually we'll want to add a function in that module that lets us tell apart a one-row- or many-row-per-patient `orm_class` objects – but that can wait until we've got the `assure` functionality working correctly.

milanwiedemann:
Before, data for (1) one-row-per-patient and (2) many-rows-per-patient tables was added in the same way – a list of dictionaries. This was confusing because (1) should only be able to take one dictionary (i.e., one row per patient).

Closes #1846

### DIFF
--- a/docs/how-to/test-dataset-definition.md
+++ b/docs/how-to/test-dataset-definition.md
@@ -6,7 +6,11 @@ The [assure](../../reference/cli/#assure) command works like a suite of unit tes
 You can write assurance tests that help you and others to understand and review the expected behaviour of your dataset definition.
 Assurance tests also provide confidence that existing functionality remains unchanged when reworking your dataset definition.
 
-In this guide we first set up an example dataset definition and then demonstrate the steps needed to test the behaviour of the ehrQL queries in the dataset definition.
+In this guide we will demonstrate how to test ehrQL queries:
+
+1. Create dataset definition
+2. Specify test data and expectations
+3. Run tests
 
 ## Example dataset definition
 
@@ -42,32 +46,27 @@ dataset.has_asthma_med = latest_asthma_med.exists_for_patient()
 dataset.latest_asthma_med_date = latest_asthma_med.date
 ```
 
-## Specifying data and expectations
+## Specifying test data and expectations
 
 Next, you need to provide (1) data for test patients and (2) specify the data that you expect to see in the dataset for each patient after applying your ehrQL queries.
-Test data and expectations are defined in a nested dictionary called `test_data`.
-The key in the outermost dictionary specifies the patient id.
+Test data and expectations are both defined in a nested dictionary called `test_data`.
 
 ### Data for test patients
 
-In the example below we have created one test patient with the patient id `1` and specified more data for this patient in two lists called `patient` and `medications`.
-The names of the inner lists match the names of the tables you are using in your dataset definition.
-To explore how these tables are structured you can look at the column names in the [core schema](../../reference/schemas/core) documentation.
+To set up test patients and their data you need to use the following structure:
 
-The `patients` list contains one dictionary with two keys, one key for each column (`date_of_birth` and `sex`) that we want to populate.
-Note that you don't have to specify a value for each column in the underlying table.
-For example we did not specify `date_of_death` in the dictionary inside the `patients` list so the column will be missing with the value `None`.
-
-The `medications` list contains two dictionaries (one for each row we add to the table) with two keys (one for each column: `date` and `dmd_code`).
-Note that you have to specify a list for each table you use in your dataset definition, but this could also be an empty list.
+* **Patient ID**: Outermost dictionary keys represent patient IDs.
+* **Table names**: Second-level dictionary keys denote table names from OpenSAFELY backends.
+    * *One-row-per-patient* tables (e.g., [patients](../../reference/schemas/core/#patients)) are specified in a single dictionary
+    * *Many-rows-per-patient* tables (e.g., [medications](../../reference/schemas/core/#medications)) are specified in a list that can contain multiple dictionaries, where each dictionary adds one row for the patient to the `medications` table.
+* **Column names**: Third-level dictionary keys indicate column names in the tables.
 
 ```py
 test_data = {
     1: {
-        "patients": [{
+        "patients": {
             "date_of_birth": date(2020, 1, 1),
             "sex": "female"},
-            ],
         "medications": [
             {
                 # First prescription of asthma medication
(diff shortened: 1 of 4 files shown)


ehrql/assurance.py
```diff
@@ -51,7 +51,11 @@ def validate(variable_definitions, test_data):
     input_data = []
     for patient_id, patient in test_data.items():
         for orm_class in orm_classes.values():
-            for record in patient[orm_class.__tablename__]:
+            if hasattr(orm_class, "row_id"):
+                for record in patient[orm_class.__tablename__]:
+                    input_data.append(orm_class(patient_id=patient_id, **record))
+            else:
+                record = patient[orm_class.__tablename__]
                 input_data.append(orm_class(patient_id=patient_id, **record))
```
iaindillingham, line 59: Appending to `input_data` in two locations seems cumbersome. We could also tidy away the loop. How about we arrange our data into a sequence, if it isn't already arranged into a sequence, and then extend `input_data`? Something like:

```suggestion
            if hasattr(orm_class, "row_id"):
                records = patient[orm_class.__tablename__]
            else:
                records = [patient[orm_class.__tablename__]]
            input_data.extend([orm_class(patient_id=patient_id, **r) for r in records])
```

Notice that the listcomp isn't necessary: we could remove the `[` and `]` and this would still work, because what remains would be a genexp. However, if the call to `orm_class` errored, then we would want to know sooner rather than later.

iaindillingham: I'm also not entirely sure about calling `hasattr` to tell us whether `patient[orm_class.__tablename__]` is a sequence of records or a single record. I'd like to know what a more experienced Pythonista thinks; @evansd? @inglesp?

docs/how-to/test-dataset-definition.md
```diff
@@ -86,9 +85,20 @@ test_data = {
 }
 ```
 
+In the example above we have created one test patient with the patient ID `1` and added test data for two tables: `patient` and `medications`.
```
iaindillingham, line 88:
```suggestion
In the example above we have created one test patient with the patient ID `1` and added test data for two tables: `patients` and `medications`.
```

docs/how-to/test-dataset-definition.md
```diff
@@ -86,9 +85,20 @@ test_data = {
...
+The keys of the second-level dictionary match the names of the tables in the dataset definition.
+To explore how these tables are structured you can look at the column names in the [table schemas](../../reference/schemas) documentation.
+As mentioned above, adding data is different for *one-row-* and *many-rows-per-patient* tables:
+
+* `patients` is a *one-row-per-patient* table, so you can only define one dictionary with one key for each column (`date_of_birth` and `sex`) that you want to populate.
+    Note that you don't have to specify a value for each column in the underlying table.
+    For example we did not specify `date_of_death` in the dictionary so the column will be missing with the value `None`.
```
iaindillingham, line 95: Out of interest, what happens when the column has a "Never `NULL`" constraint? (e.g. `open_prompt.ctv3_code`).

milanwiedemann: It *just works* without complaining, e.g., [patients.sex](https://docs.opensafely.org/ehrql/includes/generated_docs/schemas/tpp/#patients.sex) also has the 'Never `NULL`' constraint.

iaindillingham: I'm not sure whether working without complaining is a good idea. We'd like a user's conceptual model of test data to be the same as that of production data.

inglesp: Iain's right.  But fixing that is outside the scope of this ticket.  (New ticket Milan?)

ehrql/assurance.py
inglesp: It's a slightly leaky abstraction, but I'd say it's fine for now.

(Ideally we'd have a function in `orm_utils.py` that told us whether a particular `orm_class` was one- or many-rows-per-patient, but there's already a `has_one_row_per_patient` function in that module (imported from `ehrql.query_model.nodes`) and so it's not clear what we'd call our new function.)

inglesp: The use of `extend` is neat.

docs/how-to/test-dataset-definition.md
milanwiedemann: yes, good next ticket, I also agree! But it also means that a user always needs to add values for all 'Never NULL' columns (even if they are not used by the user).

Further to the user's conceptual model, I think this improvement is also useful because a user also needs to meet all the other constraints that a column might have, and therefore the test data will match the real data :)

milanwiedemann: I'll write the ticket up

ehrql/assurance.py
milanwiedemann: thanks! I added Iain's `extend` suggestion

evansd: > Ideally we'd have a function in `orm_utils.py` that told us whether a particular orm_class was one- or many-rows-per-patient

We already have exactly this function!
https://github.com/opensafely-core/ehrql/blob/5ba76862b05ca4a1dac1a975eea8398d6012ced8/ehrql/utils/orm_utils.py#L104-L107

iaindillingham: Oh, we should use `table_has_one_row_per_patient` rather than `hasattr`, then!

milanwiedemann: thanks, that's exactly what I needed (but didn't see)!

I replaced the `hasattr` with `table_has_one_row_per_patient`!

docs/how-to/test-dataset-definition.md
milanwiedemann: Issue with description of the problem here: https://github.com/opensafely-core/ehrql/issues/1858

iaindillingham (approved):

PR #1857, opened 2023-12-21
feat: Add graph-query command

inglesp:


evansd (approved):
Glorious fun! I've just been playing around with it locally.

PR #1870, opened 2024-01-03
Allow supplying explicit NULLs in `case()`

### LINKED ISSUE #1637: `KeyError` when `.then` is passed `None`
Until we pick up #1548, missing data in `addresses.imd_rounded` are represented as `-1` rather than `None`. When grouping rounded IMD into quintiles, I could handle missing data like this

```py
imd_rounded = addresses.for_patient_on(index_date).imd_rounded
max_imd = 32844

dataset.imd_quintile = case(
    when((imd_rounded >= 0) & (imd_rounded < int(max_imd * 1 / 5))).then(1),
    when(imd_rounded < int(max_imd * 2 / 5)).then(2),
    when(imd_rounded < int(max_imd * 3 / 5)).then(3),
    when(imd_rounded < int(max_imd * 4 / 5)).then(4),
    when(imd_rounded <= max_imd).then(5),
)
```

but I thought that handling missing data like this was easier to reason about

```py
dataset.imd_quintile = case(
    when(imd_rounded < 0).then(None),
    when(imd_rounded < int(max_imd * 1 / 5)).then(1),
    when(imd_rounded < int(max_imd * 2 / 5)).then(2),
    when(imd_rounded < int(max_imd * 3 / 5)).then(3),
    when(imd_rounded < int(max_imd * 4 / 5)).then(4),
    when(imd_rounded <= max_imd).then(5),
)
```

However, the latter raises

```
  File "/app/ehrql/query_language.py", line 933, in _wrap
    cls = REGISTERED_TYPES[type_, is_patient_level]
          ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: (<class 'NoneType'>, True)
```

I think there probably are other times when a researcher wants to pass `None` to `.then`, even after we've picked up #1548.

evansd:
As part of this change we end up making the type system a bit more powerful and can throw away some of the workarounds we previously needed.

Closes #1637

### DIFF
--- a/docs/includes/generated_docs/specs.md
+++ b/docs/includes/generated_docs/specs.md
@@ -2978,6 +2978,37 @@ returns the following patient series:
 
 
 
+#### 10.1.4 Case with explicit null
+
+This example makes use of a patient-level table named `p` containing the following data:
+
+| patient|i1 |
+| - | - |
+| 1|6 |
+| 2|7 |
+| 3|8 |
+| 4|9 |
+| 5| |
+
+```python
+case(
+    when(p.i1 < 8).then(None),
+    when(p.i1 > 8).then(100),
+    otherwise=200,
+)
+```
+returns the following patient series:
+
+| patient | value |
+| - | - |
+| 1| |
+| 2| |
+| 3|200 |
+| 4|100 |
+| 5|200 |
+
+
+
 ### 10.2 Case expressions with single condition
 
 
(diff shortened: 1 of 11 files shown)


iaindillingham (approved):
I still don't think I would have arrived at this solution, even after your comments on #1637. However, I appreciate a set of well described, well sequenced, commits. Thank you!

PR #1873, opened 2024-01-05
Wrap `DomainMismatchError` to be less scary

evansd:
This isn't the easiest error to explain but I've done my best and am very open to bikeshedding on the text here:
```
Error loading file 'example_dataset.py':

  File "/home/dave/projects/ebmdatalab/ehrql/example_dataset.py", line 6, in <module>
    dataset.v = medications.date > clinical_events.date
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ehrql.query_language.InvalidOperationError: 
Cannot combine series which are drawn from different tables and both
have more than one value per patient.

Hint: try reducing one series to have only one value per patient by
using an aggregation like `maximum_for_patient()` or pick a single
row from the table using `first_for_patient()`.
```

I would have been possibly to implement this by just changing the error raised in the query model, but I think it's important that we have a mechanism in place for translating query model errors into ehrQL errors.

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -242,7 +242,7 @@ class BaseSeries:
             # `CombineAsSet` query model object which doesn't have a representation in
             # the query language.
             other_as_set = qm.AggregateByPatient.CombineAsSet(_convert(other))
-            return _wrap(qm.Function.In(_convert(self), other_as_set))
+            return _wrap(qm.Function.In, _convert(self), other_as_set)
         elif isinstance(other, PatientSeries):
             raise TypeError(
                 "Argument must be an EventSeries (i.e. have many values per patient); "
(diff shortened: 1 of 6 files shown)


ehrql/query_language.py
```diff
@@ -1022,6 +1029,23 @@ def _wrap(qm_cls, *args, **kwargs):
...
+        return qm_cls(*args, **kwargs)
+    except qm.DomainMismatchError:
+        raise InvalidOperationError(
+            "\n"
+            "Cannot combine series which are drawn from different tables and both\n"
+            "have more than one value per patient.\n"
+            "\n"
+            "Hint: try reducing one series to have only one value per patient by\n"
+            "using an aggregation like `maximum_for_patient()` or pick a single\n"
+            "row from the table using `first_for_patient()`."
```
inglesp, line 1044: a single row for each patient?

inglesp (approved):
A definite improvement.

ehrql/query_language.py
evansd: Oh yes, good point.

PR #1876, opened 2024-01-08
Ignore "sort by constant" errors from MSSQL

evansd:
This is all extremely tedious. The changes added in #1871, which were designed to actually fix the issue which #1869 failed to fix, also failed to fix the issue.

A proper fix here is achievable but not worth the effort, and therefore we tell the gentests to ignore this issue and accept that there are valid query model constructions we can't execute on MSSQL.

In the worst case, a user tries to do something which is in fact a no-op although they don't realise it and then get a runtime error. This seems acceptable for now.

### DIFF
--- a/docs/includes/generated_docs/specs.md
+++ b/docs/includes/generated_docs/specs.md
@@ -464,9 +464,6 @@ e.sort_by(
     p.i1,
     # Literal constant
     0,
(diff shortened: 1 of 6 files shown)


Jongmassey (approved):
Seems like an acceptable end to this sorry saga!

PR #1877, opened 2024-01-08
Add checks for column constraints in test data for `assure`

### LINKED ISSUE #1858: Check column constraints in `test_data` for `assure`
As @iaindillingham pointed out, we currently don't check the constraints of columns when creating test data for assure, see https://github.com/opensafely-core/ehrql/pull/1854#discussion_r1432873684. For example, we can currently add the following test patient, where none of the values meet the constraints specified in the [patients](https://docs.opensafely.org/ehrql/reference/schemas/core/#patients) table, without a warning or error:

```
    1: {
        "patients": {"date_of_birth": None,
                     "sex": 1,
                     "date_of_death": "not-available",
                     },
        "expected_in_population": False
    },
```

We should check that the data for the test patients meets the constraints. This will (1) ensure users add test patients that match the data in the database and (2) help users to have a consistent conceptual model of the test/production data.

milanwiedemann:
This builds on the existing structure that tests the output of ehrQL queries and now also checks if the provided test data meets the constraints. We return details about test data that does not meet the constraints along with the results of testing the ehrQL query instead of returning an error. This makes it possible for users to specify test data that does not meet the constraints. 

Closes #1858

### DIFF
--- a/ehrql/assurance.py
+++ b/ehrql/assurance.py
@@ -4,50 +4,30 @@ from ehrql.query_model.introspection import get_table_nodes
 from ehrql.utils.orm_utils import orm_classes_from_tables, table_has_one_row_per_patient
 
 
+UNEXPECTED_TEST_VALUE = "unexpected-test-value"
 UNEXPECTED_IN_POPULATION = "unexpected-in-population"
 UNEXPECTED_NOT_IN_POPULATION = "unexpected-not-in-population"
-UNEXPECTED_VALUE = "unexpected-value"
+UNEXPECTED_OUTPUT_VALUE = "unexpected-output-value"
 
 
 def validate(variable_definitions, test_data):
-    """Given some input data, validates that the given variables produce the given
-    expected output.
+    """Validates that the given test data
+    (1) meet the constraints in the tables and
+    (2) produce the given expected output.
 
-    test_data contains two parts: data for each patient to be inserted into the
-    database, and the expected values in the dataset for each patient.
+    Returns two dictionaries mapping IDs of patients with details of
+    (1) test data that did not meet the constraints of the tables and
+    (2) unexpected data in the output dataset.
 
-    For instance:
-
-        test_data = {
-            1: {
-                "patients": {"date_of_birth": date(1999, 12, 31)},
-                "events": [],
-                "expected_in_population": False,
-            },
-            2: {
-                "patients": {"date_of_birth": date(2010, 1, 1)},
-                "events": [{"date": date(2020, 1, 1), "code": "11111111"}],
-                "expected_columns": {
-                    "has_matching_event": True,
-                },
-            },
-        }
-
-    This indicates that two rows should be inserted into the patients table (one for
-    each patient), and one row should be inserted into the events table (for patient 2).
-
-    This also indicates that patient 1 should not be in the dataset, and that patient 2
-    should be, with True in the "has_matching_event" column.
-
-    Returns a dict mapping IDs of patients with unexpected data in the dataset, to
-    details of the unexpected data.
-
-    See tests/unit/test_assurance.py for more examples.
+    For more see docs/how-to/test-dataset-definition.md
     """
 
     # Create objects to insert into database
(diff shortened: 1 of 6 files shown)


ehrql/query_model/table_schema.py
```diff
@@ -22,15 +23,24 @@ def __post_init__(self):
         def description(self):
             return f"Possible values: {', '.join(f'`{v}`' for v in self.values)}"
 
+        def validate(self, value):
+            return self.values == value if value else False
```
inglesp, line 27: This is not quite right.  A categorical column is one where values in the column should come from a given tuple (ie `self.values`).  So we should be validating whether the given value is in `self.values`.

ehrql/query_model/table_schema.py
```diff
@@ -22,15 +23,24 @@ def __post_init__(self):
...
+
     class NotNull(BaseConstraint):
         description = "Never `NULL`"
 
+        def validate(self, value):
+            return value is not None
+
     class Unique(BaseConstraint):
         description = "Always unique"
```
inglesp, line 36: This class could have a `validate` method that returns `True`, with a comment explaining that we can't actually validate a single value.

ehrql/assurance.py
```diff
@@ -69,11 +80,31 @@ def validate(variable_definitions, test_data):
...
+            constraints = table.schema.schema[column].constraints
+            for constraint in constraints:
+                is_valid = constraint.validate(value)
+                if not is_valid:
+                    error_msg = f"Validate test data: Found errors for patient {patient_id}. Column '{column}' did not meet the '{constraint}' constraint, got '{value}'."
+                    errors.append(error_msg)
+    if len(errors) == 0:
+        return errors
+    else:
+        raise DataValidationError(error_msg)
```
inglesp, line 107: As you said in Slack, raising an exception here means that users would have to fix constraint validation errors one by one.

So instead of raising an exception, why not just return `errors`?  You're already (on lines 66-8) collecting all the constraint validation errors.

The question then becomes what to do with the collected errors.  I should think you could separate out the `present` function into two parts: one that presents any constraint errors, and one that presents any test errors.

milanwiedemann: thx! instead of returning an `DataValidationError`, we have decided to return both (1) all data validation errors and (2) all result validation errors in a message.

ehrql/assurance.py
```diff
@@ -69,11 +56,33 @@ def validate(variable_definitions, test_data):
...
+        for column, value in record.items():
+            constraints = table.schema.schema[column].constraints
+            for constraint in constraints:
+                is_valid = constraint.validate(value)
+                if not is_valid:
+                    unexpected_test_values.append(
+                        {
+                            "column": column,
+                            "constraint": f"{constraint}",
+                            "value": f"{value}",
```
milanwiedemann, line 82: I should probably use `str()` here instead of f string here for simplicity?

evansd: Yeah, maybe. I think the f-string version is pretty clear though – I wouldn't have picked it out as a thing to change myself.

ehrql/query_model/table_schema.py
```diff
@@ -41,6 +55,9 @@ def __post_init__(self):
         def description(self):
             return f"Matches regular expression: `{self.regex}`"
 
+        def validate(self, value):
+            return bool(match(self.regex, value)) if value else True
```
evansd, line 59: And same again here otherwise the empty string will always pass the test.

evansd (approved):
This is great, Milan, thanks for doing this.

The only changes that need making are checking for `if value is not None` rather than `if value` – a classic source of bugs in Python.

ehrql/assurance.py
milanwiedemann: thx, I'll keep it as it is then!

milanwiedemann:
ohh, thanks! that's good to know, I changed `if value` to `if value is not None` where necessary and also added another `assert not c.validate(...)` to the tests.

evansd (approved):
:+1: All looks good!

PR #1879, opened 2024-01-10
Improve error messages

evansd:
This is a follow on from #1873 and attempts to address many of the issues noted in #505.

As part of this work I discovered that there's an alternative ISO date format (`YYYYMMDD` without hyphens) and so I've updated the parser to only accept the hyphenated version to avoid any possibility of confusion.

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -1079,9 +1079,26 @@ def _convert(arg):
     # If it's an ehrQL series then get the wrapped query model node
     elif isinstance(arg, BaseSeries):
         return arg._qm_node
-    # Otherwise it's a static value and needs to be put in a query model Value wrapper
-    else:
+    # If it's a static value then we need to be put in a query model Value wrapper
+    elif isinstance(
+        arg, bool | int | float | datetime.date | str | BaseCode | frozenset
+    ):
         return qm.Value(arg)
+    #
+    # Handle various kinds of user error
+    #
+    elif isinstance(arg, BaseFrame):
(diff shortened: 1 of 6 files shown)


tests/unit/test_query_language.py
```diff
@@ -74,6 +74,17 @@ class events(EventFrame):
 events_schema = TableSchema(event_date=Column(date), f=Column(float))
 
 
+def assert_not_chained_exception(excinfo):
+    # Including chained exception details in the traceback is the default Python
+    # behaviour but we often want to hide internal details from the user where these are
+    # not helpful
+    traceback_str = "\n".join(traceback.format_exception(excinfo.value))
```
bloodearnest, line 82: How about `assert excinfo.value.__context__ is None` instead or as well?

bloodearnest (approved):
This makes some sense to be at a high level, although I don't really understand the details of some of the specific errors.

ehrql/query_language.py
```diff
@@ -30,12 +30,17 @@
 REGISTERED_TYPES = {}
 
 
-class InvalidOperationError(Exception):
+class LanguageError(Exception):
```
inglesp, line 33: Could we call this `EhrqlError`?

inglesp (approved):

ehrql/query_language.py
evansd: Yeah, I considered that but got put off by the question of how to capitalise it! I think it's the more natural name though, if we can get over my hang-ups.

PR #1883, opened 2024-01-10
Try to add helpful context to bitwise operator errors

### LINKED ISSUE #1401: Raise a more helpful error message following an operator precedence issue
Because [bitwise OR has higher precedence than `==`][1], this dataset definition fails with a `TypeError`:

```py
from ehrql import Dataset
from ehrql.tables.beta.tpp import patients


dataset = Dataset()
dataset.define_population(patients.sex == "male" | patients.sex == "female")
dataset.sex = patients.sex
```

```
  File "/Users/iaindillingham/Code/opensafely-core/ehrql/dataset_definition.py", line 6, in <module>
    dataset.define_population(patients.sex == "male" | patients.sex == "female")
                                              ~~~~~~~^~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'StrPatientSeries'
```

But this doesn't:

```py
from ehrql import Dataset
from ehrql.tables.beta.tpp import patients


dataset = Dataset()
dataset.define_population((patients.sex == "male") | (patients.sex == "female"))
dataset.sex = patients.sex
```

Whilst we can't do anything about operator precedence, we could raise a more helpful error message.

[1]: https://docs.python.org/3/reference/expressions.html#operator-precedence

evansd:
Because `and`, `or` and `not` are control-flow primitives in Python they are not overridable and so we're forced to use the bitwise operators for logical operations. However these have different precedence rules from those governing the standard operators and so it's easy to accidentally do the wrong thing.

Here we attempt to identify errors resulting from this precedence confusion and add some context to steer the user in the right direction.

Using the example given in the [original issue][1], we now get an error like this:
```
2024-01-10 18:16:08 [info     ] Compiling dataset definition from example_dataset.py
Error loading file 'example_dataset.py':

  File "/home/dave/projects/ebmdatalab/ehrql/example_dataset.py", line 8, in <module>
    dataset.define_population(patients.sex == "male" | patients.sex == "female")
                                              ~~~~~~~^~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'StrPatientSeries'

WARNING: The `|` operator has surprising precedence rules, meaning
you may need to add more parentheses to get the correct behaviour.

For example, instead of writing:

    a == b | x == y

You should write:

    (a == b) | (x == y)

```

The implementation isn't particularly lovely (we have to match the text of the `TypeError` with a regex) but it does the job.

Closes #1401

[1]: https://github.com/opensafely-core/ehrql/issues/1401


ehrql/query_language.py
```diff
@@ -1525,3 +1525,65 @@ def minimum_of(value, other_value, *other_values):
...
+    elif operator == "~":
+        name = "not"
+        example_bad = "~ a == b"
+        example_good = "~ (a == b)"
+    else:
+        assert False
+    return (
+        f"\n"
+        f"WARNING: The `{operator}` operator has different precedence rules from the "
+        f"normal `{name}` operator\n"
```
inglesp, line 1579: I'm on the fence about whether mentioning the rules for eg `or` is helpful.  Maybe:

> WARNING: The `{operator}` operator has surprising precedence rules, meaning...

?

inglesp (approved):

ehrql/query_language.py
evansd: Yeah, genuinely unsure myself. I wonder if I should ask for some feedback from researchers?

evansd: Have asked for feedback. Will probably regret doing so.
https://bennettoxford.slack.com/archives/C02HJTL065A/p1704971443630719

PR #1886, opened 2024-01-12
Add `test-data-file` parameter to `generate-dataset`

### LINKED ISSUE #1851: Add `--test-data` parameter to `generate-dataset`
As an experiment, we should try adding a `--test-data` parameter to the `generate-dataset` command.

To implement:

* Add a new argument to the parser in [`ehrql.__main__.add_generate_dataset`](https://github.com/opensafely-core/ehrql/blob/307f928f8cb5b2338a8f3452febdb5ae8e9cb764/ehrql/__main__.py#L139-L175)
* Add a new parameter to [`ehrql.main.generate_dataset`](https://github.com/opensafely-core/ehrql/blob/307f928f8cb5b2338a8f3452febdb5ae8e9cb764/ehrql/main.py#L41-L52)
* After we've loaded the dataset definition [here](https://github.com/opensafely-core/ehrql/blob/307f928f8cb5b2338a8f3452febdb5ae8e9cb764/ehrql/main.py#L54-L56), check to see whether the user has passed the path to some test data.  If so, then copy the relevant code from [`ehrql.main.assure`](https://github.com/opensafely-core/ehrql/blob/307f928f8cb5b2338a8f3452febdb5ae8e9cb764/ehrql/main.py#L322-L327).

For now, this will print a warning if the tests fail.  We should think about whether we want the command to completely fail when this happens.

milanwiedemann:
This allows users to test their dataset definition every time they run their `generate-dataset` action by providing a test file through the `--test-data-file` flag. It makes it possible for users to integrate tests into their OpenSAFELY pipeline.

Closes #1851

### DIFF
--- a/ehrql/main.py
+++ b/ehrql/main.py
@@ -48,6 +48,7 @@ def generate_dataset(
     query_engine_class,
     dummy_tables_path,
     dummy_data_file,
+    test_data_file,
     environ,
     user_args,
 ):
@@ -55,6 +56,10 @@ def generate_dataset(
     variable_definitions, dummy_data_config = load_dataset_definition(
         definition_file, user_args, environ
     )
+
+    if test_data_file:
+        log.info(f"Testing dataset definition with tests in {str(definition_file)}")
+        assure(test_data_file, environ=environ, user_args=user_args)
 
     if dsn:
         generate_dataset_with_dsn(
--- a/docs/how-to/test-dataset-definition.md
+++ b/docs/how-to/test-dataset-definition.md
@@ -172,12 +172,35 @@ test_data = {
 
 ## Running the tests
 
-Finally, you can run your assurance tests through the terminal using the command below to verify if your expectations were successful or failed.
-The results of your tests will be displayed in your terminal.
+Finally you can run your assurance tests to verify if your expectations were successful or failed.
+
+### Option 1: Running tests through the terminal
+
+To run your tests through the terminal, use the following command:
 
 ```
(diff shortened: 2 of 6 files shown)


ehrql/main.py
```diff
@@ -56,6 +57,10 @@ def generate_dataset(
         definition_file, user_args, environ
     )
 
+    if test_data_file:
+        log.info(f"Testing dataset definition with tests in {str(definition_file)}")
+        assure(test_data_file, environ=environ, user_args=user_args)
```
evansd, line 62: It looks like the `assure` command never actually triggers an error: it just prints some stuff and exists successfully regardless of the whether the tests passed or failed:
https://github.com/opensafely-core/ehrql/blob/e863a3409c6c0473109dbfe3d1567c50eacb9b30/ehrql/main.py#L323-L328

I would have expected the command to fail if the test failed.

---

Update: I've just read this:

> For now, this will print a warning if the tests fail. We should think about whether we want the command to completely fail when this happens.

evansd (approved):
I'm approving this on the basis that it does implement the behaviour described in the ticket.

I'm not sure I'm in favour of things which look like tests but never actually fail and just print warnings though: that feels like quite unexpected behaviour to me.

I'm also on the fence about whether bundling testing and dataset generation up together like this is a good idea. I can see the upside in terms of not needing a separate action and ensuring the tests always get run. But it does muddy the conceptual waters a bit. I'd need to spend a bit more time thinking about the options here.

PR #1889, opened 2024-01-15
Improve error messages from `measures`

evansd:
This refactors some of the validation code in `ehrql.query_language` so that we can re-use it in measures and benefit from the more helpful error messages.

### DIFF
--- a/ehrql/measures/measures.py
+++ b/ehrql/measures/measures.py
@@ -11,6 +11,8 @@ from ehrql.query_language import (
     IntPatientSeries,
     Parameter,
     PatientSeries,
+    validate_patient_series,
+    validate_patient_series_type,
 )
 from ehrql.query_model.nodes import Series
 
@@ -233,15 +235,7 @@ class Measures:
     def _validate_num_denom(self, value, name):
         if value is None:
             return
-        if not isinstance(value, PatientSeries):
-            raise Error(
-                f"`{name}` must be a one row per patient series,"
-                f" got '{type(value)}': {value!r}"
-            )
-        if value._type not in (bool, int):
-            raise Error(
-                f"`{name}` must be a boolean or integer series, got '{value._type}'"
-            )
+        validate_patient_series_type(value, types=(bool, int), context=name)
 
     def _validate_intervals(self, intervals):
         if intervals is None:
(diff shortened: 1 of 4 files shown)


ehrql/measures/measures.py
```diff
@@ -165,19 +164,17 @@ def define_measure(
         # Ensure all required arguments are provided (either directly or by defaults)
         for key, value in kwargs.items():
             if value is None:
-                raise ValidationError(
-                    f"No value supplied for '{key}' and no default defined"
-                )
+                raise Error(f"No value supplied for '{key}' and no default defined")
```
rebkwok, line 167: What's the reason for not raising a ValidationError? Is it just for user-friendliness?

ehrql/query_language.py
```diff
@@ -1531,6 +1498,63 @@ def minimum_of(value, other_value, *other_values):
...
+    except TypeError as e:
+        raise TypeError(f"invalid {name}:\n{e})") from None
+    if not isinstance(arg, BaseSeries):
+        raise TypeError(
+            f"invalid {name}:\n"
+            f"Expecting an ehrQL series, got type '{type(arg).__qualname__}'"
+        )
+
+
+def validate_patient_series(name, arg):
```
rebkwok, line 1526: I found the actual values for `name` passed to this function (and the other similar `validate_...()` functions) slightly confusing (e.g. "f"`group_by` value for '{key}'"). Here I'm expecting it to be the name of a thing, and it's not really, it's more like a description. I'm not sure `description` is a better name for the argument though...perhaps just using the keyword when it's called would be enough?

rebkwok (approved):

ehrql/query_language.py
evansd: You're quite right, thanks. I've changed the name to `context` and passed it as keyword arg instead of the first arg. I think it's much clearer now. Let me know if you agree.

ehrql/measures/measures.py
evansd: Yeah, it was just a case of not needlessly creating new exception classes for user-facing errors.

ehrql/query_language.py
rebkwok: Much better, thanks :)

PR #1897, opened 2024-01-26
fix: Account for multiple dm+d mappings in TPP medication table

inglesp:
Some records in TPP's MedicationDictionary table have MULTIPLE_DMD_MAPPING in their DMD_ID column.  We have accounted for these in our custom medication dictionary, but need to account for them in ehrQL too.

Spotted by @alschaffer.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -519,17 +519,18 @@ class TPPBackend(SQLBackend):
         )
 
         # We construct a medication dictionary table which combines the table provided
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
Looks good, little embarrassed I didn't spot this deficiency when I looked last at this. Important to note that this is not a comprehensive "fix" for the issue of multiply-mapped dm+d - multilex records, which would be getting the "multiple mapping" table from TPP.

evansd (approved):

PR #1923, opened 2024-02-21
Fix categories for waiting list data

evansd:
Some of the columns we have specified as categorical have a small number of unexpected values. We need to treat these as NULL in order to ensure that our results are valid as specified. (In particular, the Arrow writer will choke if we give it an unexpected categorical value.)


iaindillingham (approved):
I recall that Andrea and I agreed to encode only the most numerous values as constraints. I can't recall why I didn't restrict the query tables to these; more than likely an oversight on my part, as values that aren't encoded as constraints are only a problem with Arrow files. Thanks for fixing this!

PR #1930, opened 2024-02-27
Update dependencies

evansd:
This updates the production dependencies (which Dependabot fails to do for us, see #964) and, for good measure, updates the dev dependencies as well.

We also remove a SQLAlchemy workaround which is no longer required and participate in some faff in order to avoid a SQLAlchemy warning triggered by upgrading Trino.


rebkwok (approved):
LGTM; the warning with the updated trino version reminds me that we still have the outstanding question of trino versions on EMIS, but that's for another time :)

PR #1931, opened 2024-02-27
Consolidate code paths for reading/writing files

evansd:
This is mainly an internal refactoring with the aim of having the necessary machinery in place to support having multiple output files, which is needed to enable event-level data extracts. However this does involve some (backwards compatible) user-facing changes:
 * The `CSVQueryEngine` is now the `LocalFileQueryEngine` and can query data in any of the supported formats (i.e. `csv.gz` and `arrow` as well as `csv`).
 * This means that both the sandbox and the `--dummy-tables` argument to `generate-dataset` and `generate-measures` can now accept files in any supported format.
 * It also means that the sandbox and `--dummy-tables` arguments now validate the data they're given against the schema defined for the tables.
 * Validation errors encountered when reading files are now displayed without a stack trace.
 * `create-dummy-tables` can now write table data in any of the supported formats (though it defaults to CSV for backwards compatibility).

On this last point, output formats can be specified using a made-up syntax which looks like:

    path/to/output/directory:arrow

i.e. with what would normally be the extension given with a colon after the directory name. I don't _love_ this kind of syntactic novelty. But I dislike it less than any of the alternatives I could think of.


ehrql/main.py
```diff
@@ -306,7 +306,7 @@ def generate_measures_with_dummy_data(
         reader = read_rows(dummy_data_file, column_specs)
         results = iter(reader)
     elif dummy_tables_path:
-        log.info(f"Reading CSV data from {dummy_tables_path}")
+        log.info(f"Reading data from {dummy_tables_path}")
```
inglesp, line 329: Nit: these two log changes are different, which is surprising.

ehrql/query_engines/csv.py
```diff
@@ -32,9 +31,21 @@ def get_results(self, variable_definitions):
...
 
     def populate_database(self, table_nodes):
-        # Populate the database using CSV files in the supplied directory
-        orm_classes = orm_classes_from_tables(table_nodes)
-        input_data = read_orm_models_from_csv_directory(
-            Path(self.csv_directory), orm_classes.values()
+        # Populate the database using the files in the supplied directory
+        self.database.setup(self.get_input_data(table_nodes))
+
+    def get_input_data(self, table_nodes):
```
inglesp: Does this belong in `orm_utils`?

ehrql/file_formats/__init__.py
```diff
@@ -91,6 +98,22 @@ def get_extension_from_directory(filename):
         return list(matching)[0]
 
 
+def split_directory_and_extension(filename):
+    # This is slightly unpleasant in that we've invented our own syntax for saying "I
+    # want a directory here with files of this type in it" e.g.
+    #
+    #     path/to/my/directory:csv
+    #
+    # But I don't seen an obvious alternative and I hope it's reasonably natural.
```
inglesp, line 107: Looks reasonable.  Did you consider requiring the user to pass in the file extension as a separate flag?

inglesp (approved):

ehrql/main.py
evansd: Ah, you're quite right. However looking at this more closely, I think the problem was that they shouldn't have been identical beforehand because they do different things.

ehrql/query_engines/csv.py
evansd: Possibly, but `orm_utils` is actually on my list to get rid of – or at least, to move into `tests/lib`. The ORM snuck in to production via a set of convenient-at-the-time changes but there's actually no need for it and it overcomplicates things. One thing at a time though!

ehrql/file_formats/__init__.py
evansd: Yeah, I did. It would involve adding an extra flag everywhere we want to specify outputs though, and then threading the value through to where it's needed.

As I'm planning to extend this to `generate-measures` for one-measure-per-file outputs, and to `generate-dataset` for event-level outputs, I felt it would get messy pretty quickly. There's something very nice about maintaining the simplicity of a single string which specifies where you want the output to go and what format you want it in.

PR #1936, opened 2024-02-28
Add docs on defining populations

### LINKED ISSUE #1935: Add "how to include/exclude patients from your population" documentation
We've had a few recent queries from users about how to exclude patients from a dataset.  ([Example](https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1708692226021599).)

We should add a short How To guide about doing this.

evansd:
Closes #1935

### DIFF
--- a/docs/how-to/define-population.md
+++ b/docs/how-to/define-population.md
@@ -1,0 +1,74 @@
+## Basic population definition
+
+You specify the patients you want to include in your dataset using the
+`define_population()` method. For example, to include all patients born
+in 1950 you would write something like this:
+```ehrql
+from ehrql import create_dataset
+from ehrql.tables.core import patients
+
+dataset = create_dataset()
+dataset.define_population(patients.date_of_birth.year == 1950)
+```
+
+
+## Combining multiple inclusion criteria
+
+You can combine multiple inclusion criteria using the [logical operators](../../reference/language/#BoolPatientSeries.and):
+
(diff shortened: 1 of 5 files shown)


docs/how-to/define-population.md
```diff
@@ -0,0 +1,74 @@
...
+and women born in 1960 you could use the `|` operator. This says that
+patients must match **either** the first condition **or** the second:
+```python
+dataset.define_population(
+    ((patients.date_of_birth.year == 1950) & (patients.sex == "male"))
+    | ((patients.date_of_birth.year == 1960) & (patients.sex == "female"))
+)
+```
+
+!!! note "What's with all the parentheses?"
```
rebkwok, line 49: This seems like an important note that maybe belongs somewhere earlier in the introduction to ehrQL (or would be worth duplicating there). Maybe in the "writing a dataset definition" bit of the tutorial?

rebkwok (approved):

docs/how-to/define-population.md
evansd: Yeah, I guess we should mention it wherever else we introduce the logical operators. The error messages here have got a lot better though (see https://github.com/opensafely-core/ehrql/pull/1883) and I know that at least some users have managed to fix their code based purely on the suggestions in the error message. So things aren't quite as bad as they were here.

PR #1937, opened 2024-02-29
Fix date-sensitive test that stated failing today

evansd:
Past-Dave helpfully observed:
> TODO: I dislike using today's date as part of the data generation because it makes the results non-deterministic.

The amusingly manifested itself at midnight last night when the below test started failing:
https://github.com/opensafely-core/ehrql/blob/804c159a99332ef70a89bc82e023bffd6b13e3a4/tests/unit/measures/test_dummy_data.py#L31-L38

This is because the test uses fixed date intervals but the dummy data generator was producing events relative to today's date, meaning that as time went on the proportion of these events falling in the required intervals got smaller. Today it reached zero in one interval and test failed.

We now support supplying the "origin date" as an argument to the dummy data generator.

Amusing side note: this problem manifested itself on 29 February but, almost unbelievably, it is _not_ a leap year bug. Given the particular random seed and exponential distribution we have chosen for dates, the tests fails 1520 days after the measure start date. And it _just so happens_ that 1520 days after 2020-01-01 (a natural choice for start date) is 29 Feb.

### DIFF
--- a/ehrql/dummy_data/generator.py
+++ b/ehrql/dummy_data/generator.py
@@ -33,14 +33,19 @@ class DummyDataGenerator:
         batch_size=5000,
         random_seed="BwRV3spP",
         timeout=60,
+        today=None,
     ):
         self.variable_definitions = variable_definitions
         self.population_size = population_size
         self.batch_size = batch_size
         self.random_seed = random_seed
         self.timeout = timeout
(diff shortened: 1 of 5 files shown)


inglesp (approved):
Objectively hilarious

evansd:
Even more hilarious given the bizarre coincidence I have now noted in the PR description.

PR #1946, opened 2024-03-05
Fix URL for documentation link check

StevenMaude:


evansd (approved):
Just to confirm my understanding here: the link itself points to the correct location, it's just that the format used doesn't play nicely with the link checker. Is that correct?

StevenMaude:
> Just to confirm my understanding here: the link itself points to the correct location, it's just that the format used doesn't play nicely with the link checker. Is that correct?

Yes. The suffix is needed for the link checker to pick up that it's a relative link to a Markdown file.

For what it's worth, the [MkDocs documentation also uses `.md` in its examples](https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages).

PR #1955, opened 2024-03-14
Fix awkward line spacing in ehrQL tutorial

Jongmassey:
This section of the tutorial caused some stumbling during a recent run-through. I'm not sure why the reference to the created asthma codelist variable needed a big code block that broke the flow of this sentence. @evansd  also spotted a similarly disjointed layout for the ethnicity codelist variable which has also been fixed.


evansd:
There's also this on the same page:
image

Is that worth changing at the same time?

Jongmassey:
Yep, good spot

evansd (approved):

PR #1993, opened 2024-04-24
Expose `consultation_id` on TPP `medications` table

### LINKED ISSUE #1983: Expose `consultation_id` column on `clinical_events` and `medications`
This is specifically needed for the Pharmacy First project:
 * https://github.com/ebmdatalab/pharmacy-first-project/issues/12

However my understanding is that this is not currently blocking progress.

Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1713446209919039

evansd:
This requires us to create a TPP specialisation of existing core table. Obviously we can just do this by copy/pasting but I worry that this risks drift in the documentation. (Drift in the table structure itself is less of a worry as the backends will naturally enforce consistency for tables sharing the same name.) So I've made a slight tweak to the `@table` decorator to support table inheritance and added a mechanism to enforce docstring consistency.

Closes #1983

### DIFF
--- a/ehrql/tables/tpp.py
+++ b/ehrql/tables/tpp.py
@@ -6,8 +6,9 @@ OpenSAFELY-TPP backend. For more information about this backend, see
 
 import datetime
 
+import ehrql.tables.core
 from ehrql import case, when
-from ehrql.codes import CTV3Code, DMDCode, ICD10Code, OPCS4Code, SNOMEDCTCode
+from ehrql.codes import CTV3Code, ICD10Code, OPCS4Code, SNOMEDCTCode
 from ehrql.tables import Constraint, EventFrame, PatientFrame, Series, table
 from ehrql.tables.core import patients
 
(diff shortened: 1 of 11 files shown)


ehrql/tables/tpp.py
```diff
@@ -659,6 +672,46 @@ class household_memberships_2020(PatientFrame):
     household_size = Series(int)
 
 
+@table
+class medications(ehrql.tables.core.medications.__class__):
+    """
+    The medication table provides data about prescribed medications in primary care.
```
rebkwok, line 678:
```suggestion
    The medications table provides data about prescribed medications in primary care.
```

rebkwok (approved):
Nice. I had an attempt at doing the table inheritance a while ago but couldn't figure out how to get it to work. There are probably some other tables that should be updated with it too (I think it was the clinical events table that i wanted to do it for before.

PR #1994, opened 2024-04-25
Expose `MedicationStatus` column in `medications` raw tables

### LINKED ISSUE #1991: Expose `MedicationStatus` column in `medications` table
This is needed for the Pharmacy First project. We assume that the `MedicationStatus` has [integers](https://github.com/opensafely-core/ehrql/blob/5ec8f6ab852ff214fe5812f7843553c461888209/tests/lib/tpp_schema.py#L4942) but we have not tested this and we do not know what they mean. We may want to consider adding this to the `raw.core` schema (similar to `ons_deaths`) for data exploration first.

https://github.com/ebmdatalab/pharmacy-first-project/issues/10

Slack thread: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1709544965518179

milanwiedemann:
This adds raw `medications` tables for all backends and exposes the `MedicationStatus` field. I decided not to map the values to the descriptions from the data dictionary (see https://github.com/opensafely-core/ehrql/issues/1991#issuecomment-2076828972) yet as I think we should explore this more thoroughly first, but have added this to the description of the column.

Initially, I think we only need this in TPP and I'm not even sure if EMIS provides identical information. However, because this is a core table I just added this this to EMIS schema and pretend it exises. I assume this is not the best approach but couldn't find a better way to stop the tests failing. Also, I dont know what our current approach is for updating EMIS when we make changes to core tables?  

Closes #1991

### DIFF
--- a/docs/includes/generated_docs/schemas/raw.core.md
+++ b/docs/includes/generated_docs/schemas/raw.core.md
@@ -12,9 +12,114 @@ and data curation purposes.
 
 ``` {.python .copy title='To use this schema in an ehrQL file:'}
 from ehrql.tables.raw.core import (
+    medications,
     ons_deaths,
 )
 ```
+
+<p class="dimension-indicator"><code>many rows per patient</code></p>
+## medications
+
+The raw medication table provides data about prescribed medications in primary care together with the medication status.
+
+Prescribing data, including the contents of the medications table are standardised
+across clinical information systems such as SystmOne (TPP). This is a requirement
+for data transfer through the
+[Electronic Prescription Service](https://digital.nhs.uk/services/electronic-prescription-service/)
+in which data passes from the prescriber to the pharmacy for dispensing.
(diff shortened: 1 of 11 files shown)


evansd:
Thanks Milan. We're only just starting to establish the process for adding additional provider-specific fields to core tables. In fact, I've just done the first such PR here:
 * https://github.com/opensafely-core/ehrql/pull/1993

But I think this case is slightly different because there isn't currently a `medications` table in `raw.core` that we need to extend. So I think we should just be able to add it to the `raw.tpp` schema. Or am I missing something?

milanwiedemann:
I think my problem was that I started by changing
`ehrql/tables/raw/core.py` and subsequently the integration tests wouldnt pass without adding the same to emis

milanwiedemann:
closing as this is now covered in #2002

PR #2002, opened 2024-05-01
Expose `medication_status` on `raw.tpp.medications`

### LINKED ISSUE #1991: Expose `MedicationStatus` column in `medications` table

evansd:
Closes #1991

### DIFF
--- a/docs/includes/generated_docs/schemas/raw.tpp.md
+++ b/docs/includes/generated_docs/schemas/raw.tpp.md
@@ -16,6 +16,7 @@ from ehrql.tables.raw.tpp import (
     apcs_historical,
     covid_therapeutics_raw,
     isaric,
+    medications,
     ons_deaths,
     wl_clockstops,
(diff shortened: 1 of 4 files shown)


milanwiedemann (approved):
Thanks, this makes the medications query much clearer to me and I understand how you added a provider-specific field to a core table. Looks good to me!

PR #2008, opened 2024-05-10
Use batched inserts for `InsertMany`

### LINKED ISSUE #1934: Optimise `InsertMany` performance
The `InsertMany` class describes itself thus:
https://github.com/opensafely-core/ehrql/blob/03899032749c01e91e2852d59f82c63ab0f746d0/ehrql/utils/sqlalchemy_query_utils.py#L203-L210

But in fact the implementation is far from optimal:
https://github.com/opensafely-core/ehrql/blob/03899032749c01e91e2852d59f82c63ab0f746d0/ehrql/utils/sqlalchemy_query_utils.py#L222-L232

For relatively small inserts like codelists this isn't particularly significant, but for anything using the `@table_from_file` feature where row counts can get very large this becomes much more problematic.

Observing the logs of Cohort Extractor jobs which use the equivalent feature shows that they can spend significant amounts of time (many hours) on INSERTs. We should fix this in ehrQL before it becomes a problem.

There are a couple of relevant links in the comments:
https://docs.sqlalchemy.org/en/20/core/connections.html#engine-insertmanyvalues
https://docs.sqlalchemy.org/en/20/tutorial/dbapi_transactions.html#tutorial-multiple-parameters

Also note that `self.rows` here is an iterator, but not necessarily a list and therefore any implementation should avoid loading the whole thing into memory (though it may need to load batches into memory).

evansd:
The `InsertMany` construct is most frequently used for codelists, which are small enough that it's not worth optimising. However it is also used for the `@table_from_file` feature where the row counts are much higher – and I have seen such inserts take 2 hours 40 minutes.

Using SQLAlchemy's batched inserts is about 10x faster on MSSQL, and about another order of magnitude faster again on Trino.

Closes #1934

### DIFF
--- a/ehrql/utils/sqlalchemy_query_utils.py
+++ b/ehrql/utils/sqlalchemy_query_utils.py
@@ -1,5 +1,6 @@
 import collections
 import graphlib
+from itertools import islice
 
 import sqlalchemy
 from sqlalchemy.ext.compiler import compiles
@@ -209,9 +210,14 @@ class InsertMany:
     Acts enough like a SQLAlchemy ClauseElement for our purposes.
     """
 
-    def __init__(self, table, rows):
+    # The batch_size here doesn't make a huge difference as SQLAlchemy does its own
+    # internal batching optimisied for the specific database dialect. It just needs to
+    # be big enough that it gives SQLAlchemy's batching enough to work with, but not so
+    # big that we need to worry about memory consumption.
+    def __init__(self, table, rows, batch_size=10000):
         self.table = table
         self.rows = rows
+        self.batch_size = batch_size
 
     def get_children(self):
         return [self.table]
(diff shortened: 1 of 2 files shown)


Jongmassey (approved):
Looks great.

re:

> The batch_size here doesn't make a huge difference ...

Given on l233 a batch_size's worth of rows are loaded into memory (AFAICT) - could 10k cause problems with some _very_ wide tables-from-files? 

> ... does its own internal batching optimisied for the specific database dialect

The SQLAlchemy docs seem quite vague about actual numbers here, so I guess a best guess is all we've got?

evansd:
> Given on l233 a batch_size's worth of rows are loaded into memory (AFAICT) - could 10k cause problems with some very wide tables-from-files?

Yeah, we do load a full batch into memory (SQLAlchemy refuses to work with it otherwise), but to use even 1GB of RAM we'd be talking 100KB per row, which seems _very_ wide and likely to cause problems in other ways.

> The SQLAlchemy docs seem quite vague about actual numbers here, so I guess a best guess is all we've got?

I think it's vague because it's configured per dialect. For MSSSL I think it does 999 which seems to match some limit somewhere.

Issue #2023, opened 2024-05-15
Expose a "clean" therapeutics table


rebkwok:
See [slack thread](https://bennettoxford.slack.com/archives/C31D62X5X/p1715159270056439)

To make the covid therapeutics data consistent with the data cohort-extractor provided and easier for users to use.

Current `covid_therapeutics_raw` table:
- remove the onset of symptoms columns
- restrict the dataset?

Add a new non-raw table that:
- only exposes required fields
- removes fully duplicate rows
- removes "filler" words from the 3 risk group fields and joins as single risk group field (a comma-separated string)
- casts datetimes to date (already done in the raw table)
(cohort extractor applied collation to the intervention and currentstatus columns, but according to the [database report](https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#Therapeutics) they already have the applied collation, so that should be unnecessary.

Refer to cohort-extractor's implementation:
[create_therapeutics_table](https://github.com/opensafely-core/cohort-extractor/blob/main/cohortextractor/tpp_backend.py#L3024) does the removal of duplicated and the comma-separated risk groups (as separate columns). Joining the 3 groups is done [here](https://github.com/opensafely-core/cohort-extractor/blob/main/cohortextractor/tpp_backend.py#L3166).
(Note that we don't need to worry about duplicate risk groups across the 3 risk group columns because only one of those contains data in any one row)

rebkwok:
To add the new TPP table in ehrQL:

- Add to the Backend as a new QueryTable (here's where the [raw one](https://github.com/opensafely-core/ehrql/blob/main/ehrql/backends/tpp.py#L299) is defined)
- Add a table (EventFrame, as the therapeutics table contains multiple rows per patient) with docstring, to [tpp/tables.py](https://github.com/opensafely-core/ehrql/blob/main/ehrql/tables/tpp.py)
- Add a backend test similar to the [one for the raw table](https://github.com/opensafely-core/ehrql/blob/main/ehrql/tables/tpp.py) (with relevant input data to check for the collated strings etc)
-

madwort:
I think you've got a typo for the link to "backend test similar to the one for the raw table" - just checking you meant https://github.com/opensafely-core/ehrql/blob/b2c675017b046e70e136015eba6d168419f0d72b/tests/integration/backends/test_tpp.py#L666-L711

rebkwok:
Yes, that's the one

rebkwok:
[...]

PR #2025, opened 2024-05-17
Add a cleaner therapeutics table

madwort:
* risk cohort values from different sources are aggregated
* fixes https://github.com/opensafely-core/ehrql/issues/2023


ehrql/tables/raw/tpp.py
```diff
@@ -659,14 +659,6 @@ class covid_therapeutics_raw(EventFrame):
             date or a past date at the time of form submission.
         """,
     )
-    CASIM05_date_of_symptom_onset = Series(
```
rebkwok, line 662: There's also the place where these are defined in the QueryTable SQL (backends/tpp.py); if we're not including them in the queryable tables, we don't need to extract them from the db either.

ehrql/backends/tpp.py
```diff
@@ -296,6 +296,32 @@ def modify_query_variables(self, variables):
...
+            Patient_ID AS patient_id,
+            COVID_indication AS covid_indication,
+            Count AS count,
+            CurrentStatus AS current_status,
+            Diagnosis AS diagnosis,
+            FormName AS form_name,
+            Intervention AS intervention,
+            CASIM05_date_of_symptom_onset,
+            MOL1_onset_of_symptoms,
+            SOT02_onset_of_symptoms,
```
rebkwok, line 311: Hmm...I think we do want to do the select distinct over all the fields so we're only discarding completely duplicate rows, but we don't need to include all these other fields that we don't use in the queryable table. Or does it make no different, efficiency-wise?

We could check the duplicates in the actual data and see if there's a smaller subset of fields that we need in order to remove duplicates (there were only ~35; possibly just the fields we do need to expose are enough)

ehrql/backends/tpp.py
```diff
@@ -296,6 +296,32 @@ def modify_query_variables(self, variables):
...
+            STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(CASIM05_risk_cohort, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, '') as CASIM05_risk_cohort,
+            STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(MOL1_high_risk_cohort, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, '') as MOL1_high_risk_cohort,
+            STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(SOT02_risk_cohorts, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, '') as SOT02_risk_cohorts,
+            STUFF(coalesce(',' + NULLIF(MOL1_high_risk_cohort, ''), '') + coalesce(',' + NULLIF(SOT02_risk_cohorts, ''), '') + coalesce(',' + NULLIF(CASIM05_risk_cohort, ''), ''), 1, 1, '') as risk_cohort,
```
rebkwok, line 315: Does this also handle the filler word replacing?

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,157 @@ class clinical_events_ranges(EventFrame):
...
+    assume that they usually do). They may have another form completed for another
+    treatment, either because it was decided to give them a different treatment, or for
+    some other reason. They may in theory also have another form completed some months
+    later for another instance of infection.
+
+    Treatment dates may be in the past or future at the point when the form is
+    submitted.
+
+    Note that this dataset contains some **duplicate** rows – some full duplicates and
+    some partial duplicates.
```
rebkwok, line 590: It shouldn't contain any fully duplicate rows; it will contain more than 1 row per patient, but mostly those should be valid - different visits/treatments. It probably does contain some partial duplicates.

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,157 @@ class clinical_events_ranges(EventFrame):
...
+
+
+    **More Information**
+
+    * [Treatment guidelines](https://www.nice.org.uk/guidance/ta878)
+    * [Draft Data Report](https://docs.google.com/document/d/15o4x9sqHEO-sLm2dTqgm3PyAh72cdgOOmZC4AB3BTNk/) (currently only available to internal staff)
+    """
+
+    age_at_received_date = Series(
```
rebkwok, line 599: We don't want to expose all of these fields - will need to check with Millie which fields should be exposed. From cohort-extractor, I think it would be:
- covid_indication
- intervention
- current_status
- received
- risk_cohort
- treatment_start_date
- region

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,157 @@ class clinical_events_ranges(EventFrame):
...
+    )
+    region = Series(
+        str,
+        description="""
+            NHS England region in which the CMDU submitting the form is located.
+        """,
+    )
+    risk_cohort = Series(
+        str,
+        description="""
```
rebkwok, line 672: I think this description needs to refer to the fact that it's a combined field of comma-separated values and comes from one of the 3 intervention risk groups (also possibly the fact that the data doesn't include risk groups for Paxlovid and Remdesivir - need to check with Millie). Also we should double-check the list of diseases (might be able to just do a release from Airlock for the therapeutics-risk-conditions workspace).

tests/integration/backends/test_tpp.py
```diff
@@ -663,6 +663,134 @@ def test_clinical_events_ranges(select_all_tpp):
...
+        Therapeutics(
+            Patient_ID=1,
+            COVID_indication="a",
+            Count=3,
+            CurrentStatus="b",
+            Diagnosis="c",
+            FormName="d",
+            Intervention="e",
+            CASIM05_date_of_symptom_onset="f",
+            CASIM05_risk_cohort="g",
```
rebkwok, line 698: Can you make the risk cohort values ones that include the text that we expect to the replaced and joined?
e.g.
"Patients with a haematological diseases and liver disease" should become "haematological diseases, liver disease"
"Solid cancer" --> "solid cancer"
"Patients with immune deficiencies and HIV or AIDS and solid organ recipients" --> "immune deficiencies, HIV or AIDS, solid organ recipients"

ehrql/tables/raw/tpp.py
madwort: oh, yeah, true, removed.

ehrql/tables/tpp.py
madwort: sorry, yes, updated to this, wdyt?
```
    Note that this dataset may contain **duplicate** rows – full duplicates are removed
    but there may remain some partial duplicates.
```

ehrql/tables/tpp.py
madwort: ok, if you could check with Millie & let me know for definite I will remove the unnecessary fields.

ehrql/backends/tpp.py
madwort: do you mean doing something like "select fields_we_want from (select distinct * from Therapeutics)"? I wouldn't guess that this would be very much more efficient than just "(select * from Therapeutics)" as the db server probably has to load & check the same amount of data, although you can never know with databases unless you actually try it for real.

Alternatively, it would seem weird to apply distinct to more columns that we're returning in a select, as you could still get duplicates in your output (if they differ only by columns that are not returned by the select).

If it's possible to get more knowledge of the source data, though, and we don't need these columns to eliminate the duplicates in this particular dataset then let me know & I'll remove them.

ehrql/tables/tpp.py
madwort: oh, right, the ticket said 

> Note that we don't need to worry about duplicate risk groups across the 3 risk group columns because only one of those contains data in any one row

so my expectation was that this would never return comma-separated values to the researcher (all commas would be introduced & also removed during the cleaning) - but I haven't verified this for real, maybe we should warn researchers here?

madwort: oh, I see, it's only of these named fields, but the field can contain multiple groups. So the previous docs were inaccurate?

tests/integration/backends/test_tpp.py
madwort: have added these to tests - see other comment.

ehrql/backends/tpp.py
madwort: good spot - it actually didn't do what I thought it was doing - the final aggregation column was using the raw columns not the cleaned columns (they were selected as the same name, which misled me). Now fixed.

madwort: (and tested!)

tests/integration/backends/test_tpp.py
```diff
@@ -663,6 +663,134 @@ def test_clinical_events_ranges(select_all_tpp):
...
+        },
+    ]
+
+
+@register_test_for(tpp.covid_therapeutics)
+def test_covid_therapeutics_risk_cohort_aggregation(select_all_tpp):
+    results = select_all_tpp(
+        Therapeutics(
+            Patient_ID=1,
+            CASIM05_risk_cohort="Patients with a haematological diseases and liver disease",
```
madwort, line 734: test risk

tests/integration/backends/test_tpp.py
```diff
@@ -663,6 +663,134 @@ def test_clinical_events_ranges(select_all_tpp):
...
+            "covid_indication": None,
+            "current_status": None,
+            "diagnosis": None,
+            "form_name": None,
+            "intervention": None,
+            "load_date": None,
+            "patient_id": 1,
+            "received": None,
+            "region": None,
+            "risk_cohort": "haematological diseases,liver disease",
```
madwort, line 758: test risk shortened correctly here

ehrql/backends/tpp.py
rebkwok: Yes, I think that's what I meant. Also, I think in the cohort-extractor version we were trying to be careful to only deduplicate entirely duplicate rows. Maybe it is OK to deduplicate rows that are exact duplicates in the columns of interest; if received date, treatment date, intervention and risk group are all identical, it's probably safe to assume the row is a duplicate. It would inflate `.count_for_patient()` results. Probably also a question for Millie though.

madwort: Thinking a bit more - as we're planning to have the intervention column in the output, we should only need the aggregated risk_cohort column to deduplicate (i.e. remove CASIM05_date_of_symptom_onset, MOL1_onset_of_symptoms, SOT02_onset_of_symptoms)?

ehrql/tables/tpp.py
rebkwok: So a single risk group column can have multiple values in it - they're joined with "ands" in the raw values, like "Liver disease and cancer". The cleaning on the 3 separate columns is removing those and joining with commas (with all the REPLACE and COALESCE and STUFFing), so the value in that column becomes "Liver disease, cancer". In cohort-extractor it does that part when it creates the temporary table, and then it selects from that and joins across the 3 columns, but it'll still need to clean up the values (or make the combined column in a subsequent query). 
(I think the tests might make this clearer)

ehrql/backends/tpp.py
madwort: i.e. if we had a patient that had two interventions on the same day - so received date, treatment date, and risk group are the same but intervention differs & the *_risk_cohorts differ but follow the value of intervention.

madwort: removed

ehrql/backends/tpp.py
```diff
@@ -296,6 +296,32 @@ def modify_query_variables(self, variables):
...
+            Diagnosis AS diagnosis,
+            FormName AS form_name,
+            Intervention AS intervention,
+            CASIM05_date_of_symptom_onset,
+            MOL1_onset_of_symptoms,
+            SOT02_onset_of_symptoms,
+            CASIM05_risk_cohort,
+            MOL1_high_risk_cohort,
+            SOT02_risk_cohorts,
+            STUFF(coalesce(',' + NULLIF(STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(CASIM05_risk_cohort, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, ''), ''), '') + coalesce(',' + NULLIF(STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(MOL1_high_risk_cohort, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, ''), ''), '') + coalesce(',' + NULLIF(STUFF(coalesce(',' + NULLIF(REPLACE(REPLACE(REPLACE(SOT02_risk_cohorts, 'Patients with a ', ''),  'Patients with ', ''), ' and ', ','), ''), ''), 1, 1, ''), ''), ''), 1, 1, '') as risk_cohort,
```
rebkwok, line 312: urgh this makes my head hurt, and I wrote the previous incantation of it in cohort-extractor
Do you think it could be any clearer by splitting the repeated bit out into a helper function and using it as an f-string value? Or putting some comments in about what it's doing?

madwort: https://github.com/opensafely-core/ehrql/pull/2025/commits/412b2121df1268cb9d7dce0e1b4226e0a1df1411 better?

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,160 @@ class clinical_events_ranges(EventFrame):
...
+    intervention = Series(
+        str,
+        description="""
+            Intervention or therapeutic name. Expected to be one of:
+
+             * Paxlovid
+             * Sotrovimab
+             * Molnupiravir
+             * Remdesivir
+             * Casirivimab and imdevimab
```
rebkwok, line 653: There is also now Sarilumab and Baricitinib
This should probably be a Categorial Constraint

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,160 @@ class clinical_events_ranges(EventFrame):
...
+            e.g. `liver disease,rare neurological conditions`.
+
+            This is an aggregation of `CASIM05_risk_cohort`, `MOL1_high_risk_cohort`
+            and `SOT02_risk_cohorts`.
+
+            The available groups as at the time of writing are listed below. However
+            note that the precise wording used has changed over time and so filtering by
+            a specific disease name may not be reliable.
+
+             * `Downs syndrome`
```
rebkwok, line 684: Check this list against the latest one (see therapeutic-risk-groups workspace in Airlock)

evansd (approved):

ehrql/tables/tpp.py
rebkwok: Confirmed by Millie; just expose these fields :point_up:

ehrql/tables/tpp.py
```diff
@@ -552,6 +553,160 @@ class clinical_events_ranges(EventFrame):
...
+    assume that they usually do). They may have another form completed for another
+    treatment, either because it was decided to give them a different treatment, or for
+    some other reason. They may in theory also have another form completed some months
+    later for another instance of infection.
+
+    Treatment dates may be in the past or future at the point when the form is
+    submitted.
+
+    Note that this dataset may contain **duplicate** rows – full duplicates are removed
+    but there may remain some partial duplicates.
```
rebkwok, line 590: We need to document that the risk group only applies to Sotroviman, Molnupiravir, and Casirivimab & imdevimab interventions (see comments on issue #2023)
It's probably a good idea to get Millie to review the docs for this table?

rebkwok (approved):
Couple more comments about docs mostly, otherwise lgtm

ehrql/tables/tpp.py
madwort: I have `risk_cohort` in place of `risk_group`, but otherwise have limited to this list of fields. I've kept all the fields in the raw table because of `SELECT DISTINCT` - they /probably/ could be removed, but I think the safest thing is to keep them?

ehrql/tables/tpp.py
madwort: I mentioned this on the [series description](https://github.com/opensafely-core/ehrql/pull/2025/files#diff-a3e087910f839341b26b122527c775d2bf29f55802a35228b3307436da676065R651-R652) although I'm going to revise that wording to mention intervention names rather than db column names. But do you think it should be in the table description here as well, or is just in the series description sufficient?

ehrql/tables/tpp.py
madwort: Just checked this list against those lists in Airlock & they match (with the caveat that as discussed elsewhere there are ` and `-separated combinations which we convert to comma-separated combinations).

ehrql/tables/tpp.py
madwort: is `Sarilumab and Baricitinib` one entry on the list or two...?

ehrql/backends/tpp.py
```diff
@@ -296,6 +296,49 @@ def modify_query_variables(self, variables):
         """
     )
 
+    def _covid_therapeutics_risk_cohort():
```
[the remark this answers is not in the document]
madwort: thanks, yeah that's nice, have used that

ehrql/tables/tpp.py
rebkwok: 2; but there's another one that wasn't on the original list too (I've just checked). Full list is:
- Sotrovimab
- Molnupiravir 
- sarilumab (Note: all lower case in db, others are title case. I don't think it matters for querying because the collation is case insensitive)
- Baricitinib
- Paxlovid
- Remdesivir
- Tocilizumab
- Casirivimab and imdevimab

rebkwok: The list for the raw table also needs updating

ehrql/tables/tpp.py
rebkwok: I think the series description is probably fine :+1:

ehrql/tables/tpp.py
madwort: Thanks, will update. Out of curiosity, where did you check that?

madwort: ok, alphabetised & updated in both places.

PR #2047, opened 2024-06-18
Include archived HES tables in relevant queries

### LINKED ISSUE #2038: Create "archive" versions of secondary care data tables (APCS, EC, OPA)
The secondary care data tables (APCS, EC, OPA and their various associated tables) are getting too large to be transferred in full each time.

The current proposed solution is to split these into a "live" dataset covering 2022 onwards and an "archive" dataset covering dates prior to 2022. The live dataset would be regularly updated and the archived one would stay static once created.

Obviously the same problem is going to reoccur a few years down the line at which point we can either agree to shift the live/archive cutoff point or possibly we'll have some totally different solution.

The specific tables in question are:
```
APCS
APCS_Cost
APCS_Der

EC
EC_Comorbidities
EC_AlcoholDrugInvolvement
EC_Cost
EC_Diagnosis
EC_Investigation
EC_PatientMentalHealth
EC_Treatment

OPA
OPA_Cost
OPA_Diag
OPA_Proc
```

For each of these, there will be a table with an identical schema and the suffix `_ARCHIVED` containing data prior to 2022.

I'd suggest that, for now, we take the same approach and create mirror copies of our current [ehrQL tables](https://docs.opensafely.org/ehrql/reference/schemas/tpp/) with an `_archive` suffix.

The docstring for each table should include a clear callout (e.g. a warning box) explaining the date limitations for that table.

### Implementation notes

For the test tables (in [`tests/lib/tpp_schema.py`](https://github.com/opensafely-core/ehrql/blob/main/tests/lib/tpp_schema.py)) there's a slight circular dependency here in that we'll want to have this code working before TPP include the new tables in the build where they'll get automatically reflected in the test schema. I would suggest manually editing the test schema to create the new tables by copy-paste-renaming the existing ones. The scheduled action will create a PR to try to revert these changes, but this PR won't pass the new tests and so won't be mergable.

Once the new tables are in place we can re-run and re-publish the database schema report and the scheduled action will pick this up and fix the PR.

For the ehrQL tables, it may be possible to use inheritance to reduce the duplication here by creating base tables without the `@table` decorator and then two subclasses with the decorator e.g.
```py
class APCSBase(EventFrame):
    """
    Documentation goes here
    """
    ...
    admission_date = Series(date)


@table 
class apcs(APCSBase):
    """
    Documentation goes here

    Note: only contains data from 2022 onwards
    """


@table
class apcs_archive(APCSBase):
    """
    Documentation goes here

    Note: only contains data prior to 2022
    """
```

For the backend table definitions, it may be possible to use the `@QueryTable.from_function`  decorator to reduce duplication in the SQL definitions. For inspiration see:
https://github.com/opensafely-core/ehrql/blob/90f7eeafe9d24ff0de223153ff4b9005043867e6/ehrql/backends/tpp.py#L582-L607

### Relevant Slack threads
https://bennettoxford.slack.com/archives/C010SJ89SA3/p1716280499463089
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1716208559335039

evansd:
As described in the issue below, this updates all queries involving the HES/SUS tables to combine data from both the regularly updated tables and the soon-to-be-populated archive tables.

There are a few subtleties here:

 * The current and archived tables containing overlapping data so, to avoid duplicate rows, we pick a cutoff date and filter the data on this before combining. The stated intention is that the regularly updated tables contain data from 2022 onwards, so we use `2022-01-01` as the cutoff. Note that we will need to [double-check this](https://github.com/opensafely-core/ehrql/issues/2048) once the real data starts flowing.

 * Filtering by date involves picking a date column. The most natural column for each dataset seemed to be:
   - `APCS.Admission_Date`
   - `OPA.Appointment_Date`
   - `EC.Arrival_Date`
   
 * There are a small number of NULL values in these columns (approx 0.04% of APCS, 0.001% of OPA, and none in EC). As most research work using these tables is going to want to use these dates these rows are effectively not present in any case. But I don't want this change to suddenly make them completely invisible so I've opted to include NULL-dated rows from the current dataset but not from the archive (we can't do both or we risk duplicates).

Closes #2038

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -423,21 +423,25 @@ class TPPBackend(SQLBackend):
             "Arrival_Date",
         )
 
-    ec_cost = QueryTable(
-        """
-        SELECT
-            cost.Patient_ID AS patient_id,
-            cost.EC_Ident AS ec_ident,
-            cost.Grand_Total_Payment_MFF AS grand_total_payment_mff,
-            cost.Tariff_Total_Payment AS tariff_total_payment,
-            ec.Arrival_Date AS arrival_date,
-            ec.EC_Decision_To_Admit_Date AS ec_decision_to_admit_date,
-            ec.EC_Injury_Date AS ec_injury_date
-        FROM EC_Cost AS cost
-        LEFT JOIN EC AS ec
-        ON cost.EC_Ident = ec.EC_Ident
-    """
-    )
+    @QueryTable.from_function
+    def ec_cost(self):
+        return self._union_over_hes_archive(
+            """
+            SELECT
+                cost.Patient_ID AS patient_id,
+                cost.EC_Ident AS ec_ident,
+                cost.Grand_Total_Payment_MFF AS grand_total_payment_mff,
+                cost.Tariff_Total_Payment AS tariff_total_payment,
+                ec.Arrival_Date AS arrival_date,
+                ec.EC_Decision_To_Admit_Date AS ec_decision_to_admit_date,
+                ec.EC_Injury_Date AS ec_injury_date
+            FROM EC_Cost{table_suffix} AS cost
+            LEFT JOIN EC{table_suffix} AS ec
+            ON cost.EC_Ident = ec.EC_Ident
+            WHERE {date_condition}
+            """,
+            "ec.Arrival_Date",
+        )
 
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Tedious, thanks

PR #2060, opened 2024-07-08
Update T1OO logic

evansd:
We expand the logic to exclude de-registered patients on the basis that they may have registered a T1OO elsewhere which has not propagated to our database. Patients who died at the time of deregistration (or shortly after, to account for data flow issues) are **not** excluded on the basis that their deregistration is due to their death and not to their registering elsewhere.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -110,7 +110,7 @@ class TPPBackend(SQLBackend):
         if self.include_t1oo:
             return variables
         # Otherwise we add an extra condition to the population definition to include
-        # only patients who we know have not opted out
+        # only patients who we know have not opted out.
         patients = ehrql.tables.tpp.patients
         regs = ehrql.tables.tpp.practice_registrations
         is_registered = regs.where(regs.end_date.is_null()).exists_for_patient()
@@ -120,13 +120,16 @@ class TPPBackend(SQLBackend):
             ~t1oo.exists_for_patient()
             & (
                 # Of those, include only those who have a currently active practice
-                # registration (otherwise they may have recorded a T1OO elsewhere) ...
+                # registration. Patients who are not currently registerd with TPP may
+                # have gone on to record a t1oo at another, non-TPP practice, and we
+                # would not know about it...
                 is_registered
-                # ... or those who were deregistered at time of death therefore cannot
-                # have gone on to register a T1OO elsewhere (due to imperfections in
-                # data flows we consider deaths up to 28 days after to count as the
-                # "same time" for this purpose)
-                | ((patients.date_of_death - deregistration_date).days <= 28)
+                # ... or those who were deregistered at or after time of death therefore
+                # cannot have gone on to register a T1OO elsewhere. This is the most
+                # conservative case; due to imperfections in data flows, deaths may be
+                # registered a short time *after* deregistration date, but we err on the
+                # safe side and do not include any window after deregistration.
+                | (patients.date_of_death <= deregistration_date)
             )
         )
         return {
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -109,9 +109,26 @@ class TPPBackend(SQLBackend):
         # return it unmodified
         if self.include_t1oo:
             return variables
-        # Otherwise we add an extra condition to the population definition which is that
-        # the patient does *not* appear in the T1OO table.
-        extra_condition = ~t1oo.exists_for_patient()
+        # Otherwise we add an extra condition to the population definition to include
+        # only patients who we know have not opted out
+        patients = ehrql.tables.tpp.patients
+        regs = ehrql.tables.tpp.practice_registrations
+        is_registered = regs.where(regs.end_date.is_null()).exists_for_patient()
+        deregistration_date = regs.end_date.maximum_for_patient()
+        extra_condition = (
+            # Include only patients who do *not* have a T1OO recorded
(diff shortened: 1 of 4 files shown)


ehrql/backends/tpp.py
```diff
@@ -109,9 +109,24 @@ def modify_query_variables(self, variables):
...
+            ~t1oo.exists_for_patient()
+            & (
+                # Of those, include only those who have a currently active practice
+                # registration (otherwise they may have recorded a T1OO elsewhere) ...
+                registrations.where(
+                    registrations.end_date.is_null()
+                ).exists_for_patient()
+                # ... or those who have died and therefore cannot have registered a T1OO
+                # elsewhere
+                | patients.date_of_death.is_not_null()
```
rebkwok, line 127: Just checking - if a patient has a date of death recorded in TPP, then they must have died while registered at a TPP practice. Do we need to worry about possible data entry errors where deregistration date is before the recorded date of death?

tests/integration/backends/test_tpp.py
```diff
@@ -2825,29 +2825,78 @@ def test_is_in_queries_on_columns_with_nonstandard_collation(
...
+        # Excluded: death recorded, registered, but had opt out in place
+        Patient(
+            Patient_ID=5, DateOfBirth=date(1955, 1, 1), DateOfDeath=date(2024, 7, 1)
+        ),
+        RegistrationHistory(
+            Patient_ID=5, StartDate=date(2020, 1, 1), EndDate=date(9999, 12, 31)
+        ),
+        PatientsWithTypeOneDissent(Patient_ID=5),
+        # Included: no opt-out and a deregistration followed by a subsequent
+        # re-registration
```
rebkwok, line 2886: In this case, could a patient have opted out during their previous registration (or during the period where they were registered elsewhere)? Do we assume that if they're not opted out now, it's OK to allow access to their data from the earlier registration period?  (I think so, just asking)

evansd: > could a patient have opted out during their previous registration (or during the period where they were registered elsewhere)?

Existing opt-outs should transfer in along with the rest of the patient's records.

> Do we assume that if they're not opted out now, it's OK to allow access to their data from the earlier registration period?

Opting out (and opting back in again) covers all time periods. So it's not like some periods of patient history are covered by the opt out and others aren't: either the entire history needs to be excluded, or it doesn't.

ehrql/backends/tpp.py
evansd: Data entry errors in the sense of a patient being recorded as having died when in fact they haven't? I don't want to say that's impossible, but there's a certain point at which we just have to rely on the data we're given.

If the worry is about the exact date of death being incorrect then I don't think that matters because, as you say, the fact that death is recorded in TPP indicates they were registered with a SystmOne practice at time of death – and that's all that's relevant for our purposes.

ehrql/backends/tpp.py
```diff
@@ -109,9 +109,24 @@ def modify_query_variables(self, variables):
...
+            # Include only patients who do *not* have a T1OO recorded
+            ~t1oo.exists_for_patient()
+            & (
+                # Of those, include only those who have a currently active practice
+                # registration (otherwise they may have recorded a T1OO elsewhere) ...
+                registrations.where(
+                    registrations.end_date.is_null()
+                ).exists_for_patient()
+                # ... or those who have died and therefore cannot have registered a T1OO
+                # elsewhere
```
inglesp, line 126: What about a patient who has de-registered with TPP, opted out while elsewhere, and then died?  I think this is ok, because the presence of a date of death on the patients table indicates that they haven't de-registered before dying, but it'd be worth being really explicit about this.

inglesp (approved):

ehrql/backends/tpp.py
evansd: Yes, the comment could more precise here. It should be something like:
> ... or those whose death is recorded in TPP and therefore were registered with a SystmOne practice at time of death and cannot have gone on to register a T1OO elsewhere

inglesp: That's an improvement!

ehrql/backends/tpp.py
```diff
@@ -99,24 +109,31 @@ def modify_query_variables(self, variables):
...
+        patients = ehrql.tables.tpp.patients
+        registrations = ehrql.tables.tpp.practice_registrations
+        extra_condition = (
+            # Include only patients who do *not* have a T1OO recorded
+            ~t1oo.exists_for_patient()
+            & (
+                # Of those, include only those who have a currently active practice
+                # registration (otherwise they may have recorded a T1OO elsewhere) ...
+                registrations.where(
+                    registrations.end_date.is_null()
```
madwort, line 123: Can I verify some assumptions? In the real-world it's only valid for a patient to have one registration at any one time? Does the data reflect that? If the data contains a patient with a registration without an end date, and then a second registration with a later start date than the first registration but *with* an end date, has the first registration ended by default?

I.e. wondering if this check should be "the most recently started registration has a null end date" instead of "there exists at least one registration with a null end date".

evansd: > In the real-world it's only valid for a patient to have one registration at any one time?

I don't actually know this is the case. It's certainly _normal_ for them to only have one registration at a time, but I don't know that it's completely _invalid_ for them to have multiple.

I think if the data shows an "open" practice registration than we have to take that at face value.

evansd:
Closing in favour of #2072

PR #2072, opened 2024-07-24
Switch to inclusion table for T1OO handling

evansd:
This switches the T1OO logic to use a table of explicitly _allowed_ patients, rather than a table of exclusions. This allows TPP to apply more complex criteria in determining the appropriate set of patients.

### DIFF
--- a/tests/lib/tpp_schema.py
+++ b/tests/lib/tpp_schema.py
@@ -309,6 +309,13 @@ class APCS_JRC20231009_LastFilesToContainAllHistoricalCostData(Base):
     )
 
 
+class AllowedPatientsWithTypeOneDissent(Base):
+    __tablename__ = "AllowedPatientsWithTypeOneDissent"
+    _pk = mapped_column(t.Integer, primary_key=True)
+
+    Patient_ID = mapped_column(t.BIGINT)
+
+
 class Appointment(Base):
     __tablename__ = "Appointment"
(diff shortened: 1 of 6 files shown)


tests/lib/tpp_schema.py
```diff
@@ -309,6 +309,13 @@ class APCS_JRC20231009_LastFilesToContainAllHistoricalCostData(Base):
     )
 
 
+class AllowedPatientsWithTypeOneDissent(Base):
```
madwort, line 312: am I misunderstanding, or should this be *without*? 

```suggestion
class AllowedPatientsWithoutTypeOneDissent(Base):
```

madwort: (although if that's the tablename TPP have used, then, obvs, it's that...!)

evansd: Yeah, I think they've just prefixed "Allowed" to the previous table name to indicate the change from deny-list to allow-list – though I agree the name is now somewhat misleading!

madwort: maybe it's worth a comment above the class def?

evansd: Yeah fair point, although it can't really go in the class def because that file is autogenerated from the db schema. But I can add a comment at the point in the backend where it's used.

rebkwok (approved):

PR #2132, opened 2024-10-02
Tweak the docs generation text in DEVELOPERS

madwort:


DEVELOPERS.md
```diff
@@ -315,12 +315,13 @@ To generate this file, run:
 
     just generate-docs
 
-This generates the markdown files in `docs/includes/generated_docs`.
+This generates the markdown files in `docs/includes/generated_docs`. It is then a developer's responsibility to update the generated docs in their PR if required.
```
rebkwok, line 318:
```suggestion
This generates the markdown files in `docs/includes/generated_docs`. If any of the files have changed you should include them in your PR. 
```

DEVELOPERS.md
```diff
@@ -315,12 +315,13 @@ To generate this file, run:
...
 
-This command runs as a pre-commit hook and will fail if there are any changes to the
-generated markdown files. It is a developer's responsibility to update the generated docs in
-their PR if required. There is also a CI step that will check that the documentation is up to
-date.
+To verify the documentation is up to date, run:
+
+    just docs-check-generated-docs-are-current
+
+This command runs as a pre-commit hook and will display a diff if there are any changes to the generated markdown files. There is also a CI step that will check that the documentation is up to date.
```
rebkwok, line 324: The confusion here is that `just docs-check-generated-docs-are-current` _used_ to run as a pre-commit hook, but it was removed.  If you've run `just generate-docs` and it created new docs files, you should just add them to the PR. This check ONLY runs in CI.

So we could change this to say: "A CI step will generate the markdown files and fail if any files have changed. If this happens you should run `just generate-docs` locally and commit the result."

rebkwok (approved):
Some suggested changes; the problem wasn't the description of workflow, but that the workflow itself has changed and hadn't been updated in the docs

DEVELOPERS.md
madwort: ah right, yes, I was aware that pre-commit wasn't running for me locally but I thought I might have disabled it in my own setup for being annoying... will tweak the text some more.

PR #2143, opened 2024-10-07
Fix capitalisation in example data

### LINKED ISSUE #2106: Inconsistent capitalisation of "the" causing problems with example data
The [example data](https://github.com/opensafely-core/ehrql/blob/main/ehrql/example-data/practice_registrations.csv) reference "Yorkshire and the Humber", while our [table validation](https://github.com/opensafely-core/ehrql/blob/ad304d3dd213b2cb468c19829c1bc72b4e6e183a/ehrql/tables/tpp.py#L1333) checks for "Yorkshire and The Humber".

This causes problems in the sandbox:

```
(.venv) inglesp@malbogies:~/work/ebmdatalab/ehrql$ python -m ehrql sandbox ehrql/example-data
Python 3.11.10 (main, Sep  7 2024, 18:35:41) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from ehrql.tables.tpp import practice_registrations
>>> practice_registrations
  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 145, in parser
    return convertor(value)
           ^^^^^^^^^^^^^^^^
  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 168, in wrapper
    raise ValueError(f"{value!r} not in valid categories: {category_str}")
ValueError: 'Yorkshire and the Humber' not in valid categories: 'North East', 'North West', 'Yorkshire and The Humber', 'East Midlands', 'West Midlands', 'East', 'London', 'South East', 'South West'

During handling of the above exception, another exception occurred:

  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 83, in __iter__
    yield row_parser(row)
          ^^^^^^^^^^^^^^^
  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 110, in row_parser
    return tuple(parser(row) for parser in parsers)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 110, in <genexpr>
    return tuple(parser(row) for parser in parsers)
                 ^^^^^^^^^^^
  File "/home/inglesp/work/ebmdatalab/ehrql/ehrql/file_formats/csv.py", line 147, in parser
    raise ValueError(f"column {name!r}: {e}")
ValueError: column 'practice_nuts1_region_name': 'Yorkshire and the Humber' not in valid categories: 'North East', 'North West', 'Yorkshire and The Humber', 'East Midlands', 'West Midlands', 'East', 'London', 'South East', 'South West'

During handling of the above exception, another exception occurred:

  File "<console>", line 1, in <module>
ehrql.file_formats.base.FileValidationError: row 8: column 'practice_nuts1_region_name': 'Yorkshire and the Humber' not in valid categories: 'North East', 'North West', 'Yorkshire and The Humber', 'East Midlands', 'West Midlands', 'East', 'London', 'South East', 'South West'
```

alarthast:
Fixes #2106.

- Incorrect capitalisation in the example `practice_registrations` data leads to a ValueError being raised when used
- Capitalisation is now fixed and a new unit test is added to check that all example data files can be loaded correctly.

### DIFF
--- a/tests/unit/test_example_data.py
+++ b/tests/unit/test_example_data.py
@@ -1,0 +1,24 @@
+from pathlib import Path
+
+import pytest
+
+import ehrql
+from ehrql.query_engines.local_file import LocalFileQueryEngine
+from ehrql.query_language import EventFrame
+from ehrql.tables import core, tpp
+
+
+# Example csv files are given for all core tables (although the tutorial imports them from the tpp module)
+CORE_TABLES = [getattr(core, table) for table in core.__all__]
+# Example csv files are given for these tables that are currently only available for tpp
+TPP_TABLES = [tpp.addresses, tpp.practice_registrations]
+
+EXAMPLE_TABLES = CORE_TABLES + TPP_TABLES
+EXAMPLE_DATA_DIR = Path(ehrql.__file__).parent / "example-data"
+
+
+@pytest.mark.parametrize("ql_table", EXAMPLE_TABLES)
+def test_populate_database_using_example_data(ql_table: EventFrame):
+    # The engine populates the database with the example data and validates the column specs in the process
+    engine = LocalFileQueryEngine(EXAMPLE_DATA_DIR)
+    engine.populate_database([ql_table._qm_node])
--- a/ehrql/example-data/practice_registrations.csv
+++ b/ehrql/example-data/practice_registrations.csv
@@ -6,8 +6,8 @@ patient_id,start_date,end_date,practice_pseudo_id,practice_stp,practice_nuts1_re
 4,2019-01-02,,7199,E54000026,East
 5,2007-07-21,2018-05-23,8595,E54000024,East
 6,2020-09-14,,5045,E54000017,West Midlands
-7,2013-12-05,,7588,E54000050,Yorkshire and the Humber
-7,2017-05-12,,5121,E54000050,Yorkshire and the Humber
+7,2013-12-05,,7588,E54000050,Yorkshire and The Humber
+7,2017-05-12,,5121,E54000050,Yorkshire and The Humber
 8,2020-01-19,,1858,E54000012,West Midlands
 9,1995-03-01,2017-11-10,8189,E54000023,East
 10,2001-10-24,,7783,E54000037,South West


rebkwok (approved):

tests/unit/test_example_data.py
```diff
@@ -0,0 +1,40 @@
...
+    tpp.clinical_events,
+    tpp.medications,
+    core.ons_deaths,
+    core.patients,
+    tpp.practice_registrations,
+]
+
+
+@pytest.mark.parametrize("filename,ql_table", zip(filenames, ql_tables))
+def test_read_rows(filename, ql_table):
```
rebkwok, line 31: Could you add a comment about what this test is doing? It's not immediately obvious that CSVRowsReader is validating it against the table's column specs

tests/unit/test_example_data.py
```diff
@@ -0,0 +1,40 @@
...
+    table_nodes = get_table_nodes(ql_table._qm_node)
+    [table] = table_nodes  # There should only be one table
+    column_specs = get_column_specs_from_schema(table.schema)
+
+    CSVRowsReader(
+        Path(f"ehrql/example-data/{filename}"),
+        column_specs=column_specs,
+        allow_missing_columns=True,
+    )
```
evansd, line 40: This test ends up duplicating some of the logic from `LocalFileQueryEngine` where it might be better to use that logic directly. It would also avoid the awkwardness of having to manually specific the filenames.

You could do this with something like:
```py
LocalFileQueryEngine("path/to/example-data").populate_database([ql_table._qm_node])
```
Which will throw an error if there's anything wrong with the data for that tabe.

tests/unit/test_example_data.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.mark.parametrize("filename,ql_table", zip(filenames, ql_tables))
+def test_read_rows(filename, ql_table):
+    table_nodes = get_table_nodes(ql_table._qm_node)
+    [table] = table_nodes  # There should only be one table
+    column_specs = get_column_specs_from_schema(table.schema)
+
+    CSVRowsReader(
+        Path(f"ehrql/example-data/{filename}"),
```
evansd, line 37: It might be better to get the path from the module, rather than make assumptions about the current directory. So something like:
```py
Path(ehrql.__file__).parent / "example-data"
```

tests/unit/test_example_data.py
```diff
@@ -0,0 +1,40 @@
...
+    "practice_registrations.csv",
+]
+
+ql_tables = [
+    tpp.addresses,
+    tpp.clinical_events,
+    tpp.medications,
+    core.ons_deaths,
+    core.patients,
+    tpp.practice_registrations,
```
evansd, line 26: It's not obvious where this particular list of tables comes from. That's not a reflection on your code! I just don't think we've been systematic in deciding what tables we're providing example data for. A reasonable approach would be: everything in `core` and every table used in the tutorial.

But whatever we decide, we should be dynamically constructing the list of tables here otherwise someone could add a new core table, or add a table to the tutorial, and nothing would tell them that they had failed to add it to the example data.

evansd (commented):
Thanks Alice. I think ideally we'd split the data fixes out from the tests in separate commits as they're distinct changes.

There's a wider question here as to exactly what tables we ought to be including in the example data. We don't necessarily have to address that immediately, but we're going to need to resolve it at some point so it might make sense to think about this now.

PR #2147, opened 2024-10-08
Improve performance and shrink quality of Hypothesis tests

DRMacIver:
This is a number of small niceties and general improvements to the Hypothesis tests that I've spotted and seem useful to include.

Each of the changes should be relatively self contained and explain itself, but short summary:

1. This allows the shrinker to disable various things (tests to run, database backends) that it determines aren't necessary for the test to fail, which should make the tests much faster when shrinking.
2. The explain phase turns out to be really slow for your tests (this is downstream of data generation being slow, which is downstream of Hypothesis being slow for generating large data), so this disables them by default.
3. The max depth logic breaks shrinking, because it will often prevent it from shrinking the example to one that doesn't hit the max depth limit. This fixes that by always inserting a decision point the shrinker can use to force termination to happen.

There still seem to be some cases where this triggers bad shrinks, but I've not yet been able to reproduce them reliably (there's some nondeterminism going on there that I don't quite understand and need to figure out)

### DIFF
--- a/tests/generative/generic_strategies.py
+++ b/tests/generative/generic_strategies.py
@@ -1,0 +1,40 @@
+from hypothesis import strategies as st
+
+
+# This is a variable that will normally be set to True, but which shrinking is
+# allowed to make False.
+#
+# Examples of where this is useful:
+#
+# * Turning off expensive test operations that you want to check but don't care
+#   about running if they turn out not to be relevant to the error you're
+#   seeing.
+# * Giving the shrinker a place to terminate generation in places where you've
+#   decided to do it yourself because e.g. the data was getting too large.
+usually = st.integers(0, 255).map(lambda n: n > 0)
+
+
+@st.composite
+def usually_all_of(draw, options, min_size=1):
+    """Generates a list of distinct elements drawn from `options`, of size at least
+    `min_size`. In the normal course of things, this will usually be all of `options`,
+    but the shrinker is allowed to remove elements from it, which can speed up
+    test execution during shrinking significantly."""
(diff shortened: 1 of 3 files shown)


tests/generative/generic_strategies.py
```diff
@@ -0,0 +1,32 @@
+from hypothesis import strategies as st
+
+
+# This is a variable that will normally be set to True, but which shrinking is
+# allowed to make False. You can use this for
```
evansd, line 5: Don't leave us hanging! What can we use it for?

evansd (approved):
This is great, thanks. The `usually` trick would never have occurred to me, but now I see it I _think_ I understand why it works. And once we've got that I can see how the rest drops out quite naturally.

tests/generative/generic_strategies.py
```diff
@@ -0,0 +1,40 @@
...
+# This is a variable that will normally be set to True, but which shrinking is
+# allowed to make False.
+#
+# Examples of where this is useful:
+#
+# * Turning off expensive test operations that you want to check but don't care
+#   about running if they turn out not to be relevant to the error you're
+#   seeing.
+# * Giving the shrinker a place to terminate generation in places where you've
+#   decided to do it yourself because e.g. the data was getting too large.
```
evansd, line 13: Great, thank you! Feel free (though not obliged) to rebase this on to the original commit if you want. We tend to be quite liberal with rebasing (even mid-review, which I know is controversial) in order to prioritise the usefulness of the history.

DRMacIver: Noted! I'll rebase before merging.

PR #2148, opened 2024-10-10
Exit with specific status codes for specific types of failure

### LINKED ISSUE #1102: Exit with specific status codes for specific types of failure
Job-runner will spot these status codes and provide more helpful error messages to the user.

The list of available codes and their meanings is given [here](https://github.com/opensafely-core/job-runner/blob/75e3c53cea547bc3798658f44880389546ae93f5/jobrunner/config.py#L201-L211):
```py
# Map known exit codes to user-friendly messages
DATABASE_EXIT_CODES = {
    # Custom database-related exit codes return from cohortextractor, see
    # https://github.com/opensafely-core/cohort-extractor/blob/0a314a90/cohortextractor/cohortextractor.py#L787
    3: (
        "A transient database error occurred, your job may run "
        "if you try it again, if it keeps failing then contact tech support"
    ),
    4: "New data is being imported into the database, please try again in a few hours",
    5: "Something went wrong with the database, please contact tech support",
}
```

The exit codes are emitted in Cohort Extractor [here](https://github.com/opensafely-core/cohort-extractor/blob/52588b9518c67632ddae98780d58078c947a720d/cohortextractor/cohortextractor.py#L922-L947):
```py
        except Exception as e:
            # Checking for "DatabaseError" in the MRO means we can identify database errors without
            # referencing a specific driver.  Both pymssql and presto/trino-python-client raise
            # exceptions derived from a DatabaseError parent class

            # Ignore errors which don't look like database errors
            if "DatabaseError" not in str(e.__class__.mro()):
                raise

            traceback.print_exc()
            # Exit with specific exit codes to help identify known issues
            transient_errors = [
                "Unexpected EOF from the server",
                "DBPROCESS is dead or not enabled",
            ]
            if any(message in str(e) for message in transient_errors):
                logger.error(f"Intermittent database error: {e}")
                sys.exit(3)
            if "Invalid object name 'CodedEvent_SNOMED'" in str(e):
                logger.error(
                    "CodedEvent_SNOMED table is currently not available.\n"
                    "This is likely due to regular database maintenance."
                )
                sys.exit(4)
            logger.error(f"Database error: {e}")
            sys.exit(5)
```

I think the way to handle this in ehrQL would be to provide a hook for backends to map exceptions to exit codes, as this code is all backend specific and doesn't belong in the ehrQL core.

So `BaseBackend` would have a method called something like `get_exit_status_for_exception()` which is passed an exception object and returns either a pair of `(exit_code, log_message)` or `None`.

If the return value is `None` we continue as normal. Otherwise we log the traceback to stderr as normal (to aid debugging), then output the supplied log message and exit with the supplied status.

The default implementation is of course just `return None` and then something like the above logic can go in `TPPBackend.get_exit_status_for_exception()`.

rebkwok:
Fixes #1102 

For database tasks (generate_dataset_with_dsn, generate_measures_with_dsn), exit with specific exit codes supplied
by the backend.

### DIFF
--- a/ehrql/__main__.py
+++ b/ehrql/__main__.py
@@ -2,6 +2,7 @@ import importlib
 import logging
 import os
 import sys
+import traceback
 import warnings
 from argparse import ArgumentParser, ArgumentTypeError, RawTextHelpFormatter
 from pathlib import Path
@@ -109,6 +110,17 @@ def main(args, environ=None):
         # Handle errors encountered while reading user-supplied data
         print(f"{exc.__class__.__name__}: {exc}", file=sys.stderr)
         sys.exit(1)
+    except Exception as exc:
+        # For functions which take a `backend_class` give that class the chance to set
+        # the appropriate exit status for any errors
+        if backend_class := kwargs.get("backend_class"):
+            if exit_status := backend_class(environ).get_exit_status_for_exception(
+                exc
+            ):  # pragma: no branch
+                # log the traceback to stderr to aid debugging
+                traceback.print_exc()
+                sys.exit(exit_status)
+        raise
     finally:
         root_logger.setLevel(orig_log_level)
 
--- a/ehrql/backends/base.py
+++ b/ehrql/backends/base.py
@@ -41,6 +41,9 @@ class BaseBackend:
         """
         return query
 
+    def get_exit_status_for_exception(self, exception):
+        return None
+
 
 class SQLBackend(BaseBackend):
     query_engine_class = None
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -126,6 +126,34 @@ class TPPBackend(SQLBackend):
         )
         return variables
 
+    def get_exit_status_for_exception(self, exception):
+        # Checking for "DatabaseError" in the MRO means we can identify database errors without
+        # referencing a specific driver.  Both pymssql and presto/trino-python-client raise
+        # exceptions derived from a DatabaseError parent class
+
+        # Ignore errors which don't look like database errors
+        if "DatabaseError" not in str(exception.__class__.mro()):
+            return
+
+        # Exit with specific exit codes to help identify known issues
+        transient_errors = [
+            "Unexpected EOF from the server",
+            "DBPROCESS is dead or not enabled",
+        ]
+        if any(message in str(exception) for message in transient_errors):
+            exception.add_note(f"\nIntermittent database error: {exception}")
+            return 3
+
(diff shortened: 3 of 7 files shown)


evansd (commented):
Thanks Becky. I think all the exception inspection code here looks good. But I think there are some structural changes that should be made here which would also result in the code being a bit simpler. 

The first is that I don't think we should be calling `sys.exit()` inside `main.py`. Ideally, there's just a single entry/exit point where we handle exiting. Currently that's the function `ehrql.__main__.main()`.

Secondly, having the exception handling done inside the specific command functions themselves means we need to carefully about what parts of the code might and might not throw database exceptions. And if we add some new command nothing in the tests would tell us we were missing the relevant exception handler. So I think it would make sense to handle these at a higher level.

I've also realised we could make the API a bit neater and easier to call by not returning an `(exit_code, log_message)` tuple. Instead we could return just an exit code (or None) and then if the backed wants to add helpful messaging it can use the [`add_note()`](https://docs.python.org/3/library/exceptions.html#BaseException.add_note) method on the exception. This will get displayed as the last thing in the traceback, so it's functionally equivalent to what I first proposed. (Though we may want to add an extra leading newline before the message just to give some visual separation.)

Putting all that together, I think we could handle this all with an extra `except` clause in `ehrql.__main__.main()`, just [before the `finally`](https://github.com/opensafely-core/ehrql/blob/f62511e28c43b33a975cff7189251ac0e881c85e/ehrql/__main__.py#L112):

```py
    except Exception as exc:
        # For functions which take a `backend_class` give that class the chance to set
        # the appropriate exit status for any errors
        if backend_class := kwargs.get("backend_class"):
            if exit_status := backend_class(environ).get_exit_status_for_exception(exc):
                traceback.print_exc()
                sys.exit(exit_status)
        raise
```

That shouldn't require any changes to the inner functions at all, and it means we don't need to worry about adding new exception handlers if we add new commands.

It also simplifies the testing story because there's no need to have separate dataset/measures tests as they all go through the same code path.

rebkwok:
> The first is that I don't think we should be calling `sys.exit()` inside `main.py`. Ideally, there's just a single entry/exit point where we handle exiting. Currently that's the function `ehrql.__main__.main()`.

:facepalm: This is how I wanted to do it, but for some reason I'd decided that I couldn't get the backend class at this point.

evansd:
Well, yanking it out of the `kwargs` isn't particularly lovely. But I think it's not _too_ nasty and it makes the structure overall much cleaner.

evansd (approved):
Looks good, thanks!

evansd:
Oh, I forgot: this needs a pedantic commit prefix somewhere :facepalm:

PR #2155, opened 2024-10-14
Replace some uses of assert False with pragma: no branch

DRMacIver:
This is an extremely trivial piece of housekeeping, but I noticed that there were a couple places in the code base where `assert False` was being used to indicate to coverage that a loop would never exit normally. The `pragma: no branch` comment is a more straightforward way to do this.

### DIFF
--- a/ehrql/dummy_data/generator.py
+++ b/ehrql/dummy_data/generator.py
@@ -69,7 +69,7 @@ class DummyDataGenerator:
         )
         start = time.time()
 
-        for patient_id_batch in self.get_patient_id_batches():
+        for patient_id_batch in self.get_patient_id_batches():  # pragma: no branch
             # Generate batches of patient data (just enough to determine population
             # membership) and find those matching the population definition
             patient_batch = {
@@ -112,9 +112,6 @@ class DummyDataGenerator:
                 )
                 return data
 
-        # Keep coverage happy: the loop should never complete
-        assert False
-
     def get_patient_id_batches(self):
         id_stream = self.get_patient_id_stream()
         while True:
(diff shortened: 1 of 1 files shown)


evansd (approved):
This is obviously slightly fewer characters, but I'm not 100% convinced it's an improvement because the asserts mean we blow up immediately and loudly if anything happens to violate our assumptions. That feels like an additional benefit which the pragmas alone don't give us.

Incidentally, I'm particularly interested in your views here because I started leaning more heavily on asserts after reading the Hypothesis source, so at least partial credit/blame goes in your direction :)

DRMacIver:
I think the advantage here is really clarity rather than number of characters, in that the pragma conveys the meaning "this path isn't covered" more clearly than the `assert False`. If you prefer the assert False I don't have a very strong preference in favour of the no branch though.

In general I'm a big fan of assertions and think that assertions that have a real possibility of failure are very valuable, but assertions that are in some sense really logically impossible are somewhat less valuable, so for coverage in particular I will tend to use comments rather than assertions to indicate impossible paths unless I'm not as confident as all that that it really is impossible.

As an aside: I've started preferring `raise AssertionError()` to `assert False`. The advantage is that it won't be removed with optimisations turned on, and because there's no nontrivial expression to evaluate and it will only be hit on the unhappy path, this is basically free. Also pragmatically I've started preferring it because `shed` automatically turns `assert False` into `raise AssertionError` for me.

evansd:
OK, I think I only actually disagree on the first example here and I just didn't pay enough attention to the others the first time round.

With the `get_table_docstring()` function you _could_ hit that branch by passing in something which wasn't a `PatientFrame`/`EventFrame` subclass. So I think the actual mistake here is that the comment on the assert should say:
```py
# Should only be called with PatientFrame/EventFrame subclass
```
Rather than:
```py
# Keep coverage happy
```
i.e. the `assert` is doing more than just placating coverage.

With the other two examples you're right that there aren't reasonable paths by which you could hit them, in which case the pragma makes more sense.

PR #2159, opened 2024-10-15
Fix MSSQL retry code to work as intended

### LINKED ISSUE #2144: MSSQL retry code isn't working as intended
Having [believed](https://github.com/opensafely-core/ehrql/pull/1529) that the dreaded "DBPROCESS is dead or not enabled" errors were a thing of the past we have just hit our [first one](https://jobs.opensafely.org/comparing-disparities-in-rsv-influenza-and-covid-19/disparities-comparison-rsv-flu-c19/23632/gftsijtwub7edtgd/) in ehrQL.

This revealed that our retry code isn't working as intended because the exception gets thrown while the results iterator is being consumed and not when the query is made.
Image

That means that this exception handler here is insufficient:
https://github.com/opensafely-core/ehrql/blob/aca562c141778f15b0b01901ed1af68b0e993bb7/ehrql/utils/sqlalchemy_exec_utils.py#L68-L72

The very simplest fix would be to wrap `execute()` in `list()` i.e. consume the iterator inside the exception handler. Given the fact that this is only used in the context of the "batched" downloader which deliberately fetches only reasonable-sized chunks of results I think this would be perfectly sensible behaviour.

Before this though, we should add a test with exposes the current limitation. There are already some tests which use `mock` to inject errors and test that they're handled correctly e.g.
https://github.com/opensafely-core/ehrql/blob/971924e43dee8ea8b48c501fe124c49a688a1a1d/tests/unit/utils/test_sqlalchemy_exec_utils.py#L65-L73

We'd need to add another one which included in its results an iterator which didn't throw an error until mid-way through iteration.

alarthast:
Fixes #2144.

- Consume the iterator returned by `Connection.execute` within the while loop designed to retry downloads in case of transient DBPROCESS errors (The current code is not working as intended as it only passes the iterator along leaving for things to fail downstream) 
- Added a test that mimics the transient errors the retry code aims to catch

### DIFF
--- a/tests/unit/utils/test_sqlalchemy_exec_utils.py
+++ b/tests/unit/utils/test_sqlalchemy_exec_utils.py
@@ -62,13 +62,45 @@ def test_fetch_table_in_batches(table_size, batch_size, expected_query_count):
 ERROR = OperationalError("A bad thing happend", {}, None)
 
 
+def test_execute_with_retry():
+    tracker = {"error": True, "finished": []}
+
+    def log(msg):
+        tracker["logs"] = tracker.get("logs", []) + [msg]
+
+    num_parts = 4
+
+    def execute():  # Return an iterator to mimic Connection.execute
+        # Succeed for 0, 1, fail for 2 on the first go, then succeed for 2, 3
+        def execute_part(part, tr):
+            if "error" in tr and part == 2:
+                tr.pop("error")
+                raise ERROR
+            return part
+
+        return (execute_part(part, tracker) for part in range(num_parts))
+
+    execute_with_retry = execute_with_retry_factory(
+        execute, max_retries=3, retry_sleep=0, backoff_factor=2, log=log
+    )
+
+    # Do something with the results
+    for part in execute_with_retry():
+        tracker["finished"].append(part)
+
+    assert tracker["finished"] == list(range(num_parts))
+    assert "Retrying query (attempt 1 / 3)" in tracker["logs"]
+    assert "Retrying query (attempt 2 / 3)" not in tracker["logs"]
+
+
 @mock.patch("time.sleep")
-def test_execute_with_retry(sleep):
+def test_sleep_between_retries(sleep):
     execute = mock.Mock(side_effect=[ERROR, ERROR, ERROR, "its OK now"])
     execute_with_retry = execute_with_retry_factory(
         execute, max_retries=3, retry_sleep=10, backoff_factor=2
     )
-    assert execute_with_retry() == "its OK now"
+    # list() is always called on the successful return value
+    assert execute_with_retry() == list("its OK now")
     assert execute.call_count == 4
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_sqlalchemy_exec_utils.py
```diff
@@ -62,13 +62,45 @@ def execute(self, query):
...
+    execute_with_retry = execute_with_retry_factory(
+        execute, max_retries=3, retry_sleep=0, backoff_factor=2, log=log
+    )
+
+    # Do something with the results
+    for part in execute_with_retry():
+        tracker["finished"].append(part)
+
+    assert tracker["finished"] == list(range(num_parts))
+    assert "Retrying query (attempt 1 / 3)" in tracker["logs"]
```
evansd, line 92: Asserting on the log message too is a nice addition. I think these were untested before so thanks for including this.

You can do this slightly more simply though like this:
```py
log_messages = []

execute_with_retry = execute_with_retry_factory(
    ...    
    log=log_messages.append,
)

File: tests/unit/utils/test_sqlalchemy_exec_utils.py [hunk 2]
```diff
@@ -62,13 +62,45 @@ def execute(self, query):
...
+    assert "Retrying query (attempt 2 / 3)" not in tracker["logs"]
+
+
 @mock.patch("time.sleep")
-def test_execute_with_retry(sleep):
+def test_sleep_between_retries(sleep):
     execute = mock.Mock(side_effect=[ERROR, ERROR, ERROR, "its OK now"])
```
evansd, line 98: I think there's a slight simpler way of testing the behaviour here which is to replace one of the "immediate error" cases with an iterator that triggers an error part way through e.g.
```suggestion
    def iterate_with_error():
        yield 1
        yield 2
        raise ERROR

    execute = mock.Mock(side_effect=[ERROR, iterate_with_error(), ERROR, ["its", "OK", "now"])
```

So maybe modifying this test to cover that case and adding your log-message-asserting code would be a simpler way to go?

tests/unit/utils/test_sqlalchemy_exec_utils.py
```diff
@@ -62,13 +62,45 @@ def execute(self, query):
...
     execute_with_retry = execute_with_retry_factory(
         execute, max_retries=3, retry_sleep=10, backoff_factor=2
     )
-    assert execute_with_retry() == "its OK now"
+    # list() is always called on the successful return value
+    assert execute_with_retry() == list("its OK now")
```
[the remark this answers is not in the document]
alarthast: Got it - made the value a tuple and assert turned to list

evansd:
I think I just remember is that in order to generate a new version at least one commit in the PR needs to have a "semantic prefix" which means one of `fix:` or `feat:`. I think `fix:` is the appropriate one here, and you've nearly got it already – just missing the colon.

alarthast:
Thank you - have now added the colon :)

PR #2171, opened 2024-10-17
Derive extra column constraints for dummy data from population queries

DRMacIver:
This is a first pass at getting the dummy data generator good enough to reliably create full datasets.

The basic idea is that we take the population query and turn it into an extra set of constraints to pass to the Generator for the column.

I don't think this is actually going to be good enough to get the full range of dummy data we want, but it's already able to significantly improve the amount of data we can generate for many hard to satisfy queries, because it boosts the probability of a member of a population passing the population definition query (sometimes all the way to 1, sometimes only enough to improve the odds).

Apologies for the giant pattern match statement, but I think unironically something like it is going to be the best way to handle this.

This pull request also contains a few niceties for the generative tests, that I can pull out into their own thing if you want.

### DIFF
--- a/ehrql/dummy_data/generator.py
+++ b/ehrql/dummy_data/generator.py
@@ -186,19 +186,34 @@ class DummyPatientGenerator:
         self.rnd.seed(f"{self.random_seed}:{patient_id}")
         # TODO: We could obviously generate more realistic age distributions than this
 
-        dob_column = self.get_patient_column("date_of_birth")
-        if dob_column is not None and dob_column.get_constraint(
-            Constraint.GeneralRange
-        ):
-            self.events_start = self.today - timedelta(days=120 * 365)
-            self.events_end = self.today
-            date_of_birth = self.get_random_value(dob_column)
-        else:
-            date_of_birth = self.today - timedelta(
-                days=self.rnd.randrange(0, 120 * 365)
-            )
-        age_days = self.rnd.randrange(105 * 365)
-        date_of_death = date_of_birth + timedelta(days=age_days)
+        while True:
+            # Retry until we have a date of birth and date of death that are
+            # within reasonable ranges
+            dob_column = self.get_patient_column("date_of_birth")
+            if dob_column is not None and dob_column.get_constraint(
+                Constraint.GeneralRange
+            ):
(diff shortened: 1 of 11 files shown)


evansd (commented):
Thanks for this, David. There's definitely an interesting idea here worth pursuing. In fact, this is fleshing out exactly the original aspiration for the `query_info` module which was to answer the question: given a particular query, what can we learn about the shape of the underlying data to guide dummy data generation?

However that never got any further than the absolute minimum required to get us unblocked. So it's great to see what you've managed to do here.

What I'm not sure about is whether it makes sense to merge this as is now. On the plus side it does, as you say, make some cases much better without making anything worse. On the other hand, I'm not sure that the cases it makes better bear any resemblance to population definitions our users would actually write. So the concrete gains may be hypothetical at this stage. And of course there's always a cost to having code in production.

If you think it really is important to get the early work merged sooner rather than later then I'm certainly open to that. There's probably some small refactoring I'd like to do to the _existing_ code to make some of the things your doing easier. But that would only be worth doing at this stage if the intention was to merge early.

DRMacIver:
> What I'm not sure about is whether it makes sense to merge this as is now. On the plus side it does, as you say, make some cases much better without making anything worse. On the other hand, I'm not sure that the cases it makes better bear any resemblance to population definitions our users would actually write. So the concrete gains may be hypothetical at this stage. And of course there's always a cost to having code in production.

I'm certainly open to keeping this in a long-running branch or something. I'm generally on team merge-early merge-often, as I think it makes it easier to keep code in sync and catch problems early, and don't like to diverge too far from main, but if you'd prefer to keep it separate until it's a bit more fully fleshed we can do it that way instead.

evansd:
> I'm generally on team merge-early merge-often, as I think it makes it easier to keep code in sync and catch problems early

Yes, quite right. And it would be good to get things in front of users as soon as we can as well. Having reflected on this a bit more I think there might be another way forward here. Will start a conversation on Slack.

DRMacIver:
OK @evansd this now only makes changes to the `dummy_data_nextgen` package rather than `dummy_data`.

evansd (approved):

PR #2173, opened 2024-10-18
fix: trigger release for previous commit

alarthast:
Awkwardly, the commit message in https://github.com/opensafely-core/ehrql/pull/2145 did not have a `fix:` tag and hence did not trigger a release.
This is an "empty" PR that adds the tag so that the previous change can be deployed.

### DIFF
--- a/tests/unit/utils/test_sqlalchemy_exec_utils.py
+++ b/tests/unit/utils/test_sqlalchemy_exec_utils.py
@@ -72,7 +72,7 @@ def test_execute_with_retry(sleep):
         raise ERROR
 
     execute = mock.Mock(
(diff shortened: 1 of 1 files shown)


evansd:
Because the change was a docs change you don't actually need to release a new version here. This is all very awkward and unsatisfactory. We had a plan to do something about it at one stage but it sort of fell by the wayside. We should probably pick it up again :smile:

alarthast:
Oh - thank you Dave! Can indeed see that the docs change is live!

rebkwok:
@evansd I asked Alice to make an empty commit (but neglected to explain how) - but it wasn't for the docs change, it was for the mssql retry code. So I think we do need this.

rebkwok:
No, I am confusing PRs and actions. Nvm!

PR #2177, opened 2024-10-21
Promote `practice_registrations` to a "core" table

evansd:
As well as being obviously deserving of a place here this means we can provide richer and more complex examples in the documentation while respecting our self-imposed rule to only use `core` tables.


ehrql/tables/core.py
```diff
@@ -99,6 +100,57 @@ def age_on(self, date):
...
+        description="Pseudonymised practice identifier.",
+    )
+
+    def for_patient_on(self, date):
+        """
+        Return each patient's practice registration as it was on the supplied date.
+
+        Where a patient is registered with multiple practices we prefer the most recent
+        registration and then, if there are multiple of these, the one with the longest
+        duration. If there's stil an exact tie we choose arbitrarily based on the
```
rebkwok, line 130:
```suggestion
        duration. If there's still an exact tie we choose arbitrarily based on the
```

rebkwok (approved):
The tests for `practice_registrations.spanning` and `practice_registrations.for_patient_on` are in tests/integrations/backends/test_tpp.py - should they most to test_core.py now?

evansd:
Good catch, thank you!

Issue #2169, opened 2024-10-16
Come up with a set of coherent example data


inglesp:
The existing set of example data is anaemic, which limits its value as an aid for learning ehrQL.

We should come up with a set of interesting and coherent properties that we would like the example data to have.  This will allow us add some interesting narrative structure to the tutorial, to a quiz (see #2168), and to other learning materials.

This ticket does not include any implementation work.

inglesp:
See implementation notes in #2215.

PR #2180, opened 2024-10-23
Evansd/tutorial examples

### LINKED ISSUE #2146: Define criteria for tables/columns to be provided in `example-table` CSV files
- See [Slack conversion](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1728308865306139) and [this comment](https://github.com/opensafely-core/ehrql/pull/2143#discussion_r1791471961) in #2143.
- Currently it is unclear what tables should be included as CSV files in the example data. @evansd suggested that all core tables + all tables used in the tutorial would make sense
  - If so, we should add a CSV file representing the `tpp.apcs` table
- Once the criteria is set, the [test file](https://github.com/opensafely-core/ehrql/blob/main/tests/unit/test_example_data.py) for example data should be updated to remove the hard-coding of tables. 
- We should also look at whether our example CSV files ought to include all the columns of the tables they represent
  - If so, add the columns and use `allow_missing_columns=False` in the tests

evansd:
This reworks the more complex dataset example in the tutorial to only use core tables rather than TPP tables. In particular that means:

 * Dropping use of the `addresses` table and removing Index of Multiple Deprivation from the dataset entirely.
 * Replacing `age` with `age_band` so as to still have an example of using `case()` to map values ranges to categories (which Index of Multiple Deprivation used to provide).
 * Dropping use of the `apcs` table to get hospital admissions as an outcome and replacing it with astham exacerbations drawn from the `clinical_events` table.
 *  Using the SNOMED-CT rather than CTV3 version of the ethnicity codelist.

See discussion with researchers here:
https://bennettoxford.slack.com/archives/C02HJTL065A/p1729168468507629

Following this we [adopt the policy](https://github.com/opensafely-core/ehrql/issues/2146#issuecomment-2428906428) that our bundled example data should cover all tables in core, and all columns in those tables, and enforce this in a test.

We have **not** yet expanded this example data to support a more coherent narrative as per #2169.

Closes #2146


docs/tutorial/writing-a-more-complex-dataset-definition/index.md
```diff
@@ -254,14 +254,31 @@ dataset.sex = patients.sex
...
 Remember that to be included in the population,
 a patient was an adult on the index date.
 
 ```python
-dataset.age = patients.age_on(index_date)
+age = patients.age_on(index_date)
 ```
 
+We then use the `age` variable to match the patient to one of five age
+bands using `case()` expression and assign this to `dataset.age_band`.
```
rebkwok, line 266: missing "the" before `case()`?

docs/tutorial/writing-a-more-complex-dataset-definition/index.md
```diff
@@ -254,14 +254,31 @@ dataset.sex = patients.sex
...
+    when((age >= 50) & (age < 65)).then("age_50_64"),
+    when((age >= 65) & (age < 75)).then("age_65_74"),
+    when((age >= 75) & (age < 85)).then("age_75_84"),
+    when((age >= 85)).then("age_85_plus"),
+)
+```
+
+The `case()` expression consists of a list of `when(<condition>).then(<value>)`
+expressions, where `<condition>` is always a Boolean and `<value>` can
+be any type as long as it's consistent with the `case()` expression.
```
rebkwok, line 280: Is it obvious what we mean by "as long as it's consistent with the case() expression"? What would be inconsistent?

docs/tutorial/writing-a-more-complex-dataset-definition/index.md
```diff
@@ -383,39 +383,53 @@ Notice that we:
...
 
 ```python
-from ehrql.tables.tpp import apcs
+asthma_exacerbations_codelist = codelist_from_csv(
+    "codelists/bristol-asthma-exacerbations.csv",
+    column="code",
+)
 ```
 
 Finally, we query the table
```
rebkwok, line 406: It reads a bit oddly to say "First..." on line 397, and then "Finally,..." as the next thing (I suspect this was copy/pasted originally from the Ethnicity or Number of medication sections which both have a First (codelist), Next, Finally pattern.

docs/tutorial/writing-a-more-complex-dataset-definition/index.md
evansd: Missing "a" I think, but good catch – thanks.

docs/tutorial/writing-a-more-complex-dataset-definition/index.md
evansd: I think I meant to write "within the case expression". Does that make enough sense on its own?

docs/tutorial/example-study/codelists/codelists.json
```diff
@@ -11,6 +11,12 @@
       "url": "https://www.opencodelists.org/codelist/opensafely/ethnicity/2020-04-27/",
       "downloaded_at": "2023-10-02 10:13:20.621822Z",
       "sha": "3d2f3998b71863a80b24d60f6f4282e14191ae13"
+    },
+    "bristol-asthma-exacerbations.csv": {
+      "id": "bristol/asthma-exacerbations/73c4eace",
+      "url": "https://www.opencodelists.org/codelist/bristol/asthma-exacerbations/73c4eace/",
```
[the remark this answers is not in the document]
evansd: Yeah, I did wonder about that. It's certainly not as neat as having everything be from the OpenSAFLEY org. But I couldn't think of a specific problem it would create. We're linking to a fixed version so it's not like they can change it out from under us.

PR #2185, opened 2024-10-24
Plumbing for next-gen dummy data

rebkwok:
Allows users to configure dummy data to opt in to the next-gen package, by calling
```
dataset.configure_next_gen_dummy_data()
```
instead of 
```
dataset.configure_dummy_data()
```

Currently the signature is identical for the two methods; only population size is configurable.

Note that users could configure dummy data twice (as they could before); whichever definition is last is the one used.

### DIFF
--- a/docs/includes/generated_docs/language__dataset.md
+++ b/docs/includes/generated_docs/language__dataset.md
@@ -72,4 +72,18 @@ dataset.configure_dummy_data(population_size=10000)
 ```
 </div>
 
+<div class="attr-heading" id="Dataset.configure_next_gen_dummy_data">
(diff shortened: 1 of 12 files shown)


DRMacIver (approved):

evansd (approved):
Looks good to me, just a few surface level (and fully ignorable) comments.

Is `next_gen` a bit "inside baseball" as a phrase to expose to end users? I'd wondered about calling the method `configure_experimental_dummy_data()` or something. But maybe that's not any better.

Is it worth putting the warning about the experimental nature of this method at the top of the docstring and in a markdown `!!! warning` block?

Should we emphasise that not only is this not documented but that the API might change at any time and break your code and you should not it for any real research code?

PR #2187, opened 2024-10-24
Make dummy_tables_path optional for starting the ehrQL sandbox

### LINKED ISSUE #2184: Make `dummy_tables_path` optional for starting the ehrQL sandbox
- Currently for codespaces users, the command for starting the ehrQL sandbox is `opensafely exec ehrql:v1 sandbox [dummy_tables_path]`
- In the tutorial, `dummy_tables_path` is `example-data`, which is a directory already dumped into the base directory when a codespace is loaded
- For repos created from `research-template`, `example-data` is not dumped into the base directory by default. For users to use the sandbox in those repos, they would either have to call `dump-example-data` or create their own dummy tables to use in the sandbox.
- Make `dummy_tables_path` optional and default to `example-data` when not provided. If `example-data` does not exist, ask the user for permission to dump it into the current directory. 
- The alternative approach is to default to `ehrql/example-data`, which is invisible to the user. If they happen to have an `example-data` folder in their base directory, they might try to edit those csv files thinking those changes would be reflected in the sandbox. This might cause additional confusion so is not done.

alarthast:
Fixes #2184.

- Make the sandbox command default to using `example-data` if `dummy_tables_path` is not provided. 
- If `example-data` does not exist, ask to dump it in the current directory (rationale in the ticket).
- To preserve the existing syntax, `dummy_tables_path` is an optional *positional* argument, so update `cli.py` to accommodate this.
- Add a banner message at the beginning of the sandbox stating the location of the dummy tables used, now that this is not an explicit argument for starting the sandbox. This replaces the default `InteractiveConsole` banner message.

### DIFF
--- a/docs/explanation/running-ehrql.md
+++ b/docs/explanation/running-ehrql.md
@@ -30,17 +30,18 @@ by allowing you to interactively query some dummy tables.
 :computer:
 To start the sandbox, from the `learning-ehrql` directory, run:
 
-    opensafely exec ehrql:v1 sandbox example-data
+    opensafely exec ehrql:v1 sandbox
 
 You will now be in a session with an interactive Python console,
 and you should see something like this:
 
 ```pycon
-Python 3.11.3 (main, Apr  5 2023, 14:15:06) [GCC 9.4.0] on linux
-Type "help", "copyright", "credits" or "license" for more information.
-(InteractiveConsole)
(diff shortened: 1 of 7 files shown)


alarthast:
Closed since we are likely going to be replacing the sandbox with marimo notebooks

Issue #2178, opened 2024-10-22
Extend `generate_example_data` script to generate `date_of_death` and `practice_registrations`


evansd:
There is currently a very basic script on this branch which generates a minimal and simplistic `patients.csv` file:
https://github.com/opensafely-core/ehrql/blob/example-data-generator/scripts/generate_example_data.py

You can run it using:
```
python scripts/generate_example_data.py output_dir
```
Which will write CSV files into `output_dir`.

We would like to extend this script to generate more data. Eventually we would like to use it to generate all the example data which we bundle with ehrQL and use in the tutorial and the sandbox.

### 1. Adding `date_of_death` to the `patients` table

The first extension we'd like to make involves adding a randomly generated `date_of_death` to the `patients` table. There are a few key conditions we need to satisfy here:

1. Unlike `date_of_birth`, not every patient should have `date_of_death` defined. For some proportion (let's say roughly 50% to start with) `date_of_death` should be `None`.
2. `date_of_death` must always come after `date_of_birth`.
3. `date_of_death` should never be after the `TODAY` date.
4. Unlike `date_of_birth`, the date should **not** be rounded to the first of the month.
5. The `medications` dates we generate for a patient should not occur after their date of death.

### 2 Adding a `practice_registrations` table

The second extension is to add a `practice_registrations` table. This records which practice a patient was registered with when by recording the start and end date of each registration.

It has the following structure:

column | type | description
-- | -- | --
`start_date` | date | Should never be `None`. Should not be before patient’s date of birth or after patient’s date of death.
`end_date` | date | Can be `None` if it is the patient’s last registration (this indicates that is in ongoing). Can be after patient’s date of death, but never after `TODAY`.
`practice_pseudo_id` | integer | This is a meaningless number which lets us identify when patients are registered to the same practice without telling us which practice that is for privacy reasons.

We should generate, for each patient, one or more `practice_registration` records (every patient must have at least one because we wouldn't have their data if they had never been registered at a practice). For now, let's say a maximum of 10 registrations for any patient.

It's OK for the registrations to have gaps in. That is, it's OK if the `start_date` for one registration comes quite a bit later than the `end_date` of the previous registration.

Registrations shouldn't _generally_ overlap. That is, the `start_date` for one registration shouldn't come before the `end_date` of the previous registration. We might later decide to deliberately generate a small number of overlapping registrations, but we can treat that as a bonus feature.

[...]

PR #2195, opened 2024-10-29
Example data generator

Providence-o:
connected to #2178 

changes made in scripts/generate_example_data_modified.py


scripts/generate_example_data.py
```diff
@@ -58,25 +60,82 @@ def main(output_dir):
...
 
+    # Check when the patient registered in the practice
+    # set the max number of registrations per patient
+    registrations_count = random.randrange(1, 11)
+
+    # generate data for registrations
+    generated_dates = []
+    for _ in range(registrations_count):
+        practice_pseudo_id = random.choice(pseudo_id)
+        if date_of_death == None:
```
inglesp, line 101: Instead of two big blocks that are very similar, I'd do:

```
if is_dead:
    last_possible_date = date_of_death
else:
    last_possible_date = TODAY
```

and then just have one block, and do:

```
    start_date = random_date(date_of_birth, last_possible_date)
    end_date = random_date(start_date, last_possible_date)
```

scripts/generate_example_data.py
```diff
@@ -123,20 +113,23 @@ def generate_patient_data():
...
-                        break
-                if not (overlap or repeat):
+                if not date_overlap_repeat(start_date, end_date, generated_dates):
                     generated_dates.append((start_date, end_date))
                     yield "practice_registrations", {"start_date": start_date, "end_date": end_date,
                                                      "practice_id": practice_pseudo_id}
                     break
 
+# check for overlap and repeat
+def date_overlap_repeat(start_date, end_date, generated_dates):
```
inglesp, line 123: Please could you add a docstring comment giving examples of inputs where this returns True and where it returns False?

inglesp (approved):
We've had a couple of good discussions.  Thanks for making the suggested changes!

PR #2196, opened 2024-10-29
Prevent pip-compile showing absolute constraint paths

rebkwok:
If the --constraints path is included in a requirements.*.in file, the absolute path is shown in the comments in the requirements.*.txt output. This appears to be related to a pip upgrade. Specifying the --constraints path as a command line arg instead results in the expected relative paths.

We can likely change back when this issue is addressed: https://github.com/jazzband/pip-tools/issues/2131

### DIFF
--- a/justfile
+++ b/justfile
@@ -40,7 +40,7 @@ pip-compile *ARGS: devenv
 
 update-dependencies: devenv
   just pip-compile -U requirements.prod.in
-  just pip-compile -U requirements.dev.in
+  just pip-compile -U requirements.dev.in -c requirements.prod.txt
 
 # Ensure dev and prod requirements installed and up to date
 devenv: virtualenv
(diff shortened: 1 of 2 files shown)


bloodearnest (approved):
Have approved, but I am not 100% sure this change is worth it or not.

Having the absolute path is not great, but it's not really a big problem, and as you point out, will probably be fixed.

However, I think including the constraints in the .txt is cleaner than having to pass them.

But then again, we may switch to uv soonish, which will make this all moot.

rebkwok:
@bloodearnest Yeah, on second thoughts I'm not sure if it's worth it  either. Shall we just leave it with the absolute paths and assume it'll go back to the way it was at some point? Also means if you run just pip-compile on its own, you need to remember to pass the constraints path

Also no idea why coverage has decided that a line is no longer covered on this PR

Issue #1359, opened 2023-06-28
Handle duplicates in `ons_deaths`


iaindillingham:
The `ons_deaths` table contains duplicates, and so is an `EventFrame`. However, as @hendersonad [observed][1], this can be confusing.

[We discussed][2]:

* The SQL approach: Modifying `TPPBackend.ons_deaths` (i.e. replacing the `MappedTable` with a `QueryTable`)
* The ehrQL approach: Renaming `ons_deaths` to `ons_deaths_raw` and creating an `ons_deaths` that handled duplicates

I think we decided on the second option, as it's easier for a researcher to reason in ehrQL than in SQL. (Eventually, there may be non-SQL backends; it's easier for a researcher to reason in on language than in several languages.)

[1]: https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1687966907932439?thread_ts=1687434431.458539&cid=C04DVD1UQC9
[2]: https://bennettoxford.slack.com/archives/C03FB777L1M/p1678790050286699

iaindillingham:
There's more discussion in [this Slack thread][1]. @inglesp suggested making either `ons_deaths.first_for_patient` or `ons_deaths.last_for_patient` a noop, having first sorted `ons_deaths`. We could then convert it from an `EventFrame` to a `PatientFrame` later.

[1]: https://bennettoxford.slack.com/archives/C04DVD1UQC9/p1695291953108589

iaindillingham:
Changes to code, tests, and docs:

- [ ] Rename `ons_deaths` ehrQL table to `ons_deaths_raw` (`ehrql.tables.beta.core`)
- [ ] Rename `TPPBackend.ons_deaths` backend table to `TPPBackend.ons_deaths_raw` (`ehrql.backends.tpp`)[^1]
- [ ] Create `ons_deaths` ehrQL table (`PatientFrame`) and an `TPPBackend.ons_deaths` backend table (`QueryTable` using [`ROW_NUMBER`](https://learn.microsoft.com/en-us/sql/t-sql/functions/row-number-transact-sql?view=sql-server-ver16))
- [ ] Search for repos with code that will need to be updated. Is it worth opening (draft) PRs?
- [ ] Announce the change in [Platform news](https://www.opensafely.org/changelog/)

[^1]: This will remain an instance of `MappedTable`.

milanwiedemann:
A quick search on github indicates that the original ons_deaths (EventFrame) was used in these dataset definitions:

[...]

PR #2200, opened 2024-11-04
Add helper for querying `ons_deaths` "cause of death" fields

### LINKED ISSUE #1363: Add helper method for querying multiple `cause_of_death` fields

evansd:
This adds a method: `ons_deaths.cause_of_death_is_in(codelist)` which matches a codelist against the `underlying_cause_of_death` field and all 15 `cause_of_death_NN` fields.

This resurrects work which was nearly completed in #1366 but abandoned due to the complexity caused by duplicate records (which was addressed separately in #1359). I was reminded of this by the recent review of all existing ehrQL, which identified that a copy-pasted Python helper was still being used for this.

Closes #1363

### DIFF
--- a/docs/how-to/examples.md
+++ b/docs/how-to/examples.md
@@ -152,6 +152,26 @@ dataset.define_population(patients.exists_for_patient())
 
 :notepad_spiral: There are currently [multiple](https://github.com/opensafely-core/ehrql/blob/d29ff8ab2cebf3522258c408f8225b7a76f7b6f2/ehrql/tables/beta/core.py#L78-L92) cause of death fields. We aim to resolve these to a single feature in the future.
 
+
+### Finding patients with a particular cause of death
+
+The `ons_deaths` table has multiple "cause of death" fields. Using the
+[`cause_of_death_is_in()`](../reference/schemas/core.md#ons_deaths.cause_of_death_is_in)
+method we can match a codelist against all of these at once.
+
+```ehrql
+from ehrql import create_dataset, codelist_from_csv
+from ehrql.tables.core import ons_deaths, patients
+
+dataset = create_dataset()
+
+cause_of_death_X_codelist = codelist_from_csv("XXX", column="YYY")
+
+dataset.died_with_X = ons_deaths.cause_of_death_is_in(cause_of_death_X_codelist)
+dataset.define_population(patients.exists_for_patient())
+```
+
+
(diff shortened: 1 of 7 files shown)


rw251 (approved):
Looks fine to me. I'm curious why the core docs and the emis docs were updated but not the tpp ones - but I suspect the issue is my understanding.

evansd:
> I'm curious why the core docs and the emis docs were updated but not the tpp ones - but I suspect the issue is my understanding.

@rw251 No, this is in fact a very good catch! It looks like we have a separate `ons_deaths` table in the TPP schema which shadows but does not extend the core one. I'm assuming this is because we have a [place of death](https://docs.opensafely.org/ehrql/reference/schemas/tpp/#ons_deaths.place) field on the TPP table which wasn't available in EMIS. And I guess this predates the slightly shonky table inheritance mechanism that we've since added.

So this fix here is to make the TPP table extend, rather than shadow, the core table and then it will have this method defined automatically. I'll do that tomorrow.

Thanks for being curious! You've saved us a good deal of confusion and annoyance had this been deployed as is.

PR #2203, opened 2024-11-05
fix: remove structlog and just use stdlib logging

bloodearnest:
We were not using most of the features of structlog in ehrql log
outputs.  The two we were using were coloured output, which we think we
want to remove anyway, as it causes issue using logs in productions.
The other was k=v formatted extra values, which we only use in one
place.

stdlib logging can do the same job, and we just include the k=v tags
a bit more manually when needed.  This removes a dependency, which is
always nice, but also means that the ehrql sandbox can work without
*any* external dependencies, which is quite a nice property.

We may end up reintroducing coloured outputs again, but hopefully not by
embedding ANSI colour codes in log files.

We also tweak the formatting, since we can do more easily now, removing timestamps and some whitespace.

### DIFF
--- a/ehrql/utils/log_utils.py
+++ b/ehrql/utils/log_utils.py
@@ -15,7 +15,7 @@ CONFIG = {
     "formatters": {
         "formatter": {
             "()": EHRQLFormatter,
-            "format": "{asctime} [{levelname_lower:<7}] {message}",
+            "format": "[{levelname_lower:<7}] {message}",
             "datefmt": "%Y-%m-%d %H:%M:%S",
             "style": "{",
         }
--- a/ehrql/utils/log_utils.py
+++ b/ehrql/utils/log_utils.py
@@ -1,30 +1,12 @@
+import logging
 import logging.config
 import os
 
-import structlog
 
-
-pre_chain = [
-    structlog.stdlib.add_log_level,
-    structlog.stdlib.add_logger_name,
-    structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S"),
-]
-
-structlog.configure(
-    processors=[
-        structlog.stdlib.filter_by_level,
-        structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S"),
-        structlog.stdlib.add_log_level,
-        structlog.stdlib.PositionalArgumentsFormatter(),
-        structlog.processors.StackInfoRenderer(),
-        structlog.processors.format_exc_info,
-        structlog.processors.UnicodeDecoder(),
-        structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
-    ],
-    logger_factory=structlog.stdlib.LoggerFactory(),
-    wrapper_class=structlog.stdlib.BoundLogger,
-    cache_logger_on_first_use=True,
-)
+class EHRQLFormatter(logging.Formatter):
+    def format(self, record):
+        record.levelname_lower = record.levelname.lower()
+        return logging.Formatter.format(self, record)
 
 
 def init_logging():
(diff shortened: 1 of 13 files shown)


ehrql/utils/log_utils.py
```diff
@@ -34,11 +16,10 @@ def init_logging():
...
             "formatters": {
                 "formatter": {
-                    "()": structlog.stdlib.ProcessorFormatter,
-                    "processor": structlog.dev.ConsoleRenderer(
-                        pad_event=0, sort_keys=False
-                    ),
-                    "foreign_pre_chain": pre_chain,
+                    "()": EHRQLFormatter,
+                    "format": "{asctime} [{levelname_lower:<9}] {message}",
+                    "datefmt": "%Y-%m-%d %H:%M:%S",
```
madwort: wot, no timezone?

ehrql/utils/log_utils.py
```diff
@@ -59,3 +40,8 @@ def init_logging():
             },
         }
     )
+
+
+def kv(kv_pairs):
+    """Generate a string of kv pairs in space separated k=v format."""
+    return " ".join("{}={}".format(k, v) for k, v in kv_pairs.items())
```
madwort, line 48: this is covered by an existing unit test, right?

madwort (approved):

ehrql/utils/log_utils.py
bloodearnest: Executed, yes, but it wasn't being checked explicitly, due to us faking the logging in the test.  Have added a commit to use it there.

ehrql/utils/log_utils.py
bloodearnest: Also, I've been somewhat reluctant to use the Z prefix, as python doesn't parse it :(. And +NN:00 is a lot of noise.

```
>>> from datetime import datetime
>>> datetime.fromisoformat("2024-11-06 12:34:56Z")
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2024-11-06 12:34:56Z'
>>> datetime.fromisoformat("2024-11-06 12:34:56+00:00")
datetime.datetime(2024, 11, 6, 12, 34, 56, tzinfo=datetime.timezone.utc)
```

ehrql/utils/log_utils.py
evansd: I'd be tempted to stick in a very basic unit test in the existing unit test file as it's such an easy thing to test. Possibly doesn't add much, but I've found having 100% unit test coverage of the utils modules makes them easier to work on by increasing the speed of the feedback loop.

ehrql/utils/log_utils.py
evansd: I know you've said you weren't trying to alter existing behaviour, but I wonder if now is the time to think about this stuff simply because it's going to involve a visual change for users in any case (dropping the colourisation) so it might make sense to do this all at once.

What do you think about removing the timestamps entirely? It's redundant noise in the production logs where docker adds them for us. And I'm not sure they have any use locally.

bloodearnest: I am for removing them. I don't think they add value for the user, probably just noise. And it sidesteps the UTC question too.

evansd (approved):
Great! This feels like a win in terms of simplicity, and also in terms of how easy it is to control our own log output. It feels like a lot of the benefits of structlog don't really apply to ehrQL in any case in the way they would with a traditional application or service.

I was going to complain that you hadn't given any detail about the context of this decision, until I realised you had given loads! – just all hidden in the first commit. Could you yank that out into the PR body to surface it better? I'm really not a fan of `gh`'s default thing of repeating the commit titles as bulletpoints.

It's a shame our initial discussion about this was in DMs because otherwise I would have suggested linking to the Slack thread as well. I'm a big fan of leaving as many breadcrumbs/cross-links as possible for the sake of our future selves.

evansd:
@bloodearnest Oh, one more thing. Given that this involves user-facing changes we'll want a new release so one of the commits needs a pedantic prefix.

bloodearnest:
> I'm really not a fan of `gh`'s default thing of repeating the commit titles as bulletpoints.

Yeah, its not great.

But I'm not a huge fan of the details of motivation being in the merge commit and not the commit that added the change.

evansd:
> I'm not a huge fan of the details of motivation being in the merge commit and not the commit that added the change.

Having it in both is great! But the Github UI makes it pretty easy to jump from a commit to the PR which merged it, and there's always going to be context there which isn't available on the commit itself (review comments etc.) so if it _is_ only going to be in one place I'd prioritise the PR body.

evansd:
BTW, the pedantic prefix needs to be on a commit rather than the PR title (because its pedantry knows no bounds).

PR #2214, opened 2024-11-08
Add a `quiz` module with a `check_answer` function

### LINKED ISSUE #2212: Add a `quiz` module containing functionality to check answers provided by users against model answers
- See the parent issue for context.
- Our quiz questions would likely require the user to write syntax to give a dataset, frame or series. 
- We should write a function that checks whether the user's answer is correct and provide helpful error messages when the user's answer does not match the expected answer.
- After discussing with @inglesp, we would use the sandbox engine to evaluate the datasets / frames / series and check the results, as opposed to checking e.g. query graphs. 
- The above means that quiz questions should be well-designed such that it is unlikely for a user to accidentally arrive at the correct answer without understanding the syntax they wrote.

alarthast:
Closes #2212. Context provided in the ticket.

- Uses two commits from @evansd 's experimental marimo branch
- In the new `quiz` module, there is current only a `check_answer` function and the smaller functions that it calls. It is expected for `check_answer` to be used in some larger quiz framework. 

- Is a draft PR at the moment as the tests might benefit from some improvements using hypothesis.

### DIFF
--- a/ehrql/quiz.py
+++ b/ehrql/quiz.py
@@ -1,19 +1,142 @@
 from typing import Any
 
-from ehrql.query_engines.sandbox import SandboxQueryEngine
+from ehrql.query_engines.in_memory_database import PatientColumn, PatientTable
+from ehrql.query_engines.sandbox import EmptyDataset, SandboxQueryEngine
 from ehrql.query_language import BaseFrame, BaseSeries, Dataset
 
 
 def check_answer(
     engine: SandboxQueryEngine, answer: Any, expected: Dataset | BaseFrame | BaseSeries
 ) -> str:
-    if type(answer) is not type(expected):
-        return (
-            f"Expected {type(expected).__name__}, got {type(answer).__name__} instead."
-        )
+    if not isinstance(answer, type(expected)):
+        answer_type_name = get_type_name(answer)
+        expected_type_name = get_type_name(expected)
+        return f"Expected {expected_type_name}, got {answer_type_name} instead."
 
-    ev_answer = engine.evaluate(answer)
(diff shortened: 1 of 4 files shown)


DRMacIver:
The Hypothesis bits LGTM, although I'd be tempted to figure out which of the checks it can actually trigger. Maybe add asserts into each of the functions and add `@example` decorators for ones that fire?

ehrql/quiz.py
```diff
@@ -0,0 +1,264 @@
...
+
+def evaluate(
+    engine: SandboxQueryEngine, answer: Dataset | BaseFrame | BaseSeries
+) -> Any:
+    if isinstance(answer, Dataset):
+        return engine.evaluate_dataset(answer)
+    return engine.evaluate(answer)
+
+
+def check_dataset_not_empty(ev_ans: Any, ev_exp: Any) -> str | None:
```
eli-miriam, line 87: Nitpick: it wasn't immediately obvious to me what `ev_ans` and `ev_exp` meant. How annoying would it be to expand them out to `ev_answer` and `ev_expected` / `ev_expectation` (here and elsewhere)?

eli-miriam (commented):
Looks good! I like the checking utils functions in particular; it feels like a good level of parameterisation :)

eli-miriam (approved):

ehrql/quiz.py
alarthast: Done! Thanks :)

alarthast:
Merge conflicts and broken tests fixed in #2228 and brought in here by rebasing.

PR #2216, opened 2024-11-12
Give context-appropriate hint for domain mismatch errors

evansd:
Since #1873 we attempt to give a helpful error message for "domian mismatch" errors with some hints as to how to go about fixing it. However we fail to mention the possible solution of using `is_in()` which we've supported since #1795.

This is actively unhelpful for users who might reasonably assume that the options suggested are the only ones available. For example, see:
https://bennettoxford.slack.com/archives/C33TWNQ1J/p1731320032663639

This is only relevant where the problematic operation is an equality comparison, so we only add the hint in that context.

The error now looks like this:

    Cannot combine series which are drawn from different tables and both
    have more than one value per patient.

    To address this, try one of the following:

     * Use `x.is_in(y)` instead of `x == y` to check if values from
       one series match any of the patient's values in the other.

     * Reduce one series to have only one value per patient by using
       an aggregation like `maximum_for_patient()`.

     * Pick a single row for each patient from the table using
       `first_for_patient()`.

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -1068,14 +1068,26 @@ def _build(qm_cls, *args, **kwargs):
     try:
         return qm_cls(*args, **kwargs)
     except qm.DomainMismatchError:
+        hints = (
+            " * Reduce one series to have only one value per patient by using\n"
+            "   an aggregation like `maximum_for_patient()`.\n\n"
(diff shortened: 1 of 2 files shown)


Providence-o (approved):
Walked through with Peter and we're happy with this.

Issue #2217, opened 2024-11-12
Productionise VS Code extension


inglesp:
We've been prototyping a VS Code extension that helps users debug their ehrQL while they write their dataset definition.

I (Peter) am satisfied that this meets many of the needs that the existing tools (such as the sandbox) do not meet, and as such, I think it's time to commit to developing this.

This ticket covers getting the current code into a state where other people can make tweaks (such as showing just the head/tail of a large table) but does not cover making those tweaks.

There is a hard deadline of our workshop at the Symposium on 26/11, so we should we can deprioritise thorough testing in order to meet that.  However, we should not finish work on the ticket until it's tested to the standard of the rest of the codebase.

#2218 covers publishing the extension.

PR #2224, opened 2024-11-14
Add a debug ehrQL command

rebkwok:
Adds an ehrQL debug command, which uses dummy tables to alllow users to evaluate variables, tables and datasets in a dataset definition.  Called from the command line like:

```
ehrql debug <dataset definition file> --dummy-tables /path/to/dummy/tables
```

In a dataset definition, users use the new `show()` and `stop()` functions:

```
...
from erhql.debug import show, stop

age = patients.age_on(index_date)
show(age)

dataset = create_dataset()
dataset.define_population(...)
stop()

dataset.age = age
```

`show(age)` prints the contents of the age variable; `stop()` means we stop loading the dataset definition at that point and evaluate and display the dataset as it is (i.e. in the above example, the dataset only has patient_id on it as age hasn't been added yet.

Both `show` and `stop` take optional `head` and `tail` kwargs, which display on the the first and last rows respectively. When passed to `stop()` they affect the rows shown on the final output dataset.

Example terminal output:
```
Debug line 14:
patient_id        | value
------------------+------------------
1                 | 1973
2                 | 1948
...               | ...
10                | 1979
Stopping at line 21
patient_id        | age               | year
------------------+-------------------+------------------
1                 | 49                | 1973
2                 | 74                | 1948
3                 | 19                | 2003
6                 | 28                | 1994
7                 | 69                | 1953
8                 | 30                | 1992
10                | 43                | 1979
```

The command can be passed an optional `--display-format` arg, which can be either "ascii" (the default, used for terminal output) or "html" (used for the vscode extension).

Example VS code output:
Screenshot from 2024-11-12 10-41-42

### DIFF
--- a/docs/includes/generated_docs/cli.md
+++ b/docs/includes/generated_docs/cli.md
@@ -89,6 +89,14 @@ Internal command for testing code isolation support.
 Output the dataset definition's query graph
 </p>
 
+<div class="attr-heading">
+  <a href="#debug"><tt>debug</tt></a>
+</div>
+<p class="indent">
+Internal command for getting debugging information from a dataset
+definition; used by the [OpenSAFELY VSCode extension][opensafely-vscode].
+</p>
+
 </div>
 
 <div class="attr-heading" id="ehrql.help">
@@ -638,6 +646,7 @@ Database connection string.
 ```
 ehrql serialize-definition DEFINITION_FILE [--help]
       [--definition-type DEFINITION_TYPE] [--output OUTPUT_FILE]
+      [--dummy-tables DUMMY_TABLES_PATH] [--display-format RENDER_FORMAT]
       [ -- ... PARAMETERS ...]
 ```
 Internal command for serializing a definition file to a JSON representation.
@@ -667,7 +676,7 @@ show this help message and exit
   <a class="headerlink" href="#serialize-definition.definition-type" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Options: `dataset`, `measures`, `test`
+Options: `dataset`, `measures`, `test`, `debug`
 
 </div>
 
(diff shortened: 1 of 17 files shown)


evansd (approved):
This is a big leap forward in terms of our UX around developing and debugging dataset definitions. Really glad you've done this.

I think Peter's already discussed (hopefully!) some of the questions around the exact API here and the desire to unblock the tutorial work and so hopefully it's not a shock that I've pushed some commits to this branch.

The changes I've made are:
 * To make `show()` importable directly from `ehrql`, because I think we want to keep that as the primary public namespace.
 * To make `show()` accept multiple elements.
 * To document the `debug` subcommand as internal to the VSCode extension because we might want to make changes here.
 * To remove some stray files that got committed (` dataset_definition.py`, `dd.py`, `simplify_query_model.py`)

I think there's still a question about the role of `stop()` so I haven't made that part of the public namespace. If we do decide that we need this then I think there's a cleaner, exception-based implementation we could do which wouldn't require writing out temporary files. But what you've got clearly works for now.

rebkwok:
@evansd  All changes look fine to me. 

Whether we have an actual stop() thing or not, I think we do want there to be a way for people to work their way through a dataset definition bit by bit, even if later parts are broken. Maybe that just means ignoring any errors if you're running the debug command command or sending them to a different webview panel in vs code (I did that with stdout/stderr separately for a while when i was working on the extension).

We'd also need some way of using head/tail with the final output dataset. I did wonder about not showing the dataset by default, so that users have to say `show(dataset)` to see it, but that would mean if you didn't have any `show()` calls, pressing the Debug ehrQL button wouldn't output anything. Possibly that's fine, and it could have some user message so you know what happened.  For later though :)

rebkwok:
Fixes #2217

PR #2225, opened 2024-11-18
Remove all dependencies from core ehrql functionality

bloodearnest:
This means we can run the sandbox w/o pyarrow, sqlalchemy, or the various graphviz dependencies installed.

Introduces a new fixture that installs ehrql in an isolated venv to test
the dependency-not-installed code paths work.

### DIFF
--- a/ehrql/file_formats/__init__.py
+++ b/ehrql/file_formats/__init__.py
@@ -1,10 +1,6 @@
 import contextlib
 import urllib.parse
 
-from ehrql.file_formats.arrow import (
-    ArrowRowsReader,
-    write_rows_arrow,
-)
 from ehrql.file_formats.base import FileValidationError
 from ehrql.file_formats.csv import (
     CSVGZRowsReader,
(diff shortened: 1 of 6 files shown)


bloodearnest:
>? Is it definitely a pytest bug, and not something to do with the non-standard way we add xdist groups?

I think so. I don't think it should use the controller it does unless you are running with -n.   Basically, its hardcoded detection of xdist being used is wrong.

evansd (commented):
I'm slightly hesitant about merging this to be honest, because at things stand it adds complexity without giving us any direct benefit. It's good to know how we'd do this kind of optional import if we needed to, but we don't need to right now.

I'm also a bit surprised that `pyarrow` is the only troublesome dependency here. I thought we end up importing the SQLite engine inside `ehrql.main` which ends up dragging in `sqlean`? But if you've tested that this does all work in Pyodide with just `pyarrow` removed then I'm prepared to believe it.

bloodearnest:
> I'm slightly hesitant about merging this to be honest, because at things stand it adds complexity without giving us any direct benefit. It's good to know how we'd do this kind of optional import if we needed to, but we don't need to right now.

Understood. When this work was done, we were still exploring running it in pyodide, but that has been punted in favour of vscode extension, AIUI.

> I'm also a bit surprised that `pyarrow` is the only troublesome dependency here. I thought we end up importing the SQLite engine inside `ehrql.main` which ends up dragging in `sqlean`? But if you've tested that this does all work in Pyodide with just `pyarrow` removed then I'm prepared to believe it.

It's not, I have follow on branch that defers imports for some commands (sqlalchemy et al for dump_dataset_sql and graphviz et all for graph_dataset), which use the standalone venv fixture added here to test.

The mechanisms for making imports optional are quite different though - here we use the pre-existing machinery detect valid outputs, whereas we just more clumsily defer imports for the others, so I though best as separate PRs.

evansd:
> I have follow on branch that defers imports for some commands

Ah I see, great. I feel like it would be worth including those commits in this PR. In terms of the aims it's all of a piece, and it's not like this PR is so huge that it needs splitting up for reviewability.

That would mean all the work's in one place. And there's a separate question as to whether to leave it as a proof-of-viability for further down the line, or whether to actually merge – though we'd still need to decide how to resolve the pytest issue.

Issue #2221, opened 2024-11-13
Allow users to do a quiz by filling in gaps in an incomplete file


inglesp:
This is an alternative to #2168.

We can provide a quiz-like experience by giving users an incomplete file with gaps for them to fill in.  The file would look something like this:

```
from ehrql import quiz
from ehrql.tables.core import patients, clinical_events, medications

# Question 0: find the age of each patient on 1st January 2024
quiz.questions[0].check(patients.age_on("2024-01-01"))

# Question 1: find the number of medications given to each patient
quiz.questions[1].check(...)

# ...

quiz.summarise()
```

When they run the file, they would see output like:

```
Question 0: ☑
Question 1: ☒ The patient series you have provided has the wrong values for the following patients: 1, 3, 6
Question 2: ☐
Question 3: ☐

You have answered 1 question correctly, 1 incorrectly, and left 2 unanswered.
```

There is one line of output for every call to `.check()`, and one line from the call to `.summarise()`.

The incomplete file can be included in the tutorial repo, so that users can do the quiz after they've done the tutorial.

To run the file, we should provide a VS Code command for users to run from the command palette and/or a button they can click on.

We should add a script to the scripts directory that produces the incomplete file, based on a list of questions that looks something like this, and which uses the answer-checking functionality in #2214.

```
questions = [
    Question(
        "find the age of each patient on 1st January 2024",
        patients.age_on("2024-01-01"),
    ),
    ...
]
```

Writing real questions is out of scope for this ticket.

alarthast:
[Slack conversation](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1731515095239259)

Preliminary thoughts about the first steps towards implementing this
- [ ] Get the `opensafely exec ehrql:v1 quiz` command do print something (anything) to the terminal.
- [ ] Allow an optional command line argument for `quiz` assuming that most users won't bother to change the default file name we give them.
- [ ] Like `dump-example-data`, have a `dump-quiz-file` command. First duplicate then refactor.

alarthast:
Closed by #2241

PR #2230, opened 2024-11-19
Allow users to do a quiz by filling in gaps in an incomplete file

alarthast:
Addresses #2221. Context is provided in the ticket. 

- Allows a user to run the quiz in the codespace by `opensafely exec ehrql:v1 quiz`, provided that a `quiz_answers.py` file already exists. This can be done easily by adding it to the ehrQL tutorial template.
- Writing quiz questions is out of the scope for this PR. It should be noted that imports made for the user are currently hard-coded strings that needs to be adjusted with the questions. Unless it is possible to run ruff on the file upon generation, I found this to be the easiest, albeit non-ideal implementation.

### DIFF
--- a/docs/includes/generated_docs/cli.md
+++ b/docs/includes/generated_docs/cli.md
@@ -32,6 +32,20 @@ Start the ehrQL sandbox environment.
 </p>
 
 <div class="attr-heading">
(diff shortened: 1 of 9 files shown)


alarthast:
Replaced by #2241

PR #2231, opened 2024-11-19
Adding helper functions `is_alive_on` and `is_dead_on`

### LINKED ISSUE #2206: Add helper function for is_alive and/or is_dead
As part of a review of user ehrql code one of the patterns often used was whether a person was alive or dead on a particular date. This ticket is to define the implementation of a helper function like `is_alive()`.

rw251:
Closes #2206 

Adding helper functions for `is_alive_on` and `is_dead_on` to replace a very common pattern in user ehrql.

### DIFF
--- a/docs/includes/generated_docs/schemas/core.md
+++ b/docs/includes/generated_docs/schemas/core.md
@@ -521,6 +521,50 @@ return (date - patients.date_of_birth).years
   </dd>
 </div>
 
+<div markdown="block">
+  <dt id="patients.is_alive_on">
+    <strong>is_alive_on(</strong>date<strong>)</strong>
+    <a class="headerlink" href="#patients.is_alive_on" title="Permanent link">🔗</a>
+    <code></code>
+  </dt>
+  <dd markdown="block">
+Whether a patient is alive on the given date, based on the date of death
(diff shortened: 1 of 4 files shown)


evansd (approved):
This looks great, thanks Richard.

An annoying feature of the current ehrQL build process (which we keep wanting to change but not getting round to) is that in order to create a new release it requires a PR to have at least one commit with a "semantic prefix". This means a commit title with the prefix `fix:` or `feat:`. If a PR doesn't have any such commit then it won't create a new release of ehrQL – though distressingly, nothing will warn you about that beforehand.

Given that you'll need to update one of the commits anyway I'd suggest squashing the history down to a single commit, rather than reflecting the local development history in the commit sequence.

Finally, descending to the minorest of minor quibbles, I've found makes reviewing easier to separate out all the changes produced by `just generate-docs` into their own commit, separate from any handmade changes to the text. In this particular case the changes are pretty small so it doesn't really make any difference but I think it's a worthwhile habit.

rw251:
> in order to create a new release it requires a PR to have at least one commit with a "semantic prefix".

Good to know - I will almost certainly forget this a few times! 

> minorest of minor quibbles

Very happy for quibbles.

PR #2233, opened 2024-11-19
Extend `debug()` function so that related series are shown together

evansd:
By "related" we mean having the same query model domain i.e. all PatientSeries, or all EventSeries drawn from the same EventFrame.

As an example, this line:
```py
debug(patients.date_of_birth, patients.date_of_birth.year)
```

Now produces this output:
```
Debug line 16:
patient_id        | series_1          | series_2
------------------+-------------------+------------------
1                 | 1973-07-01        | 1973
2                 | 1948-03-01        | 1948
3                 | 2003-04-01        | 2003
4                 | 2007-06-01        | 2007
5                 | 1938-10-01        | 1938
6                 | 1994-04-01        | 1994
7                 | 1953-05-01        | 1953
8                 | 1992-08-01        | 1992
9                 | 1931-10-01        | 1931
10                | 1979-04-01        | 1979
```

This combined output only occurs if _all_ the arguments to `debug()` are related series. Otherwise we show the reprs one after the other as previously.


ehrql/debugger.py
```diff
@@ -37,7 +45,12 @@ def debug(
     """
     line_no = inspect.getframeinfo(sys._getframe(1))[1]
     elements = [element, *other_elements]
-    element_reprs = [repr(el) for el in elements]
+
+    if hasattr(element, "__repr_related__") and elements_are_related_series(elements):
+        element_reprs = [element.__repr_related__(*other_elements)]
```
rebkwok, line 50: It would be good to separate out elements that don't have a `"__repr__related__"` method before this; then e.g. `debug("foo", patients.date_of_birth, patients.date_of_death`)` can still put together the patient table.

rebkwok (approved):

ehrql/debugger.py
evansd: Yeah, I originally suggested something similar but Peter liked the simplicity of the all-or-nothing approach. I think we can safely _expand_ the range of circumstances in which we display related series later without worrying about backwards compatibility. So I guess it makes sense to start small for now.

Issue #2232, opened 2024-11-19
Add helper function for `is_registered_on`


rw251:
A helper function so users can quickly filter to only people with an active practice registration.

rw251:
After discussion (https://bennettoxford.slack.com/archives/C069YDR4NCA/p1732015213840789) decided to implement `practice_registrations.exists_for_patient_on(date)`

PR #2234, opened 2024-11-19
Add `exists_for_patient_on` helper for `practice_registrations`

rw251:
`practice_registrations.for_patient_on(date).exists_for_patient()` is a common filter when defining a cohort. Simplifying this to `practice_registrations.exists_for_patient_on(date)`.

### DIFF
--- a/ehrql/tables/core.py
+++ b/ehrql/tables/core.py
@@ -142,6 +142,12 @@ class practice_registrations(EventFrame):
         )
         return ordered_regs.last_for_patient()
 
+    def exists_for_patient_on(self, date):
+        """
+        Returns whether a person was registered with a practice on the supplied date.
+        """
+        return self.for_patient_on(date).exists_for_patient()
+
     def spanning(self, start_date, end_date):
         """
         Filter registrations to just those spanning the entire period between
--- a/ehrql/tables/core.py
+++ b/ehrql/tables/core.py
@@ -145,8 +145,14 @@ class practice_registrations(EventFrame):
     def exists_for_patient_on(self, date):
         """
         Returns whether a person was registered with a practice on the supplied date.
+
+        NB. The implementation currently uses `spanning()`. It would also have been
+        valid to implement as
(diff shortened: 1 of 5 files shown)


ehrql/tables/core.py
```diff
@@ -142,6 +142,12 @@ def for_patient_on(self, date):
         )
         return ordered_regs.last_for_patient()
 
+    def exists_for_patient_on(self, date):
+        """
+        Returns whether a person was registered with a practice on the supplied date.
+        """
+        return self.for_patient_on(date).exists_for_patient()
```
evansd, line 149: There's no way you could have known this, but there's actually a less obvious but more efficient alternative construction we can use here:
```suggestion
        return self.spanning(date, date).exists_for_patient()
```

The reason for this is that the implementation of `for_patient_on()` uses `last_for_patient()` which will end up selecting a specific registration for each patient, whereas all we want to do is identify whether any appropriate registration exists at all.

There's no reason in principle that we couldn't spot the pattern whereby a first/last selection isn't actually being used and, internally, rewrite the query to do something more efficient. We already a have [a stage](https://github.com/opensafely-core/ehrql/blob/5a730b5ad451cc08be82bca361714da8af12d4eb/ehrql/query_model/transforms.py) in the query processing pipeline where these kind of transformations happen, we just haven't implemented this one yet.

Given that I'm not sure when we _will_ implement it, and given how frequently we expect this to be used, I think it would be worth going with the more efficient option for now.

When we do implement the optimisation we could change the implementation to your more natural one, which I agree is preferable from a user-comprehensibility viewpoint.

evansd (approved):

ehrql/tables/core.py
rw251: Thanks. I agree efficiency should win out here. But to be clearer to any users looking at the implementation in the docs, I've added a bit that explains that `for_patient_on` would also have worked but been less efficient.

rw251:
closes #2232

PR #2235, opened 2024-11-19
fix: Improve rendering of aggregations in the sandbox

inglesp:
Previously, if you filtered an event series such that not every patient had a record in the new series, then aggregating that series (with eg .exists_for_patient() or .count_for_patient()) would not show those patients in the output:

    >>> clinical_events.where(clinical_events.date >= "2016-01-01").exists_for_patient()
    patient_id        | value
    ------------------+------------------
    2                 | True
    3                 | True
    4                 | True
    5                 | True
    6                 | True
    7                 | True
    8                 | True

But now:

    >>> clinical_events.where(clinical_events.date >= "2016-01-01").exists_for_patient()
    patient_id        | value
    ------------------+------------------
    1                 | False
    2                 | True
    3                 | True
    4                 | True
    5                 | True
    6                 | True
    7                 | True
    8                 | True
    9                 | False
    10                | False

This makes things easier to explain in the tutorial, and is probably less confusing in sandbox environments.

### DIFF
--- a/ehrql/query_engines/in_memory_database.py
+++ b/ehrql/query_engines/in_memory_database.py
@@ -349,10 +349,9 @@ class EventColumn:
         )
 
     def filter(self, predicate):  # noqa A003
-        patient_to_rows = {
-            p: rows.filter(predicate[p]) for p, rows in self.patient_to_rows.items()
-        }
-        return EventColumn({p: rows for p, rows in patient_to_rows.items() if rows})
+        return EventColumn(
+            {p: rows.filter(predicate[p]) for p, rows in self.patient_to_rows.items()}
+        )
 
     def sort_index(self):
         return EventColumn(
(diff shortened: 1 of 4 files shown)


rw251:
It took me a while to figure out the "-1" in `pick_at_index(-1)` in the tests but I got there in the end - and it's clearly because I'm relatively new to the code base. If we can't import `qm.Position.FIRST`, `qm.Position.LAST` into the test, maybe a quick comment explaining what the -1 is? But perhaps that's overkill.

Also any reason to just test `LAST` and not `FIRST` as well in the test?

rw251 (approved):

ehrql/query_engines/in_memory_database.py
```diff
@@ -352,7 +352,7 @@ def filter(self, predicate):  # noqa A003
         patient_to_rows = {
             p: rows.filter(predicate[p]) for p, rows in self.patient_to_rows.items()
         }
-        return EventColumn({p: rows for p, rows in patient_to_rows.items() if rows})
+        return EventColumn({p: rows for p, rows in patient_to_rows.items()})
```
evansd, line 355: This dict comprehension is redundant because it's just recreating the input dict. I think it would be clearer to write the whole function as a single line which then mirrors what we do elsewhere:
```suggestion
        return EventColumn(
            {p: rows.filter(predicate[p]) for p, rows in self.patient_to_rows.items()}
        )
```

evansd (approved):
For a while I couldn't understand how this could possibly be correct, but now I'm convinced it is correct. I think what confused me is that we filter a table by filtering each of its columns, which we do by filtering each of the rows for each patient in the column.  So all the actual filtering is being done by the bottom layer, and we're just removing a check in the middle layer which used to discard patients who had no remaining rows.

This is fully inline with the query model semantics so I'm happy with it. If it ends up triggering bugs then those are bugs elsewhere in the system and not with this change.

PR #2236, opened 2024-11-19
Render DateDifferences in sandbox

inglesp:


tests/unit/test_sandbox.py
```diff
@@ -7,11 +7,11 @@
 
 
 user_input = """
-from ehrql import Dataset
+from ehrql import create_dataset
```
evansd: Thanks. We need to do a clean-sweep of the codebase for instances of this at some point because they occur quite frequently in the tests.

ehrql/sandbox.py
```diff
@@ -16,6 +16,7 @@ def run(dummy_tables_path):
     # Overwrite __repr__ methods to display contents of frame/series.
     BaseFrame.__repr__ = lambda self: repr(engine.evaluate(self))
     BaseSeries.__repr__ = lambda self: repr(engine.evaluate(self))
+    DateDifference.__repr__ = lambda self: repr(engine.evaluate(self.days))
```
evansd, line 19: It might be helpful to add the units here to make it more obvious that this is a different kind of value from a simple integer.

You could do that using something like this:
```py
DateDifference.__repr__ = lambda self: repr(
    format_column(engine.evaluate(self.days), "{} days")
)

```

With `format_column` defined as:
```py
from ehrql.query_engines.in_memory_database import apply_function


def format_column(column, template):
    return apply_function(template.format, column)
```

evansd (approved):
A nice improvement, thanks.

I've realised that this won't apply to the `debug()` command because when I did the refactoring there I'd forgotten that it was copy-pasted from the sandbox and not shared with it.

I think the answer is to replace the monkey-patching code in the sandbox with the `activate_debug_context` context manager as used in `ehrql/loaders.py`:
https://github.com/opensafely-core/ehrql/blob/5a730b5ad451cc08be82bca361714da8af12d4eb/ehrql/loaders.py#L266-L269

And then move your changes to the `debugger` module.

inglesp:
Closing in favour of #2240.

PR #2240, opened 2024-11-20
Render DateDifferences when debugging

inglesp:
Replaces #2236.

### DIFF
--- a/tests/unit/test_debugger.py
+++ b/tests/unit/test_debugger.py
@@ -1,5 +1,6 @@
 import json
 import textwrap
+from datetime import date
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/unit/test_debugger.py
```diff
@@ -246,12 +255,14 @@ def test_repr_related_patient_series(dummy_tables_path, capsys):
 def test_repr_related_event_series(dummy_tables_path, capsys):
     with activate_debug_context(
         dummy_tables_path=dummy_tables_path,
-        render_function=lambda value: json.dumps(list(value), indent=4),
+        render_function=lambda value: json.dumps(
+            list(value), indent=4, default=date_serializer
```
evansd, line 259: Ughh, sorry about this. I only used JSON here because it gave nicer output than either `repr()` or `pprint.pformat()`.

evansd (approved):

PR #2241, opened 2024-11-20
Support quizes as externally managed files

evansd:
This builds on the work in #2230 but moves the definition of the quiz itself, and the mechanism for executing it, outside of ehrQL.

Instead, the plan is that we define the quiz in a standalone Python file e.g. `erhql_quiz_answers.py`:
```py
from ehrql import create_dataset
from ehrql.quiz import Question, Questions
from ehrql.tables.core import clinical_events, patients


introduction = """\
Welcome to the ehrQL Quiz!
"""


questions = Questions()
questions.set_dummy_tables_path("dummy_tables")

questions[1] = Question(
    """
    Add column `age` to the dataset corresponding to the patient's age on January 1, 2024.
    """,
)
questions[1].expected = create_dataset()
questions[1].expected.age = patients.age_on("2024-01-01")


questions[2] = Question(
    """
    Filter the clinical events table to only include records with SNOMED code `60621009`.
    """,
)
questions[2].expected = clinical_events.where(
    clinical_events.snomedct_code.is_in(["60621009"])
)
```

We also provide a command `scripts/generate_quiz_from_answers.py` which takes a file like the one above and generates an empty template quiz file ready to completed e.g.
```py
# Welcome to the ehrQL Quiz!

from ehrql_quiz_answers import questions

from ehrql import create_dataset, debug
from ehrql.tables.core import clinical_events, patients


# Question 1
# Add column `age` to the dataset corresponding to the patient's age on January 1, 2024.
dataset = create_dataset()
dataset.age = patients.age_on("2000-01-01")
debug(dataset.age)
answer = dataset
questions[1].check(answer)

# Question 2
# Filter the clinical events table to only include records with SNOMED code `60621009`.
answer = ...
questions[2].check(answer)

questions.summarise()
```

This file can then be run via our VSCode extension, directly via the `debug` command, or just as a vanilla Python script.

Both the template quiz and the answers file would need to be committed alongside each other, as well as the relevant dummy tables.

To avoid users spoiling their learning experience by accidentally looking at the answers file I suggest adding a header like:
```py
################################################################
#                                                              #
#  IF YOU'RE LOOKING FOR THE EHRQL QUIZ THIS IS THE WRONG FILE #
#                                                              #
################################################################
#
# The file you want is called `ehrql_quiz.py`.
#
# This file defines the ehrQL quiz questions and provides the answers. If you don't want
# to spoil the quiz, don't look ahead!
#
#
#
```
(Followed by enough comment lines to shunt the code well off screen.)


tests/unit/test_quiz.py
```diff
@@ -340,3 +340,12 @@ def test_questions():
...
     assert questions[2].engine.dsn.name == "test_dummy_path"
+
+
+def test_set_dummy_tables_path_in_debug_context():
+    with debugger.activate_debug_context(
+        dummy_tables_path="foo", render_function=lambda v: v
+    ):
+        questions = quiz.Questions()
+        questions.set_dummy_tables_path("bar")
+        assert debugger.DEBUG_QUERY_ENGINE.dsn.name == "bar"
```
inglesp, line 351: Worth an assert that `debugger.DEBUG_QUERY_ENGINE` gets reset outside the context manager?

inglesp (approved):

PR #2255, opened 2024-11-27
Better distribution of number of events

DRMacIver:
This solves a couple of problems (as witnessed in the tests):

1. Our distribution of number of event rows creates a very skewed distribution for binary properties based on whether an event with some code exists.
2. If our population definition depends on some event existing (or not existing) we may not do a very good job of data generation.

The first is fixed by replacing the uniform distribution with a geometric one. This biases dummy data much more towards having small number of events per patient.

Set against that we also do some work to figure out if particular tables seem to be needed (or forbidden - that usually shouldn't happen in real queries I think, but might if we're bad at generating the right data) and modifies the row count distribution in future iterations if so.

### DIFF
--- a/ehrql/dummy_data_nextgen/generator.py
+++ b/ehrql/dummy_data_nextgen/generator.py
@@ -318,8 +318,9 @@ class DummyPatientGenerator:
         if table_info.has_one_row_per_patient:
             row_count = self.rnd.randint(0, 1)
         else:
-            # Geometric distribution with parameter 0.25. Will 3 events per patient.
-            row_count = math.floor(math.log(self.rnd.random()) / math.log(1 - 0.25))
+            # Geometric distribution with parameter 0.2. Will average 4 (=1/0.2 - 1) events
+            # per patient.
+            row_count = math.floor(math.log(self.rnd.random()) / math.log(1 - 0.2))
             if self.required_tables and table_info.name in self.required_tables:
                 row_count += 1
         return [{} for _ in range(row_count)]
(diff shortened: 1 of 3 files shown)


ehrql/dummy_data_nextgen/generator.py
```diff
@@ -106,6 +111,25 @@ def get_data(self):
                 if found >= self.population_size:
                     break
 
+            if generator.required_tables is None and valid_patient_ids:
```
rebkwok, line 114: Just trying to understand what's going on here; in the first batch of patients generated, we go through all of the tables in the database, and built a dict which has keys patient_id and values a set of tables that that patient appears in. Then we loop through all the patient IDs and update `required_tables` and `forbidden_tables` to be all tables that at least one patient has an entry in, and all tables that no patients have any entries in respectively.

And then on line 324 we use the existence of the required tables built in this step, so on subsequent batches of generated patient IDs, if the table is in required tables, the row count for event-level tables is increased by 1.

Tables in the database are populated by the tables in the population query (other tables that are needed for the dataset by aren't required for determining population membership are added later, but not added to the database), so if a tables isn't required for the population query, it won't be in the required tables and it will have a lower event-level row count.

I'm struggling to understand why we're doing this though.

rebkwok: Actually, maybe I do understand. The geometric distribution can be 0, so if the table is required for the population inclusion, it ensures that the rows are at least 1.

But this isn't the case on the first batch, is it? Because then self.required_tables is None, so if the row count is calculated as 0, it won't be incremented.

rebkwok (commented):
I've added some comments which are more just me trying to understand why the code is doing what it's doing. I may just have completed misunderstood it, but it might help see where I've misunderstood

rebkwok (approved):

PR #2256, opened 2024-11-28
Revert buggy in-memory engine change

evansd:
This reverts the changes in #2247 which caused the in-memory engine to give incorrect results. As the in-memory engine is used for dummy data generation I don't think we can afford to live with this.

Fixing this would involve having some mechanism whereby a PatientFrame derived from an EventFrame by picking rows can distinguish patient_ids which are actually present in the frame from those which exist just in order to give less confusing results in the sandbox/debug context.

The bug was originally [discovered](https://github.com/opensafely-core/ehrql/actions/runs/12050788107/job/33600394222) by the overnight generative tests (thanks Hypothesis!).

Slack discussion:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1732729097689479

### DIFF
--- a/ehrql/query_engines/in_memory_database.py
+++ b/ehrql/query_engines/in_memory_database.py
@@ -208,7 +208,7 @@ class EventTable:
     def pick_at_index(self, ix):
         return PatientTable(
             {
-                name: col.pick_at_index(ix, name == "patient_id")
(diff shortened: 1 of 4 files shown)


inglesp (approved):
Thanks for adding the regression test

PR #2261, opened 2024-11-28
Added 2 vscode configs for local debugging

rw251:
VSCode users will now see two extra debug options:
1. Debug the currently open test file
2. Debug the doc generation process
 
image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,21 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug: Current test file",
+      "type": "debugpy",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": "0.2.0",
```
inglesp, line 2: What's the significance of the version number?

inglesp (approved):

.vscode/launch.json
rw251: That's the version of the vscode config language i.e. what defines all the keys in the json. It gets autogenerated when you first create the config file.

PR #2263, opened 2024-11-28
Render debugged DateDifference elements with other related series

rebkwok:
DateDifference nodes are not themselves a Series. However, we
render them as the repr of their .days method, which IS a Series.
In order to determine whether a DateDifference is related to other
debugged elements, we need to compare it as days.

We also move the decision about what to render to the sandbox
query engine's evaluate method.

### DIFF
--- a/ehrql/debugger.py
+++ b/ehrql/debugger.py
@@ -5,7 +5,6 @@ import sys
 from ehrql.query_engines.in_memory_database import (
     EventColumn,
     PatientColumn,
-    apply_function,
 )
 from ehrql.query_engines.sandbox import SandboxQueryEngine
 from ehrql.query_language import BaseFrame, BaseSeries, Dataset, DateDifference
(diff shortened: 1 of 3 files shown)


evansd (commented):
I think there might be a much simpler factoring here which would be to teach `SandboxEngine.evalute()` how to handle `DateDifference` objects i.e. calling `evaluate(date_diff.days)` and using `apply_function()` to add the units.

In fact, I think it would also make sense to have `evaluate()` handle `Dataset` instances as well and do the right thing there.

That would mean the `__repr__` method for every type becomes the same:
```py
def repr_ehrql(obj):
    return query_engine.evaluate(obj)._render_(
        render_function
    )

BaseFrame.__repr__ = repr_ehrql
BaseSeries.__repr__ = repr_ehrql
DateDifference.__repr__ = repr_ehrql
Dataset.__repr__ = repr_ehrql
```

rebkwok:
@evansd Thanks - I've done as you suggested, and it has made even more things simpler that I originally thought. There's no need to keep track of which columns are date diff ones if `evaluate()` does the unit formatting.

evansd (approved):
Great! Glad that actually worked. I'm always a _bit_ nervous when I suggest a simpler approach that there'll be something silly I missed which means it won't work.

There might be some stuff you can delete from the PR description now.

I also missed last time round that it needs a pedantic commit prefix :disappointed:

rebkwok:
Ah, yes - I updated the commit message but not the pr description. And bah, semantic commit again, thanks

PR #2265, opened 2024-11-29
More principled RNG seeding in dummy data generation

DRMacIver:
The way that the dummy data generation just arbitrarily calls `seed` on its global random number generator was *stressing me out* because of how easy it makes it to inadvertently introduce bugs, so this replaces it with a more principled approach where we use a context manager to control the seed and make sure we are always seeding correctly.

### DIFF
--- a/ehrql/dummy_data_nextgen/generator.py
+++ b/ehrql/dummy_data_nextgen/generator.py
@@ -4,6 +4,7 @@ import logging
 import random
 import string
 import time
+from contextlib import contextmanager
 from datetime import date, timedelta
 
 from ehrql.dummy_data_nextgen.query_info import QueryInfo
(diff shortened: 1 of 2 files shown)


ehrql/dummy_data_nextgen/generator.py
```diff
@@ -160,21 +178,21 @@ def get_patient_data(self, patient_id, table_names):
...
+            with self.seed(f"{patient_id}:{name}"):
+                table_info = self.query_info.tables[name]
+                # Support specialised generators for individual tables, otherwise just make
+                # some empty rows
+                get_rows = getattr(self, f"rows_for_{table_info.name}", self.empty_rows)
+                rows = get_rows(table_info)
+                for row in rows:
+                    # Fill in any values that haven't already been set by a specialised
+                    # generator
+                    self.populate_row(table_info, row)
```
evansd, line 190: It might be clearer to just have the context manager wrap the section of code where it's used, which I think are just these lines here.

ehrql/dummy_data_nextgen/generator.py
```diff
@@ -187,42 +205,42 @@ def get_patient_column(self, column_name):
     def generate_patient_facts(self, patient_id):
         # Seed the random generator using the patient_id so we always generate the same
         # data for the same patient
-        self.rnd.seed(f"{self.random_seed}:{patient_id}")
-        # TODO: We could obviously generate more realistic age distributions than this
+        with self.seed(patient_id):
```
evansd, line 208: Given that this method only has a single call site it might make sense to apply the context manager there and not have to indent the entire method body.

evansd (approved):
Ah yes, great. This is obviously a good move now I see it. Thanks!

ehrql/dummy_data_nextgen/generator.py
DRMacIver: I think in general I prefer the functions to be callable without having to do anything special at the callsite.

I did wonder about making this a decorator rather than a context manager, which would remove the indentation concern.

evansd: Yes, that's fair enough. I'd leave it as is then.

PR #2270, opened 2024-12-03
Dummy patient demographics

DRMacIver:
This splits dummy patient generation up into two stages: First we generate a "demographic", essentially a set of allowed values for each column, and then we generate a patient from this demographic.

At present this doesn't do a huge amount and is basically a way of getting a slightly better distribution of some properties (e.g. it makes it easier to generate who have one but not both of two conditions).

This is mostly laying groundwork for future work where we can look at what demographics make it into the final dummy data and use that to boost demographics that are likely to satisfy the population query and penalise demographics that aren't.

### DIFF
--- a/ehrql/dummy_data_nextgen/generator.py
+++ b/ehrql/dummy_data_nextgen/generator.py
@@ -28,6 +28,46 @@ CHARS = string.ascii_letters + string.digits + ".-+_"
 get_regex_generator = functools.cache(create_regex_generator)
 
 
+class PopulationSubset:
+    """A population subset consists of some group of patients all with
+    "similar" characteristics. These don't really correspond to
+    any particularly natural category among patients, and are instead an idea
+    borrowed from a concept called "swarm testing". The idea is to
+    first draw a random population_subset, and then draw patients from the
+    conditional distribution on that population_subset. This allows us to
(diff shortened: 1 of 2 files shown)


rebkwok (approved):
I think this all makes sense to me - thanks for the explanatory comments. A minor naming nit - "demographic" to me means characteristics of the patient (age, sex, ethnicity etc) and study code often refers to things like "demographic variables". I had some initial confusion because in my head, "demographic" wouldn't be something like clinical events (helped by the example in the doc string being asthma and diabetes, but I had to shift my mental model of it when I got to that bit of the description). I'm not sure if there is a better name for this thing though

DRMacIver:
> A minor naming nit - "demographic" to me means characteristics of the patient (age, sex, ethnicity etc) and study code often refers to things like "demographic variables". I had some initial confusion because in my head, "demographic" wouldn't be something like clinical events (helped by the example in the doc string being asthma and diabetes, but I had to shift my mental model of it when I got to that bit of the description). I'm not sure if there is a better name for this thing though

Hmm. Yeah, that makes sense. How about just giving it some awkward name like `PopulationSubset` or the like?

I've called the generic term for this a `Configuration` which I don't love but could also go somewhere in that direction.

PR #2272, opened 2024-12-04
Changed function name from debug to show (#2268)

Providence-o:
* https://github.com/opensafely-core/ehrql/issues/2268


rebkwok (commented):
I would prefer that we keep the names of the internal things (the cli command `debug_dataset_definition`,  `activate_debug_context`, `load_debug_definition`) - so we're just changing the user-facing things. I think it's clearer to developers what `debug` means (and the module is still `debugger.py`), and it also means we don't have to change it in the vs code extension

madwort:
hmm, I was thinking we should also change the internal things, as I find blocks like this loader annoying when things were a consistent pattern & become _nearly_ a pattern. but I don't feel too strongly about it, as you say a developer should be able to cope with this. https://github.com/opensafely-core/ehrql/blob/a56ce7d0fc77822d900269756249aa44c8a73973/ehrql/loaders.py#L307-L312

evansd:
Thanks Providence. I think I'm with Becky here in that "show" doesn't really make sense as the name of the command or the loader functions so it's best to keep them called "debug". I'd describe the situation as `show()` being the function you call while you're debugging, although we don't require the user to know that the thing they're doing is called "debugging".

@madwort To your point, I think what Becky's suggesting would maintain the consistency there as the CLI command would still be called `debug`. Keeping that is what allows us to not change the VSCode extension.

tests/unit/test___main__.py
```diff
@@ -299,8 +299,8 @@ def test_backend_from_id_special_case_aliases(alias):
     assert backend_from_id(alias) is None
 
 
-def test_debug(capsys, tmp_path):
-    # Verify that the debug subcommand can be invoked.
+def test_show(capsys, tmp_path):
+    # Verify that the show subcommand can be invoked.
```
rebkwok, line 303: This one should stay as `test_debug` (and the comment referring to the `debug` subcommand - the subcommand is still debug (called on line 322)

tests/unit/test___main__.py
```diff
@@ -329,7 +329,7 @@ def test_debug(capsys, tmp_path):
     assert captured.out == ""
 
 
-def test_debug_rejects_unknown_display_format(capsys, tmp_path):
+def test_show_rejects_unknown_display_format(capsys, tmp_path):
```
rebkwok, line 332: This one also should stay as `test_debug...`

rebkwok (approved):
Thanks @Providence-o . There's a couple left in tests that need to be changed back, otherwise LGTM

Issue #1402, opened 2023-07-07
DRAFT: Proposal for spec tests revamp


evansd:
:warning: These are unfinished notes, which I wanted to record somewhere.

---

We should think of the spec tests as documentation that happens to be executable as test code, rather than test code from which we try to generate documentation. This means we should start by thinking about what makes most sense and is clearest from the point of view of documentation and then work out how to execute it.

It may be that some examples can only be executed on the in-memory engine because they rely on returning non-patient-level data, but that's OK – we can use standard integration tests to cover these elsewhere.

I think the tests divide into three broad classes, two easy cases and a ragbag of more awkward stuff:

 1. Tests for functions i.e. things which take a bunch of series and return a new series – everything in the `Function` query model namespace.
 2.  Tests for series aggregations i.e. things which take a many-rows-per-patient series and return a one-row-per-patient series.
 3. Everything else, which I think means `where`, `except_where`, `sort_by`, `first|last_for_patient` and the two frame-level aggregations: `exists_for_patient()` and `count_for_patient()`.

Fortunately the two easy cases (1 and 2) are the most numerous, and also the areas where we're likely to continue to add new operations. So if we have a good story for these it doesn't matter so much if we have to duplicate some of the other cases as both documentation examples (which are only checked against the in-memory engine) and more traditional integration tests.

### 1. Tests for functions

We can express these in a tabular form, for example:

`(d1 – d2).years` | `d1` | `d2`
-- | -- | --
0 | 2000-06-01 | 2000-01-01
1 | 2001-01-01 | 2000-01-01
-1 | 1999-12-31 | 2000-01-01

`b1 \| b2` | `b1` | `b2`
-- | -- | --
T | T | T
T | T | F
F | F | F
T | T | NULL
NULL | F | NULL
NULL | NULL | NULL

(Note I'm using explicit `NULL`s here rather than empty cells because I think that's clearer.)

I don't know whether we'd write these as Python data structures and generate markdown from them, or write them as markdown and parse them into something executable. But neither option seems _particularly_ challenging.

### 2. Tests for series aggregations

We can take a similar tabular approach here, with the complication that we need to indicate that there a multiple values in a given column. I'm using subscripts here to indicate that. If anyone can think of a  clearer way of doing this I'd be very interested!

`i1.sum_for_patient()` | `i1` <sub>1</sub> | `i1` <sub>2</sub> | `i1` <sub>3</sub>
-- | -- | -- | --
6 | 1 | 2 | 3
[...]

PR #2274, opened 2024-12-04
Add support for `ICD10MultiCodeString` type

### LINKED ISSUE #2010: Better validation for "list of codes shoved in a string" columns
We have some columns containing codes which we have to declare as type `str` rather than their coding system (e.g `ICD-10`) because they contain multiple codes shoved into a single column. For example:
https://docs.opensafely.org/ehrql/reference/schemas/tpp/#apcs.all_diagnoses

We query these columns using the `.contains()` method, which takes strings. But we still want to ensure that the strings supplied are valid for the coding system in question. The quick hacky way I did this when writing the original proof-of-concept study was to construct a code object and use the internal `_to_primitive_type()` method to get the string back out of the code object:
```py
    for code in codelist:
        # Pass the string through the ICD10Code to constructor to validate that it has
        # the expected format
        code_string = ICD10Code(code)._to_primitive_type()
        code_strings.add(code_string)
```

Annoyingly (but also as a positive testament to the code sharing that goes on) this has propagated itself into [various research codebases](https://github.com/search?q=org%3Aopensafely%20_to_primitive_type()&type=code).

At a minimum we should provide a non-private and less ugly way to do this. One simple way would be to define a `__str__` method on `BaseCode` so we could rewrite the above as:
```py
    for code in codelist:
        # Pass the string through the ICD10Code to constructor to validate that it has
        # the expected format
        code_string = str(ICD10Code(code))
        code_strings.add(code_string)
```

However that still leaves the study author with the responsibility to use the correct coding system when checking the codes, which is exactly what ehrQL is supposed to avoid.

The long-term solution here is [proper handling](https://github.com/opensafely-core/ehrql/issues/574) of multi-valued columns – but that's a whole lot of thinking and work and I don't propose to tackle that now.

I wonder if there's a simpler improvement here which is to define an explicit type for "bunch of codes from coding system X shoved in a string". That could catch a couple of things:
 1. Prevent equality comparisons between the string and a single code so you can't accidentally write:
    `list_of_codes_field == single_code`
 3. Ensure that the arguments to `.contains()` queries are of a format that could potentially match something in the field.

Note that for 2 we shouldn't require that the argument to `.contains()` is a complete valid code because it may just be a prefix to a code: for lexically hierarchical coding systems like ICD-10 it's useful to be able to match codes like `N171` using just the prefix `N17`.

Related Slack threads:
https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1715334831886929
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1715344398345399

rw251:
fixes #2010 

Notes
- I've currently just implemented this for the `all_diagnoses` field of `apcs`. If successful I will then also implement the `all_procedures` field as an `OPCS4MultiCodeString`.
- We don't want people to do something like `all_diagnoses == "N170"`, which would run previously (but return no results). I'm overriding `__eq__ ` (and `__ne__`) so that the error is more helpful. I've also updated the docstrings so it's clear that it doesn't make sense to use these operators.
- There are other methods that are inherited which don't make sense for this field type e.g. `is_in`, `map_values`, and `count_distinct_for_patients`. They all appear in the docs (reference/language/#MultiCodeStringEventSeries). I could have overriden them as well, but thought a better solution (in the future) would be to have a better way to hide, or remove, these methods in some way
- I don't understand the serializer.py file enough to know if my changes just made the tests pass. I wasn't sure if I needed a method with a `@marshal.register(BaseMultiCodeString)` decorator - or if the fact that it's just a string means that it's fine.

### DIFF
--- a/ehrql/codes.py
+++ b/ehrql/codes.py
@@ -142,6 +142,10 @@ class ICD10MultiCodeString(BaseMultiCodeString):
     def _code_type(cls):
         return ICD10Code
 
+    # We want to allow prefix searching, so when we check the regex we
+    # want to account for that
+    regex = re.compile(r"[A-Z][0-9]{0,3}")
+
 
 def codelist_from_csv(filename, *, column, category_column=None):
     """
--- a/ehrql/codes.py
+++ b/ehrql/codes.py
@@ -30,6 +30,26 @@ class BaseCode:
         return self.value
 
 
+# A base class for fields that are concatenated lists of clinical codes. This occurs
+# in the admitted patient care spell (apcs) table of hospital episode statistics for
+# all_diagnoses (ICD10 codes), and all_procedures (OPCS4 codes).
+#
+# This inherits from str because that's what the underlying data is, but is in this
+# file as it's sort of a code. In future a better implementation might be to parse the
+# field value into a Set of clinical codes.
+@dataclasses.dataclass(frozen=True)
+class BaseMultiCodeString(str):
+    @classmethod
+    def _code_type(cls):
+        raise NotImplementedError(
+            "BaseMultiCodeString subclasses must implement the _code_type method"
+        )
+
+    @classmethod
+    def _primitive_type(cls):
+        return str
+
+
 class BNFCode(BaseCode):
     "Pseudo BNF"
 
@@ -106,6 +126,21 @@ class DMDCode(BaseCode):
 
     # Syntactically equivalent to SNOMED-CT
     regex = SNOMEDCTCode.regex
+
+
+#
+# ICD10 codelist as concatenated string
+#
+# This is specifically for fields in the admitted patient care (APC) part
+# of the hospital episode statistics (HES) data where there are fields
+# that are a concatenation of all diagnosis codes for a patient's episode
+# or spell.
+class ICD10MultiCodeString(BaseMultiCodeString):
+    "Multiple ICD-10 codes"
+
+    @classmethod
+    def _code_type(cls):
+        return ICD10Code
 
 
 def codelist_from_csv(filename, *, column, category_column=None):
--- a/ehrql/codes.py
+++ b/ehrql/codes.py
@@ -37,7 +37,6 @@ class BaseCode:
 # This inherits from str because that's what the underlying data is, but is in this
 # file as it's sort of a code. In future a better implementation might be to parse the
 # field value into a Set of clinical codes.
-@dataclasses.dataclass(frozen=True)
 class BaseMultiCodeString(str):
     @classmethod
     def _code_type(cls):
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -1038,6 +1038,20 @@ class CodeEventSeries(CodeFunctions, EventSeries):
 
 
 class MultiCodeStringFunctions:
+    def _cast(self, value):
+        code_type = self._type._code_type()
+
+        if isinstance(value, code_type):
+            # The passed code is of the expected type, so can convert to a string
+            return value._to_primitive_type()
+        elif isinstance(value, str) and self._type.regex.fullmatch(value):
+            # A string that matches the regex for this type
+            return value
+        else:
+            raise TypeError(
+                f"Expecting a {code_type}, or a string prefix of a {code_type}"
+            )
+
     def __eq__(self, other):
         """
         This operation is not allowed because it is unlikely you would want to match the
@@ -1073,13 +1087,7 @@ class MultiCodeStringFunctions:
         all_diagnoses.contains(ICD10Code("N170"))
         ```
         """
-        code_type = self._type._code_type()
-        if isinstance(code, code_type):
-            code = code._to_primitive_type()
-        elif not isinstance(code, str):
-            raise TypeError(
-                f"Expecting a {code_type}, or a string prefix of a {code_type}"
-            )
+        code = self._cast(code)
         return _apply(qm.Function.StringContains, self, code)
 
     def contains_any_of(self, codelist):
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -1045,7 +1045,7 @@ class MultiCodeStringFunctions:
         ```python
         apcs.all_diagnoses == "||I302, K201, J180 || I302, K200, M920"
         ```
-        Instead you should use the contains or contains_any_of methods.
+        Instead you should use the `contains` or `contains_any_of` methods.
         """
         raise TypeError(
             "This column contains multiple clinical codes combined together in a single "
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -1,11 +1,12 @@
 import dataclasses
 import datetime
 import functools
+import operator
 import re
 from collections import ChainMap
 from pathlib import Path
 
-from ehrql.codes import BaseCode
+from ehrql.codes import BaseCode, BaseMultiCodeString
 from ehrql.file_formats import read_rows
 from ehrql.query_model import nodes as qm
 from ehrql.query_model.column_specs import get_column_specs_from_schema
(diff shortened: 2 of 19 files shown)


tests/integration/tables/test_tpp.py
```diff
@@ -220,3 +223,77 @@ def test_addresses_imd_quintile(in_memory_engine):
...
+
+
+def test_apcs_all_diagnoses_cant_match_string():
+    with pytest.raises(TypeError):
+        tpp.apcs.all_diagnoses == "I000"
+    with pytest.raises(TypeError):
+        tpp.apcs.all_diagnoses != "I000"
+
+
+def test_apcs_all_diagnoses_matches_icd10_prefix_or_icd10_code(in_memory_engine):
```
evansd, line 235: I don't think this is quite the right place for these tests (which obviously you couldn't be expected to know!). This file is for testing the logic of helper methods which are specific to tables in the `tpp` schema.

I think the right place for exercising this logic is in what we call the "spec" tests, which you probably haven't seen yet. They live in `tests/spec` and also get built as [part of the docs](https://docs.opensafely.org/ehrql/reference/features/). We had grand plans for how great automatically creating a readable specification from test cases was going to be but we never fully [followed through](https://github.com/opensafely-core/ehrql/issues/1402).

One of the nice things about the spec tests is that they're automatically executed over all the query engines so we can enforce consistent behaviour there.

You'll need someone to talk you through these (and I'll need to do some serious memory-refreshing beforehand!) but the basic idea would be to add a column of the new type to the [test schema](https://github.com/opensafely-core/ehrql/blob/562d5baf67107d5f4ef8156c58d6ffc954f5342a/tests/spec/tables.py). And then add a new test along the lines of [`code_series_ops.test_containment`](https://github.com/opensafely-core/ehrql/blob/562d5baf67107d5f4ef8156c58d6ffc954f5342a/tests/spec/code_series_ops/test_containment.py) which exercises the new type.

We _could_ make the argument that all we really need to test here is that the ehrQL constructs compile to the correct query model structures as we're not actually adding any new operations to the query model, and the existing query model operations are fully tested elsewhere. That could be done entirely in the unit tests without hitting the database.

But there's something more reassuring about hitting a real database and seeing it give the correct results, so I think the spec tests are probably the way to go.

ehrql/serializer.py
```diff
@@ -39,6 +39,7 @@ class SerializerError(Exception):
         DisclosureControlConfig,
         *get_all_subclasses(Node),
         *get_all_subclasses(BaseCode),
+        *get_all_subclasses(BaseMultiCodeString),
```
evansd, line 42: This is all you need to do serializer-wise, because we never need to serialize _instances_ of this type. We just need to serialize references to the type itself because they occur in the schema.

evansd (commented):
This is looking really good. I think the ratio of usefulness of feature to lines of production code is particularly high here, which is always nice to see.

I've made a few minor comments. I think the major thing we still need to do is validation for strings, but that should be reasonably straightforward to add.

There's also some messing around with exactly where tests live which isn't significant in one sense but worth getting right.

tests/integration/tables/test_tpp.py
```diff
@@ -220,3 +223,77 @@ def test_addresses_imd_quintile(in_memory_engine):
         },
         {"patient_id": 6, "imd_quintile": "unknown", "imd_decile": "unknown"},
     ]
+
+
+def test_apcs_all_diagnoses_cant_match_string():
+    with pytest.raises(TypeError):
+        tpp.apcs.all_diagnoses == "I000"
+    with pytest.raises(TypeError):
+        tpp.apcs.all_diagnoses != "I000"
```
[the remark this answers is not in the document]
rw251: Would `tests/unit/test_tables.py` be a good place for them? That already has references to `tpp` tables, while `test_query_language.py` does not.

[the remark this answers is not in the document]
rw251: I've put them in `tests/unit/test_tables.py`. Can easily move them to a better place if needed.

ehrql/query_language.py
```diff
@@ -1036,6 +1037,72 @@ class CodeEventSeries(CodeFunctions, EventSeries):
...
+        would match all acute renal failure), or a clinical code. E.g.
+        ```python
+        all_diagnoses.contains("N17")
+        all_diagnoses.contains(ICD10Code("N170"))
+        ```
+        """
+        code_type = self._type._code_type()
+        if isinstance(code, code_type):
+            code = code._to_primitive_type()
+        elif not isinstance(code, str):
```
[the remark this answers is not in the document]
rw251: I've done this by having a regex on the `ICD10MultiCodeString` class, which is checked in the `_cast` method of `MultiCodeStringFunctions`. I've also moved the logic around returning a string if passed an ICD10Code, or the string itself, into the `_cast` method as well.

rw251:
Couple of other things to note about the implementation - mainly around the fact we want a spec test table to contain `||` characters but not have any parsers split on them.

ehrql/codes.py
```diff
@@ -30,6 +30,26 @@ def _to_primitive_type(self):
...
 
 
+# A base class for fields that are concatenated lists of clinical codes. This occurs
+# in the admitted patient care spell (apcs) table of hospital episode statistics for
+# all_diagnoses (ICD10 codes), and all_procedures (OPCS4 codes).
+#
+# This inherits from str because that's what the underlying data is, but is in this
+# file as it's sort of a code. In future a better implementation might be to parse the
+# field value into a Set of clinical codes.
+@dataclasses.dataclass(frozen=True)
```
evansd, line 40: I don't think this actually needs to be a dataclass because it's not using any of the dataclass behaviour.

tests/spec/multi_code_string_series_ops/test_containment.py
```diff
@@ -0,0 +1,56 @@
...
+            3: False,
+            4: None,
+        },
+    )
+
+
+def test_contains_any_of_codelist(spec_test):
+    spec_test(
+        table_data,
+        p.m1.contains_any_of([ICD10Code("M069"), ICD10Code("A429")]),
```
evansd, line 49: You could slightly tweak this test to demonstrate the function taking a full code as a string as well as the prefix example above. And it would show that mixed types work equivalently.
```suggestion
        p.m1.contains_any_of([ICD10Code("M069"), "A429"]),
```

tests/spec/conftest.py
```diff
@@ -97,9 +98,13 @@ def parse_row(column_types, col_names, line):
...
     """
 
+    # Regex splits on any '|' character, as long as it's not adjacent
+    # to another '|' character using look-ahead and look-behind. This
+    # is to allow '||' to appear as content within a field, currently
+    # just for the all_diagnoses and all_procedures fields in apcs
     return {
         col_name: parse_value(column_types[col_name], token.strip())
-        for col_name, token in zip(col_names, line.split("|"))
+        for col_name, token in zip(col_names, re.split(r"(?<!\|)\|(?!\|)", line))
```
evansd, line 107: Given that this logic occurs twice in this PR and I think, as you mentioned, there's something similar going on in the in-memory database now might be the time to abstract this out into a "parse Peter's ad-hoc ascii table format" helper. Maybe something like `utils/ascii_table_utils.py`.

(Sorry that this PR seems to keep growing extra bits on it!)

evansd (approved):
Great! I'm happy this does all the right things. A few minor comments but nothing that's immediately blocking.

tests/integration/tables/test_tpp.py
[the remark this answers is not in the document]
rw251: Ah yes I see. Something like this:
```python
def test_multi_code_string_series_throws_on_invalid_comparison():
    @table
    class a(EventFrame):
        code_string = Series(ICD10MultiCodeString)

    # We don't allow ==
    with pytest.raises(TypeError):
        a.code_string == "I000"

    # We don't allow !=
    with pytest.raises(TypeError):
        a.code_string != "I000"

    # ICD10 string prefixes must be valid prefixes
    with pytest.raises(TypeError):
        a.code_string.contains("ZZ2")

    # Must be ICD10 code, not just any code type
    with pytest.raises(TypeError):
        a.code_string.contains(SNOMEDCTCode("11100000"))
```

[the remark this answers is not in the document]
evansd: Exactly! Yes

tests/spec/multi_code_string_series_ops/test_containment.py
rw251: Good suggestion. I'll do it.

tests/spec/conftest.py
rw251: I guess my question is whether I'm breaking anything by _not_ changing the in-memory file, or I will break something by changing it. I've got a catch-up with Peter in a few minutes, so I'll pick his brains.

ehrql/codes.py
rw251: I've removed it, and the tests still pass, so I'm assuming you're right. I'll need to read up on dataclasses to understand it better.

tests/spec/conftest.py
rw251: Ok so the ascii parsing in `in_memory_database.py` is only used in one test file, which doesn't (and likely won't ever) contain `||`s which is why nothing breaks. I propose to merge the changes without the refactor, but to raise a ticket to pick that up in deck scrubbing.

evansd: Sounds sensible to me.

PR #2275, opened 2024-12-04
Protect standard out when loading ehrql dataset

bloodearnest:
Users may reasonably use print() statment in their dataset. Currently,
because we use stdout to output the serialization from the isolated
subprocess, this will break that process.

So, we force any use of sys.stdout within the dataset definition to use
stderr instead, to avoid that happening.

Lots of detailed discussion on this in this slack thread

https://bennettoxford.slack.com/archives/C069YDR4NCA/p1733228945599029

### DIFF
--- a/ehrql/loaders.py
+++ b/ehrql/loaders.py
@@ -337,6 +337,11 @@ def load_module(module_path, user_args=()):
     # generally looks as it would had you run: `python script.py some args --here`
     original_sys_argv = sys.argv.copy()
     sys.argv = [str(module_path), *user_args]
+    # Force any user generated output (prints etc) to stderr so it doe not get
+    # mixed up with anything we might want to output ourselves
+    original_sys_stdout = sys.stdout
+    sys.stdout = sys.stderr
+
     try:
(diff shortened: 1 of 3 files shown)


ehrql/loaders.py
```diff
@@ -337,6 +337,11 @@ def load_module(module_path, user_args=()):
     # generally looks as it would had you run: `python script.py some args --here`
     original_sys_argv = sys.argv.copy()
     sys.argv = [str(module_path), *user_args]
+    # Force any user generated output (prints etc) to stderr so it doesn't mixed up with
```
evansd, line 340:
```suggestion
    # Force any user generated output (prints etc) to stderr so it doesn't get mixed up with
```

evansd (approved):
Thanks for doing this!

Tediously, you'll need a pedantic commit prefix to get a new release.

Worth linking to the Slack thread in the PR body as well as there's quite a lot of useful context there:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1733228945599029

PR #2279, opened 2024-12-05
Added a `.hint()` method to allow hints in the quiz

rw251:
I've made a default hint which reminds people about `show()`. I realise this method is currently still called `debug()` but I doubt this will affect anyone before the debug->show change is also made.

### DIFF
--- a/ehrql/quiz.py
+++ b/ehrql/quiz.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 from pathlib import Path
 from typing import Any
@@ -307,7 +308,7 @@ class Question:
         self.engine = engine
         self.attempted = False
         self.correct = False
-        self._hint = "Unfortunately we don't have a hint for this question."
+        self._hint = ""
 
     def check(self, answer: Any = ...) -> str:
         if answer is not ...:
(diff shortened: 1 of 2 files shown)


ehrql/quiz.py
```diff
@@ -321,8 +322,19 @@ def check(self, answer: Any = ...) -> str:
...
+
+        if len(self._hint) > 0:
+            message_lines.append(self._hint)
+
+        message_lines.append(
+            inspect.cleandoc("""
+            Remember that you can use show() to take a look at your
+            output. So instead of calling questions[].check(..your answer..), you
+            can call show(..your answer..) to see if you're on the right track
+        """)
```
inglesp, line 335: It'd be more idiomatic to use `textwrap.dedent` for this.

inglesp (approved):
Good stuff.  What's the plan for adding some real hints?  Once we've done that we'll also need to document this.

rw251:
> Good stuff. What's the plan for adding some real hints? Once we've done that we'll also need to document this.

I've already got some hints ready to go. I was waiting for this to merge so that the hint method is actually available, and for debug to be changed to show. Therefore once this is merged I'll tidy up my hints and raise a PR.

PR #2300, opened 2024-12-09
Update docs tests so ehrql in tutorial can be tested

### LINKED ISSUE #2245: Ensure all new tutorial code snippets are tested
There is code in [test_complete_examples.py](https://github.com/opensafely-core/ehrql/blob/main/tests/docs/test_complete_examples.py) for testing complete examples.  However, many of the code snippets in the new tutorial are not complete examples, and so are not currently testable.

rebkwok:
Fixes #2245 
Adds the minimal dataset boilerplace (if it's missing) to create a dataset
and define a population so that an ehrql example can be validated. This
means that the tutorial examples, which mostly use the show() function,
don't need to include the dataset configuration and can still be tested
to ensure they're valid ehrql.

This does the minimal things to allow the tutorial ehrql to be tested without having to add extra code to the docs examples. It doesn't yet allow snippets where imports are missing (other than the `patients` table that's used for the very simplest population definition). It also doesn't do any magic boilerplating for measures definition.

### DIFF
--- a/tests/docs/test_complete_examples.py
+++ b/tests/docs/test_complete_examples.py
@@ -13,6 +13,7 @@ import mkdocs.config
 import pytest
 
 import ehrql.main
+from ehrql.utils.string_utils import strip_indent
 
 
 def get_markdown_extension_configuration():
@@ -170,12 +171,17 @@ class EhrqlExample:
                 # Add imports and create_dataset/define_population at the beginning of the file
                 # to ensure it's available for the rest of the snippet. This results in a
                 # weird order of imports if the snippet already has imports, but it's still valid ehrql
-                self.source = (
-                    "from ehrql import create_dataset\n"
-                    "from ehrql.tables.core import patients\n"
-                    "dataset = create_dataset()\n"
-                    f"{population_definition}\n"
-                    f"{self.source}"
+                self.source = strip_indent(
+                    """
+                    from ehrql import create_dataset
+                    from ehrql.tables.core import patients
+                    dataset = create_dataset()
+                    {population_definition}
+                    {source}
+                    """
+                ).format(
+                    source=self.source,
+                    population_definition=population_definition,
                 )
                 return EhrqlExampleDefinitionType.DATASET
 
--- a/tests/docs/test_complete_examples.py
+++ b/tests/docs/test_complete_examples.py
@@ -112,8 +112,11 @@ class EhrqlExample:
     The origin of such an example may be a Markdown fence,
     or a standalone Python module.
 
-    Such examples to be tested should be complete examples,
-    not partial snippets of examples.
+    Such examples to be tested should be complete examples, with the exception of
+    the basic dataset boilerplate. If no create_dataset() call is found, we insert
+    the minumum code required to ensure the dataset is created, a valid population
+    defined and it can be evaluated. Note that this doesn't apply to measures definitions, which need to be
+    complete examples.
 
     An EhrqlExample itself is not validated as a correct ehrQL definition example at creation,
     but used as part of the tests to validate the ehrQL."""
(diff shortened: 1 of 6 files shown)


tests/docs/test_complete_examples.py
```diff
@@ -148,11 +151,31 @@ def _get_definition_type(self):
...
+                    population_definition = (
+                        "dataset.define_population(patients.exists_for_patient())"
+                    )
+                # Add imports and create_dataset/define_population at the beginning of the file
+                # to ensure it's available for the rest of the snippet. This results in a
+                # weird order of imports if the snippet already has imports, but it's still valid ehrql
+                self.source = (
+                    f"from ehrql import create_dataset\nfrom ehrql.tables.core import patients\n"
+                    f"dataset = create_dataset()\n{population_definition}\n"
+                    f"{self.source}"
```
inglesp, line 176: I've a mild preference for a multiline string with `textwrap.dedent()`.

inglesp (approved):
Looks good.

Does this change mean that these commits can be reverted?

* https://github.com/opensafely-core/ehrql/commit/e2dd46080277d49073a3edf80f2537424d9c9120
* https://github.com/opensafely-core/ehrql/commit/777b8bb937f07ddb6f93b54593da9446613f8f0d

rebkwok:
@inglesp I don't think we can revert  https://github.com/opensafely-core/ehrql/commit/e2dd46080277d49073a3edf80f2537424d9c9120 until ew've reinstated whatever file it is that the commit message refers to:
> We've removed the one .py file that's referred to in the docs.  I expect we'll want to reinstate soon.

Do you know what that is?

tests/docs/test_complete_examples.py
rebkwok: I tried dedent, but I think I'm doing it wrong

This doesn't indent a tutorial example properly:
```
self.source = textwrap.dedent(
    f"""from ehrql import create_dataset
    from ehrql.tables.core import patients
    dataset = create_dataset()
    {population_definition}
    {self.source}"""
)
```

An example from the tutorial that has the boilerplate added ends up indented like this:
```
from ehrql import create_dataset
                    from ehrql.tables.core import patients
                    dataset = create_dataset()
                    dataset.define_population(patients.exists_for_patient())
                    from ehrql import show
from ehrql.tables.core import patients, practice_registrations, clinical_events, medications

index_date = "2024-03-31"

show(
    patients.age_on(index_date),
    patients.is_alive_on(index_date)
)
```
I think the problem is that `self.source` has linebreaks in it, so `textwrap.dedent` doesn't recognise the leading whitespace as common to the other lines.

Happy to update if you know the way to make dedent do the right thing though

evansd: The trick with `dedent` is a bit non-obvious but you need to suppress the leading newline with a `\`:
```py
self.source = textwrap.dedent(
    f"""\
    from ehrql import create_dataset
    from ehrql.tables.core import patients
    dataset = create_dataset()
    {population_definition}
    {self.source}"""
)
```

Because this is a bit non-obvious and easy to forget I also added a `strip_indent` function in the [ehrql.utils.string_utils](https://github.com/opensafely-core/ehrql/blob/4ea7bda492870bde3fa4794206109e4156d3f404/ehrql/utils/string_utils.py)
module, which you could use instead.

rebkwok: That still doesn't work (although different output) - now I get:
```
                    from ehrql import create_dataset
                    from ehrql.tables.core import patients
                    dataset = create_dataset()
                    dataset.define_population(patients.exists_for_patient())
                    from ehrql import show
from ehrql.tables.core import patients, practice_registrations, clinical_events, medications

index_date = "2024-03-31"

show(
    patients.age_on(index_date),
    patients.is_alive_on(index_date)
)
```

And if I use `strip_indent`, I get the same output as before (without the leading \).  I still think it's because `self.source` contains `\n`s

rebkwok: Yes, so if I remove all the `\n` from `self.source`, it's obviously broken, but it does dedent properly
```
source = self.source.replace("\n", "")
    self.source = strip_indent(
        f"""
        from ehrql import create_dataset
        from ehrql.tables.core import patients
        dataset = create_dataset()
        {population_definition}
        {source}
        """
    )
```
is formatted as 
```
from ehrql import create_dataset
from ehrql.tables.core import patients
dataset = create_dataset()
dataset.define_population(patients.exists_for_patient())
from ehrql import showfrom ehrql.tables.core import patients, practice_registrations, clinical_events, medicationsindex_date = "2024-03-31"show(    patients.age_on(index_date),    patients.is_alive_on(index_date))
```

rebkwok: Given that @inglesp only expressed a mild preference for using dedent, I think I'll just leave it as it is

evansd: Ah sorry, yeah! I wasn't paying enough attention here and saw one problem and assumed that was it.

I've been bitten by exactly this before, it's very annoying. Doing it the way you originally had it is probably the simplest solution. The other one is to use a template string rather than an f-string e.g.
```py
self.source = strip_indent(
    """
    from ehrql import create_dataset
    from ehrql.tables.core import patients
    dataset = create_dataset()
    {population_definition}
    {source}
    """
).format(
    source=self.source,
    population_definition=population_definition,
)
```

rebkwok: Nope, template string still doesn't work (either with `string_indent` or `textwrap.dedent`); I'm getting the same indenting as before.

evansd: I just tried the template thing locally and it did what I expected. You are removing the indent _before_ calling `.format()`, right?

rebkwok: I swear I did, and I tried it both with `textwrap.dedent` (with the newline suppression) and with `strip_indent`, and it didn't work. But I just tried it again locally and now it does.

PR #2303, opened 2024-12-10
Add fixture for testing use of random module

DRMacIver:
This fixture asserts that we never use the global random number generator in any of our tests (or, technically, that we only use it in ways that restore the previous state afterwards. Don't do that please). This ensures problems like that in #2302 are likely to be caught.

To my pleasant mild surprise it looks like we don't currently have any cases that this catches. I did revert #2302 to make sure that it would have caught that problem, and it would.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import random
 import subprocess
 import threading
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


evansd (approved):
Unknown future person at unknown future time is going to absolutely _love_ you for this!

Issue #2298, opened 2024-12-09
Ensure rows are displayed in correct order when displaying with show(dataset)


rebkwok:
The `show(dataset)` in the [full dataset](https://docs.opensafely.org/ehrql/tutorial/building-a-dataset/#datasets) in the ehrql tutorial returns a dataset with patients in an unsorted order
![Image](https://github.com/user-attachments/assets/53c49dbb-79b9-4657-8b5f-57b0e239ce3f)

It looks like this is because `PatientTable.to_records()` doesn't sort the patients it iterates over (`self.patients()` is a set, so is unordered)
https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/in_memory_database.py#L101

Compare to other `to_records()` methods that do sort, e.g. [EventTable](https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/in_memory_database.py#L184)

Issue #2297, opened 2024-12-09
Don’t render anything in `show()` for null values


rebkwok:
When using `show()` on an ehrql expression or dataset, show nothing instead of "None".

The `show()` command uses `SandboxQueryEngine.evaluate` [method](https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/sandbox.py#L26) to obtain the output of an ehrql query.

For a column or table, the query object is extracted [here](https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/sandbox.py#L38) where we call `self.visit()` (note that the variable is confusing named `column`, when it can actually be a table - that's a leftover from previous code and should be renamed). At this point the query element can be a:
- PatientTable
- PatientColumn
- EventTable
- EventColumn
(all defined in https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/in_memory_database.py)

The `show()` command uses the `_render_` function on each of these, which calls their `to_records()` method to to get the table/column as a list of dicts. We should be able to change how None values appear by updating the `to_records()` methods to replace None with empty strings. E.g. for PatientTable
```
def to_records(self):
    for p in self.patients():
        yield {name: col[p] for name, col in self.name_to_col.items()}
```
becomes:
```
def to_records(self):
    for p in self.patients():
        yield {name: "" if col[p] is None else col[p] for name, col in self.name_to_col.items()}
```

`DateDifference` is a special case in `show()`, because it's not a thing that can actually be added to a dataset, but we still let users `show()` it.  
In the evaulate() method, we call `.days` on DateDifference objects to convert them into a column, and [apply a format function ](https://github.com/opensafely-core/ehrql/blob/main/ehrql/query_engines/sandbox.py#L40)to add " days" to any value when we show it.  This should still work, but the function will need to be updated to handle empty strings instead of None values.

To evaluate a dataset in the sandbox, we call `evaluate_dataset()` method, which calls `InMemoryEngine.get_results_as_table()`, and returns the dataset as a PatientTable; so updating the PatientTable.to_records method should render dataset None values as empty strings also.

Finally, one last quirk: in `debugger.py` we also combine columns together to show a single table if we can. When we do this, we don't use the usual `to_records()` methods, instead we have two custom methods to combine records from the columns `related_patient_columns_to_records()` and `related_event_columns_to_records()` - these will also need to replace None values with empty strings (or update to use each column's `to_records()`.
[...]

PR #2309, opened 2024-12-11
Fix dataset display order and format

Providence-o:
* https://github.com/opensafely-core/ehrql/issues/2297
* https://github.com/opensafely-core/ehrql/issues/2298

### DIFF
--- a/ehrql/debugger.py
+++ b/ehrql/debugger.py
@@ -5,6 +5,7 @@ import sys
 from ehrql.query_engines.in_memory_database import (
     EventColumn,
     PatientColumn,
+    render_value,
 )
 from ehrql.query_engines.sandbox import SandboxQueryEngine
 from ehrql.query_language import BaseFrame, BaseSeries, Dataset, DateDifference
@@ -147,7 +148,7 @@ def related_patient_columns_to_records(columns):
     for patient_id in columns[0].patient_to_value.keys():
         record = {"patient_id": patient_id}
         for i, column in enumerate(columns, start=1):
-            record[f"series_{i}"] = render_value(column[patient_id])
+            record[f"series_{i}"] = render_value(column[patient_id], convert_null=True)
         yield record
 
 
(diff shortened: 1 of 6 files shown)


ehrql/query_engines/in_memory_database.py
```diff
@@ -99,7 +99,7 @@ def __getitem__(self, name):
         return self.name_to_col[name]
 
     def to_records(self):
-        for p in self.patients():
+        for p in sorted(self["patient_id"].patients()):
```
rebkwok, line 102: This can be `sorted(self.patients())`, I think - `self.patients()` just returns `self["patient_id"].patients()`

ehrql/query_engines/in_memory_database.py
```diff
@@ -258,7 +261,7 @@ def __getitem__(self, patient):
 
     def to_records(self):
         return (
-            {"patient_id": p, "value": v}
+            {"patient_id": render_value(p), "value": render_value(v)}
```
rebkwok, line 264: `patient_id` should never be None, so no need to format `p`

ehrql/query_engines/in_memory_database.py
```diff
@@ -319,7 +322,11 @@ def __getitem__(self, patient):
 
     def to_records(self):
         return (
-            {"patient_id": p, "row_id": k, "value": v}
+            {
+                "patient_id": render_value(p),
+                "row_id": render_value(k),
+                "value": render_value(v),
+            }
```
rebkwok, line 329: I also don't think `patient_id` or `row_id` should ever be None here

ehrql/debugger.py
```diff
@@ -156,5 +156,11 @@ def related_event_columns_to_records(columns):
...
             for i, column in enumerate(columns, start=1):
-                record[f"series_{i}"] = column[patient_id][row_id]
+                record[f"series_{i}"] = render_value(column[patient_id][row_id])
             yield record
+
+
+def render_value(value):
+    if value is None:
+        value = ""
+    return value
```
inglesp, line 166: Rather than reimplement this, it would be good to import the version from `in_memory_database.py`.

inglesp (approved):
Looks good.  Have made one small suggestion.

Also, would you be able to squash the last two commits together?  The second-to-last commit introduces a line that we remove again in the last commit.

PR #2312, opened 2024-12-11
Added consistent example ons_death data

### LINKED ISSUE #2273: Add `ons_deaths` back into the example data we provide
[Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1733311372037229)
When closing #2146 we had a policy of including all core tables in our provided example data. Following the tutorial revamp, we now have better example data for four of the core tables, but do not provide a file for `ons_deaths`. 

We have also switched to using [scripts/generate_example_data.py](https://github.com/opensafely-core/ehrql/blob/main/scripts/generate_example_data.py) to generate example data. 

To keep the same policy, we can either put [the minimal ons_deaths.csv file](https://github.com/opensafely-core/ehrql/blob/main/tests/fixtures/example-data/ons_deaths.csv) we previously had back into the `example-data` folder, or extend the script to generate `ons_deaths` as well (albeit it's not being used in the tutorial). 

After that, we should amend the [tests](https://github.com/opensafely-core/ehrql/blob/main/tests/unit/test_example_data.py) such that they test for the new example data. The tests make sure that the LocalFileQueryEngine has no problem loading the example data.

rw251:
Everyone with a death date in the `patients.csv` file has an entry in the new example ons death data. I've also added 2 people who don't have a death date in the patients file to reflect the real world where ons death data is typically recorded before it appears in the primary care record.

Also updated the example data test to use the `./ehrql/example-data` directory as its source - this was changed recently to use example data in the test fixtures directory because of the missing ons_deaths data. Now that we have that this can revert to its previous behaviour.

I haven't removed the test fixtures example-data as it is still used by the `test_quiz.py` file. I suspect we can revert this as well and get rid of the test fixture example-data directory, but maybe in a separate PR.

fixes #2273


alarthast:
The `tests/fixtures/example-data` bit was unclear ticket writing on my part. I linked to it only to point out what the old file used to look like (I could have equally linked to [this](https://github.com/opensafely-core/ehrql/blob/9364bc34530f2f21142dca0d54470127e854e531/ehrql/example-data/ons_deaths.csv) instead).

There is probably little worth in removing `tests/fixtures/example-data` as that move would require rewriting all the assertions in `test_quiz.py` for the `check_answer` function

evansd:
Yes, I agree that the quiz tests should use their own standalone dummy data fixture. It might be worth renaming that `quiz-example-data` or something to make it clear that the fixture is only for that purpose.

I think the mistake in [this commit here](https://github.com/opensafely-core/ehrql/commit/254f71c507e0fab39b1b41c749c13e7f0aa86509) was to change the path in `test_example_data.py` to point to the fixture. This test is supposed to enforce certain properties of the example data we ship with ehrQL. If we decided it was temporarily OK to violate those then we should have xfailed the test, not changed it to test some _other_ set of files which have nothing to do with the bundled example data.

rw251:
> Yes, I agree that the quiz tests should use their own standalone dummy data fixture. It might be worth renaming that `quiz-example-data` or something to make it clear that the fixture is only for that purpose.

In addition to the changes to the generation script, I've also renamed the quiz dummy data fixture.

evansd (approved):
Great, looks good.

Very minor thing, but I found the commit order a bit confusing which made it very slightly less clear to review. I would have expected:
 * Extend generation script
 * Regenerate example data using new script
 * Rename fixture

Combining the updated files with the modified script mixes up reviewable code with effectively unreviewable data changes. And having the new ONS file before the script makes it look like it was hand-crafted rather than an output of the script added in the next commit.

But that really is super nitpicky!

rw251:
> Combining the updated files with the modified script mixes up reviewable code with effectively unreviewable data changes. And having the new ONS file before the script makes it look like it was hand-crafted rather than an output of the script added in the next commit.

Yes fair enough. I've only ever reviewed PRs as a whole, rather than a commit story, so it's taking me a while to adjust! I'll get there.

Issue #2210, opened 2024-11-07
Automatically link from each table's reference docs to example docs for that table


evansd:
This would require adding some more structure to our [documented examples](https://docs.opensafely.org/ehrql/how-to/examples/) so that we can automatically determine which examples use which tables. We've already talked about [breaking up](https://github.com/opensafely-core/ehrql/issues/1622) the examples page in some fashion anyway.

@rw251 has pointed out that, as a new ehrQL user, it's possible to find your way to the docs for a particular table and then get stuck because you don't know what sort of things you can do with that table. The existing examples are helpful, but only if know where to go looking for them. Adding direct breadcrumbs from the table docs to the examples for that table would not only prevent this dead end, but would also help surface the existence of the examples more generally.

This is an interesting "regression" from Cohort Extractor in that because Cohort Extractor was so limited and non-composable, each table's documentation came bundled with a fixed set of "things you can do with this table". Possibly this accounts for the impression we've sometimes seen that ehrQL is less approachable than Cohort Extractor.

It also occurs to me that it might be good to separate out examples of general query patterns in ehrQL (which apply to any table) from examples specific to a give table (or tables). At the top of each section of table-specific examples we could link to the more general examples.

madwort:
decided to try adding manual links from the table reference pages to the examples, we have automated link checking that should check if they go stale. the examples page has not substantially changed in the last year & I don't think we expect it to the change in the coming year.

PR #2314, opened 2024-12-11
Reorganise examples page

madwort:
Examples page is now organised by primary table of interest
Screenshot 2024-12-12 at 14 37 40

Reference docs for the table now link to the relevant section of the examples page

Screenshot 2024-12-12 at 14 37 59

This is done manually, if we produce many more examples in the future we could build some tooling to automate (but I'm not convinced that's worth doing right now).

Fixes https://github.com/opensafely-core/ehrql/issues/2210

### DIFF
--- a/docs/how-to/examples.md
+++ b/docs/how-to/examples.md
@@ -7,6 +7,10 @@ Or you can use the navigation bar at the top-right of this page,
 to see a list of the examples,
 and then jump to a specific example of interest.
 
+The examples are organised firstly by the table which they pull data from -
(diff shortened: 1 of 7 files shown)


inglesp (approved):
I think this is a good improvement.  It would be nice to be able to link the other way, from the examples page to the schema docs, linking to the core namespace where possible.

I also don't think we need a new <h2> for Example usage.  It makes the schema table below look like it belongs to this section.  Just a link would be fine.

PR #2317, opened 2024-12-12
Added `OPCS4MultiCodeString` type and assigned to the `all_procedures` field

### LINKED ISSUE #2315: Add `OPCS4MultiCodeString` type for the `all_procedures` field of `apcs`
We've recently added a `ICD10MultiCodeString` type for `all_diagnoses`. We should also do this for the `all_procedures` field in `apcs` as it has the same format.

rw251:
fixes #2315

### DIFF
--- a/ehrql/codes.py
+++ b/ehrql/codes.py
@@ -146,6 +146,25 @@ class ICD10MultiCodeString(BaseMultiCodeString):
     regex = re.compile(r"[A-Z][0-9]{0,3}")
 
 
+#
+# OPCS4 codelist as concatenated string
+#
+# This is specifically for fields in the admitted patient care (APC) part
+# of the hospital episode statistics (HES) data where there are fields
+# that are a concatenation of all procedure codes for a patient's episode
+# or spell.
+class OPCS4MultiCodeString(BaseMultiCodeString):
+    "Multiple OPCS4 codes"
+
+    @classmethod
+    def _code_type(cls):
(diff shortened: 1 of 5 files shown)


ehrql/codes.py
```diff
@@ -146,6 +146,25 @@ def _code_type(cls):
...
+class OPCS4MultiCodeString(BaseMultiCodeString):
+    "Multiple OPCS4 codes"
+
+    @classmethod
+    def _code_type(cls):
+        return OPCS4Code
+
+    # We want to allow prefix searching, so when we check the regex we
+    # want to account for that
+    regex = re.compile(r"[A-Z][0-9]{0,3}")
```
inglesp, line 165: I don't understand the comment about checking the regex -- when would we check the regex?

inglesp (approved):

ehrql/codes.py
rw251: When people call `all_procedures.contains()` (or any column that derives from `BaseMultiStringCode`), they can pass either a code, or a string prefix e.g. "I30" to match codes like "I300", "I301" etc. If they pass a code we can check it's the right type, but if it's a string prefix we want to check whether it's a valid prefix for the field, so e.g. `all_procedures.contains("HELLO")` would throw an error, rather than failing silently by running but returning 0 records.
I'll update the comment to make that a bit clearer.

Issue #2301, opened 2024-12-09
Combining PatientSeries in `show()` does not display complete set of patients


evansd:
It should be that calling:
```py
show(patient_series_a, patient_series_b)
```
displays a table containing the union of all patients in either series A or series B.

However at present, only patients contained within series A are displayed.

The code which needs fixing is here:
https://github.com/opensafely-core/ehrql/blob/6219d37a576df9d99c697946280923253a99227b/ehrql/debugger.py#L146-L151

Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1733753206863339

PR #2318, opened 2024-12-13
Show combined patient series

madwort:
Upgrade related_patient_columns_to_records to do a "full join"

* fixes https://github.com/opensafely-core/ehrql/issues/2301

### DIFF
--- a/ehrql/debugger.py
+++ b/ehrql/debugger.py
@@ -148,7 +148,7 @@ def related_patient_columns_to_records(columns):
     patients_all = set()
     for c in columns:
         patients_all.update(c.patient_to_value.keys())
-    for patient_id in patients_all:
+    for patient_id in sorted(patients_all):
         record = {"patient_id": patient_id}
         for i, column in enumerate(columns, start=1):
             record[f"series_{i}"] = render_value(column[patient_id], convert_null=True)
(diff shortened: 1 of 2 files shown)


ehrql/debugger.py
```diff
@@ -145,7 +145,11 @@ def related_columns_to_records(columns):
 
 
 def related_patient_columns_to_records(columns):
-    for patient_id in columns[0].patient_to_value.keys():
+    patients_all = set()
+    for c in columns:
+        for k in c.patient_to_value.keys():
+            patients_all.add(k)
```
evansd, line 151: I think you're better off using `.update` here e.g
```suggestion
    for c in columns:
        patients_all.update(c.patient_to_value.keys())
```

ehrql/debugger.py
```diff
@@ -145,7 +145,11 @@ def related_columns_to_records(columns):
...
+    for patient_id in patients_all:
```
evansd, line 152: I think we should probably sort here for consistency, otherwise we'll get the patients in an arbitrary order.

evansd (approved):

ehrql/debugger.py
madwort:
```suggestion
    for patient_id in sorted(patients_all):
```

PR #2319, opened 2024-12-13
Update behaviour of `show()` to just display ehrql objects

rw251:
New `show` behaviour
- if not debug mode just display ignore message
- if debug mode and single ehrql object then display
- if debug mode and multiple series from same domain then display
- otherwise throw a helpful message

### DIFF
--- a/ehrql/debugger.py
+++ b/ehrql/debugger.py
@@ -26,11 +26,14 @@ def show(
     tail: int | None = None,
 ):
     """
-    Show the output of the specified element within a dataset definition
+    Show the output of the specified element or elements
 
     _element_<br>
-    Any element within the dataset definition file; can be a string, constant value etc,
-    but will typically be a dataset variable (filtered table, column, or a dataset itself.)
+    Any ehrql object, such as a series, frame, or dataset
+
+    _other_elements_<br>
+    0 or more series, but only if the _element_ was also a series, and only if they are all
+    from the same domain
 
     _label_<br>
     Optional label which will be printed in the show output.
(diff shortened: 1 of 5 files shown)


ehrql/debugger.py
```diff
@@ -50,7 +53,39 @@ def show(
...
+            "\n\nshow() can be used in two ways. Either:\n"
+            " - call it with a single argument such as:\n"
+            "     show(dataset)\n"
+            "     show(patients)\n"
+            '     show(clinical_events.date.is_after("2024-01-01"))\n'
+            " - or call it with multiple columns/series:\n"
+            "     show(patients.sex, patients.date_of_birth)\n"
+            '     show(patients.sex == "male", patients.date_of_birth.is_before("2006-01-01"))\n\n'
+            "If passing in multiple arguments, they must all be 'of the same domain', meaning "
+            "you can't combine series from different tables."
```
rebkwok, line 87: This isn't strictly true - I can combine e.g. `patients.date_of_birth` with `clinical_events.exists_for_patient()` and `ons_deaths.date`; as long as the series are all one row per patient series, they can be drawn from different tables. If they're multiple rows per patient series, they do need to come from the same table.

ehrql/debugger.py
```diff
@@ -161,3 +194,11 @@ def related_event_columns_to_records(columns):
                     column[patient_id][row_id], convert_null=True
                 )
             yield record
+
+
+def is_ehrql_object(element):
+    isDataset = isinstance(element, Dataset)
+    isSeries = isinstance(element, BaseSeries)
+    isDateDifference = isinstance(element, DateDifference)
```
rebkwok, line 202: `isinstance` can take a tuple, so this can be one `isinstance(element, (Dataset, BaseSeries, DateDifference))`

ehrql/debugger.py
```diff
@@ -161,3 +194,11 @@ def related_event_columns_to_records(columns):
...
+    isFrame = isinstance(getattr(element, "_qm_node", None), qm.Frame)
```
rebkwok, line 203: nit: rename variables in snake case (it's the python way)

rebkwok (commented):
This looks fine, just a question about the error we show to users if they try to combine the wrong things. Minor nits otherwise

ehrql/debugger.py
rw251: bah - yes still tripping me up from JavaScript days.

ehrql/debugger.py
rw251: Yes good point - I've updated the error message.

ehrql/debugger.py
rw251: Thanks. I've changed it to `isinstance(element, BaseSeries | DateDifference | Dataset)`.

rebkwok (approved):

Issue #2260, opened 2024-11-28
Query retry code doesn't work as expected due to transaction state


evansd:
This has happened several times recently, always with measures and always during the downloading of results. I think there are a few things we need to do here, in increasing order of difficulty and usefulness:

1. Update the [code](https://github.com/opensafely-core/ehrql/blob/15524551a6c27d3ba36dcdb8c7da34dacac5cc55/ehrql/backends/tpp.py#L140-L147) which reports transient database errors to check the full exception chain not just the most recent exception. That way we will at least correctly report to the user that it was a transient database error and not a problem with their code.

2. Try to reproduce this transaction error locally by injecting real database errors in the [test code](https://github.com/opensafely-core/ehrql/blob/15524551a6c27d3ba36dcdb8c7da34dacac5cc55/tests/integration/query_engines/test_mssql.py#L34-L40) and getting the retry to fail.

3. Try to make the retry code work correctly, possibly by fiddling the [transaction mode](https://github.com/opensafely-core/ehrql/blob/15524551a6c27d3ba36dcdb8c7da34dacac5cc55/ehrql/query_engines/mssql.py#L176-L179) we run in. It's possible that we just can't do this. It's also possible that we can do this for some classes of error but not others and so we won't know if we've really addressed this until the next time we hit one of these connection errors in production.

4. If we can't retry mid-query then it's possible we could do something to help specifically for measures. These are far more likely to be the victim of intermittent problems simply because they run for longer and perform more queries. But we also have the option to retry the entire data extract for the specific interval that failed without throwing away the results for previous intervals.

5. We should try to work out _why_ we're getting these connection errors. Given our limited diagnostic tools with respect to the database this is likely to hard and will probably require the assistance of TPP.

---

As mentioned in [this Slack thread](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1732807789124149?thread_ts=1732283490.781759&cid=C01D7H9LYKB) and seen in the logs of [this job](https://jobs.opensafely.org/impact-and-inequalities-of-winter-pressures-in-primary-care-providing-the-evidence-base-for-mitigation-strategies/wp-work-package-1/23757/umvbp5gg2os3oj64/):

![Image](https://github.com/user-attachments/assets/b1dd5230-8496-4c23-8b2a-9d2146080aa8)
![Image](https://github.com/user-attachments/assets/bb7fac82-8f75-4baa-a46e-b71836876846)

[...]

PR #2322, opened 2024-12-17
Attempt to improve database error handling

evansd:
This partially addresses (points 1, 2 and 3 of):
 * https://github.com/opensafely-core/ehrql/issues/2260

By injecting a synthetic database error into the tests we can reproduce the stack trace we see in production and confirm that the new rollback behaviour fixes this. Whether or not this actually addresses the issue is impossible to say because we can't reproduce, and don't know the cause of, the original error. The only way to know if this has helped is to search the logs of ehrQL jobs to identify instances of the retry code in operation. I have set myself a reminder to do this early in the new year.

Additionally, we now inspect the full exception chain when identifying database errors which now means we'll at least correctly report this _as_ database errors. Hopefully this will reduce user confusion and wasted time to some degree.


tests/unit/utils/test_sqlalchemy_exec_utils.py
```diff
@@ -84,18 +92,22 @@ def error_during_iteration():
...
+    assert connection.execute.call_count == 4
     assert sleep.mock_calls == [mock.call(t) for t in [10, 20, 40]]
     assert "Retrying query (attempt 3 / 3)" in log_messages
 
 
 @mock.patch("time.sleep")
 def test_execute_with_retry_exhausted(sleep):
-    execute = mock.Mock(side_effect=[ERROR, ERROR, ERROR, ERROR])
+    connection = mock.Mock(
+        **{
```
bloodearnest, line 104: TIL about this slightly obtuse way of using dotted notation to define nested attributes

ehrql/utils/sqlalchemy_exec_utils.py
```diff
@@ -75,6 +75,12 @@ def execute_with_retry(*args, **kwargs):
                     raise
                 retries += 1
                 log(f"{e.__class__.__name__}: {e}")
+                # Note that as we're running with AUTOCOMMIT isolation level, this
+                # rollback should be purely internal to SQLAlchemy and should have no
+                # effect on the database itself. See:
+                # https://docs.sqlalchemy.org/en/20/core/connections.html#understanding-the-dbapi-level-autocommit-isolation-level
```
bloodearnest, line 81: Good grief.  This both pragmatic and horrid.

tests/integration/query_engines/test_mssql.py
```diff
@@ -68,8 +68,17 @@ def wrap_select_queries():
     mocked = mock.Mock()
 
     def wrapper(self, *args, **kwargs):
-        if args and isinstance(args[0], Select):
-            mocked(*args, **kwargs)
+        if args and isinstance(args[0], sqlalchemy.sql.Select):
+            try:
+                mocked(*args, **kwargs)
+            except Exception:
+                # Force the database connection to revalidate. Ideally we would be able
```
bloodearnest, line 75: "revalidate" doesn't quite scan right for me here, am I missing something?

ehrql/query_engines/mssql.py
```diff
@@ -173,9 +173,10 @@ def get_results(self, variable_definitions):
         setup_queries, cleanup_queries = get_setup_and_cleanup_queries(results_query)
 
         with self.engine.connect() as connection:
-            # Previously, when using the temporary database to store results, it was
-            # important to use AUTOCOMMIT. Possibly we don't need this anymore, but in
-            # the spirit of not changing too much at once I'm leaving it in place.
+            # All our queries are either (a) read-only queries against static data, or
+            # (b) queries which modify session-scoped temporary tables. For this reason
+            # we can avoid the overhead of transaction integrity and can run in "YOLO"
+            # isolation mode.
```
bloodearnest, line 179: The nittest of nits: I wonder if we should be clearer on language here.

"isolation mode" to me means the DB level isolation, e.g. READ COMMITTED, etc.  AIUI this is the client side DB-API transaction mode, which feels like it maybe needs a different name? Or does it also set a DB level isolation mode?

bloodearnest (approved):
Clear improvements, and worth getting in before the holidays.

tests/integration/query_engines/test_mssql.py
evansd: Oh yes, fair point. I used that term because I'd been knee-deep in SQLAlchemy's internals for a while and the effect of nulling this value out is to force SQLAlchemy to call a `revalidate_connection()` method which ends up raising the error we were seeing.

I think the comment ought to be:
> Simulate the effect of getting an error from the underlying database driver.

ehrql/utils/sqlalchemy_exec_utils.py
evansd: IKR? :disappointed:

ehrql/query_engines/mssql.py
evansd: It's definitely worth being as clear as we can on language. I don't fully understand all the details here, but I'll at least make sure I mirror SQLAlchemy's description exactly.

PR #2323, opened 2024-12-18
Improve some links in examples page 

StevenMaude:
For the examples page:

* Use `.md` in URLs to appease the OpenSAFELY documentation link checker, which also facilitates using the correct relative path to the linked file
* Use relative links instead of links to `docs.opensafely.org`


madwort (approved):
thanks for picking this up!

PR #2325, opened 2024-12-18
fix: Pass appropriate kwargs to dummy data generators

rebkwok:
This lets dataset definitions run with the cli use the new `additional_population_constraints`

### DIFF
--- a/ehrql/main.py
+++ b/ehrql/main.py
@@ -133,11 +133,7 @@ def generate_dataset_with_dummy_data(
         query_engine = LocalFileQueryEngine(dummy_tables_path)
         results = query_engine.get_results(variable_definitions)
     else:
-        generator = get_dummy_data_class(dummy_data_config)(
-            variable_definitions,
-            population_size=dummy_data_config.population_size,
-            timeout=dummy_data_config.timeout,
-        )
+        generator = get_dummy_data_generator(variable_definitions, dummy_data_config)
         results = generator.get_results()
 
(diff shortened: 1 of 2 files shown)


evansd (approved):
Thanks for catching this. I'd like to have a think about what kind of testing approach would have stopped this making it into prod, but that's a wider question.

Now that we're returning `kwargs` I think there might be a slightly more straightforward factoring here, along the lines of:
```py
def get_dummy_data_generator(variable_definitions, dummy_data_config):
    if dummy_data_config.legacy:
        return DummyDataGenerator(variable_definitions, ...)
    else:
        return NextGenDummyDataGenerator(variable_definitions, ...)
```

Returning a reference to the class and the constructor kwargs feels like unnecessary indirection at this point.

rebkwok:
> Thanks for catching this. I'd like to have a think about what kind of testing approach would have stopped this making it into prod, but that's a wider question.
> 
> Now that we're returning `kwargs` I think there might be a slightly more straightforward factoring here, along the lines of:
> 
> ```python
> def get_dummy_data_generator(variable_definitions, dummy_data_config):
>     if dummy_data_config.legacy:
>         return DummyDataGenerator(variable_definitions, ...)
>     else:
>         return NextGenDummyDataGenerator(variable_definitions, ...)
> ```
> 
> Returning a reference to the class and the constructor kwargs feels like unnecessary indirection at this point.

Yeah, I'd thought about doing it that way, happy to change it

PR #2326, opened 2024-12-19
Improves distribution of dummy data

DRMacIver:
This improves the distribution of dummy data through the strategy of oversampling: We produce a larger population of patients than requested, and then sample down to a subset of it according to some weighted distribution of values.

This weighting scheme is calculated in two ways:

1. We try to force the distribution to be "more uniform" than whatever the basic dummy data generation produces.
2. We allow the user to provide an arbitrary weighting function to deviate from that.

It's also a good place to insert heuristics we might want to add (e.g. we could add default age and sex distributions, we could choose what proportion of nullable columns should be null, etc). This doesn't do any of that.

It does come with the cost of making dummy data slower. I think this is an OK tradeoff, and will be improved by anything that improves dummy data generation (in particular future work on making constraints satisfied more often)

### DIFF
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -177,14 +177,14 @@ class Dataset:
         else:
             self.dummy_data_config.additional_population_constraint = None
         if patient_weighting is not None:
+            if legacy:
+                raise ValueError("Cannot provide patient weighting in legacy mode.")
             validate_patient_series_type(
                 patient_weighting,
                 types=[float],
                 context="patient weighting",
             )
             self.dummy_data_config.patient_weighting = patient_weighting._qm_node
-        else:
-            self.dummy_data_config.patient_weighting = Value(1.0)
 
         self.dummy_data_config.oversample = oversample
 
@@ -192,8 +192,6 @@ class Dataset:
             raise ValueError(
                 "Cannot provide an additional population constraint in legacy mode."
             )
-        if legacy and patient_weighting:
-            raise ValueError("Cannot provide patient weighting in legacy mode.")
 
     def __setattr__(self, name, value):
         if name == "population":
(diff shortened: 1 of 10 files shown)


ehrql/query_language.py
```diff
@@ -129,6 +133,12 @@ def configure_dummy_data(
         _timeout_<br>
         Maximum time in seconds to spend generating dummy data.
 
+        _oversample_<br>
+        Dummy data generation will generate a larger population and then sample from it
+        to improve the distribution of patients. This parameter controls how much larger.
+        Lower values will be faster to generate, while larger values will get closer to
+        the target distribution.
```
rebkwok, line 140: I think it'd be helpful to users to give the default and an idea of what a small/large value is. i.e. 1 means no oversampling, 2 means oversampling by up to 2x the specified population size

ehrql/query_language.py
```diff
@@ -142,6 +152,12 @@ def configure_dummy_data(
         patients.age_on(some_date) < 80)`` would give you dummy data consisting of only men
         and women who were under the age of 80 on some particular date.
 
+        _patient_weighting_<br>
+        Defines a "weight" expression that lets you control the distribution of patients.
+        Ideally a patient row will be generated with probability proportionate to its weight,
+        although this ideal will be imperfectly realised in practice. The higher the value of
+        ``oversample`` the closer this ideal will to being realised.
```
rebkwok, line 159: An example as we have for additional_population_constraints would be helpful. Will this always be a `case()`?

inglesp:
We're closing for now.  See discussion [here](https://bennettoxford.slack.com/archives/C07MZ5WCJV6/p1738678082171959).

PR #2328, opened 2024-12-19
Allow measures to specify dummy data additional population constraint

rebkwok:


ehrql/query_language.py
```diff
@@ -50,6 +50,23 @@ class DummyDataConfig:
...
+            validate_patient_series_type(
+                self.additional_population_constraint,
+                types=[bool],
+                context="additional population constraint",
+            )
+            self.additional_population_constraint = (
+                self.additional_population_constraint._qm_node
+            )
+        else:
+            self.additional_population_constraint = None
```
madwort, line 64: do we need this branch? (the other branch tested for `is not None`)

ehrql/query_language.py
```diff
@@ -50,6 +50,23 @@ class DummyDataConfig:
...
+    def validate_additional_population_constraint(self):
+        if self.additional_population_constraint is not None:
+            validate_patient_series_type(
+                self.additional_population_constraint,
+                types=[bool],
+                context="additional population constraint",
+            )
+            self.additional_population_constraint = (
+                self.additional_population_constraint._qm_node
+            )
```
evansd, line 62: I don't think this is quite right because now the "validate" method is changing the value of `self.additional_population_constraint`. So, for example, it will fail if you call it twice in a row.

I think you want to make it a `set_additional_population_constraint()` method which takes the value, validates it, and then updates the attribute if it's valid.

evansd (approved):
Thanks for doing this!

ehrql/query_language.py
rebkwok: Huh that's how I did it first. Will change it back, hadn't considered calling it twice

PR #2337, opened 2024-12-20
Improved autocomplete for most ehrql things

rw251:
The type checker/autocomplete now knows:
- the correct type of all column series (e.g. `patients.date_of_birth` is a `DatePatientSeries`)
- the return type of almost all methods on the series (e.g. `clinical_events.numeric_value.sum_for_patient()` is a `FloatPatientSeries`
- properties on series have the correct type (e.g. `patients.date_of_birth.day` is a `IntPatientSeries`)
- the return type for most operators like `+`, `-`, `>`, `/` etc.

### DIFF
--- a/ehrql/docs/language.py
+++ b/ehrql/docs/language.py
@@ -19,6 +19,7 @@ EXCLUDE_FROM_DOCS = {
     ql.when,
     ql.DummyDataConfig,
     ql.Error,
+    ql.int_property,  # Internal thing for type hints and autocomplete
 }
 
 
--- a/ehrql/query_language.py
+++ b/ehrql/query_language.py
@@ -4,8 +4,9 @@ import functools
 import operator
 import re
 from collections import ChainMap
+from collections.abc import Callable
 from pathlib import Path
-from typing import Generic, TypeVar, overload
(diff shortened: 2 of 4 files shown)


evansd (approved):
This is great!

The only part that gave me any pause was the `@int_property` stuff just because I'm not completely clear on what problem it solves and how it does it. (I'm quite confident there _is_ a problem and that it _does_ solve it, I'd just like to understand it better.)

But it's a small and simple enough bit of code that I'd be happy to merge it now and understand it later.

In terms of testing I'll make a separate ticket to think about options there.

rw251:
> The only part that gave me any pause was the `@int_property` stuff just because I'm not completely clear on what problem it solves and how it does it. (I'm quite confident there _is_ a problem and that it _does_ solve it, I'd just like to understand it better.)

Having not looked at it since before my break I'm not sure I can remember! It was something to do with how the autocomplete interacted with the `@property` decorator. I tried a few things, and this seemed to be the simplest. I also considered making it more like a `@autocomplete_property[T]`, but given we currently only have `int` properties it seemed easier to just do that and not make it more complicated than it needed to be.

PR #2348, opened 2025-01-09
Expose decision support values table in ehrQL

### LINKED ISSUE #2264: Expose `DecsionSupportValues` table in ehrQL
At present this table contains only electronic frailty scores pre-calculated at a particular date. But it may in future contain other types of decision support value. We used to expose this in Cohort Extractor and so we should do so in ehrQL although we were waiting for a specific use case, which has now [arisen](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1732274106719799).

The table itself is `DecisionSupportValue`:
https://github.com/opensafely-core/ehrql/blob/a61b4cd37f8aa3ab57003c6c3baba15cce501faa/tests/lib/tpp_schema.py#L562-L569

And there is also `DecisionSupportValueReference` which would need to be joined on `AlgorithmType`:
https://github.com/opensafely-core/ehrql/blob/a61b4cd37f8aa3ab57003c6c3baba15cce501faa/tests/lib/tpp_schema.py#L572-L583

We'd want to expose this as a single, denormalised table (presumably `erhql.tables.tpp.decision_support_values`) with the column names appropriately snake-cased, and the DATETIME cast to a DATE. We can do this by defining it in the TPP backend as a `QueryTable` (see e.g. [addresses](https://github.com/opensafely-core/ehrql/blob/0484d0f060afbb7b69555aa44d72c9a5b034a806/ehrql/backends/tpp.py#L166-L199)).

I'm not sure about including the `AlgorithmSourceLink`. It's good that it exists for our own reference but it feels a bit superfluous to have it in with the patient data.

You'll find that [`tests/integration/backends/test_tpp.py`](https://github.com/opensafely-core/ehrql/blob/0484d0f060afbb7b69555aa44d72c9a5b034a806/tests/integration/backends/test_tpp.py) will fail until you add a test which exercises this table i.e. writes some data and checks that the SQL returns what's expected.

We should also include a helper method which filters it to just the eFI values, so something like:
```py
def electronic_frailty_index(self):
    return (
        self.where(self.algorithm_description == "UK Electronic Frailty Index (eFI)")
        .where(self.algorithm_version == "1.0")
    )
```

For reference, here's the equivalent Cohort Extractor code:
https://github.com/opensafely-core/cohort-extractor/blob/eebf49133434c3b42a7dcca94dab8c423f6d54f7/cohortextractor/tpp_backend.py#L1343

And docs:
https://docs.opensafely.org/legacy/study-def-variables/#cohortextractor.patients.with_these_decision_support_values

And here's the associated Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1732870106005209

rw251:
Closes #2264 

Some thoughts:
- The `AlgorithmType` column is used to join the tables, but I've also exposed it to users. Given it's an internal TPP thing I'm not sure that's necessary - but then again to pick a particular algorithm you either need to search by `algorithm_type` or `algorithm_description` - both of which are defined by TPP so perhaps not much difference
- Are there any checks missing? It feels a bit uncomfortable exposing a new table but not being able to actually test the code on the actual database. Especially because the way we identify the EFI is by searching for a hard-coded string definition and version number.
- For the EFI, I've copied code from @evansd which looks for descriptions matching "UK Electronic Frailty Index (eFI)". I just want to double check this is the correct string. In cohort extractor the search was actually done by looking for an `AlgorithmType` of `1`, so perhaps we should do that.
- Do we have any way of checking the possible values in the `Type`, `Description` and `Version` columns. I think at the moment it is just v1 of the EFI, but it would be good to know if TPP change anything or add new ones.

### DIFF
--- a/docs/how-to/examples.md
+++ b/docs/how-to/examples.md
@@ -923,3 +923,27 @@ dataset.prescription_date = case(
 )
 dataset.define_population(patients.exists_for_patient())
 ```
+
+## Decision support values
+
+Examples for the [TPP decision support values table](../reference/schemas/tpp.md#decision_support_values).
+
+### Finding the most recent decision support value
+
+#### Finding each patient's EFI (electronic frailty index)
+
+```ehrql
+from ehrql import create_dataset
+from ehrql.tables.tpp import decision_support_values
+
+dataset = create_dataset()
+latest_efi_record = (
+  decision_support_values
+    .electronic_frailty_index()
+    .sort_by(decision_support_values.calculation_date)
+    .last_for_patient()
+)
+dataset.latest_efi = latest_efi_record.numeric_value
+dataset.latest_efi_date = latest_efi_record.calculation_date
+dataset.define_population(decision_support_values.exists_for_patient())
+```
--- a/docs/includes/generated_docs/schemas/tpp.md
+++ b/docs/includes/generated_docs/schemas/tpp.md
@@ -15,6 +15,7 @@ from ehrql.tables.tpp import (
     clinical_events,
     clinical_events_ranges,
     covid_therapeutics,
+    decision_support_values,
     ec,
     ec_cost,
     emergency_care_attendances,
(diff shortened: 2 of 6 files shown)


evansd (approved):
This all looks good to me, thanks.

The questions you raise are all good ones, and I think there are some reasonable answers which I'd like to go into but don't have time before the end of the day. I just thought it was worth at least approving the PR now and then we can discuss tomorrow.

evansd:
On the question of filtering on `AlgorithmType` vs `AlgorithmDescription`, the way I see it is this. Neither field has been formally specified to us as having stable values so there are, in principle, two failures modes:
 1. we try to return EFI results but something has changed and we don't return any results;
 2. we try to return EFI results but something has changed and we return _the wrong_ results.

Of the two, it seems to me that 1 is far more preferable. It's likely to be noticed immediately by the researcher, and even if it isn't missing values are likely to be overall less harmful than incorrect values. And filtering by the description is, I think, far more likely to fail in first way whereas filtering by the type number runs a greater risk of failing in the second way. That is, it's very unlikely that the string "UK Electronic Frailty Index (eFI)" will refer to anything other than the thing we want. But it's not inconceivable that "Algorithm Type 1" might do.

In fact, I might be minded to not expose AlgorithmType at all to users just to avoid the risk that they start relying on it.

On the question of what the values are, we have now [included these](https://github.com/opensafely/tpp-database-schema/pull/190) in the TPP database schema reports we generate. However, I don't think we've yet re-run the report. And we haven't included it in the set of things we [automatically fetch](https://github.com/opensafely-core/ehrql/blob/32f477b265a004826a3e1cd98be616871bc66099/tests/lib/update_tpp_schema.py#L73-L90) and include in the ehrQL repo. We should do that.

PR #2350, opened 2025-01-10
test and not-quite-right fix for assertion error in specialize()

rebkwok:


evansd (approved):
Thanks!

I don't think we need to wait for David's response before merging, although I still think it's worthwhile checking our reasoning with him.

Also, very sadly, you're going to need a pedantic commit prefix here to actually get the fix out :(

PR #2351, opened 2025-01-13
Upgrade dependencies and reformat after ruff update to 2025 style guide

rebkwok:
ruff 0.9.0 shipped a new style guide that introduces some [formatting changes](https://astral.sh/blog/ruff-v0.9.0)

Rule A005 (stdlib-module-shadowing) is no longer in preview as of v0.9.0, and now complains about file_formats/csv.py now. We ignore it since we're not using relative imports and this should not cause
problems.
https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#090


ehrql/docs/__main__.py
```diff
@@ -60,9 +60,9 @@ def render(docs_data, output_dir):
     reference_index = Path(__file__).parents[2] / "docs/reference/language.md"
     reference_index_text = reference_index.read_text()
     for filename in section_filenames:
-        assert (
-            filename in reference_index_text
-        ), f"{filename} not included by {reference_index}"
+        assert filename in reference_index_text, (
+            f"{filename} not included by {reference_index}"
+        )
```
evansd, line 65: I'm so pleased that assertions are now getting formatted "properly". The old style always made we wince a bit whenever I saw it.

tests/integration/file_formats/test_main.py
```diff
@@ -132,9 +132,7 @@ def test_read_rows_validates_categories(test_file):
     # the schema but with CSV we can only validate individual values
     errors = {
         "dataset.arrow": (
-            "Unexpected categories in column 'c'\n"
-            "  Categories: A, B\n"
-            "  Expected: X, Y"
+            "Unexpected categories in column 'c'\n  Categories: A, B\n  Expected: X, Y"
```
evansd, line 135: Ah, this is a bit of shame. I think the string is quite a lot less readable now. I wonder if we should use `textwrap.dedent` here?

tests/integration/utils/test_mssql_log_utils.py
```diff
@@ -58,7 +58,7 @@ def log(event, **kwargs):
 
     assert results == [(1,), (3,)]
 
-    assert log_lines[0] == ("[info   ] SQL:\n" "          SELECT 1")
+    assert log_lines[0] == ("[info   ] SQL:\n          SELECT 1")
```
evansd, line 61: Could lose the brackets now I think. It's just a bit confusing to have them when they're not doing anything.

tests/integration/file_formats/test_main.py
```diff
@@ -156,11 +154,7 @@ def test_read_rows_validates_categories_on_non_categorical_column(test_file):
     column_specs = TEST_FILE_SPECS.copy()
     column_specs["s"] = ColumnSpec(str, categories=("X", "Y"))
 
-    error = (
-        "Unexpected categories in column 's'\n"
-        "  Categories: a, b\n"
-        "  Expected: X, Y"
-    )
+    error = "Unexpected categories in column 's'\n  Categories: a, b\n  Expected: X, Y"
```
evansd, line 157: Maybe `textwrap.dedent` again here?

tests/unit/utils/test_sqlalchemy_query_utils.py
```diff
@@ -201,10 +201,7 @@ def test_generated_table_from_query():
     )
     table = GeneratedTable.from_query("some_table", query, schema="some_schema")
     assert str(sqlalchemy.schema.CreateTable(table)).strip() == (
-        "CREATE TABLE some_schema.some_table (\n"
-        "\tnumber INTEGER, \n"
-        "\tstring VARCHAR\n"
-        ")"
+        "CREATE TABLE some_schema.some_table (\n\tnumber INTEGER, \n\tstring VARCHAR\n)"
```
evansd, line 204: And `textwrap.dedent` again here?

evansd (approved):
Thanks! Really pleased they've sorted out the formatting of assertions.

There are a couple of places where I should probably have used `textwrap.dedent` but used multiple strings instead and now Ruff has put them all on one line. I think it's worth fixing these as it does make them quite a lot harder to read.

tests/unit/utils/test_sqlalchemy_query_utils.py
rebkwok: I've changed the others you pointed out (and a couple more), but I've left this one because with dedent it ends up with trailing spaces at the end of the lines, which `just fix` removes, and breaks the test

evansd: Uggh, yeah. Thanks/sorry!

PR #2352, opened 2025-01-13
Fix graph query test argument to be str not Path

bloodearnest:


evansd (approved):
Thanks for catching this!

PR #2355, opened 2025-01-14
Add markers for html table output.

bloodearnest:
We will then be able to use these markers to deliniate pre-formatted
html output from normal console output when rendering in the vscode
extension.


evansd (approved):
:+1: Nice and straightforward!

Sadly you'll need a `fix:` or `feat:` prefix on the commit to get a new release here though.

PR #2356, opened 2025-01-14
Refactor debug rendering

bloodearnest:
The sandbox method of evaluating and rendering ehrql objects was via
monkey patching various repr functions  The initial debug implementation
re-used this method, and manually called `repr()` from within the show()
call.

Now that we don't have a sandbox, all rendering can be done explicitly
by the show() function, so there is no need for repr monkey patching.
This refactor switches the evaluate/render steps to be done explicitly
in show(), rather than going via repr.

However, the repr monkey patching implementation used a closure to
capture the supplied renderer and engine details in the monkeypatched
function. As `show()` is a user imported static function, it cannot use
a closure. Instead, we formalize the DEBUG_QUERY_ENGINE global into
a new DEBUG_CONTEXT global, which is set via `activate_debug_context`,
and then available to use inside `show()`. This DEBUG_CONTEXT stores the
engine instance and the renderer, and knows how to render and object
with those.

Also includes some follow refactors around truncation and tests.

### DIFF
--- a/ehrql/renderers.py
+++ b/ehrql/renderers.py
@@ -27,7 +27,11 @@ def headtail(sequence: typing.Sequence, head: int = 0, tail: int = 0):
 
 def records_to_html_table(records: list[dict], head: int = 0, tail: int = 0):
     rows = []
-    headers = "".join([f"<th>{header}</th>" for header in records[0].keys()])
+    headers = (
+        "".join([f"<th>{header}</th>" for header in records[0].keys()])
+        if records
+        else ""
+    )
     head_rows, ellipsis, tail_rows = headtail(records, head, tail)
 
     def html_row(row):
@@ -54,7 +58,7 @@ def records_to_ascii_table(records: list[dict], head: int = 0, tail: int = 0):
     width = 17
     lines = []
 
-    headers = records[0].keys()
+    headers = records[0].keys() if records else []
 
     lines.append(" | ".join(name.ljust(width) for name in headers))
     lines.append("-+-".join("-" * width for _ in headers))
--- a/ehrql/renderers.py
+++ b/ehrql/renderers.py
@@ -1,134 +1,75 @@
-import re
+import typing
 
 
 START_MARKER = "<!-- start debug output -->"
 END_MARKER = "<!-- end debug output -->"
 
 
-def records_to_html_table(records: list[dict]):
+def headtail(sequence: typing.Sequence, head: int = 0, tail: int = 0):
+    if head == tail == 0:
+        return sequence, False, []
+
+    # Check we have enough rows to truncate to head and tail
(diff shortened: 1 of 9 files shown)


ehrql/renderers.py
```diff
@@ -1,134 +1,75 @@
...
 
-    tail_rows = rows[-tail:] if tail is not None else [ellipsis_row]
-    truncated_rows.extend(tail_rows)
 
-    # re-join the truncated rows with <tr>
-    truncated_rows = "<tr>" + "<tr>".join(truncated_rows)
-    return start + truncated_rows + end
+def records_to_html_table(records: list[dict], head: int = 0, tail: int = 0):
+    rows = []
+    headers = "".join([f"<th>{header}</th>" for header in records[0].keys()])
```
rw251, line 30: Could records ever be an empty list? If so then I think this throws an error now which wouldn't have happened before.

ehrql/renderers.py
```diff
@@ -1,134 +1,75 @@
...
 
-    # Build the list of rows we need to include, with ellipsis rows where necessary
-    truncated_rows = table_rows[:headers]
+    html_rows = "".join(rows)
 
-    head_rows = (
-        table_rows[headers : headers + head] if head is not None else [ellipsis_row]
-    )
-    truncated_rows.extend(head_rows)
+    return f"{START_MARKER}<table><thead><tr>{headers}</tr></thead><tbody>{html_rows}</tbody></table>{END_MARKER}"
```
rw251, line 49: Do we still need the `START_MARKER` and `END_MARKER`? Or were they just so the previous regex would find the table?

tests/unit/test_debugger.py
```diff
@@ -21,36 +20,59 @@ def date_serializer(obj):
...
     expected_output = textwrap.dedent(
         """
-        Show line 32:
+        Show line 3:
 
         """
     ).strip()
 
-    show("Hello")
+    exec(
```
rw251, line 36: Nice

rw251 (approved):
Some good code simplification there.

ehrql/renderers.py
bloodearnest: Yes, they are not for the regex, but for the vscode extension to know where pre-formatted html tables start and end (this bit is what I am working on now).

ehrql/renderers.py
bloodearnest: Good catch, will fix.  It might be better to supply the header info independently of the rows, so we can show headers for an empty table, but I am not sure how to do that, so I will go for just not breaking.

ehrql/renderers.py
rw251: Ah of course. I'd naively just searched the ehrql code space for the marker and forgot about the extension.

PR #2359, opened 2025-01-16
Support as_int on bool series

### LINKED ISSUE #1313: Expand `CastToInt` operation to accept `Series[bool]`
For the purposes of calculating measures we sometimes need to convert a boolean series into an integer series (where `False -> 0` and `True -> 1`). We currently do this with a `Case` expression:
https://github.com/opensafely-core/ehrql/blob/11e94d97067cba1aac810bafb61cbcc549c8bd5a/ehrql/measures/calculate.py#L159-L167

But it would be more efficient to do this in SQL with a simple CAST expression.

We'd need to change the type signature here:
https://github.com/opensafely-core/ehrql/blob/6749f86f98622a44004e3b93946056e5069c45ff/ehrql/query_model/nodes.py#L336-L337

To something like:
```py
class CastToInt(Series[int]):
    source: Series[Numeric] | Series[bool]
```

We'd need to copy the `as_int()` method to the `BoolFunctions` class:
https://github.com/opensafely-core/ehrql/blob/6749f86f98622a44004e3b93946056e5069c45ff/ehrql/query_language.py#L293-L294

The `cast_to_int` generator in the generative tests will need updating so that it will generate boolean series as an input, not just numeric series:
https://github.com/opensafely-core/ehrql/blob/6749f86f98622a44004e3b93946056e5069c45ff/tests/generative/variable_strategies.py#L224-L227

And then finally we'll need to write some appropriate spec tests.

It may well be that the existing query engine implementations will work out of the box without any further changes. But if not, the spec tests should tell us.

rebkwok:
Fixes #1313 

- support using `is_int` on a bool series
- add spec test
- Update trino engine (has its own custom cast_to_int method to round floats towards zero, which isn't applicable for booleans)
- Replace the case statement for casting bools to ints in the measures framework

### DIFF
--- a/ehrql/query_model/nodes.py
+++ b/ehrql/query_model/nodes.py
@@ -376,7 +376,7 @@ class Function:
 
     # Casting numeric types
     class CastToInt(Series[int]):
-        source: Series[Numeric]
+        source: Series[Numeric] | Series[bool]
 
     class CastToFloat(Series[float]):
         source: Series[Numeric]
--- a/tests/spec/bool_series_ops/test_conversion.py
+++ b/tests/spec/bool_series_ops/test_conversion.py
@@ -1,7 +1,7 @@
 from ..tables import p
 
 
-title = "Convert a float value"
+title = "Convert a boolean value to an integer"
 
 table_data = {
     p: """
@@ -16,7 +16,7 @@ table_data = {
 
 def test_bool_as_int(spec_test):
     """
-    Booleans are converted to 0 or 1.
+    Booleans are converted to 0 (False) or 1 (True).
     """
     spec_test(
         table_data,
--- a/tests/spec/bool_series_ops/test_conversion.py
+++ b/tests/spec/bool_series_ops/test_conversion.py
@@ -1,0 +1,25 @@
+from ..tables import p
+
+
+title = "Convert a float value"
+
+table_data = {
+    p: """
+         | b1
+       --+----
+       1 |  T
+       2 |
+       3 |  F
+       """,
+}
+
+
+def test_bool_as_int(spec_test):
+    """
+    Booleans are converted to 0 or 1.
+    """
+    spec_test(
+        table_data,
+        p.b1.as_int(),
+        {1: 1, 2: None, 3: 0},
+    )
--- a/docs/includes/generated_docs/language__series.md
+++ b/docs/includes/generated_docs/language__series.md
@@ -119,6 +119,14 @@ status = status_code.map_values(
 ```
 </div>
 
+<div class="attr-heading" id="BoolPatientSeries.as_int">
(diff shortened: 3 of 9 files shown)


tests/spec/bool_series_ops/test_conversion.py
```diff
@@ -0,0 +1,25 @@
...
+
+table_data = {
+    p: """
+            | b1
+        --+----
+        1 |  T
+        2 |
+        3 |  F
+        """,
+}
```
madwort, line 14: tiny nit, but can we line this up?

```suggestion
    p: """
          | b1
        --+----
        1 |  T
        2 |
        3 |  F
        """,
}
```

madwort (approved):
Code looks good, although because it's so end-user focused I'd like to see some more documentation - e.g. it should definitely be mentioned https://docs.opensafely.org/ehrql/reference/features/#7-operations-on-boolean-series & maybe also as an alternative option here https://docs.opensafely.org/ehrql/reference/features/#1113-case-with-boolean-column ?

rebkwok:
> Code looks good, although because it's so end-user focused I'd like to see some more documentation - e.g. it should definitely be mentioned in https://docs.opensafely.org/ehrql/reference/features/#7-operations-on-boolean-series & maybe also as an alternative option in https://docs.opensafely.org/ehrql/reference/features/#1113-case-with-boolean-column ?

Thanks, I'd forgotten to add the new spec test file to the toc.py so the reference docs weren't being pulled in. Also added a comment in the case spec test.

madwort:
wow, not sure I'd really seen that test/docs tooling, that's brill (& the docs look much better now thanks)

ehrql/query_model/nodes.py
```diff
@@ -376,7 +376,7 @@ class FloorDivide(Series[int]):
 
     # Casting numeric types
     class CastToInt(Series[int]):
-        source: Series[Numeric]
+        source: Series[Numeric] | Series[bool]
```
evansd, line 379: You could save **eight whole characters** by writing this as:
```py
Series[Numeric | bool]
```
_Eight_, Becky! What were you thinking?

evansd (commented):
Great, thanks for doing this!

One other thing it might be worth doing (non-urgently) is extending this bit of the generative tests so it produces examples of bools being cast to ints:
https://github.com/opensafely-core/ehrql/blob/b4adab29f73edfeb9b788cd6f77878eb4f6f5efd/tests/generative/variable_strategies.py#L290-L293

We've got code in the tests which will moan at you if you add a completely new operation and don't include it in the generative tests. But we don't have anything which reminds you to update the tests if you extend an existing operation (I don't quite know how we'd do that to be honest) so it's one of those things which we inevitably forget.

PR #2362, opened 2025-01-17
Allow `fetch_table_in_batches` to work without unique key

evansd:
One significant thing which will change when we add support for event-level data is that not every set of results we want to download will have a unique `patient_id` column which we can use for efficient batching. We will still have an indexed `patient_id` column, but we can't depend on it being unique.

This PR adds an alternative approach which will work with such tables. It's a little bit more involved, particularly given that it needs to handle unstable sort orders, but I'm reasonably confident it's correct (especially since I've let Hypothesis do its worst on it).


rebkwok (approved):
Thanks for the detailed comments, made reviewing this much easier!

PR #2363, opened 2025-01-17
Update QueryEngine API to support multiple results tables

evansd:
This PR refactors the query engine API and the plumbing around it such that it _could_ in principle support returning event-level data if only we had some datasets which returned more than one query.

The one remaining piece of plumbing that needs doing is upgrading the `show()` debug command to output multiple tables when called on a dataset with event-level data. But that's not a total blocker so I'm going to defer working on that for now.

### Query Engine API

The major change to the query engine API is replacing the `get_results()` method, which returned an iterator over a single table of results, with `get_results_tables()` which returns a nested iterator over multiple tables of results.

To avoid massive changes to the test suite (and particularly to the under-development dummy data tests) we retain a `get_results()` method which asserts that we only have a single table of results and returns it as previously.

The SQL engines also change their `get_query()` method to `get_queries()` to support multiple queries.

### Plumbing

The major change here is supporting writing more than one table of output. If ehrQL sees that your dataset definition is going to produce multiple output tables then it will require that you give it an output path like `path/to/directory/:arrow`, with the final suffix indicating the type of file to produce.

If you don't supply any output path at all then it will write the tables to the console, one after the other. This involves adding a dedicated `console` output handler which at the moment is just a tiny wrapper around the CSV handler but could be updated to produce something more user-friendly like pretty-printed ASCII tables.

### DIFF
--- a/ehrql/file_formats/main.py
+++ b/ehrql/file_formats/main.py
@@ -35,9 +35,7 @@ def write_rows(filename, rows, column_specs):
     # whole thing into memory. So we wrap it in a function which draws the first item
     # upfront, but doesn't consume the rest of the iterator.
     rows = eager_iterator(rows)
-    # We use None for stdout
-    if filename is not None:
-        filename.parent.mkdir(parents=True, exist_ok=True)
+    filename.parent.mkdir(parents=True, exist_ok=True)
     writer(filename, rows, column_specs)
 
 
@@ -123,10 +121,7 @@ def write_tables(filename, tables, table_specs):
 
 
 def get_file_extension(filename):
-    if filename is None:
-        # If we have no filename we're writing to stdout, so default to CSV
-        return ".csv"
-    elif filename.suffix == ".gz":
+    if filename.suffix == ".gz":
         return "".join(filename.suffixes[-2:])
     else:
         return filename.suffix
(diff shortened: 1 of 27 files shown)


ehrql/file_formats/main.py
```diff
@@ -48,6 +48,30 @@ def read_rows(filename, column_specs, allow_missing_columns=False):
 
 
 def read_tables(filename, table_specs, allow_missing_columns=False):
+    # If we've got a single-table input file and only a single table to read then that's
+    # fine, but it needs slightly special handling
+    if not input_filename_supports_multiple_tables(filename):
```
rebkwok, line 55: Should this also check if `filename` exists and raise a different error? It could be a non-existent directory, which would return `False` from `filename.is_dir()`, and then you'd get a misleading error if you were trying to read multiple tables

ehrql/file_formats/main.py
```diff
@@ -23,6 +24,9 @@
 
 
 def write_rows(filename, rows, column_specs):
+    if filename is None:
```
rebkwok, line 27: There's a check for `if filename is not None` on row 38 which should never be False now, I think

rebkwok: Ah, removed in next commit, nvm

rebkwok (approved):

ehrql/file_formats/main.py
evansd: Yep, very good point. I'll fix this.

evansd:
Thanks @rebkwok, all good catches as always!

PR #2367, opened 2025-01-21
Refactored autocomplete tests to increase speed

rw251:
The existing approach to testing autocomplete via the language server was:

1. Create a temporary file on disk with dummy lines
2. Notify the language server that we "opened" a file
3. For each test
    - Notify the language server that a line had changed to the thing we want to test
    - Ask the language server for completion, or type.

This was pretty slow because presumably the language server needs to re-parse the file each time.

This is now refactored so it's possible to:
1. Create a temporary file with all the content for a parameterized test
2. Notify the language server that we "opened" a file
3. For each test
    - Ask for completion, or type, for a given line, cursor position

This makes the run time go from 2-3 minutes, to 2-3 seconds.


evansd (approved):
> This makes the run time go from 2-3 minutes, to 2-3 seconds.
:rocket:

PR #2375, opened 2025-01-24
Ensure date of death dummy data values is weighted towards None

### LINKED ISSUE #2374: Next gen dummy data for measures produces smaller counts over time
For this measures definition:
https://github.com/opensafely/sodium-valproate-prescribing/blob/main/analysis/measures_base.py

See [slack thread](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1737131611370159)

rebkwok:
Fixes #2374 

- Ensure that None is always a possible value in a population subset
- When picking a date of death from a list of possible values, and one of those values is None, make None the most highly weighted choice (around 90%, with no other scientiifc basis other than that it's more or less consistent with the None rate in the legacy dummy data)

### DIFF
--- a/ehrql/dummy_data_nextgen/generator.py
+++ b/ehrql/dummy_data_nextgen/generator.py
@@ -66,6 +66,8 @@ class PopulationSubset:
             if n < len(result):
                 indices = self.random.sample(range(0, len(result)), n)
                 indices.sort()
+                if result[0] is None and 0 not in indices:
+                    indices = [0, *indices]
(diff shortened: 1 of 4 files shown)


evansd (approved):
These both seem like pretty sensible changes to me. I agree that it's still not 100% clear what the theoretically correct fix is (it might be this) but this definitely seems like a pragmatic improvement on the status quo. Thanks! (Both for the fix, and for tracking down the root of the problem in the first place.)

PR #2376, opened 2025-01-24
Various improvements to the ehrQL autocomplete tests

rw251:
Quite a few changes but hopefully each commit is comprehensible.

Changes include:
- A refactor to use the `ast` and `tokenizer` modules to parse the `autocomplete_definition.py` file rather than my custom parser
- Ensure that all table methods (e.g. `patients.age_on()`) are either tested or ignored
- Ensure that all methods defined in `query_language.py` are either tested or ignored. This includes methods on `Series` such as `Base.Series.is_in()` and `DateFunctions.is_after()`, and ensures that the method is tested for both a `PatientSeries` and an `EventSeries` (if it can be used on both). It also includes other methods such as `Duration.starting_on()`
- Replaced the tests that manually checked the dropdown completion for items on the `core` tables, with an exhaustive test that looks for all items on all tables.

### DIFF
--- a/tests/autocomplete/test_autocomplete.py
+++ b/tests/autocomplete/test_autocomplete.py
@@ -1,4 +1,5 @@
 import ast
+import inspect
 import re
 from enum import Enum
 from io import BytesIO
@@ -7,13 +8,17 @@ from tokenize import tokenize
 
 import pytest
 
-from ehrql import tables
+from ehrql import query_language, tables
 from ehrql.query_language import (
     BaseSeries,
+    EventSeries,
+    PatientSeries,
+    days,
     get_tables_from_namespace,
+    int_property,
 )
 from ehrql.query_model.nodes import SelectColumn
-from ehrql.tables import core
+from ehrql.tables import core, emis, tpp
 from ehrql.utils.module_utils import get_submodules
 
 from .language_server import LanguageServer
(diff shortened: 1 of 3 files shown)


tests/autocomplete/test_autocomplete.py
```diff
@@ -377,3 +382,335 @@ def test_things_in_autocomplete_definition_file(
...
+    """
+    class_parents = {}
+    class_methods = {"object": []}
+    class_int_props = {"object": []}
+    series_methods = set()
+    series_int_props = set()
+    other_methods = {}
+    ql_classes = inspect.getmembers(query_language, inspect.isclass)
+    for class_name, class_obj in ql_classes:
+        if class_obj.__module__ == "ehrql.query_language":
```
evansd, line 448: This might be slightly more readable written as a guard clause i.e.
```suggestion
        if class_obj.__module__ != "ehrql.query_language":
            continue
```
And then de-indenting the inner loop.

It  makes immediately clear the intention that you just want to ignore anything which isn't part of the query language module, which otherwise isn't obvious until you get to the end of the (quite long) conditional clause.

evansd (approved):
This is really great. As you say, there's quite a lot of new code here. But it's all so clearly laid out and well-commented that I had no problem following along.

I also really appreciated the clarity of the asserts that get raised when things are missed out. We definitely going to hit these when we start adding stuff to the query language and so it's worth the effort you've put in to make it obvious what to do next.

I made one tiny, stylistic comment for the sake of having something to say :grinning:

tests/autocomplete/test_autocomplete.py
rw251: Makes sense - thanks.

PR #2379, opened 2025-01-28
Monitor changes in TPP decision support ref table

rw251:
Added the values from the `DecisionSupportValuesReference` table to the script that creates a pull request if the TPP schema or data dictionary changes.

I haven't committed the csv file that will be created by the github action to double check that it works as expected - e.g. a PR will be raised after this is committed.

### DIFF
--- a/.github/workflows/update-tpp-schema.yml
+++ b/.github/workflows/update-tpp-schema.yml
@@ -30,14 +30,20 @@ jobs:
         id: create_pr
         uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # v7.0.6
         with:
-          add-paths: tests/lib/tpp_schema.*
+          add-paths: tests/lib/*
           branch: bot/update-tpp-schema
           base: main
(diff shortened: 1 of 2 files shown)


.github/workflows/update-tpp-schema.yml
```diff
@@ -30,14 +30,20 @@ jobs:
         id: create_pr
         uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # v7.0.6
         with:
-          add-paths: tests/lib/tpp_schema.*
+          add-paths: tests/lib/*
```
madwort, line 33: on the face of it, this looks like it will catch too many files?

rw251: That's what I thought! (https://bennettoxford.slack.com/archives/C069YDR4NCA/p1737395525574229) but because this only gets called after `update_tpp_schema.py`, the only things that will have changed are related to that, and we want to commit everything that's changed.

madwort: aha yes!

evansd (approved):

PR #2384, opened 2025-01-29
Sum and group measures results in SQL

rebkwok:
Moves the summing and grouping for measures results into the SQL queries, so we only extract the summarised results, not every patient row. Makes serendipitous use of the new ability to return multiple results tables.

Also adds a test for the wrong assumption I made in #2381 (that we can extract a non-varying variable using the union of all interval denominators as the population), This is not specifically related to this PR, but would catch the error if anyone was to make that assumption again in future.

### DIFF
--- a/ehrql/measures/calculate.py
+++ b/ehrql/measures/calculate.py
@@ -115,21 +115,67 @@ class MeasureCalculator:
             measures=self.grouped_sums,
         )
 
+        all_groups = {}
+        for measure in self.measures:
+            for k, v in measure.group_by.items():
+                if k not in all_groups:
+                    all_groups[k] = v
+        self.all_groups = all_groups
+
+        self.grouping_levels = {
+            measure.name: self.get_grouping_level(measure) for measure in self.measures
+        }
+
     def get_results(self, query_engine):
         for interval in self.intervals:
             results = self.get_results_for_interval(query_engine, interval)
+
             for measure, numerator, denominator, *groups in results:
-                group_dict = dict(zip(measure.group_by.keys(), groups))
+                group_dict = dict(zip(self.all_groups.keys(), groups))
                 yield measure, interval, numerator, denominator, group_dict
 
     def get_results_for_interval(self, query_engine, interval):
         # Build the query for this interval by replacing the interval start/end
         # placeholders with actual dates
         dataset = substitute_interval_parameters(self.placeholder_dataset, interval)
+        groups_count = len(self.all_groups)
 
-        for i, table in enumerate(query_engine.get_results_tables(dataset)):
+        for table in query_engine.get_results_tables(dataset):
             for row in table:
-                yield self.measures[i], *row
+                grouping_level = row[-1]
+                row = row[:-1]
+
+                for j, measure in enumerate(self.measures):
+                    # To determine which measure(s) this row applies to, we look at its
+                    # grouping level which is a numeric representation of the subset of
+                    # all group_bys that are applied to this measure
+                    # If its level matches the row's grouping level, values for this measure
+                    # should be extracted from the row
+                    if self.grouping_levels[measure.name] != grouping_level:
+                        continue
+                    # Each row contains a pair of numerators and denominators for every
+                    # measure, in the same order as self.measures, so we can use the order to
+                    # extract the correct numerator and demoninator.
+                    numerator = row[j * 2]
+                    denominator = row[(j * 2) + 1]
+                    yield (
+                        measure,
+                        numerator,
+                        denominator,
+                        *row[(len(row) - groups_count) :],
+                    )
+
+    def get_grouping_level(self, measure):
+        # Calculate the level of grouping for each measure in the same way as the
+        # grouping ID is calculated - i.e. integer representation of a string of
+        # 0s and 1s for each column, where a 1 indicates that the column is NOT a grouping column
+        # https://learn.microsoft.com/en-us/)sql/t-sql/functions/grouping-id-transact-sql?view=sql-server-ver16
+        if not self.all_groups:
+            return 0
+        return int(
+            "".join(["0" if gp in measure.group_by else "1" for gp in self.all_groups]),
+            2,
+        )
 
     def add_measure(self, measure):
         # Record denominator and intervals from first measure
@@ -153,9 +199,7 @@ class MeasureCalculator:
         # Return the name and position of `variable` in the variables dict, adding it if not
         # already present
         try:
-            return next(
-                k for k, v in enumerate(self.variables.items()) if v == variable
-            )
+            return next(k for k, v in self.variables.items() if v == variable)
         except StopIteration:
             index = len(self.variables)
             key = f"column_{index}"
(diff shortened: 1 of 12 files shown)


ehrql/query_engines/base_sql.py
```diff
@@ -85,6 +86,72 @@ def get_next_id(self):
...
+            ]
+            group_cols = [
+                gp if gp in group_by_cols else sqlalchemy.null
+                for gp in all_group_by_cols
+            ]
+            grouping_id = get_grouping_level_as_int(all_group_by_cols, group_by_cols)
+
+            measure_queries.append(
+                sqlalchemy.select(*sum_overs, *group_cols, grouping_id).group_by(
+                    *group_by_cols
```
rebkwok, line 149: I don't think we actually need to include the group by columns anymore, we can infer them from the grouping id when we fetch the results. Not sure if that makes any difference though

ehrql/utils/sequence_utils.py
```diff
@@ -0,0 +1,18 @@
+def ordered_set(sequence):
+    """
+    Deduplicates a sequence, maintaining order
+    """
+    return list(dict.fromkeys(sequence))
+
+
+def get_grouping_level_as_int(all_groups, group_subset):
```
rebkwok, line 8: This isn't a sequence util really, but I've put it here until I work out how to categorise it

ehrql/query_engines/mssql.py
```diff
@@ -151,27 +153,86 @@ def create_inline_table(self, columns, rows):
...
+            grouping_sets.append(grouping_set)
+        all_group_bys = ordered_set(iter_flatten(grouping_sets))
+        grouping_sets = [
+            sqlalchemy.tuple_(*grouping_set) for grouping_set in grouping_sets
+        ]
+        measures_query = sqlalchemy.select(
+            *all_sum_overs,
+            *all_group_bys,
+            sqlalchemy.func.grouping_id(*all_group_bys).label("grp_id"),
+        ).group_by(sqlalchemy.func.grouping_sets(*grouping_sets))
```
evansd, line 194: :ok_hand: It's amazing that this all actually works. So good seeing this coming together!

ehrql/query_engines/mssql.py
```diff
@@ -151,27 +153,86 @@ def create_inline_table(self, columns, rows):
...
+
+        measures_results_table = temporary_table_from_query(
+            # Write the measures results to a global temporary table
+            f"##measures_results_{self.global_unique_id}",
+            measures_query,
+            index_col="num_0",
+        )
+        return [sqlalchemy.select(measures_results_table)]
```
evansd, line 202: I _think_ that if you moved the call to `get_measure_results()` inside `get_results_queries()` as discussed somewhere above then you'd get this behaviour for free.

ehrql/query_engines/mssql.py
```diff
@@ -151,27 +152,84 @@ def create_inline_table(self, columns, rows):
...
+            index_col = "patient_id"
             results_table = temporary_table_from_query(
                 # The double `##` prefix here makes this a global temporary table, i.e.
                 # one accessible to other sessions, hence we need a globally unique
                 # name. This allows us to use a separate connection to retrieve results
                 # which gives us more robust retries.
                 f"##results_{self.global_unique_id}_{n}",
                 results_query,
-                index_col="patient_id",
+                index_col=index_col,
```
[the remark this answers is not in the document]
evansd: I think we could actually change it to `index_col = 0` i.e. index by the first column regardless of what it's called. I only had `patient_id` in there for the sake of being explicit.

In fact, you could leave it out entirely as `index_col = 0` is the default, but it's probably better to be clear that an index is being created here.

ehrql/query_engines/mssql.py
```diff
@@ -151,27 +153,86 @@ def create_inline_table(self, columns, rows):
...
         results_queries = self.get_queries(dataset)
 
+        if dataset.measures:
+            assert len(results_queries) == 1, (
+                "Measures queries can only be applied to a single results table"
+            )
+            results_query_table = results_queries[0].get_final_froms()[0]
+            results_queries = self.get_measure_queries(
+                dataset.measures, results_query_table
+            )
```
evansd, line 234: I think you can probably get away without doing anything special here and just call `subquery()` on the results query as you do for the SQLite version. It does make the eventual query a bit convoluted because we're doing something like:
```sql
SELECT complicated_grouping_stuff() FROM
(
  SELECT * FROM some_table 
) AS subquery
```
But I'm pretty sure the query planner will see straight through that and will behave no differently than if we'd written:
```sql
SELECT complicated_grouping_stuff() FROM some_table
```

If it does turn out to make a difference then we can always add code to the BaseSQL engine which spots this specific pattern and turns it into the simpler one.

ehrql/query_engines/mssql.py
```diff
@@ -210,23 +270,35 @@ def get_results_stream(self, dataset):
                 )
 
                 for i, results_table in enumerate(results_tables):
+                    # fetch_table_in_batches needs a key column to sort on in order to
```
evansd, line 273: Ah yeah, this is exactly what my new PR lets us address (if we add a new `QueryType.AGGREGATE` for these sorts of queries).

ehrql/utils/sequence_utils.py
evansd: Finding appropriate locations for the random utility functions you end up needing is one of the genuinely hard problems in computer science.

You could arguably stick it in `math_utils` I mean, it's got numbers in ,right?

ehrql/measures/calculate.py
```diff
@@ -97,99 +104,111 @@ def __init__(self, measures):
...
         self.placeholder_dataset = Dataset(
-            population=self.population, variables=self.variables, events={}
+            population=self.population,
+            variables=self.variables,
+            events={},
+            measures=self.grouped_sum,
         )
 
+        self.all_groups = dict(
+            ordered_set(
```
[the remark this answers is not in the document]
rebkwok: In my head I needed the ordered set to maintain the order, because the last instance of a key will update the dict...but it'll still maintain order from the first instance, right. Will update

ehrql/measures/calculate.py
```diff
@@ -97,99 +104,111 @@ def __init__(self, measures):
...
+                    # grouping level which is a numeric representation of the subset of
+                    # all group_bys that are applied to this measure
+                    # If its level matches the row's grouping level, values for this measure
+                    # should be extracted from the row
+                    if self.grouping_levels[measure.name] != grouping_level:
+                        continue
+                    # Each row contains a pair of numerators and denominators for every
+                    # measure, in the same order as self.measures, so we can use the order to
+                    # extract the correct numerator and demoninator.
+                    numerator = row[j]
```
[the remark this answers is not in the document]
rebkwok: Yes, I realised this, but it'd need some extra work to track which numerator goes with which measure (and group by), because at the moment it just relies on the order of everything. Probably not much though

ehrql/query_engines/mssql.py
```diff
@@ -151,27 +153,86 @@ def create_inline_table(self, columns, rows):
         ]
         return table
 
+    def get_measure_queries(self, grouped_sum, results_query):
```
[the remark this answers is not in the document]
rebkwok: It didn't occur to me to check whether trino supported it or not :)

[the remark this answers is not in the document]
evansd: I learned about it as a standard feature of modern SQL so the surprise to me was that MSSQL supported it (especially the version we're using)!

rebkwok:
superceded by #2422

PR #2387, opened 2025-01-30
Expose the `Discharge_Destination` column of the APCS table

### LINKED ISSUE #2378: Expose the `APCS.Discharge_Destination` column
The column exists in the underlying data:
https://reports.opensafely.org/reports/opensafely-tpp-database-schema/#APCS

But is not currently exposed in ehrQL:
https://docs.opensafely.org/ehrql/reference/schemas/tpp/#apcs

This is a categorical column and we should map the opaque numbers it contains to meaningful categories. The meaning of the codes is given in [our documentation](https://docs.opensafely.org/data-sources/apc/#discharge-destination-reference-table), which is derived from the values given in the [NHS data dictionary](https://www.datadictionary.nhs.uk/attributes/discharge_destination.html?hl=discharge%2Cdestination). Additionally we should add this column to the [report](https://github.com/opensafely/tpp-database-schema/blob/d098887624f37321148375f1be07bc1d1e0565da/analysis/query_tpp_categorical_columns.sql#L51-L63) on categorical columns so we can be sure that it contains the values we expect.

It's needed for a short data report to compare mortality capture between ONS, primary care data, and HES. See related Slack thread:
https://bennettoxford.slack.com/archives/C33TWNQ1J/p1737740156805009

---

**Update:** Consider adding `Method_of_Discharge` at the same time, it's not loads of extra work (see thread).

rw251:
Closes #2378 

The ticket suggested we map the destination codes (e.g. 19, 29, 80) to something more meaningful. I think we should actually do that as a separate exercise. We currently don't do this mapping for the other codes in this table (e.g. admission method, patient classification) and just point people [here](https://docs.opensafely.org/data-sources/apc/#reference-tables). I think it's a hardish problem to come up with names that are suitably terse, appropriate and meaningful - and also whether to provide any level of grouping. I think it's reasonable to always expose the underlying code so I don't think we'd ever want to hide it.

I've called it `discharge_destination_code` so that in future we can define a property like `discharge_destination` which is the mapped names. But that does mean if we mapped the existing `admission_method` field then we might introduce a breaking change by renaming it `admission_method_code`. Alternatively we could call it `discharge_destination`, and then think of a better name for the mapped values.


evansd (approved):
Yep, you're quite right! I hadn't considered (or had forgotten) that we're not doing any mapping for the other fields. I agree that doing the mapping should be a separate task, and something we should do consistently for all the fields when we do it.

And on that note, I think your alternative suggestion of calling it `discharge_destination` rather than `discharge_destination_code` is the right one here for the sake of consistency. Maybe we'll end up calling the mapped fields `X_name` or something. But whatever we do will be easier if we've got a consistent base to start from.

Issue #506, opened 2022-05-27
Better autocomplete behaviour in ehrQL


evansd:
The dream of having a fully statically typed query language almost certainly isn't possible due to the limitations of Python's type system and the complexities of the query model. But we should still try to make the editing experience as helpful as possible.

Our target here should be VS Code with the default Python extension installed. That's what many users have locally, but it's also what's provided in online IDEs like Gitpod and Github Codespaces.

I think it's acceptable here for the autocomplete behaviour here to be a bit "optimistic" in that it completes method names that _probably_ exist on the object in question, even if it turns out that they don't always. At least, I think that's a lot better than not showing anything because we can't be certain exactly what methods exist. Especially if we combine that with decent error reporting as in #505.

I don't know exactly how we'd achieve that though. Have some type which has the all the methods and then use that as the type signature on everything?

It would also be really good to get column name completion if that's at all possible.

benbc:
I had a spike query language which used parameterized types to give some of what we'd want here. I eventually ran into a limitation that type variables and parameterized types don't compose as you'd like in Python.

It's possible that this will be a bit easier now we've got parallel hierarchies and mixins in the ehrQL definition.

evansd:
Following on from the changes in #663 I have a proof-of-concept for how autocomplete and type checking could be made to work in ehrQL. In fact it works in the purely in-browser version of VSCode, which was slightly surprising to me.

You should be able to open it directly in the Github editor [here](https://github.dev/opensafely-core/databuilder/blob/evansd/autocomplete-demo/autocomplete_demo.py). Or you can also visit [vscode.dev](https://vscode.dev) and copy-past the example from:
https://github.com/opensafely-core/databuilder/blob/evansd/autocomplete-demo/autocomplete_demo.py

In either case, you should be prompted to install the Python extension and click "Install anyway" when it warns you that functionality will be limited in the browser. If you've just installed it then you might need to wait 10 seconds or so before autocomplete becomes available.

You should find that the appropriate columns are offered as completions on each frame and, for each of those, the methods offered are just the appropriate ones for the type and dimension of the series.

The key trick here was using a descriptor to access series from frames and then adding lots of signature `@overloads` to the descriptors `__get__` method which say things like "given an initial type of X and an EventFrame, return a series of type XEventSeries".

There's obviously a combinatorial explosion of overloads here as we add more types, and as we start extending this approach to the rest of the ehrQL methods.

[...]

PR #2388, opened 2025-01-30
Added autocomplete for `first_for_patient()`/`last_for_patient()`

rw251:
This is still a work in progress. We still need to add some sort of way of checking that the things in `./patient_level_table_definitions` match 1-2-1 with the `EventFrame` tables.
- could be a check in the `@table` property, but would need some gymnastics to exclude things like tables created in tests
- maybe better as a test
- would need to exclude things like `tpp.medications` which extends from `core.medications` (rather than `EventFrame`) and adds a column. This doesn't work with the existing implementation. Could be added in future.
- maybe the `just update-patient-level-table-definitions` could be part of the `devenv` command so it would get called e.g. when tests were run


rw251:
Discussion [here](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1738083494891479) on the awkward circularity of adding new tables of this approach.

Possible better alternative and write up of findings [in this ticket](https://github.com/opensafely-core/ehrql/issues/506). Closing for now.

PR #2389, opened 2025-01-30
Expose the APCS `discharge_method` field

rw251:
Requested in the same slack thread as the `discharge_destination`: https://bennettoxford.slack.com/archives/C33TWNQ1J/p1737740156805009


ehrql/tables/tpp.py
```diff
@@ -284,6 +284,16 @@ class apcs(EventFrame):
...
+    discharge_method = Series(
+        str,
+        description=(
+            "The method of discharge from a hospital provider spell. "
+            "Currently the code is a 1 character string (e.g. 1 = discharged on clinical advice, "
+            "4 = died). "
+            "Refer to [this documentation](https://www.datadictionary.nhs.uk/attributes/discharge_method.html) "
+            "for a full list of codes."
+        ),
+    )
```
rebkwok, line 296: Do we know if these codes can change? If not, would it be worth adding a `Constraint` (possibly for discharge_destination too)? If the constraint is a complete list of valid codes, dummy data would be produced with valid codes. Or if it was a regex constraint, it'd at least be valid-ish rather than just a random string.

rw251: That's a good point. They are fixed but could change in the future. However, until the tpp report runs I don't know whether TPP have the old (1-5) or new (1-7) categories. You're right we should add them, but none of the other similar fields (admission method, discharge destination) have constraints so I wonder if we should do it as a separate, single exercise. Therefore I've raised a ticket (#2391) to do 2 things consistently across all fields like this:
- adding constraints
- adding helpers so that people can interact with meaningful strings instead of the category codes
Do you think that's ok?

rebkwok: Sounds good, I've added that ticket to the triage list

rebkwok (approved):

Issue #2345, opened 2025-01-07
Investigate `DBPROCESS is dead or not enabled` errors


evansd:
This is a generic client-side error which is raised when the client can no longer talk to the database. It is [unrecoverable](https://github.com/opensafely-core/ehrql/issues/2260) for us because we rely on session-scoped temporary tables which are not available if we reconnect.

This causes frustration and confusion for users because their jobs fail for no discernible reason (often after having run for a long time). The impact is even worse for measures jobs because they do multiple data extractions over an extended period of time and failure of any one will cause the whole job to fail.

Some initial thoughts and observations:

 * We only see this error during the final "fetch results" stage of the data extraction.
 * We previously execute many long-running queries (some _very_ long-running) which write their results to temporary tables without issue.
 * Fetching results is the first stage at which we start pulling down significant amounts of data over the connection.
 * We fetch results in (relatively) small batches precisely to avoid this kind of issue, but it doesn't seem to help.
 * These errors are often correlated in time in that separate but concurrently running jobs will often experience them at approximately the same time.
 * We'll often also see a cluster of them over a period of a few hours.
 * This suggests it might be an issue with either the database server or the network, rather than anything related to the specific queries being run.

Related Slack conversations:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1734085399746199
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1734703267369369

evansd:
Here are the timings of all the errors we have seen over the last year:
![Image](https://github.com/user-attachments/assets/c9824f8f-d498-4f95-91a5-dd4ff5cd2b43)

```sh
cd /srv/high_privacy/logs
rg DBPROCESS -IN --glob '202(4,5}-*/**' | cut -c-19 | sort > dbprocess-errors.txt
column dbprocess-errors.txt
```

And OCRed results:
```
2024-01-14T00:57:01	2024-04-03T12:36:26	2024-04-22T16:19:46	2024-06-16T18:52:04	2024-11-07T17:27:44
2024-01-14T00:57:01	2024-04-03T12:36:26	2024-04-22T16:19:46	2024-06-16T18:52:04	2024-11-07T17:27:44
2024-03-20T18:21:04	2024-04-03T12:39:08	2024-04-23T19:49:44	2024-06-16T21:07:42	2024-11-07T17:35:25
[...]

PR #2392, opened 2025-02-04
Use global temporary table for MSSQL results

evansd:
This allows us to use a separate connection for downloading results which can safely reconnect and resume on error without losing the temporary table. This gives us the robustness we used to get when writing results to a persistent table but without the locking and performance problems which that gave us (and with the benefit of automatic cleanup should a job get interrupted).

This functions as a workaround to the problem described in #2345.

There are various bits of database behaviour which we've had to confirm empirically:
 * Running a `SELECT * INTO ##some_table ...` query doesn't take a global lock which prevents other such queries running simultaneously (as long as they are not run inside an explicit transaction).
 * Querying the table from a separate connection works, and allows us to successfully retry following a "DBPROCESS" error.
 * The table is automatically dropped if the ehrQL job is killed.

Associated Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1738586172380799

### DIFF
--- a/tests/integration/query_engines/test_mssql.py
+++ b/tests/integration/query_engines/test_mssql.py
@@ -78,7 +78,7 @@ def wrap_select_queries():
                 # itself. But I've tried and failed to do so, partly because the
                 # `pymssql` driver we use is a compiled library and so less amenable to
                 # monkey patching.
-                self._dbapi_connection = None
+                self.invalidate()
                 raise
         return original(self, *args, **kwargs)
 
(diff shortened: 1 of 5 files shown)


tests/integration/query_engines/test_mssql.py
```diff
@@ -78,7 +78,7 @@ def wrapper(self, *args, **kwargs):
                 # itself. But I've tried and failed to do so, partly because the
                 # `pymssql` driver we use is a compiled library and so less amenable to
                 # monkey patching.
-                self._dbapi_connection = None
+                self.invalidate()
```
bloodearnest, line 81: This looks completely reasonable, but its not clear why its included in this change?

bloodearnest (approved):
Remarkably small change :+1:

tests/integration/query_engines/test_mssql.py
evansd: Ah yeah, sorry. It's because calling `invalidate()` causes the connection to actually get closed rather than left in the connection pool. With this in place the old retry code fails, but the new code succeeds.

PR #2394, opened 2025-02-10
Add a link to the walkthrough video

eli-miriam:
Turns out embedding is hard, so I've opted for adding in the thumbnail and linking it to the YouTube directly.


rebkwok (approved):
Seems fine to me. We do do youtube embeds in the main docs (e.g. https://github.com/opensafely/documentation/blob/main/docs/index.md?plain=1#L16) but it involves its own extra css and js so I'm assuming not worth it just for the tutorial link

PR #2395, opened 2025-02-13
Fix and simplify gentest randomisation toggle

evansd:
The comment correctly described the behaviour we wanted, but the code logic was backwards. This is easy enough to do when you've got an env var called "RANDOMIZE" toggling a config value called "derandomize" so we switch the name of the env var to match.

This makes it clear that there's exactly one circumstance where we don't want randomisation and that's in the CI tests.

Related Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1739440049238879

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -74,27 +74,21 @@ results are the same.
 
 To get the most out of our generative tests we want them to run for a
 long time and to explore different parts of the query space each time
-they're run. But neither of these qualities are desirable in CI or in
-local test runs. For this reason, the default configuration generates
-only a small number of examples, of limited query depth, and in a
-determinstic fashion. They thus function more as a test that the
-generative test machinery is working correctly than as a serious attempt
-to do generative testing.
+they're run. But neither of these qualities are desirable in CI.
 
-By default, CI generates more examples than the local tests on the basis
-that the extra time is less noticeble there and it gives us a better
-chance of actully catching some novel bugs. But the examples are still
-deterministic for reasons of sanity preservation.
+For this reason, the CI configuration generates only a small number of
+examples, of limited query depth, and in a determinstic fashion. They
+thus function more as a test that the generative test machinery is
+working correctly than as a serious attempt to do generative testing.
 
-To do some "proper" generative testing you can run the command:
+When running locally, although randomisation is enabled by default,
+the number of examples is still very small (to keep test runtimes
+reasonable). To do some "proper" generative testing you can run the
+command:
 ```
 just test-generative
 ```
-which increases the example count even further and enables randomisation
-by setting:
-```
-GENTEST_RANDOMIZE=t
-```
+which increases the example count by setting `GENTEST_EXAMPLES`.
 
 It would be worth running these locally when adding new query model
 operations or making significant changes to a query engine. You may even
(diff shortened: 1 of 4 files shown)


tests/generative/test_query_model.py
```diff
@@ -89,7 +89,7 @@
     # In development we never want to run with derandomize because it's
     # less effective at finding bugs and, more importantly, has very slow
     # replay due to turning off the test database
-    derandomize=bool(os.environ.get("GENTEST_RANDOMIZE"))
+    derandomize=(not os.environ.get("GENTEST_RANDOMIZE"))
```
rebkwok, line 92: This will do what we want in CI, but only for the long-running generative tests, because `GENTEST_RANDOMIZE` is set in the `generative-tests.yml`. It isn't set for the subset of tests run with the standard `just test-all` in the main.yaml workflow, so presumably it is using derandomise there. It seems a bit weird to set it to the opposite of whatever it's set to when we're running in one specific CI workflow, and it means it's not doing what the comment says, and never running with derandomize in development (that's maybe just a typo in the comment which should say CI).  
Also the double-negative is making my head hurt, so I'm not convinced I've got the true/false-ness the right way round there anyway.

evansd: I agree the double-negative is weird here, but I think the behaviour is what we want, which is: never derandomize in development, always derandomize in CI unless explicitly told not to be the long-running tests.
```
python -c '\
for CI in ("t", None):
    for GENTEST_RANDOMIZE in ("t", None):
        derandomize=(not GENTEST_RANDOMIZE) if CI else False
        print(f"{CI=}, {GENTEST_RANDOMIZE=}, {derandomize=}")'
```

```
CI='t', GENTEST_RANDOMIZE='t', derandomize=False
CI='t', GENTEST_RANDOMIZE=None, derandomize=True
CI=None, GENTEST_RANDOMIZE='t', derandomize=False
CI=None, GENTEST_RANDOMIZE=None, derandomize=False
```

rebkwok: So we _do_ want to derandomize for the non-long-running generative tests?  Maybe it's the comment that's confusing then.  Should we add something to the developer docs to point out that in non-CI, this env variable has no effect? Currently there are docs that tell you that you can use it locally and suggest that you should when you add new query model operations or change query engines

evansd: Uh yeah you're right, the docs are misleading then, and also the behaviour is fundamentally confusing. I think maybe a better thing would be to invert this completely and have a `GENTEST_DERANDOMIZE` variable that we set explicitly in `just test-all` and then we don't need to set it anywhere else. That would be much simpler.

DEVELOPERS.md
```diff
@@ -74,27 +74,21 @@ results are the same.
...
-
-To do some "proper" generative testing you can run the command:
+they're run. But neither of these qualities are desirable in CI.
+
+For this reason, the CI configuration generates only a small number of
+examples, of limited query depth, and in a determinstic fashion. They
+thus function more as a test that the generative test machinery is
+working correctly than as a serious attempt to do generative testing.
+
+When running locally, although randomisation is enabled by default by
```
rebkwok, line 84:
```suggestion
When running locally, although randomisation is enabled by default,
```

rebkwok (approved):

Issue #2369, opened 2025-01-21
Improve the helpfulness of the ehrQL docstrings


evansd:
Docstrings like this one:
https://github.com/opensafely-core/ehrql/blob/ace50725d670cc379c183df6468b6849cff01976/ehrql/query_language.py#L329-L334

Are rendered as part of the reference documentation:
https://docs.opensafely.org/ehrql/reference/language/#CodePatientSeries.is_in

And are also surfaced in some code editors, particularly VSCode.

These haven't all be written in a consistent user-focused voice and it would be good to revisit them and make sure they're as helpfully written as possible.

Slack thread:
https://bennettoxford.slack.com/archives/C63UXGB8E/p1737472147630349

alarthast:
Tracked in [this spreadsheet](https://docs.google.com/spreadsheets/d/1F8yvhLlupx1a7GdFHLxkiDYCeuV6G-wsaIq8GgN2yZI/edit?gid=0#gid=0). At the moment only did a "first pass" to make minor wording changes and add examples where it's straightforward to do so.

PR #2397, opened 2025-02-20
Tidy ehrQL docstrings on the language reference page

alarthast:
Progresses #2369; see comment in ticket for context.

### DIFF
--- a/docs/includes/generated_docs/language__measures.md
+++ b/docs/includes/generated_docs/language__measures.md
@@ -5,8 +5,18 @@
 <div markdown="block" class="indent">
 A measure definition file must define a collection of measures called `measures`.
 
-```py
+```python
 measures = create_measures()
+```
+
+Add measures to the collection using [`define_measure`](#Measures.define_measure):
+
+```python
+measures.define_measure(
+    name="adult_proportion",
+    numerator=patients.age_on(INTERVAL.start_date) >=18,
+    denominator=patients.exists_for_patient()
+)
 ```
 </div>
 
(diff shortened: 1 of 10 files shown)


docs/includes/generated_docs/language__measures.md
```diff
@@ -26,7 +26,7 @@ measures.define_measure(
 </h4>
 
 <div markdown="block" class="indent">
-Create a collection of measures with [`create_measures`](#create_measures).
+The preferential way to create a collection of measures is with [`create_measures`](#create_measures).
```
evansd, line 29: Same point as with the `Dataset` docstring.

ehrql/query_language.py
```diff
@@ -78,7 +78,7 @@ def set_additional_population_constraint(self, additional_population_constraint)
 
 class Dataset:
     """
-    Create a dataset with [`create_dataset`](#create_dataset).
+    The preferential way to create a dataset is with [`create_dataset`](#create_dataset).
```
evansd, line 81: I see what you mean about avoiding the imperative mood here, but I worry that talking about "the preferential way" just creates questions (why is it preferential? what are the other ways? what are their disadvantages?). What do you think about something like:
```
To create a dataset use the [`create_dataset`](#create_dataset) function.
```

evansd (approved):
This is great and a huge improvement on what we had before. I had one question about one bit of language, but I'd be happy to merge this exactly as is.

I think the suggestion you made about overriding methods so we can have type-specific examples in the docstrings is almost certainly the way to go, but we can tackle that in a subsequent PR.

docs/includes/generated_docs/language__measures.md
alarthast: Changed to "To create a collection of measures use the [`create_measures`](#create_measures) function."

ehrql/query_language.py
alarthast: Changed to the suggested.

alarthast:
To preserve the commit hashes (as some are referenced in the ticket), reviewer comments (+ an additional typo fix) are implemented by pushing a post-approval commit rather than by rebasing.

PR #2419, opened 2025-03-14
Rework query ordering to minimise temporary table lifetime

### LINKED ISSUE #2411: Refactor `get_setup_and_cleanup` queries to allow interleaved queries
This is pulling out some a change proposed as part of the [measures performance ticket](https://github.com/opensafely-core/ehrql/issues/2294#issuecomment-2671611199) into a standalone ticket. These changes give us a better, more flexible API and allow us to address other pain points so are worth doing even if we don't immediately take advantage of them in the measures system.

> I think we can solve 2 by changing the way we handle setup and cleanup queries in a way that will have other benefits. Instead of having three lists of queries (setup_queries, results_queries, cleanup_queries) we'd have a single list of (query_type, query) pairs, where the query types could be SETUP, CLEANUP, RESULTS. This would allow us to interleave the query types at will, rather than only doing clean up after the results have been generated.
>
> The function which currently determines the setup and cleanup lists can be modified to cleanup eagerly i.e. to clean up a given table as soon as it's no longer referenced in any subsequent queries. And there would need to a small-ish change to the query engines to handle this. But overall I don't think it would be a massively invasive change.
>
> It would also allow us to fix an awkwardness in the current system now that we have multiple different types of results query requiring different approaches to batch download. We've moved from always having one-row-per-patient results to also supporting many-rows-per-patient, and with the proposed measures changes we'll also have results which don't have a patient ID at all. We could handle these by having multiple different sub-types of the RESULTS query type which gives us a nice mechanism to thread this information through to the parts of the code which need it.

evansd:
This takes advantage of the changes made in #2416 to change the order in which we run queries in order to minimise the lifetime of temporary tables and therefore to avoid putting undue pressure on temporary table storage even in the face of very long query chains (as we hope to do shortly with measures).

Where previously we would have executed queries in this order:

```sql
CREATE TABLE temp_table_1 ...
INSERT INTO temp_table_1 ...

CREATE TABLE temp_table_2
INSERT INTO temp_table_2 SELECT * FROM temp_table_1 ...

CREATE TABLE temp_table_3 ...
INSERT INTO temp_table_3 SELECT * FROM temp_table_1 ...

SELECT * FROM temp_table_2
SELECT * FROM temp_table_3

DROP TABLE temp_table_3
DROP TABLE temp_table_2
DROP TABLE temp_table_1
```

We now execute them in this order:
```sql
CREATE TABLE temp_table_1 ...
INSERT INTO temp_table_1 ...

CREATE TABLE temp_table_2
INSERT INTO temp_table_2 SELECT * FROM temp_table_1 ...

SELECT * FROM temp_table_2
DROP TABLE temp_table_2

CREATE TABLE temp_table_3 ...
INSERT INTO temp_table_3 SELECT * FROM temp_table_1 ...
DROP TABLE temp_table_1

SELECT * FROM temp_table_3
DROP TABLE temp_table_3
```

That is, we avoid creating temporary tables until they're actually needed, and we clean them up as soon as they're no longer needed.

Closes #2411

### DIFF
--- a/ehrql/utils/sqlalchemy_query_utils.py
+++ b/ehrql/utils/sqlalchemy_query_utils.py
@@ -14,6 +14,18 @@ from sqlalchemy.sql.elements import (
 from sqlalchemy.sql.expression import ClauseElement, Executable
 
 
+def clause_as_str(clause, dialect):
+    """
+    Return a SQL clause as a string in the supplied SQL dialect with any included
+    parameters interpolated in
+    """
+    compiled = clause.compile(
+        dialect=dialect,
+        compile_kwargs={"literal_binds": True},
+    )
+    return str(compiled).strip()
+
+
 def is_predicate(clause):
     """
     Some boolean expressions are guaranteed to be boolean-typed by virtue of their
@@ -148,18 +160,6 @@ def get_generated_tables(clause):
 
 def flatten_iter(nested_iters):
     return [i for sub_iter in nested_iters for i in sub_iter]
-
-
-def clause_as_str(clause, dialect):
-    """
-    Return a SQL clause as a string in the supplied SQL dialect with any included
-    parameters interpolated in
-    """
-    compiled = clause.compile(
-        dialect=dialect,
-        compile_kwargs={"literal_binds": True},
-    )
-    return str(compiled).strip()
 
 
 def iterate_unique(clause):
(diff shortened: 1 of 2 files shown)


rebkwok (approved):
Took me a while to get my head round the ordering, but the detailed comments were very helpful, thanks :)

PR #2424, opened 2025-03-18
Fix autocomplete tests to work with latest version of pyright

rw251:
The initial language server handshake had changed.

Previously the server output 7 messages at one point and was then ready to respond to completion requests. This changed to 8 with the newer version. So rather than just reading 7 messages and assuming it's ready, we now read messages until we get the one that indicates that it is ready.

Ideally we'd just keep reading until the buffer was empty, but when I initially set all this up I couldn't quite figure how to do that cleanly without setting timeouts, so I haven't tried this time either.


evansd (approved):
Ah great, thanks!

Issue #2445, opened 2025-04-30
Return value of `.to_category` isn't compatible with `.is_null`


iaindillingham:
The return value of `CodeFunctions.to_category`, which is inherited by `CodePatientSeries` and `CodeEventSeries`, is a `Series` that wraps a `Case`. The default for `to_category` is `None`. However, `Series.is_null` doesn't know how to handle `None`. Consequently, unexpected `None`s can appear in a `Series`.

@ZoeMZou found this when running a measures definition. Their "ethnicity information is missing" measure was based on converting SNOMED CT codes to categories, and then on counting the number of nulls. However, doing so wasn't counting the number of `None`s. Ultimately, the unexpected `None`s resulted in ehrQL raising a `TypeError` here:

https://github.com/opensafely-core/ehrql/blob/f332932ad60c349a34751cdbea7e25964b0d0d8f/ehrql/measures/calculate.py#L38

The workaround was to provide an explicit default (opensafely/WinterPressuresDescriptive#10). Easy when you know how! You can find out more in [the Slack thread][1].

[1]: https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1745850451388559

evansd:
There _is_ a bug in ehrQL, but it's nothing to do with the interaction between `to_category` and `is_null`. The bug is that the measures framework wasn't correctly handling cases where the numerator was NULL.

Looking at [this block of code](https://github.com/opensafely/WinterPressuresDescriptive/blob/c72590c0e9bca7a8530df63f8e9df2a885279337/analysis/dataset_definition/variables_dataset.py#L74-L88) here:
```
tmp_exp_cat_ethnicity = (
    clinical_events.where(
        clinical_events.snomedct_code.is_in(ethnicity_snomed)
    )
    .where(clinical_events.date.is_on_or_before(cohort_start))
    .sort_by(clinical_events.date)
    .last_for_patient()
    .snomedct_code.to_category(ethnicity_snomed)
)
exp_bin_eth_missing = tmp_exp_cat_ethnicity.is_null()
exp_bin_eth_white = (tmp_exp_cat_ethnicity == "White")
exp_bin_eth_mixed = (tmp_exp_cat_ethnicity == "Mixed")
```

[...]

PR #2446, opened 2025-05-02
Handle NULL numerators in measures

evansd:
This used to be handled by the old measures code:
https://github.com/opensafely-core/ehrql/blob/13de8adb9a3159697e83d050137117e510ad18fb/ehrql/measures/calculate.py#L133-L137

However somewhere along the way this turned instead into a check for a NULL or zero-valued denominator:
https://github.com/opensafely-core/ehrql/blob/f332932ad60c349a34751cdbea7e25964b0d0d8f/ehrql/measures/calculate.py#L38

The lack of NULL numerator handling recently resulted in an error for a user (see [thread][1]) which was misdiagnosed as an issue with `is_null()` (see #2445).

I'm _fairly_ confident that it's not possible to get either a NULL or a zero denominator value given the way measure results are generated. Certainly I haven't managed to construct a test case which produces one. So I've added an assertion to that effect along with hopefully enough debugging information to help us out if we ever do hit it.

[1]: https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1745850451388559

### DIFF
--- a/ehrql/measures/calculate.py
+++ b/ehrql/measures/calculate.py
@@ -35,7 +35,15 @@ def get_measure_results(query_engine, measures, timeout=259200.0):
         calculator = MeasureCalculator(measure_group)
         results = calculator.get_results(query_engine)
         for measure, interval, numerator, denominator, group_dict in results:
-            ratio = numerator / denominator if denominator else None
+            # Given the way results are constructed we should never get a zero-valued or
+            # NULL denominator: we should just get no results at all for that measure
+            # row
+            assert denominator, (
+                f"{denominator!r} denominator in {measure.name} at {interval}"
+            )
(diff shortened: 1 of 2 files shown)


rebkwok (approved):
I feel like I remember looking at this bit of the old code, and consciously not including it. I am pretty sure it was before I added the grouping sets, when it was just returning individual grouped results from SQL, and if the numerator was null, there would be no row returned for that result, so the `if numerator is None` was irrelevant (and resulted in an uncovered branch because it was never not None.  But then I added grouping sets, so a row returns multiple grouped results, and so now it will return null results for numerators.

rebkwok:
Hmm. No, I don't sorry.

evansd:
No worries!

PR #2464, opened 2025-05-21
Fix variable name typo in example data script

alarthast:
This does not actually affect anything, but a typo is a typo.


evansd (approved):
A typo is indeed a typo! Thanks :grinning:

PR #2467, opened 2025-05-22
Don't attempt to fetch measure results in batches

### LINKED ISSUE #2466: Hit a "batch_size too small to make progress" assertion error in measures code
This was the job in question:
https://jobs.opensafely.org/impact-and-inequalities-of-winter-pressures-in-primary-care-providing-the-evidence-base-for-mitigation-strategies/wp-work-package-1/24593/3p4dnvos5imyzzv7/

Which ran this command:
```
ehrql:v1 generate-measures analysis/wp_measures.py --output output/demograph_measures/demograph_measures_2017-04-10.arrow -- --demograph_measures --start_intv 2017-04-10
```

Against this file:
https://github.com/opensafely/winter-pressures-phase-II/blob/4b88a3d26cf3c3542462f0bec6a634ce943b4ffd/analysis/wp_measures.py

Resulting in this traceback
Image

I've [informed the user](https://bennettoxford.slack.com/archives/C02HJTL065A/p1747916551929649) that this isn't their fault and that we'll investigate.

evansd:
Batched fetching was introduced to work around the unreliability of the database connection as it allows us to retry part way through the download without starting again from the beginning. The problem is particularly acute for patient and event-level results which number in the millions and where there is an obvious column (`patient_id`) to page on. However this is less of an issue for aggregated results (specifically measures) as there are fewer rows to deal with, and these also lack a natural paging column.

Attempting to use batched fetching on measures results is now causing issues (see linked ticket) and so the simplest solution is just to stop doing it.

See also discussion at:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1747925981081259

Closes #2466

### DIFF
--- a/ehrql/query_engines/mssql.py
+++ b/ehrql/query_engines/mssql.py
@@ -172,20 +172,10 @@ class MSSQLQueryEngine(BaseSQLQueryEngine):
         execute_with_log(connection, query, log.info, query_id=query_id)
 
     def execute_query_with_results(self, connection, query, query_id):
-        # We're expecting queries in a very specific form which is "select everything
-        # from one table"; so we assert that each query has this form and retrieve a
-        # reference to the table
-        results_table = query.get_final_froms()[0]
-        assert str(query) == str(sqlalchemy.select(results_table))
-
-        # The query type tells us whether or not we can depend on a unique patient_id
-        # column. We need to pass this information to the batch fetcher as it changes
-        # the algorithm we can use.
-        has_unique_key = {
-            self.QueryType.PATIENT_LEVEL: True,
-            self.QueryType.EVENT_LEVEL: False,
-        }
-        key_is_unique = has_unique_key[query._annotations["query_type"]]
+        # The query type tells us what sort of method we can use for fetching results.
+        # We prefer a batched approach using a unique key, but that's not always
+        # possible.
+        query_type = query._annotations["query_type"]
 
         # We use a separate connection to retrieve the results. Our intial connection
         # keeps the temporary table "alive" and then this connection can be safely reset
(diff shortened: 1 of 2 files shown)


ehrql/query_engines/mssql.py
```diff
@@ -201,16 +191,48 @@ def execute_query_with_results(self, connection, query, query_id):
...
+                )
+            else:
+                assert False, f"Unhandled query type: {query_type}"
+
+    def fetch_results_batched(self, execute, query, key_is_unique):
+        # We're expecting queries in a very specific form which is "select everything
+        # from one table"; so we assert that each query has this form and retrieve a
+        # reference to the table
+        results_table = query.get_final_froms()[0]
+        assert str(query) == str(sqlalchemy.select(results_table))
```
rebkwok, line 219: Don't we still want this check that the query is in the correct form for `fetch_results_in_one_go`?

ehrql/query_engines/mssql.py
```diff
@@ -201,16 +191,48 @@ def execute_query_with_results(self, connection, query, query_id):
...
+
+        return fetch_table_in_batches(
+            execute,
+            results_table,
+            key_column_index=0,
```
rebkwok, line 224: This was changed from explicitly using "patient_id" as the key column for measures - it could go back now. Or maybe just a comment that the column is generally expected to be patient_id?

rebkwok (approved):

ehrql/query_engines/mssql.py
evansd: Ah yeah, you're right: it would be better to be explicit here. I'll add that back.

ehrql/query_engines/mssql.py
evansd: I don't think so, because we just execute the query exactly as supplied and don't try to do any transformations on it. The only reason I'm still extracting the table name was that I thought it would be better for logging (we could just log the query itself but it will name every column explicitly which feels a bit noisy).

PR #2480, opened 2025-06-18
Remove flit config from pyproject.toml

inglesp:
We used to use flit to build databuilder


inglesp:
> everything is more complicated than you think it will be and computers were a mistake, really don't want to get sidetracked into this right now

PR #2484, opened 2025-06-19
Remove references to coding system types from docs

### LINKED ISSUE #2452: "Language features" docs contain references to non-public names
There are references to `SNOMEDCTCode` here:
https://docs.opensafely.org/ehrql/reference/features/#9-operations-on-all-series-containing-codes

However these classes are not part of the public API.

This has created some user confusion, see:
https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1747056013856089

evansd:
These are not part of our public API but if we use them in either the spec tests or in docstrings then they end up leaking into the public docs and causing confusion.

I have confirmed there are no more references in the docs using:
```
rg '(CTV3|BNF|DMD|ICD10|OPCS4|SNOMEDCT)Code' docs/
```

Closes #2452

### DIFF
--- a/docs/includes/generated_docs/language__series.md
+++ b/docs/includes/generated_docs/language__series.md
@@ -3219,15 +3219,14 @@ school_year = patients.age_on("2020-09-01").map_values(
   <a class="headerlink" href="#MultiCodeStringPatientSeries.contains" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Check if the multi code field contains a specific code string and
-return the result as a boolean series. `code` can
-either be a string (and prefix matching works so e.g. "N17" in ICD-10
-would match all acute renal failure), or a clinical code.
+Check if the multi code field contains a specific code string and return the
+result as a boolean series. `code` can be a partial match (so e.g. "N17" in
+ICD-10 would match all acute renal failure codes), or a full clinical code.
 
 Example usages:
 ```python
 all_diagnoses.contains("N17")
-all_diagnoses.contains(ICD10Code("N170"))
+all_diagnoses.contains("N170")
 ```
 </div>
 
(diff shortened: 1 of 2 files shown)


docs/includes/generated_docs/language__series.md
```diff
@@ -3219,15 +3219,14 @@ school_year = patients.age_on("2020-09-01").map_values(
   <a class="headerlink" href="#MultiCodeStringPatientSeries.contains" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Check if the multi code field contains a specific code string and
-return the result as a boolean series. `code` can
-either be a string (and prefix matching works so e.g. "N17" in ICD-10
-would match all acute renal failure), or a clinical code.
+Check if the multi code field contains a specific code string and return the
+result as a boolean series. `code` can a partial match (so e.g. "N17" in ICD-10
```
suzannehamilton, line 3223: Typo: should `can a partial match` be `can be a partial match`?

Ditto for a few instances below.

suzannehamilton (commented):
Makes sense to me, but I'll let someone with a more ehrQL knowledge approve it.

Just noticed at typo in the updated docs.

docs/includes/generated_docs/language__series.md
evansd: Ah, thanks! The actual typo here is in the docstring from which these docs are generated, which is why it's repeated multiple times.

All the files in `docs/includes/generated_docs` are autogenerated from the code. There's a command `just generate-docs` which updates these. And CI enforces that they're always up-to-date.

We generally try to keep all the autogenerated docs changes together in a single commit to make life easier for reviewers.

rebkwok (approved):

PR #2490, opened 2025-06-26
Add option to dump example data to a custom directory

rebkwok:
The example data is a set of CSV files that are generated for use in the ehrQL tutorial. In the tutorial repo, these CSVs are already present in a `dummy_tables` directory, for use with the `show()` command via the OpenSAFELY vs code extension.

After following the tutorial, users reasonably expect to be able to use the `show()` command in their own ehrQL, but it requires dummy tables, which the vs code extension expects to be found in a directory called `dummy_tables` at the top level of the study directory.

While it doesn't solve the chicken-egg problem with the `show()` command (users need dummy tables to help them write ehqrl, but the best way we have of creating dummy tables is to write ehrql), we can at least give them something to get started with as a default, by having the vs code extension dump the tutorial example data if there's no dummy tables available. In order to do that, we need to be able to dump the dummy tables to somewhere other that just the hard-coded `example-data` directory.

### DIFF
--- a/docs/includes/generated_docs/cli.md
+++ b/docs/includes/generated_docs/cli.md
@@ -382,7 +382,7 @@ Dotted import path to Backend class, or one of: `emis`, `tpp`
   dump-example-data
 </h2>
 ```
-ehrql dump-example-data [--help]
+ehrql dump-example-data [--help] [--dst-dir DST_DIR]
 ```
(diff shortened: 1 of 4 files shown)


evansd:
Just realised it needs a pedantic commit prefix!

Issue #2423, opened 2025-03-17
Document how to pass parameters to ehrQL scripts


inglesp:
When ehrQL is invoked, anything after `--` is stuffed into `sys.argv` so that it can be accessed by researchers in their ehrQL scripts.  (Implementation [here](https://github.com/opensafely-core/ehrql/blob/9bba6b379b65346ad62d34b428ef747c5d60de82/ehrql/loaders.py#L342-L345).)

This is documented in passing in the [CLI reference](https://docs.opensafely.org/ehrql/reference/cli/#generate-dataset.user_args), but we don't document how to access these parameters via `sys.argv`.

We should write a short how-to guide.

evansd:
A related ticket with some history here:
 * https://github.com/opensafely-core/ehrql/issues/1766

PR #2492, opened 2025-06-27
Friendlier interface to user-provided parameters

### LINKED ISSUE #1766: Make parameterizing a dataset definition less like writing Python
To parameterize a dataset definition, a user needs to do something [like this][1] in their `project.yaml`

```yaml
  model_data_structure_from_dataset_definition:
    run: ehrql:v0
      generate-dataset
      --output output/model_data_structure/dataset.csv
      analysis/model_data_structure/dataset_definition.py
      --
      --day=0
    outputs:
      highly_sensitive:
        rows: output/model_data_structure/dataset.csv
```

and something [like this][2] in their `dataset_definition.py`

```py
parser = argparse.ArgumentParser()
parser.add_argument(
    "--day",
    type=int,
    help="The day a survey was completed, relative to the day the first survey was completed",
)
parser.add_argument(
    "--window",
    type=int,
    default=0,
    help="The number of days +/- the day a survey was completed",
)
args = parser.parse_args()
```

We've had feedback that writing ehrQL is too much like writing Python. To make it a bit less like writing Python, we should consider wrapping the `ArgumentParser` instance in a helper function.

[1]: https://github.com/opensafely/airmid-data-development/blob/5b29a5360353edb1b1a431bd2d1398c9b4ea47e9/project.yaml#L79-L88
[2]: https://github.com/opensafely/airmid-data-development/blob/5b29a5360353edb1b1a431bd2d1398c9b4ea47e9/analysis/model_data_structure/dataset_definition.py#L8-L20

rebkwok:
Fixes #1766 
Partially addresses #2423 (includes reference docs although we will also want to add to the how-tos)

### DIFF
--- a/ehrql/__init__.py
+++ b/ehrql/__init__.py
@@ -16,6 +16,7 @@ from ehrql.query_language import (
     when,
     years,
 )
+from ehrql.user_parameters import get_parameter
 from ehrql.utils.log_utils import init_logging
 
 
@@ -32,6 +33,7 @@ __all__ = [
     "create_dataset",
     "create_measures",
     "days",
+    "get_parameter",
     "show",
     "maximum_of",
     "minimum_of",
--- a/ehrql/docs/language.py
+++ b/ehrql/docs/language.py
@@ -107,6 +107,9 @@ def build_language():
             "Measures": namespace["Measures"],
             "INTERVAL": namespace["INTERVAL"],
         },
+        "parameters": {
+            "get_parameter": namespace["get_parameter"],
+        },
     }
 
     # Check that the documentation is complete
--- a/docs/includes/generated_docs/cli.md
+++ b/docs/includes/generated_docs/cli.md
@@ -204,7 +204,7 @@ This argument is ignored when running against real tables.
   <a class="headerlink" href="#generate-dataset.user_args" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Parameters are extra arguments you can pass to your Python definition file. They must be
+Parameters are [extra arguments](language.md#parameters) you can pass to your Python definition file. They must be
 supplied after all ehrQL arguments and separated from the ehrQL arguments with a
 double-dash ` -- `.
 
@@ -334,7 +334,7 @@ This argument is ignored when running against real tables.
   <a class="headerlink" href="#generate-measures.user_args" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Parameters are extra arguments you can pass to your Python definition file. They must be
+Parameters are [extra arguments](language.md#parameters) you can pass to your Python definition file. They must be
 supplied after all ehrQL arguments and separated from the ehrQL arguments with a
 double-dash ` -- `.
 
@@ -464,7 +464,7 @@ Dotted import path to Backend class, or one of: `emis`, `tpp`
   <a class="headerlink" href="#dump-dataset-sql.user_args" title="Permanent link">🔗</a>
 </div>
 <div markdown="block" class="indent">
-Parameters are extra arguments you can pass to your Python definition file. They must be
+Parameters are [extra arguments](language.md#parameters) you can pass to your Python definition file. They must be
 supplied after all ehrQL arguments and separated from the ehrQL arguments with a
 double-dash ` -- `.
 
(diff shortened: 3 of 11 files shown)


ehrql/dataset_parameters.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Parameter:
+    name: str
+    type: callable = str
+    default: Any | None = None
```
evansd, line 15: I think this is left over from the previous implementation and no longer used.

ehrql/__init__.py
```diff
@@ -32,6 +33,7 @@
     "create_dataset",
     "create_measures",
     "days",
+    "get_dataset_parameter",
```
evansd, line 36: What do you think about just `get_parameter` as a name? I'm not sure "dataset" adds much and its potentially confusing and you can use these with measures (or whatever other things we come up with in future).

ehrql/dataset_parameters.py
```diff
@@ -0,0 +1,95 @@
...
+    max_age = get_dataset_parameter(name="max_age", type=int, default=20)
+    sex = get_dataset_parameter(name="sex", default="male")
+
+    dataset = create_dataset()
+    age = patients.age_on("2025-01-01")
+
+    dataset.define_population((age < max_age) & (patients.sex==sex))
+    ```
+    """
+    # find matching user args
```
evansd, line 72: By default argparse supports both `--name value` and `--name=value` syntax and I think it might be helpful if we could do the same. We might get this for free if we were able to use `parse_known_args()` (see note below).

ehrql/dataset_parameters.py
```diff
@@ -0,0 +1,95 @@
...
+                break
+            user_args.append(arg)
+
+    parser = ArgumentParser()
+    # We use nargs="+" and action="extend" so we can return either a single value or a list,
+    # depending on what values have been provided
+    parser.add_argument(f"--{name}", type=type, action="extend", nargs="+", default=[])
+
+    # Note that we could use parse_known_args(sys.argv[1:]) which won't raise errors if there are
+    # undefined args. However, parse_known_args does prefix-matching
```
evansd, line 89: I think it might be possible to disable prefix matching by passing `allow_abbrev=False` to `ArgumentParser()`:
https://docs.python.org/3.10/library/argparse.html#prefix-matching

Not actually tried this though!

ehrql/dataset_parameters.py
```diff
@@ -0,0 +1,95 @@
...
+    # https://docs.python.org/3.10/library/argparse.html#argparse.ArgumentParser.parse_known_args
+    # This means that for user args `--foo=1 --foobar=2`, --foobar will prefix-match --foo and will
+    # be parsed incorrectly as foobar=1
+    namespace = parser.parse_args(user_args)
+    value = getattr(namespace, name)
+    if not value:
+        if default is not None:
+            return default
+        raise SystemExit(
```
evansd, line 98: `SystemExit` seems a bit nuclear here and won't show the user the line where they call `get_dataset_parameter()` in their own code. If we raise our own exception (or event `argpase.ArgumentError` then ehrQL's traceback handling should hide all the internal stack frames and just show the user the line of their own code which triggered the error.

ehrql/docs/language.py
```diff
@@ -75,6 +75,7 @@ def build_language():
         "dataset": dict(
             create_dataset=namespace["create_dataset"],
             Dataset=namespace["Dataset"],
+            get_dataset_parameter=namespace["get_dataset_parameter"],
```
evansd, line 78: I wonder if this belongs in it's own section, "Parameters" or something. It's not specific to just datasets.

evansd (commented):
I've added various thoughts on the implementation and a bit of bikeshedding on the name, but fundamentally I think this is the right approach and it's going to be so much nicer than all the ad hoc stuff people are doing at the moment.

ehrql/__init__.py
rebkwok: Sounds good. I originally didn't call it just `get_parameter` because we had other just Parameter things, but those aren't exposed to users. Hadn't occurred to me that they're usuable for measures too.

ehrql/dataset_parameters.py
rebkwok: It works!

PR #2501, opened 2025-07-08
Split larger joins into smaller batches

evansd:
Empirically we've seen this make a huge improvement to overall runtime on MSSQL. For example a single huge join, which previously took 52 hours to run, took 1 hour when split into 10 smaller joins plus a final combined join.

We have various hypotheses as to why this might be the case but as yet no access to the kind of query planner or log data which would confirm the source of the difference. Nevertheless, the improvement seems so marked as to be worth trying "blind", as it were.

It is possible that splitting the joins like this will add some overhead to jobs where the join previously ran in reasonable time. But it's hard to see how this overhead could be hugely significant. And in general, even if runtime wasn't improved, we'd expect better overall system behaviour by breaking up "monster queries" into more manageable chunks.

We should keep a close eye on ehrQL runtimes following this change, and we can always revert if things are unexpectedly problematic.

In order to ensure that Hypothesis exercises this code path I've expanded the generative tests to produce examples with multiple columns. This seems like a reasonable thing to do in any case (see commit message for more detail). I tried introducing a silly bug into `split_joins()` and confirmed that Hypothesis found it. And to give Hypothesis a real run at it I temporarily changed these values to `min_size=8, max_size=8`:
https://github.com/opensafely-core/ehrql/blob/48778afb4d0f1b345def8e1f4c675c74354223ae/tests/generative/variable_strategies.py#L629

And then ran a load of examples with:
```
GENTEST_EXAMPLES=6000 GENTEST_QUERY_ENGINES="in_memory mssql" \
  pytest tests/generative/test_query_model.py::test_query_model
```

Related thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1751959832519489

### DIFF
--- a/tests/generative/test_query_model.py
+++ b/tests/generative/test_query_model.py
@@ -232,6 +232,8 @@ def run_with(engine, instances, dataset):
                 # In order to exercise the temporary table code path we set the limit
                 # here very low
                 "EHRQL_MAX_MULTIVALUE_PARAM_LENGTH": 3,
+                # Likewise for the join splitting code
+                "EHRQL_MAX_JOIN_COUNT": 3,
             },
         )
     except Exception as e:
--- a/tests/generative/variable_strategies.py
+++ b/tests/generative/variable_strategies.py
@@ -626,16 +626,16 @@ def dataset(patient_tables, event_tables, schema, value_strategies):
     return st.builds(
         make_dataset,
         valid_population(),
-        valid_patient_variable(),
+        st.lists(valid_patient_variable(), min_size=1, max_size=5),
         # Event series is optional
         st.one_of(st.none(), valid_event_series()),
     )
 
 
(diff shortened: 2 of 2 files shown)


ehrql/query_engines/base_sql.py
```diff
@@ -1038,6 +1084,65 @@ def apply_patient_joins(query):
...
+    current_batch = [id_column, first_column]
+    current_tables = id_tables | first_tables
+
+    column_batches = [current_batch]
+
+    for next_column, next_tables in column_tables:
+        if (
+            # If the next column uses no new tables then always include it in the
+            # current batch
+            next_tables <= current_tables
```
rebkwok, line 1128: huh. Why did I not know you could compare sets like this?

rebkwok (approved):

PR #2503, opened 2025-07-09
Minor improvements to get_parameter reference docs

rebkwok:


docs/includes/generated_docs/language__parameters.md
```diff
@@ -29,10 +29,10 @@ For each parameter, in the dataset definition, call `get_parameter()`, e.g.:
 ```
 
 Parameters are parsed as strings, unless a `type` is explicitly specified.
-`type` can be any callable that takes a single string argument;
+Usually this will be one of the standard python types, such as integers or floats.
```
madwort, line 32: heh, went to double-check the canonical name for these in [the python docs](https://docs.python.org/3/library/stdtypes.html#) - the url is stdtypes, the title is "built-in types" and the first sentence says "the standard types"....

madwort (approved):

Issue #2500, opened 2025-07-07
Some dataset definitions seem to take a very long to compile


evansd:
I noticed this while investigating the performance of this job:
https://jobs.opensafely.org/comparing-disparities-in-rsv-influenza-and-covid-19/disparities-comparison-rsv-flu-c19/24742/ihpvs2qid2wrvcgq/

In the context of the absolutely enormous runtime, the compilation time is pretty trivial but it was still far longer than I expected and so worth investigating what's going on here. (Could it be linked to the [test speed regression](https://github.com/opensafely-core/ehrql/issues/2489)?).

For the below action:
https://github.com/opensafely/disparities-comparison/blob/50b66155b1ebfbaa325eb0f462eae09861bd0d9b/project.yaml#L7603-L7611

There's an over 11 minute gap between `compiling` and `generating` and then a further 26 minutes gap before the first queries start executing:
![Image](https://github.com/user-attachments/assets/96da049b-7bd0-48fb-9a8f-95b17d3e0bb3)

I would be very interested to know what's in doing in this time.

To avoid cloning the huge repo history you can get the relevant commit with:
```
git clone --depth 1 git@github.com:opensafely/disparities-comparison.git
cd disparities-comparison
git fetch --depth 1 50b66155b1ebfbaa325eb0f462eae09861bd0d9b
git checkout 50b66155b1ebfbaa325eb0f462eae09861bd0d9b
```

And then reproduce the slow compilation with:
```
ehrql dump-dataset-sql --backend tpp analysis/dataset_definition.py \
  -- adults season4_start_date season4_end_date sensitive primary
```

inglesp:
Using py-spy, we see that most of the time is in `apply_transforms`, although there's also non-negligible time spent in unmarshalling.

![Image](https://github.com/user-attachments/assets/0ea93785-1d45-4314-b5f6-0001eaad8476)

inglesp:
Urgh, in uploading the flamegraph to GH we've lost all interactivity.  If you save the SVG locally and open that in Firefox, you can click around and dig into the slow bits.

inglesp:
Yeah, it's the hashing (#2489):

```
(.venv) inglesp@malbogies:~/work/ebmdatalab/research/disparities-comparison$ git -C ../../ehrql rev-parse HEAD
cb8f552eba2e85b4324bc98236e295f021c0857e
[...]

PR #2512, opened 2025-07-14
Fix hashing

### LINKED ISSUE #2489: Changes to node hashing caused test performance regression
This PR appears to have caused a noticeable uptick in the time in takes tests to run:
 * https://github.com/opensafely-core/ehrql/pull/2399/

Image

Image

It was always expected that this would be slower than what went before, but with the advantage of no longer requiring specialised hash functions for certain nodes. In production, when we generally construct just a single large query graph, the cost is largely irrelevant as the value is only computed once per node and then cached. However in tests we're constantly constructing and discarding query graphs and so the extra runtime becomes noticeable.

The `dataclasses` approach to making fast `__hash__` methods is to dynamically construct the code as a string and then `eval` it into a method:
https://github.com/python/cpython/blob/dd59c786cfb1018eb5abe877bfa7265ea9a3c2b9/Lib/dataclasses.py#L891-L897

Maybe we should go down this route. Or maybe there's something less extreme we could do to improve performance here.

Related Slack thread:
https://bennettoxford.slack.com/archives/C0926N4NR7T/p1750775427802659

inglesp:
Fixes #2489 and #2500.


ehrql/query_model/nodes.py
```diff
@@ -87,22 +87,26 @@ class Node:
...
         # Calculating the hash of an object requires recursively calculating the hashes
         # of its children. In a deeply nested query graph this can take some time and,
         # given how frequently `__hash__()` is called, this can end up completely
         # dominating ehrQL's execution time. Given that these are immutable
         # objects we can cache the hash value instead of recalcuting it each time.
+        # self._retrieved_from_cache is set to allow verification of hashing in tests.
         try:
-            return self.__hash__cache
+            value = self._hash_cache
+            object.__setattr__(self, "_retrieved_from_cache", True)
```
evansd, line 105: I think there's a less intrusive way of testing this, which is to make use of the fact that `Value` nodes can take arbitrary objects. You can give it an object with some sort of spy attached to _its_ `__hash__` method and then confirm that the child hash method is called when expected.

evansd (approved):
I have feelings. But I can't decided whether horror at the original bug or relief at the neatness of the fix predominates. Anyway, thank you for noticing the problem in the first place and for applying typical Inglespian sleuthing to it.

PR #2528, opened 2025-08-21
Serialize tables by reference

### LINKED ISSUE #2513: Generate table definitions in trusted code
Currently, ehrQL adds table definitions to the JSON which is generated by user code.

We are planning to add [stricter restrictions to some datasets](https://github.com/opensafely-core/job-runner/issues/1094), which means the agent should be using a more trustworthy source of the table definitions. For example, the generated JSON could just include the name of the table which needs to be queried. The agent would then look up a trusted definition of that table.

evansd:
This modifies the serialization code so that rather than entire structure of the table being encoded in the serialized content we just encode a reference to the table. The primary driver for this is that it prevents user code from being able to modify the table definition prior to serialization. There is not _currently_ anything additional a malicious user could do by altering the definitions (the backend object fixes what tables and columns are available); but we want to start encoding fine-grained access permissions as part of the table definitions and when we do this it's important that the table definitions can be trusted.

As an example, where previously part of a serialized query might look like this:
```json
"10": {
    "SelectColumn": {"source": {"ref": "11"}, "name": "snomedct_code"}
},
"11": {
    "SelectTable": {
        "name": "clinical_events",
        "schema": {
            "TableSchema": {
                "date": {
                    "Column": {
                        "type_": {"type": "date"},
                        "constraints": {"tuple": []},
                    }
                },
                "snomedct_code": {
                    "Column": {
                        "type_": {"type": "SNOMEDCTCode"},
                        "constraints": {"tuple": []},
                    }
                },
                "numeric_value": {
                    "Column": {
                        "type_": {"type": "float"},
                        "constraints": {"tuple": []},
                    }
                },
            }
        },
    }
},
```

It would now look like this:
```json
"10": {
    "SelectColumn": {
        "source": {
            "external_ref": {
                "module": "ehrql.tables.core",
                "name": "clinical_events",
            }
        },
        "name": "snomedct_code",
    }
},
```

There's an additional benefit to this new approach in that we can now include elements in table definitions (e.g. custom dummy data generation methods) which might otherwise be hard to serialize. Note though that we still need to be able to serialize `InlinePatientTables` (i.e. the results of the `@table_from_file` decorator) so we can't altogether avoid handling table definitions. But we don't need to handle everything that any table definition might include, which gives us more freedom.

Closes #2513

### DIFF
--- a/ehrql/serializer.py
+++ b/ehrql/serializer.py
@@ -1,14 +1,22 @@
 import datetime
 import functools
+import importlib
 import json
 import pathlib
 
+from ehrql import serializer_registry
 from ehrql.codes import BaseCode, BaseMultiCodeString
 from ehrql.file_formats.base import BaseRowsReader
 from ehrql.measures.measures import DisclosureControlConfig, Measure
 from ehrql.query_language import DummyDataConfig
 from ehrql.query_model.column_specs import ColumnSpec
-from ehrql.query_model.nodes import Node, Position, Value
+from ehrql.query_model.nodes import (
+    Node,
+    Position,
+    SelectPatientTable,
+    SelectTable,
+    Value,
+)
 from ehrql.query_model.table_schema import BaseConstraint, Column, TableSchema
 from ehrql.utils.module_utils import get_all_subclasses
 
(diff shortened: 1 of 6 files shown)


ehrql/serializer.py
```diff
@@ -283,6 +334,19 @@ def unmarshal_for_object(self, type_, value):
...
 
+    @unmarshal_for.register(SelectTable)
+    @unmarshal_for.register(SelectPatientTable)
+    def unmarshal_for_prohibited(self, type_, value):
+        # We want any table nodes in the resulting query graph to be trustworthy (i.e.
+        # not user-supplied) so that any permission restrictions they impose can be
+        # relied upon. This means that such nodes can only be serialized as "external
+        # references" i.e. references to nodes defined outside of the user's query
+        # graph.
+        raise SerializerError(
```
bloodearnest, line 345: Just checking my understanding: Is this the main security barrier that prevents user code from bypassing permissions?

Would the query engine still understand/execute an inline table ref if the usr manage to return a serialized qm that contained one?

evansd: Yeah, it is. If the user manages to get their own `SelectTable` or `SelectPatientTable` object into the graph then they could define it as having different permissions.

As a belt-and-braces we could add an extra assert/check in here:
https://github.com/opensafely-core/ehrql/blob/491cc911cb5d3f8170e34fa5e08e1b52bff5ffd8/ehrql/serializer.py#L282-L284

Could be as simple as:
```py
assert not issubclass(type_, SelectTable | SelectPatientTable)
```

PR #2537, opened 2025-09-09
Parse exceptions from ehrQL logs

evansd:
This is needed for investigating unexpected database errors.

As per the [OTel spec](https://opentelemetry.io/docs/specs/otel/trace/exceptions/) exceptions are included as an event on the span in which they occur
```json
{
  "name": "select_into",
  "start": "2025-09-08T15:04:15.648265397Z",
  "end": "2025-09-08T16:41:37.837685167Z",
  "attributes": {
    "...": "..."
  },
  "events": [
    {
      "name": "exception",
      "timestamp": "2025-09-08T16:41:37.837685167Z",
      "attributes": {
        "exception.type": "pymssql._mssql.MSSQLDatabaseException",
        "exception.message": "(20047, b'DB-Lib error message 20047, severity 9:\\nDBPROCESS is dead or not enabled\\n')",
        "exception.stacktrace": " ... full stacktrace as text ... "
      }
    }
  ]
}
```

### DIFF
--- a/scripts/parse_logs.py
+++ b/scripts/parse_logs.py
@@ -129,16 +129,34 @@ def parse_fetch_complete(match):
     }
 
 
+@register_parser(r"^Traceback \(most recent call last\):")
+def parse_exception(match):
+    name, args = get_exception_name_and_args(match.string)
+    return {
+        "type": "exception",
+        "exception_type": name,
+        "message": args,
+        "stacktrace": match.string,
+    }
+
+
+def get_exception_name_and_args(text):
+    for line in text.splitlines():
+        if not line or line.startswith("Traceback") or line.startswith("  "):
+            continue
+        name, _, args = line.partition(": ")
+        return name, args
+
+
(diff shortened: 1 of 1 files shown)


scripts/parse_logs.py
```diff
@@ -293,6 +328,24 @@ def format_for_otel_fetch(group):
     }
 
 
+def format_for_otel_exception(group):
+    prefix = "exc"
+    exception = get_one(group["exception"])
+    start = exception["timestamp"]
+    end = start
+
+    return {
```
bloodearnest, line 337: so, in otel land, exceptions are usually added as events to a span, rather than spans in their own right, as they typically don't have a duration. And usually the span is marked as having errored, which highlights it in honeycomb.

However, we can take this format, as if its another event, and do that in the job-runner code?

bloodearnest (approved):

bloodearnest (approved):

Issue #2514, opened 2025-07-14
Prevent jobs from querying restricted datasets without permission


suzannehamilton:
Part of the work for preparing for [non-COVID research](https://docs.google.com/document/d/1IPyyUiUIRw9bbkDFFSEtoscWRLjoShpIFxI5-uAa7ac/edit).

Before a job runs a query against a restricted dataset, check that the project has permission to access that dataset. This work depends on a [job-runner issue](https://github.com/opensafely-core/job-runner/issues/1094) which will add the project's permissions to the agent.

It also depends on the refactoring in #2513, which will change the ehrQL dataset code.

This is a replacement for the existing repo permissions [stored in the CLI](https://github.com/opensafely-core/opensafely-cli/blob/main/repository_permissions.yaml). Those permissions only generated warnings, but these will be used to restrict access to specific datasets in the backend.

**Which datasets?**

For now, add the same restrictions that are [defined in the CLI](https://github.com/opensafely-core/opensafely-cli/blob/main/repository_permissions.yaml). We may need to update them later.

suzannehamilton:
Someone made the suggestion that we could make these checks before starting the job, which means we give fast feedback to the user rather than running the job for hours and only failing when it gets to the restricted dataset query.

Issue #2515, opened 2025-07-14
Allow users to claim local dev permissions for restricted datasets


suzannehamilton:
Part of the work for preparing for [non-COVID research](https://docs.google.com/document/d/1IPyyUiUIRw9bbkDFFSEtoscWRLjoShpIFxI5-uAa7ac/edit). See the [restricted datasets proposal](https://docs.google.com/document/d/1QuFuqjTRS5jlT8sh4dSdIwvndPG9I3OWBtTiFCy-Ek0/edit?tab=t.0) for more context.

When a user runs an ehrQL job locally which accesses a restricted dataset, the job should fail with an error explaining that the dataset is restricted. The user should be able to override this locally and grant themselves permission to the dataset, for example by adding a line of Python code like dataset.claim_permissions() (or some alternative configuration within the project repo).

It should be clear to the user that these permissions are purely local. They will have to ask for approval to https://github.com/opensafely-core/job-runner/issues/1094.

PR #2543, opened 2025-09-19
Add a permissions system

evansd:
This adds a simple, generic permissions system to ehrQL. In this first instance this is only used to restrict access to certain tables, but I anticipate extending it to e.g. gate access to Event Level Data and for the various opt-out controls.

ehrQL table definitions can now be annotated with metadata using an internal `_meta` class. This can included a `required_permission` attribute e.g.
```py
@table
class some_table(EventFrame):
    class _meta:
        required_permission = "some_restricted_dataset"
```

A check at the documentation build stage ensures that every restricted table references the required permissions somewhere in the table documentation.

### In Production

The RAP Controller [already](https://github.com/opensafely-core/job-runner/pull/1172) supplies an `EHRQL_PERMISSIONS` environment variable with a comma-separated list of permissions that the current job has. If a dataset or measure definition uses a restricted table without the relevant permission then this results in an immediate error, before any queries are executed. For example:
```
EHRQLPermissionError: You do not currently have all the permissions needed for this action.

Missing permissions are:

    * appointments: required for access to the `tpp.appointments` table
    * waiting_list: required for access to the `tpp.wl_clockstops` and `tpp.wl_openpathways` tables

If you think this is a mistake and that you should have these permissions please contact OpenSAFELY support.
```

ehrQL then exits with a specific status code which the RAP Agent can [identify](https://github.com/opensafely-core/job-runner/pull/1188) and use to inform the user that their code failed due to a permissions error.

### Locally

Running the same code locally currently results in a warning in the logs:
```
Some of the tables or features you are using require special permission to use with real
patient data. The permissions needed are:

    * appointments: required for access to the `tpp.appointments` table
    * waiting_list: required for access to the `tpp.wl_clockstops` and `tpp.wl_openpathways` tables   

You can continue to work on your code using dummy data by “claiming” the required permisions:

    from ehrql import claim_permissions
    claim_permissions("appointments", "waiting_list")
```

The intention is that once this change has been circulated and users have had time to made the necessary changes we can make this a hard error rather than a warning so it will be impossible to miss in future.

See the documentation preview here:
https://evansd-table-permissions.databuilder.pages.dev/reference/language/#permissions

Note that the `claim_permissions()` function acts globally, rather than being a dataset/measure specific method. This seems like a better fit with the project-wide nature of permissions and less fiddly for the user who just has to stick the relevant line somewhere in their Python file.

Note also the deliberate decision to use strings for permissions rather than a dedicated enum type. The autocomplete benefits of enums are largely obviated by the fact that the error message gives the user exactly the code they need to copy. And the typo-resistance is less important given that mistyped permissions will result in immediate permissions failures anyway. Attempting to use a dedicated `Permission` type turned out to be architecturally awkward and made for a more cumbersome user story as well, so strings felt like the better compromise.

Closes: #2514 #2515


ehrql/query_model/nodes.py
```diff
@@ -245,11 +245,13 @@ class Parameter(OneRowPerPatientSeries[T]):
 class SelectTable(ManyRowsPerPatientFrame):
     name: str
     schema: TableSchema
+    required_permission: str | None = None
```
suzannehamilton, line 248: Is it safer for the default permission to be some placeholder value rather than `None`, so that all tables have to override it? Or is this something we'll definitely consider when we add a new table?

ehrql/permissions.py
```diff
@@ -0,0 +1,145 @@
...
+            f"the required permisions:\n"
+            f"\n"
+            f"    from ehrql import claim_permissions\n"
+            f"    claim_permissions({claim_list})\n"
+            f"\n"
+            f"Note that you will only be able to run your code against real data if you actually have these\n"
+            f"permissions assigned by the OpenSAFELY team. For more information see:\n"
+            f"https://docs.opensafely.org/ehrql/reference/language/#permissions"
+        )
+        # For the inital rollout of this feature we issue a warning locally but continue
```
suzannehamilton, line 86: Typo: `inital`

ehrql/permissions.py
```diff
@@ -0,0 +1,145 @@
...
+        # running. Eventually we want to make this a hard error so that it can't be
```
suzannehamilton, line 87: :+1: 

Is doing this part of #2515 or should we make a new issue to make sure we don't forget it?

bloodearnest (approved):
This broadly makes sense to me.  Interesting to see how it is plumbed in to the table definitiones and qm.

Left a couple of questions, more for my own understanding of future intent.

ehrql/query_language.py
```diff
@@ -2019,8 +2019,13 @@ def table(cls: type[T]) -> T:
     else:
         raise Error("Schema class must subclass either `PatientFrame` or `EventFrame`")
 
+    try:
+        table_name = cls._meta.table_name
```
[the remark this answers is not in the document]
evansd: Good question. I haven't got as far as thinking what the API might look like. But there are a couple of tables which are currently special-cased in the dummy data generator:
https://github.com/opensafely-core/ehrql/blob/29f56df92c1fbfe00259cd34f6c76b3d3de60580/ehrql/dummy_data/generator.py#L203-L223

And my thought was that we can move some of this logic to the table definitions themselves, and then apply it more broadly to different kinds of table.

evansd:
Prompted by Simon's question, I've switch the format of the `EHRQL_PERMISSIONS` variable to be JSON rather than comma-separated strings. This gives us a bit more future-proofing and generally seems more robust. This means we need to merge and deploy this RAP Controller change first:
 * https://github.com/opensafely-core/job-runner/pull/1190

PR #2546, opened 2025-09-26
Remove reference to Gallagher paper

alschaffer:
This paper is not a comparison of death recorded in ONS vs primary care, but rather a comparison of ONS deaths with an algorithm based on primary care data used by CPRD which relies on information from across the primary care record (not just recorded date of death). The algorithm is not shared (either in this paper or elsewhere). Therefore, I don't think it is a useful reference about accuracy of primary care date of death.


evansd:
Actually, maybe post anyway for team visibility. I just happened to see this PR come in as I was looking anyway.

evansd (approved):

PR #2551, opened 2025-10-01
Add hints for errors which look like "accidental tuples" caused by trailing commas

### LINKED ISSUE #2532: Investigate adding more helpful errors for the "trailing comma / accidental tuple" problem
Leaving a trailing comma on a value and thereby inadvertently creating a tuple is a classic Python footgun.

For example, this fails

```py
was_alive = patients.is_alive_on(start_date),

dataset.define_population(
    was_alive
    & was_registered
    & has_known_sex
)
```
with an error like:
```
TypeError: Expected type `Series[bool]` but got 'tuple[Series[bool]]'
```

Which is extremely confusing to anyone who is not a seasoned Python programmer – which we expect to include the majority of our users.

It may well be possible to spot when this is likely to have happened and to give some helpful hints to the user.

## Implementation suggestion

This would involve extending `TypeValidationError` to record the `value` as well as the `field_name` in case of errors:
https://github.com/opensafely-core/ehrql/blob/491cc911cb5d3f8170e34fa5e08e1b52bff5ffd8/ehrql/query_model/nodes.py#L801-L806

And then, in the below error handling branch, we can check for instances of this issue:
https://github.com/opensafely-core/ehrql/blob/491cc911cb5d3f8170e34fa5e08e1b52bff5ffd8/ehrql/query_language.py#L1715-L1723

I think the logic we'd want would be something like:
```py
if exc.received == tuple[exc.expected] and len(exc.value) == 1:
    ... # raise extremely clear and helpful error message
```

This is on the basis that accidental trailing commas will always result in single membered tuples and therefore if we have a larger tuple then some other mistake has been made and showing the hint will only cause more confusion. That might be an incorrect assumption though.

## Links

For reference, here's a recent example of this issue being reported on Slack:
https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1756203322218699

And the associated support thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1756205498499269

evansd:
This has bitten more than one user in the past (see linked ticket) and feels like the sort of thing it's worth having in place before lots of new users start.

Closes #2532

### DIFF
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -284,6 +284,11 @@ def test_dataset_setattr_rejects_invalid_variables(variable, error):
         Dataset().v = variable
 
 
+def test_dataset_setattr_gives_hint_for_accidental_tuple():
+    with pytest.raises(TypeError, match="trailing comma"):
+        Dataset().d = patients.date_of_birth,  # fmt: skip
+
+
 def test_accessing_unassigned_variable_gives_helpful_error():
     with pytest.raises(AttributeError, match="'foo' has not been defined"):
         Dataset().foo
--- a/tests/unit/test_query_language.py
+++ b/tests/unit/test_query_language.py
@@ -1124,7 +1124,17 @@ def test_query_model_type_errors():
             "WARNING: The `|` operator has surprising precedence rules",
         ),
         (
+            lambda: patients.i == 1 | (patients.i == 2),
+            TypeError,
+            "WARNING: The `|` operator has surprising precedence rules",
+        ),
+        (
             lambda: patients.i == 1 & patients.i == 2,
+            TypeError,
+            "WARNING: The `&` operator has surprising precedence rules",
+        ),
+        (
+            lambda: patients.i == 1 & (patients.i == 2),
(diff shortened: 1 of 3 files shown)


tests/unit/test_query_language.py
```diff
@@ -284,6 +284,11 @@ def test_dataset_setattr_rejects_invalid_variables(variable, error):
         Dataset().v = variable
 
 
+def test_dataset_setattr_gives_hint_for_accidental_tuple():
+    with pytest.raises(TypeError, match="trailing comma"):
+        Dataset().d = (patients.date_of_birth,)
```
rebkwok, line 289: Should this be without the parentheses (and with a format skip comment) so it's more like an actually "accidental" tuple? It's all the same, really, so not too bothered either way

rebkwok (approved):
tiny nit, feel free to ignore

tests/unit/test_query_language.py
evansd: Oh yes, thanks. I think I wrote it that way originally but left off the format skip and so Ruff reformatted it.

PR #2584, opened 2025-11-05
Use dict rather than UserDict for dummy data

madwort:
* We don't appear to need the conveniences of UserDict
* dict is generally faster (as it is C rather than pure Python)
* in a benchmark case the overall dummy data generation runs in 32 secs rather than 40 secs


madwort:
oh actually I want to check something first

PR #2614, opened 2025-12-05
Minimal viable change to investigate changes to values

madwort:
* only use values from population, to see impact on found count


madwort:
In testing across the 700+ current dataset definitions, this appears to work extremely well for my initial test case, work well for 20 others, and work poorly or disasterously for all others. In particular [this study](https://github.com/opensafely/prophy_effects_Sotro_Molnup/tree/bf991811a0514dfde66815c4a723e8d4a6c65654/analysis) reported a reduction of 100% in the count! I'm investigating further...

PR #2620, opened 2025-12-15
Mock the DirectionsAcknowledged column for performance testing

rebkwok:
We want to test the performance of the gp activation filtering, but we don't yet have the new column. We temporarily mock it based on the Organisation_ID; this algorithm will return a determinate proportion of activated practices (75%, which can be modified to test different activation scenarios).

Temporarily create an organisation QueryTable to do this; in order to test just the effect of the activation filtering, this is used in all other tables that make use of the Organisation table (practice_registrations, all_practice_registrations, activated). We can test the same study code with and without activations by using this branch with and without the relevant EHRQL_PERMISSIONS env variable.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -170,21 +170,37 @@ class TPPBackend(SQLBackend):
                 "Mock GP activation rate: %s", self.environ.get("PCT_ACTIVATED", 0.75)
             )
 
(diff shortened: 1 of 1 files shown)


evansd (approved):

evansd (approved):
Ah yeah, sorry. I didn't consider that this would then bleed out into all the other tables that reference this table. Changes look good though.

PR #2636, opened 2026-01-14
Move the activation filtering config to table schema

rebkwok:
Add new `_meta` attributes to the tpp tables to specify the date field to be used for filtering, or that filtering is not applicable. 

Tests are passing locally, other than the ones that check for inclusion of an activated-related string in docs. (CI failing because docs need updated, which I don't want to do as part of this PR)

### DIFF
--- a/docs/includes/generated_docs/schemas/core.md
+++ b/docs/includes/generated_docs/schemas/core.md
@@ -26,6 +26,10 @@ Note that event codes do not change in this table. If an event code in the codin
 system becomes inactive, the event will still be coded to the inactive code.
 As such, codelists should include all relevant inactive codes.
 
+By default, only events with a consultation `date` on or before the date of the patient's
+last de-registration from an activated GP practice (a practice that has acknowledged the
+new non-COVID directions) are included.
+
 [Example ehrQL usage of clinical_events](../../how-to/examples.md#clinical-events)
 <div markdown="block" class="definition-list-wrapper">
   <div class="title">Columns</div>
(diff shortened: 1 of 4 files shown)


ehrql/backends/tpp.py
```diff
@@ -263,52 +190,154 @@ def modify_dataset(self, dataset):
...
+                            name=table.activation_filter_field,
+                        ),
+                        qm.SelectColumn(source=activated_table_node, name="end_date"),
+                    ),
+                )
+            dataset = replace_source(dataset, table.name, filtered_table)
+
+        return dataset
+
+    def _filtered_practice_registrations_table(self, table):
```
evansd: I was thinking about different ways of doing this depending on whether we want to expose the `activated` field to users in the first instance. We could do it as part of the definition of the `practice_registrations` query table. So something like:
```python
@QueryTable.from_function
def practice_registrations(self):
    if self.apply_gp_activations:
        filter_condition = "WHERE activated"
    else:
        filter_condition = ""
    f"""
    SELECT
        patient_id,
        ...
    FROM
        ...
    {filter_condition}
    """
```

The only real advantage of that is that it's one less user-facing thing we have to change and document. But I don't know how much that matters.

ehrql/backends/tpp.py
```diff
@@ -263,52 +190,154 @@ def modify_dataset(self, dataset):
...
+            elif table.name == "all_practice_registrations":
+                filtered_table = self._filtered_all_practice_registrations_table(
+                    table, activated_table_node
+                )
+            elif not table.activation_filter_field:
+                continue
+            else:
+                filtered_table = qm.Filter(
+                    table,
+                    qm.Function.LT(
```
evansd: We should work out whether we want `<` or `<=` here.

evansd (commented):
Great, this pretty much matches what I was thinking independently.

ehrql/tables/tpp.py
```diff
@@ -69,6 +69,9 @@ class addresses(EventFrame):
     [Example ehrQL usage of addresses](../../how-to/examples.md#addresses)
     """
 
+    class _meta:
+        _activation_filtered = False
```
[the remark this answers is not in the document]
rebkwok: I did think about that, but I didn't like having a field that could take either a boolean or a string value. I wasn't entirely happy about having two separate fields, though, so maybe that's the least bad option

ehrql/backends/tpp.py
rebkwok: Have asked Will in slack

PR #2646, opened 2026-01-19
Support materializing `QueryTable` s to temporary tables

evansd:
It's now possible to supply a `materialize=True` argument to `QueryTable` both in the standard form:
```python
my_table = QueryTable(
    " ... SQL query ...",
    materialize=True,
)
```

And in the decorator form:
```python
@QueryTable.from_function(materialize=True)
def my_table(self):
    return " ... SQL query ..."
```

This will cause the results of the query to be written to a temporary table meaning that the query is only executed once.

Whether it's a good idea to materialize any given query depends on multiple factors: how expensive it is to execute; how often it's expected to be used in a given dataset; and how large the results are expected to be.

---

### Historical note

The implementation here involves what looks like an unnecessary layer of subquery wrapping. In `QueryTable.get_expression()` we wrap the initial query in a subquery to make it table-like:
```python
table_expr = query.alias(table_name)
```

And then in `BaseSQLQueryEnginge.get_table_select_table()` we select columns from this table expression to create another query:
```python
return self.reify_query(sqlalchemy.select(*table_expr.columns))
```

My original implementation just retrieved the original query and passed it through:
```python
assert table_expr._is_subquery
return self.reify_query(table_expr.element)
```

This gave the correct results, but it triggered mysterious SQLAlchemy warnings when run against the Trino engine:
```
tests/integration/test_query_engines.py::test_materialized_query_table[trino]
  /home/dave/projects/bennett/ehrql/ehrql/query_engines/base_sql.py:1016:
    SAWarning: Number of columns in textual SQL (4) is smaller than number of columns requested (1)
    connection.execute(query)
```

As well as being obviously backwards (4 is traditionally considered _larger_ than 1) it proved very difficult to track down why the Trino engine, and only the Trino engine, was hitting this particular branch. I wasted far too much time on this before discovering that the alternative approach didn't trigger the warning and just deciding to go with that.

### DIFF
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -819,7 +819,11 @@ class BaseSQLQueryEngine(BaseQueryEngine):
     @get_table.register(SelectTable)
     @get_table.register(SelectPatientTable)
     def get_table_select_table(self, node):
-        return self.backend.get_table_expression(node.name, node.schema)
+        table_expr = self.backend.get_table_expression(node.name, node.schema)
+        if table_expr._annotations.get("materialize"):
+            return self.reify_query(sqlalchemy.select(*table_expr.columns))
+        else:
+            return table_expr
 
     # We ignore Filter and Sort operations completely at this point in the code and just
     # pass the underlying table reference through. It's only later, when building the
--- a/tests/integration/test_query_engines.py
+++ b/tests/integration/test_query_engines.py
@@ -6,6 +6,7 @@ import pytest
 import sqlalchemy
 
 from ehrql import create_dataset, maximum_of, minimum_of, when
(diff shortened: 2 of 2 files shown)


ehrql/backends/base.py
```diff
@@ -156,15 +156,23 @@ def get_expression(self, backend, table_name, schema):
...
         self.query = query
+        self.materialize = materialize
         self.implementation_notes = implementation_notes or {}
 
     @classmethod
-    def from_function(cls, fn):
-        instance = cls(query=None)
-        instance.get_query = fn
-        return instance
+    def from_function(cls, fn=None, materialize=False):
```
rebkwok: Do/would we ever call a `QueryTable.from_function` without a function?

PR #2666, opened 2026-02-10
Apply per-project GP activation filtering in the TPP backend

rebkwok:
By default, GP activation filtering will be applied. On a per-project basis, projects can be permitted to access data without applying the filtering. In order to do so, they must have the "include_gp_unactivated" filter, which is configured in job-server:
https://github.com/opensafely-core/job-server/blob/cd23b23c1a79dfa6576a8f78fe08abc83d7cc299/jobserver/permissions/population_permissions/gp_activations.py

As of 2026-02-10, all currently approved projects (all operating under the COVID directions) have this permission, so in production, GP activation filtering is not yet applied to any project, so although we're still waiting on database changes in order to apply the filtering for real, we can remove the overall feature flag.

Required some updating of tests to add in the extra permission. Applying GP activations is the default, and we don't want to have to include setup for tests that aren't testing it specifically.

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import re
 from urllib import parse
 
@@ -15,6 +16,9 @@ from ehrql.query_engines.mssql import MSSQLQueryEngine
 from ehrql.query_model import nodes as qm
 from ehrql.query_model.introspection import get_table_nodes
 from ehrql.query_model.transforms import replace_nodes
+
+
+logger = logging.getLogger(__name__)
 
 
 class TPPBackend(SQLBackend):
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -343,13 +344,31 @@ def random_should_not_be_used():
...
+
+
 @pytest.fixture
 def call_cli(capsys):
     """
     Wrapper around the CLI entrypoint to make it easier to call from tests
     """
 
-    def call(*args, environ=None):
+    def call(*args, environ=None, gp_activation_filtering_disabled=True):
```
evansd: I guess we could make this slightly less specific to this one use case by doing:
```python
def call(*args, environ=None, permissions=("include_gp_unactivated",)):
```

And then a function like:
```python
def add_permissions_to_environ(environ, permissions):
    ...
```

evansd:
Oh, and of course we need a pedantic commit prefix...

Issue #2595, opened 2025-11-13
Schema table metadata to enhance dummy data generation


madwort:
Look into this idea 

> Add extra, handcrafted metadata to our schema tables which influence the dummy data generation. For example, the hospital admissions tables should “know” not to generate discharge dates before admission dates and not to generate overlapping admissions. We don’t need to come up with some beautiful, abstract declarative system for expressing these constraints: this is a purely internal API and we can evolve it as our dummy data needs evolve.

from [document](https://docs.google.com/document/d/1yXRe6pFufd39TQvmM1plnhy-jDTVE_jdwYrkqfRaDGY/edit?tab=t.0)

madwort:
In the course of investigating this, I came across this: https://github.com/opensafely-core/ehrql/pull/2612

PR #2676, opened 2026-02-13
Generate date values in the same row in expected chronological order

alarthast:
Part of #2595; supercedes #2637.
The only substantial change from #2637 is the final commit with the docs change. Everything else is small tweaks to formatting / wording.

This PR focuses on this part of the ticket:

> For example, the hospital admissions tables should “know” not to generate discharge dates before admission dates ...

Having related date columns (e.g. a `start_date` and an `end_date`) being "out of order" is [something that has been brought up by a user](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1755010408886709), so hopefully eliminating out of order dates might make dummy data and dummy tables easier to use.

Here, we add a `dummy_data_constraints` field to `Series`. 

We also add a new type of constraint (`Constraint.DateAfter`), which allows us to store information about the expected chronological order of columns. `DateAfter` constraints are picked up when building the `QueryInfo` for dummy data generation, such that we can swap values around post-generation until dates are in the order we expect. 

### Profiling
Manual profiling done with the repos highlighted in yellow in [this spreadsheet](https://docs.google.com/spreadsheets/d/1WOqgBBuB0nAhasygD-N6biPE3YHsw2933KvdBRXPY54/edit?gid=191444011#gid=191444011). 
None of the runs showed significant decrease in the cumulative time per call of the `get_patient_data` method. 
Eyeballing, it seems like the number of patients generated is also around the same in most cases. 
In the slowest case (no. 5 in the spreadsheet), there seems to be a small improvement (89 patients vs 120 patients).

### DIFF
--- a/docs/tutorials/dummy-data/limitations-of-native-ehrQL-dummy-data/index.md
+++ b/docs/tutorials/dummy-data/limitations-of-native-ehrQL-dummy-data/index.md
@@ -8,6 +8,14 @@
     code = clinical_events.sort_by(clinical_events.date).first_for_patient().snomedct_code
     ```
     (To some extent building up your dummy data tables can work around this sort of thing.)
+- The date logic in generated dummy data might be too simplistic and hence may not reflect idiosyncrasies
+in the real data.
+    - For example, when ehrQL generates the `addresses` dummy table, it will ensure that the `start_date` occurs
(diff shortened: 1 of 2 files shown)


alarthast:
Thanks very much! Changes made :smiley:

PR #2678, opened 2026-02-17
Limit dates in dummy data by data availability

### LINKED ISSUE #2664: Constrain date range in dummy data generation for some tables
This ticket originated from [this Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1769789655526199)

Some of our tables only have data going back to a certain date, but the dummy data generator does not know about that. Due to this, a user recently had some code that ran successfully on dummy data but did not return any results when run on the real data.

Given the current work being done in #2595, we could annotate the date columns in certain tables such that the dummy data only generates dates in a narrower range that matches the availability of the real data. This work will include surveying what columns we would like to annotate, and what date range to use for each.

We could annotate the date columns of interest by adding `GeneralRange` constraints as dummy data constraints to them. It would be good to first test out whether such added constraints have an impact on performance. 

If we do observe significant impact on performance, we may consider looking at the generator code that concerns generating base values for date columns, and modifying it to take column constraints into account.
https://github.com/opensafely-core/ehrql/blob/072a463abdd1727943f6ff1e162d4d28ba38c743/ehrql/dummy_data_nextgen/generator.py#L515-L520

alarthast:
Closes #2664.

**The commits will be squashed post approval into a single commit; they are separated for the time being for easier review (e.g. if a reviewer suggests to drop certain changes).**

* Limit `apcs` dates to after Apr 2016

* Limit `decision_support_values` calculation date to 2020-12-08
(i.e. the date for which all the values were calculated), and exclude
nulls

* Limit `ec` dates to after Oct 2017

* Limit `opa` dates to after Apr 2019 in dummy data

* Limit `sgss_covid_all_tests` dates to after Jan 2020

* Limit `ons_deaths` dates to after Feb 2019

The `decision_support_values.calculation_date` field is nullable (#1523), but make it non-nullable in the dummy data since:
- if any real data row is missing a `calculation_date`, it presumably should be 2020-12-08 anyway
- if `None` is allowed, the generated data will be skewed towards `None` which probably ends up being less realistic than having no `None` values at all.

The TPP database report shows that
- `ec.arrival_date` has values from late 2017 (assumed Oct 1st) onwards
- `opa.appointment_date` has values from Apr 2019 onwards

It could be possible that `ec.injury_date` has values before 2017-10-01, and/or that `opa.appointment_date` has values before 2019-04-01, but as a simplification, we will constrain all the date columns in each table to be after the same date.

In the TPP backend, we only have data going back until Feb 2019 for `ons_deaths`, which is a core table.
We unfortunately have to apply this constraint in `core.py` even though the value is based on data availability in the TPP backend.

### DIFF
--- a/docs/explanation/dummy-data.md
+++ b/docs/explanation/dummy-data.md
@@ -73,7 +73,7 @@ Since the configured `population_size` is 10, ehrQL outputs a dummy dataset cont
     beyond those required by your dataset or measures definition.
 
     The dummy data generator incorporates the additional constraint when identifying matching
-    patients.
+    patients, but see [note on `additional_population_constraint`](#note-on-additional_population_constraint).
 
 ## ehrQL generates dummy patients in batches until `population_size` is met (or timeout)
 ehrQL generates dummy patients in batches.
@@ -119,8 +119,15 @@ values for certain tables. You should not expect the same "clean" data in the re
 
 ## Appendix: Simplifications made in ehrQL's natively generated dummy data
 
+### Note on `additional_population_constraint`
+- Note that you cannot circumvent these simplifications by defining an `additional_population_constraint`.
+- For example, requesting negative values of `practice_pseudo_id` via `additional_population_constraint`
+will lead to the generator failing to generate any matching patients.
+
+
 ### General
-- All dates are always on or between `1900-01-01` and today's date
+- All dates are always on or between `1900-01-01` and today's date, except when there is a column-specific date
+minimum. Column-specific date minima are listed in the sections below and are based on the real data availability.
 - A patient's events always occur after their `date_of_birth` (and before their `date_of_death` if they have one)
 - A patient is only registered with a single practice from birth
 


rebkwok (commented):
This looks fine, and the chosen date limits make sense. There may be out-of-range dates in the real data, but the limits should make the dummy data generally more sensible. I wonder if it would make sense to add the date range to the overall table metadata in some way (maybe an _-prefixed `_meta` attribute so that it only needs to be updated in one place. At least atm, where there are multiple date fields in a table, we're applying the same minimum date to all of them.

The dummy data explanation page in the docs that pull in the description of constraints will be super useful for keeping track of these data date constraints too.

One thing that occurred to me - how would these date limits interact with `additional_population_constraints`? If a user wanted to simulate messy data by adding an additional population constraint that specified that admission date could be before 2016-04-01, would we be able to generate any dummy data that met those constraints?

PR #2684, opened 2026-02-23
Catch column name mismatch in assurance testing

### LINKED ISSUE #2681: Provide better error message for expected column key error in assurance testing
Reported on Slack by a user in [this thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1771317683666779).

If a key in the user's provided `expected_columns` does not appear in the actual dataset, Python raises a `KeyError` with an error message that might not be immediately helpful to the user:

```
File "/app/ehrql/assurance.py", line 71, in validate
         if (validation_result := validate_patient(patient_id, patient, query_results))
                                  ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       File "/app/ehrql/assurance.py", line 139, in validate_patient
         if expected[k] != actual[k]
                           ~~~~~~^^^
     KeyError: 'reps_ids_differ'
```

We might want to catch the mismatch earlier and return a custom error message.

alarthast:
Fixes #2681

If the user's provided expected_columns contains a key that does not match any of the actual column names in the events, provide a more informative error than Python's default KeyError message.

Like other errors in assurance testing, only report the error at the end, for the relevant patient(s).

Although this is a test definition error rather than results error, the mismatch blocks results validation, so report it as a error
in the "Validate results" section.

~~Looking at this, I'm wondering if it's worth waiting for this error to be hit by a user a second time to put in these changes,
as a test definition error like this doesn't sit neatly within the "data" / "result" error categories we currently have.
(To be fair though, there might be users hitting this and having to google the KeyError without telling us, but still.)~~

### DIFF
--- a/ehrql/assurance.py
+++ b/ehrql/assurance.py
@@ -11,6 +11,7 @@ UNEXPECTED_COLUMN = "unexpected-column"
 UNEXPECTED_ROW_COUNT = "unexpected-row-count"
 UNEXPECTED_IN_POPULATION = "unexpected-in-population"
 UNEXPECTED_NOT_IN_POPULATION = "unexpected-not-in-population"
+UNEXPECTED_KEY_IN_EXPECTED_COLUMNS = "unexpected-key-in-expected-columns"
 UNEXPECTED_OUTPUT_VALUE = "unexpected-output-value"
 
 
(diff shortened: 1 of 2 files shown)


alarthast:
> Oh - you need a pedantic semantic commit otherwise this won't trigger a new image version

Thank you - I'll start remembering to do this any day now :sweat_smile:

PR #2687, opened 2026-02-26
Extend create-dummy-tables to measures

rebkwok:
Allow create-dummy-tables to accept a dataset or a measures definition. Both use the same dummy data generators, so it's relatively trivial to dump the tables generated for a measures definition too.

### DIFF
--- a/tests/functional/test_create_dummy_tables.py
+++ b/tests/functional/test_create_dummy_tables.py
@@ -34,10 +34,10 @@ def trivial_dataset_definition_legacy_dummy_data():
 
 @function_body_as_string
 def trivial_measures_definition():
-    from ehrql import INTERVAL, Measures, years
+    from ehrql import INTERVAL, create_measures, years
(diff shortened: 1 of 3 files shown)


tests/functional/test_create_dummy_tables.py
```diff
@@ -32,25 +32,42 @@ def trivial_dataset_definition_legacy_dummy_data():
     dataset.configure_dummy_data(population_size=10, legacy=True)
 
 
+@function_body_as_string
+def trivial_measures_definition():
+    from ehrql import INTERVAL, Measures, years
+    from ehrql.tables.core import patients
+
+    measures = Measures()
```
alarthast: Nit:

```suggestion
    from ehrql import INTERVAL, create_measures, years
    from ehrql.tables.core import patients

    measures = create_measures()
```

to be consistent with the other fixtures.

PR #2715, opened 2026-03-16
Automatically include all spec tests in docs

evansd:
As discussed elsewhere, I'm not wild about these as user-facing docs; but while we have them they ought at least to be comprehensive.

I've taken the approach of automatically inserting any missing specs into the table of contents. Alternatives are:

 * Auto-generating the ToC entirely – but then we lose the ordering which was obviously deliberately chosen to have more fundamental concepts first.
 * Requiring a fully specified manual table of contents – but that just creates extra overhead for a feature that we're not convinced is worth the effort.

Retaining the existing order and inserting missing sections/modules seems like a reasonable compromise.

### DIFF
--- a/docs/includes/generated_docs/specs.md
+++ b/docs/includes/generated_docs/specs.md
@@ -441,6 +441,34 @@ returns the following patient series:
 
 
 
+### 2.5 Sort extends to all columns when underspecified to ensure that sort order is consistent
+
+
+#### 2.5.1 Sorting extends to selected column
+
+This example makes use of an event-level table named `e` containing the following data:
+
+| patient|i1|i2|i3 |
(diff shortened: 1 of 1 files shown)


rebkwok (approved):
Huh, there were more missing from the docs than I'd expected.

Is it worth having some sort of individual test-specific marker that allows a spec test to be deliberately excluded and not pulled in automatically? Are there features that could be usefully tested by a spec test, but which don't format well (I removed the table_from_rows spec tests a few weeks back mostly because they didn't format properly - not a great example though because it's not a user-facing feature so didn't really belong in the spec tests).  Perhaps the dummy module is sufficient for those though.

PR #2743, opened 2026-04-08
Convert EMIS v2 schema (from spreadsheet) to a Python dictionary

Providence-o:
Closes: #2726 

This script primarily exists so that we can have foreign key information about the EMIS v2 schema. 

It reads the [schema spreadsheet](https://docs.google.com/spreadsheets/d/1Lx6exxIJqGv50n0wXZpCCxuRMTK2Sidy3IXCNtA4pnw/edit?gid=512859146#gid=512859146), extracts each sheet into CSV files and stores the data in a dictionary which will be used to [generate a relationship diagram](https://github.com/opensafely-core/ehrql/issues/2728).


pyproject.toml
```diff
@@ -16,21 +16,18 @@ dependencies = [
...
   "sqlean.py",
-
   # For graphing query graphs
   "networkx",
   "pydot",
-
   # speed up for dummy data next generation
-  "numpy"
+  "numpy",
+  "openpyxl>=3.1.5",
```
evansd: I think this should go as a dev dependency because we don't need this in the production image.

They're slightly hidden away at the bottom of the `pyproject.toml` file:
https://github.com/opensafely-core/ehrql/blob/949eebd1d15a30f0a86d87b8560edeb9f0970b74/pyproject.toml#L139-L140

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+RAW_FILES_DIR = Path(__file__).parent / "raw_emis_csv"
+EXCEL_INPUT_DIR = RAW_FILES_DIR / "raw_emisv2_schema"
+
+EXCEL_INPUT_DIR.mkdir(parents=True, exist_ok=True)
+
+
+def extract_sheets_from_excel_file(excel_file, file_dir):
+    workbook = load_workbook(EXCEL_INPUT_DIR / excel_file)
```
evansd: I don't think you need this separate `EXCEL_INPUT_DIR` constant. You can just let the user pass in the path to the Excel file wherever they happen to have downloaded it to.

That means you can get rid of the extra gitignore config as well.

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,109 @@
...
+            ],
+            "medication": [
+                ...
+            ],
+        }
+
+    """
+
+    if file_dir is None:  # pragma: no cover
+        file_dir = RAW_FILES_DIR
```
evansd: You could make `RAW_FILES_DIR` the default for this argument and then you wouldn't need to do this check. i.e.
```python
def get_schema_from_csv(excel_file, file_dir=RAW_FILES_DIR):
```

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,109 @@
...
+
+    extract_sheets_from_excel_file(excel_file, file_dir)
+
+    schema = {}
+
+    # Iterate over sub directories in 'raw_emis_csv' directory
+    for file in file_dir.iterdir():
+        # Confirm that it's a csv file
+        if file.suffix == ".csv":
```
evansd: Checks like this are often called "guard clauses" and it can make this easier to read if you invert them and use `continue` rather than indenting the whole body of the loop i.e.
```python
if file.suffix != ".csv":
    continue
```

In fact, in this case there's an even simpler alternative which is to use:
```python
for file in file_dir.glob("*.csv"):
    ...
```
Which will only find CSV files so there's no filtering to do.

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,109 @@
...
+
+RAW_FILES_DIR = Path(__file__).parent / "raw_emis_csv"
+EXCEL_INPUT_DIR = RAW_FILES_DIR / "raw_emisv2_schema"
+
+EXCEL_INPUT_DIR.mkdir(parents=True, exist_ok=True)
+
+
+def extract_sheets_from_excel_file(excel_file, file_dir):
+    workbook = load_workbook(EXCEL_INPUT_DIR / excel_file)
+    file_dir.mkdir(exist_ok=True)
```
evansd: I think you'll need an extra step here which is to delete all the existing CSV files. If EMIS remove or rename a table then we need to make sure that the corresponding CSV file no longer exists and the easiest way to do that is to just to delete all of them before we create the new ones. Something like:
```python
for filename in file_dir.glob("*.csv"):
    filename.unlink()
```

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def run():  # pragma: no cover
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "input_file", type=str, help="Raw file as received from EMIS for v2 backend"
+    )
+    args = parser.parse_args()
+
+    get_schema_from_csv(args.input_file)
```
evansd: I think it would be good to have two separate commands here: one for unpacking the XLSX into CSV and another for generating the schema output. We may well want to make changes to the schema generation without having the original XLSX around to unpack again.

This is similar to what we do with the TPP schema generation here:
https://github.com/opensafely-core/ehrql/blob/949eebd1d15a30f0a86d87b8560edeb9f0970b74/tests/backend_schemas/tpp/update_schema.py#L245-L252

evansd (commented):
Thanks Providence, this is good and I think very close to being ready. I've made a few comments below.

One other small thing which doesn't matter much in this case but is a good thing to bear in mind in future is that it can be helpful to have changes which introduce a large amount of external data (like adding the CSV files) in a separate commit from the code changes. It makes it easier for a reviewer to make sure they've looked at all the code changes without being overwhelmed by all the other content. (Obviously in this case the CSV files are all a single directory so it's easy to skip over, but I think it's a good habit to get into.)

Providence-o:
> it can be helpful to have changes which introduce a large amount of external data (like adding the CSV files) in a separate commit from the code changes.

Thanks Dave, I'll take note of this for the future

tests/backend_schemas/emisv2/script.py
```diff
@@ -0,0 +1,111 @@
...
+from pathlib import Path
+from pprint import pprint
+
+from openpyxl import load_workbook
+
+
+RAW_FILES_DIR = Path(__file__).parent / "raw_emis_csv"
+
+
+RAW_FILES_DIR.mkdir(parents=True, exist_ok=True)
```
evansd: If you added the `parents=True` flag to line 33 i.e.:
```python
file_dir.mkdir(parents=True, exist_ok=True)
```
then I think you could get rid of this line entirely.

There's no real harm in having it, but in general it's better to avoid what are called "import time side-effects" i.e. when just importing a module but not calling any functions in it still causes some change in external state.

evansd (approved):
Looks great, Providence, thanks!

I've made one small, optional comment on something I missed last time. It's another one of those things which isn't wrong as such, but might not be considered best practice.

PR #2746, opened 2026-04-09
Add a minimal EMISV2 backend

### LINKED ISSUE #2744: Implement a minimal EMIS v2 backend to support the smoketest schema
The [smoketest schema](https://docs.opensafely.org/ehrql/reference/schemas/smoketest/) implements a slim `patients` table with only the patient ID and a `date_of_birth` column.

Implement a minimal EMIS v2 backend that supports the [smoketest dataset definition](https://github.com/opensafely-core/ehrql/blob/main/tests/acceptance/external_studies/test-age-distribution/analysis/dataset_definition.py), which queries `smoketest.patients`.

[ehrql#2735](https://github.com/opensafely-core/ehrql/pull/2735) should be used as a starting point for this ticket; it already contains an implementation of the backend that only needs to be covered by tests.

Note the following caveat in the description of #2735:

> We don't have permission to create tables in the [staging] db, hence the change to the Trino query engine; we should assume we will have that permission eventually

Hence, do not include the changes to `query_engines/trino.py` in #2735: only short-circuit `reify_query` before table creation when connecting to the staging environment locally. Tests that run against our `trino_database` fixture should be creating tables. 

To complete this ticket, one would need to:
- [ ] Connect the backend to the staging environment locally, and verify that the smoketest dataset definition runs successfully (instructions in #2735)
- [ ] Add a test in `tests/integration/backends/test_emisv2.py` that executes an inline dataset definition that queries `smoketest.patients` against the EMISv2 backend and the test Trino database. (Also see [this Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1775659053703899).) 
- [ ] Ensure that the backend is excluded from the docs (i.e. the docs should only show the existing EMIS backend, not the EMIS v2 backend)
- [ ] Ensure that CI passes

alarthast:
Closes #2744. Previous version: #2735.

The added EMISV2 backend implements the smoketest patients table as a bare minimum.

Alias it as "emisv2" so that ehrql commands can be run with it using `--backend emisv2` for development purposes.
This means that "emisv2" will appear in the cli help text, which is deemed to be acceptable.

We however do not want to display the backend prominently on the docs site, so exclude the backend from the docs.

We also add an integration test for the backend.
This uses the `Patient` orm object in tests/backend_schemas/emisv2/schema.py.
Since the integration test is in place, we can remove the temporary test for the orm object definition.


evansd (commented):
Sorry for drive-by reviewing! Just wanted to get these comments in before I forget.

evansd (approved):

PR #2749, opened 2026-04-09
Generate EMIS v2 backend schema diagram 

Providence-o:
Closes #2728

Generates a diagram showing the relationship between the tables, which helps us (the developers) understand the foreign key linkages in the EMIS v2 schema.

Because of the complexity of the schema diagram, the tables and connecting lines have been colour coded for easier visualisation


tests/backend_schemas/emisv2/schema_diagram.py
```diff
@@ -0,0 +1,128 @@
...
+]
+
+
+def get_colour(index):
+    return DISTINCT_COLOURS[index % len(DISTINCT_COLOURS)]
+
+
+def rename_schema_table(table_name_from_foreign_key_field):
+    """
+    Some table names are labelled incorrectly in the foreign key fields. This is known behaviour for two tables:
```
rebkwok, line 52: Huh, interesting. This definitely implies the spreadsheet is manually crafted, doesn't it?

tests/backend_schemas/emisv2/schema_diagram.py
```diff
@@ -0,0 +1,128 @@
...
+    1. mapping_dmdpreparation.dmd_product_code_id in the 'medication_drug_record' table
+    2. mkb_mapping_attribute.code_id in the 'patient' table
+    """
+
+    if table_name_from_foreign_key_field == "mapping_dmdpreparation":
+        return "mkb_mapping_dmdpreparation"
+    elif table_name_from_foreign_key_field == "mkb_mapping_attribute":
+        return "mkb_mapping_attributes"
```
rebkwok: This is fine, but you could also consider defining a dict mapping name: corrected name if instead of the if/elif (it would be worth it if there were more than 2 mappings to do)

tests/backend_schemas/emisv2/schema_diagram.py
```diff
@@ -0,0 +1,128 @@
...
+
+    graph = pydot.Dot("my_graph", nodesep="2", ranksep="2", splines="compound")
+    graph.set_type("digraph")
+
+    # List of tuples which holds the foreign key links between tables
+    foreign_key_links = []
+
+    table_colours = {}
+
+    for index, (table_name, table_content) in enumerate(emis_schema.items()):
```
rebkwok, line 76: It doesn't look like script.py does any whitespace stripping from the rows read from the xls , might be worth doing a `.strip()` on `table_name` here just in case. Ditto for column_name below

tests/backend_schemas/emisv2/schema_diagram.py
```diff
@@ -0,0 +1,128 @@
...
+        table_items = [table_name]
+
+        for row in table_content:
+            if row["Foreign Key"]:
+                column_name = row["Column"]
```
rebkwok: Could pull this line up to before the `if`, as it's used whether the row is a FK or not, and then you don't need to repeat it on line 95 below

tests/backend_schemas/emisv2/schema_diagram.py
rebkwok: oh, you do do table_name.strip() in the foreign key links; and then on line 101 you reassign it with 
```table_name = table_name.strip()```
I'd just do that here, and then you can use `table_name` as it is without having to strip in multiple places

File: tests/backend_schemas/emisv2/schema_diagram.py [hunk 5]
```diff
@@ -0,0 +1,128 @@
+"""
+To generate the relationship diagram:
+
+    # Install the graphviz system binary (this is different from the graphviz python package)
+    sudo apt install graphviz
+
+    python3 tests/backend_schemas/emisv2/schema_diagram.py
```
rebkwok: nit; I don't think this necessarily needs to be a justfile recipe, but `uv run python ...` is more consistent with how we do other things and means it should work even if you don't have your venv activated

rebkwok (approved):
A few minor nits, but I think this looks fine, and the code was nice and readable.

I'm not sure I find the actual diagram useful, but that's nothing to do with your code!

tests/backend_schemas/emisv2/schema_diagram.py
Providence-o: Yes it does

PR #2750, opened 2026-04-10
Use correct temporary schema for EMISV2 backend

evansd:
EMIS have configured things such that each user has a writable schema whose name matches the username.

With this PR in place we can successfully run queries against the staging database using:
```
ehrql generate-dataset \
  tests/acceptance/external_studies/test-age-distribution/analysis/dataset_definition.py \
  --backend emisv2 \
  --dsn trino://USERNAME:TOKEN@explorerplus.stagingemisinsights.co.uk:443/hive/explorer_open_safely
```

### DIFF
--- a/tests/support/trino/setup.sql
+++ b/tests/support/trino/setup.sql
@@ -1,1 +1,4 @@
--- SQL to setup the database goes here
+-- In production each user has a writable schema named with their username in which to
+-- create temporary tables; we want to mirror that set up here so we create a "trino"
+-- schema (named after the default user) in the default "trino" catalog.
+CREATE SCHEMA trino.trino;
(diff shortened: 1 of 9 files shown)


tests/support/trino/setup.sql
```diff
@@ -0,0 +1,4 @@
+-- In production each user has a writable schema named with their username in which to
+-- create temporary tables; we want to mirror that set up here so we create a "trino"
+-- schema (named after the default user") in the default "trino" catalog.
```
alarthast:
```suggestion
-- schema (named after the default user) in the default "trino" catalog.
```

alarthast (approved):
Verified that:
- This code allows ehrql to create tables in the alice.wong schema when I connect to the staging environment
- Doing `just connect-to-trino` and running `SHOW CREATE SCHEMA trino;` gives me a valid output (and `SHOW CREATE SCHEMA unknown;` fails with a "Schema 'trino.unknown' does not exist".

PR #2751, opened 2026-04-13
Implement `core.patients`,  `core.clinical_events` and  `core.medications` in EMIS v2 backend

alarthast:
Closes #2725. Replaces #2736.

Following the discussion in [this Slack thread](https://bennettoxford.slack.com/archives/C080S7W2ZPX/p1775144942900229?thread_ts=1775124493.243089&cid=C080S7W2ZPX), here we:
- Add ORM object definitions for EMIS v2 observations and medications in the test schema
- Implement the core `patients`, `clinical_events` and `medications` tables on the EMIS v2 backend
- Add integration tests for the EMIS v2 backend that validate the implemented tables using the test schema objects

I think I do not need a `feat:` prefix for these changes? Please correct me if I'm wrong!

### ORM object definitions
As done in #2739, use column types obtained by connecting to EMIS's staging environment using a SQLAlchemy session. (See PR description of #2739 for details on how this was done.)

### Query tables
These were mostly straightforward mappings with some `CAST`ing to appropriate types that ehrQL expects. The exception is that the `patients.sex` column requires a case statement. As noted in the commit message, I have written the case statement based on values found in the staging environment (`M`,`F`,`I`,`U` plus some random varchars). Here's the relevant Explorer screenshot for information:
Screenshot from 2026-04-10 13-36-46

### DIFF
--- a/ehrql/backends/emisv2.py
+++ b/ehrql/backends/emisv2.py
@@ -29,10 +29,19 @@ class EMISV2Backend(SQLBackend):
         return parts.username
 
     patients = QueryTable(
+        # Note: sex = 'U' (unknown) is a common value and
+        # is handled by the ELSE clause
         """
         SELECT
             patient_id AS patient_id,
-            CAST(date_of_birth AS date) AS date_of_birth
+            CAST(date_of_birth AS date) AS date_of_birth,
+            CASE
+                WHEN sex = 'M' THEN 'male'
+                WHEN sex = 'F' THEN 'female'
+                WHEN sex = 'I' THEN 'intersex'
+                ELSE 'unknown'
+            END AS sex,
+            CAST(date_of_death AS date) AS date_of_death
         FROM patient
         """
     )
--- a/ehrql/backends/emisv2.py
+++ b/ehrql/backends/emisv2.py
@@ -62,3 +62,17 @@ class EMISV2Backend(SQLBackend):
         FROM observation
         """
     )
+
+    medications = QueryTable(
+        # Note that we use the medical issue record's effective date here rather
+        # than the date of the linked consultation for the medical issue record
+        # (which is what we use in TPP). This is not a permanent decision -
+        # we may revise the implementation in the future.
+        """
+        SELECT
+            patient_id,
+            CAST(effective_datetime AS date) as date,
+            CAST(dmd_product_code_id AS varchar) AS dmd_code
+        FROM medication_issue_record
+        """
+    )
--- a/ehrql/backends/emisv2.py
+++ b/ehrql/backends/emisv2.py
@@ -32,7 +32,7 @@ class EMISV2Backend(SQLBackend):
         """
         SELECT
             patient_id AS patient_id,
-            date_of_birth
+            CAST(date_of_birth AS date) AS date_of_birth
         FROM patient
         """
     )
--- a/ehrql/backends/emisv2.py
+++ b/ehrql/backends/emisv2.py
@@ -1,5 +1,6 @@
 import urllib.parse
 
+import ehrql.tables.emisv2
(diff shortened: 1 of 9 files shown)


ehrql/backends/emisv2.py
```diff
@@ -32,7 +32,15 @@ def modify_temp_table_schema(self, temp_table_schema, dsn, environ):
         """
         SELECT
             patient_id AS patient_id,
-            CAST(date_of_birth AS date) AS date_of_birth
+            CAST(date_of_birth AS date) AS date_of_birth,
+            CASE
+                WHEN sex = 'M' THEN 'male'
+                WHEN sex = 'F' THEN 'female'
+                WHEN sex = 'I' THEN 'intersex'
+                WHEN sex = 'U' THEN 'unknown'
```
rebkwok: We don't need `WHEN sex = 'U' THEN 'unknown'`, the `ELSE` will catch U along with anything else. Is it included deliberately because it's a common value in the staging db?

tests/integration/backends/test_emisv2.py
```diff
@@ -45,6 +52,65 @@ def test_extract_smoketest_dataset_definition(trino_engine):
     assert results == [{"patient_id": bytes(range(16)), "age": 22}]
 
 
+def test_backend_columns_have_correct_types(trino_database):
```
rebkwok, line 57: Nice, I was hoping you were going to add this test 😄

tests/integration/backends/test_emisv2.py
```diff
@@ -45,6 +52,65 @@ def test_extract_smoketest_dataset_definition(trino_engine):
...
+        queries.append(
+            (temp_table_name, temp_table, CreateTableAs(temp_table, select_query))
+        )
+    # Create all the underlying tables in the database without populating them
+    trino_database.setup(metadata=Patient.metadata)
+    # Note: we really ought to have something like the setup/cleanup call below to
+    # handle materialized query tables (which we happen not to have any of in the EMIS
+    # backend at present). But the structure of the `queries` list doesn't currently
+    # allow for this and I think there's a wider refactor here which would make this
+    # problem go away so I'm going to punt on it for now.
```
rebkwok, line 91: We should probably make a note to look at this, since we're re-using this from the old emis backend

rebkwok (approved):

ehrql/backends/emisv2.py
alarthast: Yes, it was included deliberately since it's a common value - I thought it's better to be explicit. Should I remove the line?

rebkwok: Maybe remove it but add a comment?

tests/integration/backends/test_emisv2.py
alarthast: I can make a ticket quoting this comment so that we can put it on the board :)

ehrql/backends/emisv2.py
alarthast: Added:

https://github.com/opensafely-core/ehrql/blob/f3d0778a159069af88495314124c1a4fec9655a8/ehrql/backends/emisv2.py#L31-L43

tests/integration/backends/test_emisv2.py
alarthast: I've added the ticket here:
https://github.com/opensafely-core/ehrql/issues/2754

evansd:
Just hoping in post-review to say that this is really nice work. There's lots of fiddly stuff to deal with here but you've handled it all exactly as I would have suggested.

Refactoring that SQL type testing code has been at the back of my mind for a while. Thanks for making that into a proper ticket.

PR #2756, opened 2026-04-14
Generalise local database tooling beyond TPP

evansd:
We already had some tooling which would reuse the test machinery to set up an MSSQL Docker container, create all the tables specified in the TPP schema, and print the necessary connection string. This is very useful in being able to reproduce production behaviour locally.

Following this PR we can use this same machinery in other contexts. For instance, we can create the user-facing TPP schema in a MSSQL container:

    just create-tables mssql ehrql/tables/tpp.py

And then run a dataset definition against this:

    DATABASE_URL='<output-from-above>' \
      ehrql generate-dataset --query-engine mssql \
        analysis/dataset_definition.py

Or create the EMIS backend tables in a Trino container:

    just create-tables trino tests/backend_schemas/emisv2/schema.py

And then run a dataset definition against this:

    DATABASE_URL='<output-from-above>' \
      ehrql generate-dataset --backend emisv2 \
        analysis/dataset_definition.py


rebkwok (approved):
Nice. I've quite often wanted to do this to debug a user-reported problem with a dataset definition, and wasn't aware of how to use the `create-tpp-test-db` to get a mssql db url to pass to ehrql. I've usually ended up adding some sort of temporary test into ehrql. Can we add something similar to the commit message to DEVELOPERS.md to document it?

PR #2768, opened 2026-04-23
Fix a whole stack of error handling problems

evansd:
A user recently had a job which worked locally and failed in production in a way they found hard to debug. The root cause turned out to be user error, but it exposed a catalogue of small-ish issues with our error handling code which made this all unnecessarily confusing.

The fixed issues are:
 * Failing to hide internal sections of the stack trace (as we do elsewhere) which made the error more intimidating and harder to read than necessary.
 * Not reporting the filename of the data file which triggered the error.
 * Not using the correct exit code for the error meaning it was misreported in Job Server.
 * Reading files in places we shouldn't be, meaning the error was raised in an unexpected place in the user code. 

These weren't entirely trivial to fix. The exit code issue, in particular, exposed a bit of think-o on my part as to how errors work in ehrQL. But I think the error handling code is in a much better state now than it was before.

Slack thread:
https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1776856993059439


ehrql/exceptions.py
```diff
@@ -1,11 +1,12 @@
 class EHRQLException(Exception):
-    """Base exception for EHRQL errors of all sorts.
+    """
+    Base exception for EHRQL errors of all sorts.
 
     This is not yet reliably used everywhere it should be.
```
rebkwok, line 5: ~~With the changes in this and the last commit, are we now using it reliably everywhere?~~

rebkwok (approved):

PR #2770, opened 2026-04-27
Get EMIS v2 backend schema from Staging Explorer

### LINKED ISSUE #2761: Write script to consume EMIS v2 backend schema from Explorer
This script will need to get backend schema information from Staging Explorer. It would require: 

- Access to staging using EXA_USERNAME and EXA_TOKEN env variables
- Writing sql queries to get each tables information and parsing the results into a python dictionary using some of the logic from [the speadsheet-based version of the script](https://github.com/opensafely-core/ehrql/blob/main/tests/backend_schemas/emisv2/script.py)
- Writing the data from the python dict into a csv file which should look like [tpp's schema csv](https://github.com/opensafely-core/ehrql/blob/949eebd1d15a30f0a86d87b8560edeb9f0970b74/tests/backend_schemas/tpp/schema.csv). CSV should have tables, columns, data types, other metadata like length limits or nullability if we have it. 
- Creating a just recipe for running the command to fetch this code

Providence-o:
Closes #2761 

This PR fetches backend schema information from the EMIS staging environment into a python object which is then written into a csv file. 

It is the first step towards automatically generating EMIS schema tables.


pyproject.toml
```diff
@@ -60,6 +60,7 @@ exclude_also = [
 ]
 omit = [
     "tests/acceptance/external_studies/*",
+    "tests/backend_schemas/emisv2/update_schema.py",
```
rebkwok: This file has `# pragma: no cover file`, so we shouldn't need to omit it here

tests/backend_schemas/emisv2/update_schema.py
```diff
@@ -0,0 +1,112 @@
...
+import csv
+import os
+from pathlib import Path
+
+import urllib3
+from sqlalchemy import create_engine, inspect
+from sqlalchemy.orm import sessionmaker
+
+
+urllib3.disable_warnings()
```
rebkwok: We can avoid having to disable warnings by  removing the `"verify": False` connection arg on line 39, i.e.
```
args = {"http_scheme": "https", "request_timeout": 90}
```

tests/backend_schemas/emisv2/update_schema.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def get_table_columns(inspector):
+    tables = inspector.get_view_names(schema=SCHEMA)
+    for table in tables:
+        table_schema = inspector.get_columns(table_name=table, schema=SCHEMA)
+        yield table, table_schema
+
+
+def get_column_metadata(column):
```
rebkwok: nit: this isn't getting the metadata for a column, it's getting the metadata for a column type, so it confused me initially that it's called in `fetch_schema_rows` with `col["type"]`. Can we rename it and it's parameter `get_column_type_metadata(column_type)`

rebkwok (approved):

PR #2775, opened 2026-04-28
Return None for overflow errors in power() function

rebkwok:
Identified by [hypothesis](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1777339444442999)

In the power() function, we return None for power calculations that result in complex values, to align with SQL (which returns NULL without calculating the result). We also need to return None for OverflowErrors that result from a complex exponentiation.

### DIFF
--- a/ehrql/utils/math_utils.py
+++ b/ehrql/utils/math_utils.py
@@ -20,13 +20,22 @@ def floordiv(lhs, rhs):
 
 def power(lhs, rhs):
     """
-    Implement Python power behaviour but return None when either zero is raised to a negative
-    power or when a negative base is raised to a non-integer exponent (which would produce a complex number in Python).
+    Implement Python power behaviour but return None when:
+    - zero is raised to a negative power
+    - a negative base is raised to a non-integer exponent (which would produce a complex number in Python)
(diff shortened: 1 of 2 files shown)


ehrql/utils/math_utils.py
```diff
@@ -20,13 +20,17 @@ def floordiv(lhs, rhs):
...
+    - zero is raised to a negative power
+    - a negative base is raised to a non-integer exponent (which would produce a complex number in Python)
+    - we encounter an OverflowError
     """
     try:
         value = lhs**rhs
     except ZeroDivisionError:
         return None
+    except OverflowError:
+        return None
```
evansd: I'm probably being overly pedantic here, but I'm not sure this is exactly what we want. It's specifically `OverflowError: complex exponentiation` that we want to catch here; other kinds of overflow error I think we want to raise.

This mirrors the SQL behaviour because that tries to spot cases which might result in a complex value and returns NULL without ever computing them. The Python implementation deliberately takes a different approach (calculate the result and then check if it's complex) because that makes it a better test of the SQL behaviour. But it does mean that it's vulnerable to this specific kind of overflow.

evansd (approved):

Issue #2437, opened 2025-04-07
Non-deterministic in-memory vs SQL query engine mismatch


evansd:
This was identified, as you'd expect, by the [overnight generative tests](https://github.com/opensafely-core/ehrql/actions/runs/14287309151/job/40043963872).

The mismatch occurs with _all_ the SQL engines (i.e. they all agree on the same result but the in-memory engine returns something different) so for ease of debugging we can just concentrate on the SQLite engine. I haven't dug into this enough to have any idea what the correct result is here.

It is also non-deterministic in the sense that on some proportion on test runs with the below example (say 1 in 5) it passes.

This is all highly unsatisfactory.

Reproducing example (which Hypothesis failed to shrink):
```py
import datetime

from ehrql.query_model.nodes import (
    Case,
    Column,
    Dataset,
    Function,
    PickOneRowPerPatient,
    Position,
    SelectColumn,
    SelectTable,
    Sort,
    TableSchema,
    Value,
)
from tests.generative.test_query_model import data_setup


schema = TableSchema(
    i1=Column(int, constraints=()),
    i2=Column(int, constraints=()),
    b1=Column(bool, constraints=()),
    b2=Column(bool, constraints=()),
    d1=Column(datetime.date, constraints=()),
    d2=Column(datetime.date, constraints=()),
    f1=Column(float, constraints=()),
    f2=Column(float, constraints=()),
    s1=Column(str, constraints=()),
    s2=Column(str, constraints=()),
)

e0 = SelectTable(name="e0", schema=schema)

e0_d2 = SelectColumn(source=e0, name="d2")

dataset = Dataset(
    population=SelectColumn(
        source=PickOneRowPerPatient(
            source=Sort(
                source=e0,
                sort_by=Function.MaximumOf(
                    (Value(datetime.date(2018, 6, 27)), Function.ToFirstOfMonth(e0_d2))
                ),
            ),
            position=Position.FIRST,
        ),
        name="b1",
    ),
    variables={
        "v": Function.MinimumOf(
            (
                Function.DayFromDate(
                    Case(
                        cases={
                            Function.StringContains(
                                lhs=SelectColumn(
                                    source=PickOneRowPerPatient(
                                        source=Sort(
                                            source=e0,
                                            sort_by=Function.MaximumOf(
                                                (
                                                    Value(datetime.date(2018, 6, 27)),
                                                    Function.ToFirstOfMonth(e0_d2),
                                                )
                                            ),
                                        ),
[...]

PR #2781, opened 2026-04-30
Refactor sort tiebreaker implementation

evansd:
This takes a completely different, and overall simpler, approach to implementing our "sort tiebreaking" logic.

The previous code worked by rewriting the supplied query model to inject additional sort operations. It has been the source of [various][1] [bugs][2], attempted [bug fixes][3], and [new bugs][4] caused by those attempted fixes. It currently has a [bug][2] we don't know how to fix (although hopefully one that is very unlikely to directly impact users at the moment).

The root cause of the complexity here is that the kind of graph transform this involves is not "referentially transparent" (almost certain the wrong term here, but the one that came to mind). That is, it doesn't just involve replacing references to node X with node Y; it involves replacing X with Y1 when it occurs as a child of Z and Y2 when it doesn't.

This is much harder problem than the other kinds of transformation we do. Obviously it's not _intractable_, but neither is it straightforward; and if we can avoid it altogether than our code will be simpler and less buggy.

An alternative solution is, rather than trying to implement the behaviour we want by rewriting the query, to make it the responsibility of the query engines to do the right thing. This has some major advantages:
 * It is much simpler.
 * To the extent that it is not _entirely_ simple, the complexity is contained in the in-memory query engine.
 * Previously all the really nasty code was shared between the SQL and in-memory query engines meaning that the generative tests weren't really exercising it: if there was a bug then it would generally affect all the engines in the same way so it wouldn't be exposed. The SQL and in-memory engines now have quite divergent implementations of the same intended behaviour so it's much easier for the generative tests to expose any issues.

I have thrown Hypothesis at this locally and it hasn't found anything so far:
```
GENTEST_MAX_DEPTH=25 GENTEST_EXAMPLES=10000 \
  pytest tests/generative/test_query_model.py::test_query_model
```

I've also tried to make sure that this tiebreaking behaviour, and the reasoning behind it, is adequately documented in the codebase itself rather than just in issues/PRs.

[1]: https://github.com/opensafely-core/ehrql/commit/62d39361dc6f186da51f5e04a3997bf230bd669f
[2]: https://github.com/opensafely-core/ehrql/issues/2437
[3]: https://github.com/opensafely-core/ehrql/pull/2738
[4]: https://github.com/opensafely-core/ehrql/pull/2769


ehrql/query_model/transforms.py
```diff
@@ -72,37 +68,95 @@ class Coalesce(Series[T]):
...
+    This does exactly what we need, however it introduces an oddity into ehrQL's
+    semantics which is that you can no longer evaluate it in a bottom-up fashion. The
+    value of a pick-first-row operation depends on what columns are _going to be_
+    selected from that row.
+
+    This means we need to do some pre-processing of the query graph and annotate each
+    such operation with the set of columns that are selected from it elsewhere in the
+    query. It would be nicer not to have to do this, but given the above constraints I
+    think it's the best practical solution.
+    """
```
rebkwok, line 131: Thank you for this comment! I think I'd figured all of this out reviewing the previous PR, but it was still confusing and it was nice to read it explicitly explained.
It might be worth commenting that the "pre-processing of the query graph and annotate each such operation with the set of columns that are selected from it elsewhere in the query" is done in the query enqines' sort methods (those reference this function, but not vice versa). If I was just reading this comment in isolation, I think I'd expect for `rewrite_sorts` to be doing more than it is now.

rebkwok (approved):

PR #2787, opened 2026-05-08
fix: Add EmisV2 addresses table

### LINKED ISSUE #2785: Implement a `emisv2.addresses` table
We would like to run a study against the EMISv2 backend that queries for a patient (rounded) IMD via an `addresses` table.

We should:
- [x] Review the [tpp.addresses](https://github.com/opensafely-core/ehrql/blob/9695857fbcf73b751c604f34fbfb6e01d931e674/ehrql/tables/tpp.py#L57) table for columns that have an equivalent in the EMIS staging data (we think that these are `imd_rounded` and `msoa_code`)
- [x] Ensure the relevant EMIS columns are present in the [schema.py](https://github.com/opensafely-core/ehrql/blob/main/tests/backend_schemas/emisv2/schema.py) file and tested via integration tests
- [x] Implement a `emisv2.addresses` query table with the following columns / methods:
  - [x] `imd_rounded`
  - [ ] (Optional) `imd_quintile` and `imd_decile`
  - [x] `msoa_code`
  - [x]  `for_patient_on` (This requires adding a pair of start and end date columns. We can use `registration_start_date` and `registration_end_date` for example).
 
This work would prompt questions about the staging data that we would like to ask Optum (e.g. Is IMD the patient's IMD or the practice's?). We should ask the questions as we come up with them, but we don't need answers right away to complete this ticket.

rebkwok:
Closes #2785 

Note that the docstring for` imd_rounded` assumes that this field is based on LSOA, rounded to the nearest 100, as for TPP. This is still TBC.
    
`start_date` and `end_date` are practice registration start/end date, as this is the only information we have from the EMIS data.

The addresses table is an EventTable in order to match the syntax for TPP, so even though it can only have one row per patient (because it's derived from the patient table), `for_patient_on()` looks for rows that span the provided date, and sorts to pick the row.


ehrql/backends/emisv2.py
```diff
@@ -76,3 +76,15 @@ def modify_temp_table_schema(self, temp_table_schema, dsn, environ):
         FROM medication_issue_record
         """
     )
+
+    addresses = QueryTable(
+        """
+        SELECT
+            patient_id AS patient_id,
```
alarthast, line 83: I recall that in a previous version of this PR, there was a commit that removed `AS patient_id` from the patients table query.

Did you choose to include `AS patient_id` here to be consistent with the patients table query? (I'm not saying that this needs changing, was just curious!)

ehrql/tables/emisv2.py
```diff
@@ -1,11 +1,76 @@
...
+        Return each patient's address information as it was on the supplied registration
+        date.
+        """
+        # Note that the addresses table is an event-level table, but for EmisV2, it is
+        # derived from the patient table, so we know that there can only be at most one
+        # matching address per patient
+        spanning_addrs = self.where(self.start_date <= date).except_where(
+            self.end_date.is_not_null() & (self.end_date < date)
+        )
+        return spanning_addrs.sort_by(self.start_date).last_for_patient()
```
alarthast, line 76: Comparing this with the tpp version, I'm wondering if it's better to include the end date in the sort. 
I know that we don't need a sort at all since there is only one row, but given that we have written in a sort, we might as well be consistent with `tpp.addresses`?

tests/integration/tables/test_emisv2.py
```diff
@@ -0,0 +1,82 @@
...
+    dataset = Dataset()
+    dataset.define_population(emisv2.addresses.exists_for_patient())
+    dataset.imd_rounded = address.imd_rounded
+    results = in_memory_engine.extract(dataset)
+
+    assert results == [
+        {"patient_id": 1, "imd_rounded": 100},
+        {"patient_id": 2, "imd_rounded": 200},
+        {"patient_id": 3, "imd_rounded": None},
+    ]
```
alarthast, line 94: Is it worth covering the fourth case?
```suggestion
        },
        {
            # deregistered before 2010-01-01, not included
            emisv2.addresses: [
                dict(
                    patient_id=4,
                    start_date=date(2000, 1, 1),
                    end_date=date(2009, 1, 1),
                    imd_rounded=400,
                )
            ]
        },
    )

    address = emisv2.addresses.for_patient_on("2010-01-01")

    dataset = Dataset()
    dataset.define_population(emisv2.addresses.exists_for_patient())
    dataset.imd_rounded = address.imd_rounded
    results = in_memory_engine.extract(dataset)

    assert results == [
        {"patient_id": 1, "imd_rounded": 100},
        {"patient_id": 2, "imd_rounded": 200},
        {"patient_id": 3, "imd_rounded": None},
        {"patient_id": 4, "imd_rounded": None},
    ]
```

(I am aware that we only have the current registration in the patient table, but I think if we base the tests on that, we wouldn't even need to test the cases for non-null end dates.)

evansd (approved):
This looks good. My only question is whether we're doing the right thing by using the practice registration dates as the address dates here. I guess there's a fundamental tension between forcing users to acknowledge that the data available here is different and deciding what to do about it, and letting them write ehrQL which runs unmodified in both backends.

I guess one option is to not have `addresses` as part of `core` but instead have something like `most_recent_address` which works the same in TPP and EMIS.

Or alternatively, we could set the start and end dates to some fixed values (1900-01-01 to 3000-01-01 or something).

I'm really not sure what the best option is.

Side note: this is missing a pedantic commit prefix, but maybe that doesn't matter as we don't actually need a release of this at the moment.

alarthast (approved):
Some small questions / suggestions, but none are blocking.

ehrql/backends/emisv2.py
rebkwok: I think the previous PR was using the deduplicated_patients table that had already done the `AS patient_id` bit (but I'm not 100% sure)

ehrql/tables/emisv2.py
rebkwok: The sort is only there because it has to be, in order to use last_for_patient(). I don't think we need to include any other sorts to try and match tpp - tpp is also doing more sorts because it does have overlapping/duplicate rows

ehrql/backends/emisv2.py
alarthast: Ahh okay, that makes sense!

ehrql/tables/emisv2.py
alarthast: Understood, thanks for explaining!

Issue #2690, opened 2026-03-02
`practice_registrations.spanning()` should use inclusive bounds check


evansd:
Like the related issue #2540 this was also [pointed out](https://bennettoxford.slack.com/archives/C63UXGB8E/p1762184731489339) by @alschaffer:

> ... looks like there is a slight discrepancy between `practice_registrations.for_patient_on(date).exists_for_patient()` and `practice_registrations.exists_for_patient_on(date)`. The former will return TRUE, and the latter FALSE, if the date is equivalent to the practice registration end date, and the patient has no further registrations (e.g. they die), or if there is a gap in registrations. (FWIW I think it should be TRUE.)

The code which became the `spanning()` function was introduced in [this commit](https://github.com/opensafely-core/ehrql/pull/1729/changes/ea093a71d064ef25442cafa76a64326666d4085b) and from the comments it looks to have been copied from some earlier proof-of-concept code.

Almost certainly the right thing to do here is to update the definition of `spanning()` so that it includes the end date. Andrea's view is that:

> I find it unlikely that updating it would cause any major issues - I had a quick look on Github, and a lot of its use was to restrict to people registered for a given time period before their index date (study start date), so death isn't really relevant (this is how I've used it before).

PR #2791, opened 2026-05-12
Add `emisv2.practice_registrations`

### LINKED ISSUE #2786: Implement a `emisv2.practice_registrations` table
We would like to run a study against the EMISv2 backend that queries for patient's practice registrations.
This requires querying the `registration_start_datetime` and `registration_end_datetime` ~~and `organisation`~~ columns on the raw EMIS `patient` table. 

We should:
- [x] Ensure the relevant EMIS columns listed above are present in the [schema.py](https://github.com/opensafely-core/ehrql/blob/main/tests/backend_schemas/emisv2/schema.py) file and tested via integration tests
- [x] Implement a `emisv2.practice_registrations` query table with the following columns / methods:
  - [x] `start_date`
  - [x] `end_date`
  - [ ]  ~~`practice_pseudo_id`~~ See note below

We won't implement the `practice_pseudo_id` column for now as that work is potentially a bit involved and will be tracked in a separate issue. This is so that we can start trying to run the study, which does not use the column.

alarthast:
Closes #2786. This unblocks running a test study on the EMISv2 backend.

Similar to `emisv2.addresses`, we add `practice_registrations` as a event table even though it comes from the one-row-per-patient raw `patient` table in EMIS v2.

We do not implement the `practice_pseudo_id` column, and it is on the `core` table, so the EMIS `practice_registrations` has to be standalone.

Implementing `practice_pseudo_id` such that the table matches the core spec will be tackled separately in the future (see #2794).

### DIFF
--- a/ehrql/tables/emisv2.py
+++ b/ehrql/tables/emisv2.py
@@ -12,6 +12,7 @@ __all__ = [
     "clinical_events",
     "medications",
     "patients",
+    "practice_registrations",
 ]
 
 
@@ -74,3 +75,52 @@ class addresses(EventFrame):
             self.end_date.is_not_null() & (self.end_date < date)
         )
         return spanning_addrs.sort_by(self.start_date).last_for_patient()
+
+
+@table
+class practice_registrations(EventFrame):
+    """
+    Each record corresponds to a patient's registration with a practice.
+    """
+
+    start_date = Series(
+        datetime.date,
+        constraints=[Constraint.NotNull()],
+        description="Date patient joined practice.",
+    )
+    end_date = Series(
+        datetime.date,
+        description="Date patient left practice.",
+        dummy_data_constraints=[Constraint.Categorical([None])],
+    )
+
+    def for_patient_on(self, date):
+        """
+        Return each patient's practice registration as it was on the supplied date.
+        """
+        # Note that practice_registrations is an event-level table, but for EMISv2, it is
+        # derived from the patient table, so we know that there can only be at most one
+        # matching registration per patient
+        return self.spanning(date, date).sort_by(self.start_date).last_for_patient()
+
+    def exists_for_patient_on(self, date):
+        """
+        Returns whether a person was registered with a practice on the supplied date.
+
+        NB. The implementation currently uses `spanning()`. It would also have been
+        valid to implement as
+        `practice_registrations.for_patient_on(date).exists_for_patient()`, but for
+        internal reasons that is less efficient.
+
+        """
+        return self.spanning(date, date).exists_for_patient()
+
+    def spanning(self, start_date, end_date):
+        """
+        Filter registrations to just those spanning the entire period between
+        `start_date` and `end_date`.
+        """
+        return self.where(
+            self.start_date.is_on_or_before(start_date)
+            & (self.end_date.is_on_or_after(end_date) | self.end_date.is_null())
+        )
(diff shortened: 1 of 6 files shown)


ehrql/tables/emisv2.py
```diff
@@ -74,3 +75,55 @@ def for_patient_on(self, date):
...
+    )
+
+    def for_patient_on(self, date):
+        """
+        Return each patient's practice registration as it was on the supplied date.
+        """
+        # Note that practice_registrations is an event-level table, but for EMISv2, it is
+        # derived from the patient table, so we know that there can only be at most one
+        # matching registration per patient
+        spanning_regs = self.where(self.start_date <= date).except_where(
```
rebkwok: Why don't we just use `self.spanning(date, date)` here? I _think_ they do the same thing. (I know this is what the equivalent in the core tables does, but I'm not sure why we don't use it there either.)

tests/integration/tables/test_emisv2.py
```diff
@@ -29,6 +29,204 @@ def test_patients_age_on(in_memory_engine):
...
+            emisv2.practice_registrations: [
+                dict(
+                    patient_id=4, start_date=date(2000, 1, 1), end_date=date(2012, 1, 1)
+                ),
+                dict(
+                    patient_id=4, start_date=date(2000, 1, 1), end_date=date(2015, 1, 1)
+                ),
+            ]
+        },
+    )
```
rebkwok, line 58: These tests aren't really representative of the real data, since we assume we only have one practice reg per patient. However, I think they're all logically OK apart from the last one; since we only sort the "spanning" registrations by start date, the "choose longest" one is only coming up with the right result accidentally, I think because end date is being used as a hidden tie-break for the sort. It's probably best to just remove that one, because it's misleading

tests/integration/tables/test_emisv2.py
```diff
@@ -29,6 +29,204 @@ def test_patients_age_on(in_memory_engine):
     ]
 
 
+def test_practice_registrations_for_patient_on(in_memory_engine):
+    in_memory_engine.populate(
+        # Simple case: successive registrations
```
evansd: I'm not sure it makes sense to have these test cases given that they're not possible in the EMIS data. I'm worried they're a bit misleading and someone will end up thinking they must be there for a reason.

evansd (approved):
Looks good, just not sure about all the tests.

tests/integration/tables/test_emisv2.py
alarthast: Thank you, I've hopefully now removed the misleading tests and added the comment about only having one registration, similar to how we do for addresses :+1:

tests/integration/tables/test_emisv2.py
alarthast: Many thanks for catching that! This test has now been removed along with the other test cases that use more than one registration entry (see comment below as well) :+1:

ehrql/tables/emisv2.py
alarthast: [Discussed on Slack](https://bennettoxford.slack.com/archives/C080S7W2ZPX/p1778601647248559?thread_ts=1778591413.300969&cid=C080S7W2ZPX).

I've changed the definition of `spanning`:
https://github.com/opensafely-core/ehrql/blob/bf548b784d98e0c7f6e36c661c3ae340afe00bec/ehrql/tables/emisv2.py#L118-L126

and used it within `for_patient_on`. I've also tweaked the tests so that the edge cases are covered.

This means that the emisv2 implementation of the `spanning` helper method differs from the core one, but the amendments to the core method will be dealt with separately in #2690 .

rebkwok (approved):

Issue #2540, opened 2025-09-15
Logic in `for_patient_on()` method on `practice_registrations` /`addresses` does not match documentation


evansd:
The issue with `practice_registrations` was [pointed out](https://bennettoxford.slack.com/archives/C01D7H9LYKB/p1757947805009959) by @alschaffer:

> The underlying code orders by start date then end date - but elsewhere it says:
> 
> > Note that NULL is considered smaller than any other value, so you may wish to filter out NULL values before sorting.
> 
> So in the very rare case that someone has multiple registrations with the same start date, `for_patient_on()` will return the longest, _completed_ registration, as any active registrations (which will have a NULL end date) will appear first even if it is longest

And the parallel issue with `addresses` was [independently discovered](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1778250766095129) by @rebkwok.

The code in question is here:
https://github.com/opensafely-core/ehrql/blob/61a8241cecc5801b74854dbb7fa1b1c4d310d63f/ehrql/tables/core.py#L156-L173

And here:
https://github.com/opensafely-core/ehrql/blob/61a8241cecc5801b74854dbb7fa1b1c4d310d63f/ehrql/tables/tpp.py#L224-L246

evansd:
The original cause of this is this discrepancy is that the `for_patient_on()` helper method was added in February 2023 when we still directly exposed the `9999-12-31` "ceiling" date:
 * https://github.com/opensafely-core/ehrql/pull/1021

And then in November of 2023 we replaced ceiling dates with NULL:
 * https://github.com/opensafely-core/ehrql/pull/1706

And we obviously didn't have a test which covered this specific case.

evansd:
## Investigating affected `addresses`

We can find cases where the address selection logic would return a different address depending on whether we prefer addresses with or without an end date using the below query:
```sql
-- Find groups of addresses for a patient which start on the same day and where one has
-- a specified end date and another is unterminated
SELECT * INTO #ambiguous_addresses FROM (
    SELECT
        Patient_ID, StartDate
    FROM
        PatientAddress
    WHERE
        -- Ignore address records of zero (or negative) duration because there is no
        -- time at which we consider them current
        EndDate > StartDate
    GROUP BY
        Patient_ID,
        StartDate
    HAVING
        MAX(EndDate) = '9999-12-31T00:00:00'
        AND MIN(EndDate) != '9999-12-31T00:00:00'
[...]

PR #2800, opened 2026-05-14
Make practice_registrations.spanning use inclusive bounds

### LINKED ISSUE #2690: `practice_registrations.spanning()` should use inclusive bounds check

rebkwok:
Closes #2690 

i.e. `practice_registrations.spanning("2001-01-01", "2010-12-31")` now requires a patient to have a registration that continuously spans 2010-01-01 to 2010-12-31 _inclusive_.  This also means that `practice_registrations.exists_for_patient_on("2010-12-31")` and `practice_registrations.for_patient_on("2010-12-31").exists_for_patient()` return consistent results

### DIFF
--- a/tests/integration/tables/test_core.py
+++ b/tests/integration/tables/test_core.py
@@ -281,7 +281,7 @@ def test_practice_registrations_spanning(
                     patient_id=5,
                     practice_pseudo_id=123,
                     start_date=date(2010, 2, 1),
-                    end_date=date(2010, 12, 1),
+                    end_date=date(2010, 12, 31),
                 ),
             ]
         },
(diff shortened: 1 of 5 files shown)


tests/integration/tables/test_core.py
```diff
@@ -281,7 +281,7 @@ def test_practice_registrations_spanning(
                     patient_id=5,
                     practice_pseudo_id=123,
                     start_date=date(2010, 2, 1),
```
alarthast: Tiny suggestion, since this isn't a date of birth, it might as well be
```suggestion
                    start_date=date(2010, 1, 2),
```
to be symmetric?

alarthast (approved):
LGTM :)

Tagging #2540 since I think the docstring change for `for_patient_on` covers part of that issue, but might not fully close it.

PR #2808, opened 2026-05-19
Update docs and tests for for_patient_on() methods

### LINKED ISSUE #2540: Logic in `for_patient_on()` method on `practice_registrations` /`addresses` does not match documentation

rebkwok:
Update the docs for `tpp.addresses.for_patient_on` to reflect the actual selection behaviour in the case of multiple addresses (i.e. that in the case of multiple identical start dates, we prefer the one with the longest duration with an explicit end date).
    
`core.practice_registrations.for_patient_on` had already been updated, but we add similar warnings to the tpp table about the complexity of identifying the registration on a specific date, and a warning to consider applying custom disambiguation logic. I added this to the table doc string with a link to the `for_patient_on` methods, because it seemed to fit best there in the displayed docs.
    
Also adds explicit test cases for null end dates in overlapping addresses/registrations with the same start date.

Closes #2540

### DIFF
--- a/docs/includes/generated_docs/schemas/tpp.md
+++ b/docs/includes/generated_docs/schemas/tpp.md
@@ -53,6 +53,19 @@ from which other larger geographic representations can be derived
 
 [addresses_ukgeographies]: https://www.ons.gov.uk/methodology/geography/ukgeographies
 
+!!! warning
+
+    While questions such as "Where did a patient live at X date?" are straightforward,
+    obtaining the answer can be quite complex, as there can be multiple registered
+    addresses, which may not have clean start and end dates. E.g. Addresses may be overlapping,
+    or may have "floor" start dates (typically 1900-01-01), indicating that the start
+    date is unknown.
+
+    The [`for_patient_on` method](#addresses.for_patient_on) applies logic for determining
+    the most likely correct address for a patient on a specific date. However, studies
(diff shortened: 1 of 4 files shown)


evansd (approved):
Great! Thanks for doing this.

Issue #2801, opened 2026-05-15
Move T1OO filter to new permissions system


evansd:
The current T1OO filtering logic predates the more generalised permissions system and instead uses an extra query parameter on the database URL:
https://github.com/opensafely-core/ehrql/blob/cbb76280b88855ac2a3afb98c4e1d73a7d3fe1b5/ehrql/backends/tpp.py#L114-L118

Now that we need to implement parallel logic for EMIS this is a good point at which to move the TPP logic to use the permissions system. We certainly don't want to be trying to implement the query parameter version in EMIS, and having divergent implementations isn't ideal.

Here's an outline of a plan for doing so. I've only mention the production code changes, but each of these will require test changes as well. Where the code comments mention specific bits of documentation that need updating then that should be done as well. Each step presumes that the previous step has been merged and deployed.

## 1. Update Job Server to send the right permissions

The first step is to update Job Server so that it sends the right permissions (while also leaving the old code in place). That involves adding an `analysis_scope_for_project` function (parallel to the one for [`ndoo`](https://github.com/opensafely-core/job-server/blob/4f9cbb9e659953e9d409ab199e7b4836f912b419/jobserver/permissions/population_permissions/ndoo.py#L95-L97)) to the [`t100`](https://github.com/opensafely-core/job-server/blob/4f9cbb9e659953e9d409ab199e7b4836f912b419/jobserver/permissions/t1oo.py#L116) module. And then including the `t1oo` module in the list of "population permission" modules [here](https://github.com/opensafely-core/job-server/blob/4f9cbb9e659953e9d409ab199e7b4836f912b419/jobserver/rap_api.py#L212-L215). The permission name should be `include_t1oo`.

## 2. Update ehrQL to respect both old and new permissions

That means adding a line here:
https://github.com/opensafely-core/ehrql/blob/cbb76280b88855ac2a3afb98c4e1d73a7d3fe1b5/ehrql/backends/tpp.py#L136
```py
include_t1oo = self.include_t1oo or "include_t1oo" in self.permissions
```
And changing `self.include_t1oo` to `include_t1oo` here:
https://github.com/opensafely-core/ehrql/blob/cbb76280b88855ac2a3afb98c4e1d73a7d3fe1b5/ehrql/backends/tpp.py#L152

## 3. Remove old code from Job Server

[...]

PR #2810, opened 2026-05-20
fix: Set include_t1oo permission from environment

rebkwok:
This will be passed to ehrql from job-server (via the controller) soon, along with other permission strings.

For now, we continue to set the permission via the dsn suffix as well. This can be removed once job-server has been updated, and the t1oo permission is being passed through from the controller.

Note: in practice, we could just switch over now, as only completed Covid projects approved under the initial directive have permission, so currently running jobs should not be affected at all - include_t1oo will always be False.

Part of #2801

### DIFF
--- a/ehrql/backends/tpp.py
+++ b/ehrql/backends/tpp.py
@@ -134,6 +134,7 @@ class TPPBackend(SQLBackend):
     def modify_dataset(self, dataset):
         # Check the explicitly set permissions to determine if we can include NDOOs
         include_ndoo = "include_ndoo" in self.permissions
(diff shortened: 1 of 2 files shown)


evansd (approved):
Yeah, I wasn't sure if we actually needed to do this either. My only thought was about the opensafely-internal project. But it felt easier to do it than to work out whether we actually needed to do it.

rebkwok:
The opensafely-internal project isn't in the t1oo approved list, so we should be good there too. Definitely doesn't hurt to do it in stages though

evansd:
Oh, sorry, I think I was mixing it up with NDOO.

Issue #2802, opened 2026-05-15
Implement T1OO filtering in the EMIS backend


evansd:
The end goal here is the same behaviour as with the TPP backend, but the implementation will look different.

Note that in order for us to actually disable the filtering for real jobs then at least step 1 of the [linked issue](https://github.com/opensafely-core/ehrql/issues/2801) must be complete. However this doesn't need to block progress on this ticket. We need to retain the ability to disable the filtering when legally instructed to in case of a future public health emergency; but we have no need to disable it in the initial phases of the EMIS work.

The plan is roughly as follows:

 * We'll need to add a [`modify_dataset()`](https://github.com/opensafely-core/ehrql/blob/cbb76280b88855ac2a3afb98c4e1d73a7d3fe1b5/ehrql/backends/tpp.py#L134) method to the EMIS backend.
 * This will need to check for the presence of `"include_t1oo"` in `self.permissions` and if it is *not* present then it will need to add a condition to the population definition for the query.
 * This condition will need to include only those patients where `patient.is_consent_93c1`[/-]` is **True** or **NULL**. Note the slightly unusual semantics here which stem from the way this flag is derived. See the [discussion](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1778848021596169) on Slack and the [EMIS docs](https://docs.partner.emis-x.uk/explorer/opensafely/patient/definition/#is_consent_93c1--type-1-opt-out) for more detail.
 * We'll need to add some tests parallel to the [NDOO tests](https://github.com/opensafely-core/ehrql/blob/cbb76280b88855ac2a3afb98c4e1d73a7d3fe1b5/tests/integration/backends/test_tpp.py#L3239) for the TPP backend. (Coverage should enforce that you do this in any case.)

PR #2816, opened 2026-05-29
Filter T1OO with `WHERE` clause on `patients`

alarthast:
Draft implementation of #2802  to collect feedback.

This approach conditionally applies a `WHERE` clause to the `patients` table.
It then uses `modify_dataset` to ensure that only patients in the filtered `patients` table are included in the dataset population.

The ticket seems to expect filtering to occur all within the `modify_dataset` method, so this approach might not be what the ticket asked for.
An alternative approach using an internal list of patient IDs is explored in #2817.
https://github.com/opensafely-core/ehrql/compare/1f32b25..d76c242

Note that I have not taken extensibility for the NDOO stuff into account on purpose to keep the code simple.

### DIFF
--- a/ehrql/backends/emisv2.py
+++ b/ehrql/backends/emisv2.py
@@ -1,10 +1,16 @@
+import logging
 import urllib.parse
 
 import ehrql.tables.emisv2
 import ehrql.tables.smoketest
 from ehrql.backends.base import QueryTable, SQLBackend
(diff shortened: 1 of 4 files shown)


alarthast:
We'll go with #2817 - see [this Slack thread](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1780074913545519).

PR #2817, opened 2026-05-29
Implement T1OO filtering on EMISv2

alarthast:
Addresses #2802.

Implement T1OO filtering using an internal list of patient IDs of patients who have a type 1 opt-out, similar to the way it's done in TPP. The difference is that TPP provides the list of T1OO patients as a separate table to us, whereas for EMIS we query the EMIS `patient` table for the list.

The above implies that the EMIS implementation proposed here requires an extra scan of the `patient` table in order to construct the T1OO list - this has been [discussed on Slack](https://bennettoxford.slack.com/archives/C069YDR4NCA/p1780074913545519) to be acceptable.

I have put in a `TODO` to remind us to link to `docs.opensafely.org` when we eventually write some docs for the T1OO filtering on EMISv2. I don't think it is necessary to also write these docs now to close the ticket, but am happy to be corrected if we do need them!


tests/integration/backends/test_emisv2.py
```diff
@@ -24,19 +26,18 @@ def test_extract_smoketest_dataset_definition(trino_engine):
         # Trino DBAPI's Binary() implementation takes a string and encodes it as UTF-8
         Patient(
             _pk=1,
-            patient_id=bytes(range(16)).decode("utf-8"),
+            patient_id=bytes([1] * 16).decode("utf-8"),
```
rebkwok: This could just be `"1" * 16`

ehrql/backends/emisv2.py
```diff
@@ -98,3 +133,16 @@ def modify_temp_table_schema(self, temp_table_schema, dsn, environ):
         FROM patient
         """
     )
+
+    t1oo = QueryTable(
+        # is_consent_93c1 is based on a clinical code on the patient's record.
+        # The field is false if the patient has a type 1 opt-out.
+        # See https://docs.partner.emis-x.uk/explorer/opensafely/patient/schema/
```
rebkwok, line 141: It might be worth explicitly commenting here that both true and null = no T1OO

rebkwok (approved):
Couple of v minor comments only

tests/integration/backends/test_emisv2.py
alarthast: Many thanks! It looks much better now:

https://github.com/opensafely-core/ehrql/blob/3775a975c757cb6128b52ddada82b20e9f8caf94/tests/integration/backends/test_emisv2.py#L224-L228

https://github.com/opensafely-core/ehrql/blob/3775a975c757cb6128b52ddada82b20e9f8caf94/tests/integration/backends/test_emisv2.py#L245-L253

ehrql/backends/emisv2.py
alarthast: Added :smile: 

https://github.com/opensafely-core/ehrql/blob/3775a975c757cb6128b52ddada82b20e9f8caf94/ehrql/backends/emisv2.py#L137-L141

PR #2835, opened 2026-06-29
Update TPP schema

evansd:
The published schema now includes the `PatientsWithoutNDOO` table so we remove that from the manual additions.

Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1782315000495159?thread_ts=1781077932.079749&cid=C069YDR4NCA

### DIFF
--- a/tests/backend_schemas/tpp/categorical_columns.csv
+++ b/tests/backend_schemas/tpp/categorical_columns.csv
@@ -551,6 +551,7 @@ VaccinationReference,VaccinationName,Respiratory Syncytial Virus (RSV) vaccine
 VaccinationReference,VaccinationName,Revaxis
 VaccinationReference,VaccinationName,Rotarix
 VaccinationReference,VaccinationName,Rotavirus - Oral
+VaccinationReference,VaccinationName,Rotavirus vaccine live oral suspension 1.5ml pre-filled syringes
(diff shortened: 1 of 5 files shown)


suzannehamilton (approved):

tests/backend_schemas/tpp/categorical_columns.csv
```diff
@@ -646,7 +648,7 @@ WL_ClockStops,Waiting_List_Type,PTO
 WL_ClockStops,Waiting_List_Type,RTTI
 WL_ClockStops,Waiting_List_Type,RTTO
 WL_OpenPathways,PRIORITY_TYPE_CODE,
-WL_OpenPathways,PRIORITY_TYPE_CODE,
+WL_OpenPathways,PRIORITY_TYPE_CODE,
```
suzannehamilton, line 651: Is this new whitespace value something ehrQL needs to handle? (Nothing to do with NDOO, of course.)

evansd: Good question, but in this particular case it doesn't make any difference as all whitespace-only values will be [turned into NULLs](https://docs.opensafely.org/ehrql/reference/schemas/tpp/#wl_openpathways.priority_type_code) in any case.

PR #2836, opened 2026-07-01
Retry trino queries on TABLE_NOT_FOUND errors

### LINKED ISSUE #2830: Add retry logic for missing temporary tables in Trino engine
Trino is a multi-node, eventually consistent system in which it's possible to create a table in one query and then have a subsequent query on the same connection unable to see that table (at least, until a certain amount of time has elapsed).

The [initially proposed](https://github.com/opensafely-core/ehrql/issues/2827) solution to this isn't available to us due to lack of the relevant permissions, so we need instead to add a retry mechanism.

The basic logic we want is that if a SQL query fails with an error relating to the non-existence of one of our temporary tables then we sleep for some small amount of time and then retry. We continue to retry a small, fixed number of times, sleeping for longer each time. Finally we raise the original error.

We've already implemented very similar logic in the [`execute_with_retry_factory`](https://github.com/opensafely-core/ehrql/blob/30676327f51b9179dd8179c15953e18d5ca13bb2/ehrql/utils/sqlalchemy_exec_utils.py#L203C5-L203C31) function, however the current behaviour is quite specific to the needs of the MSSQL engine:
 * It assumes that the query returns results and calls `list()` on it.
 * It retries after any kind of `DBAPIError`.
 * It calls `connection.rollback()` before retrying.

It might be possible to abstract these things out. But the result may be too abstract and difficult to understand. Given that the core logic here is not very complicated it might be better to just write something new for the Trino engine to use.

Additionally, the approach of having a factory function which returns a callable that automatically does the retries is needed in the MSSQL case because this callable itself gets passed in to the "batch fetching" functions and allows us to separate the logic out. But for Trino we don't need this level of abstraction so that's another argument for having a separate, simpler implementation.

Related Slack thread:
https://bennettoxford.slack.com/archives/C069YDR4NCA/p1781001266292789

alarthast:
Closes #2830

### DIFF
--- a/ehrql/query_engines/trino.py
+++ b/ehrql/query_engines/trino.py
@@ -1,7 +1,10 @@
 import logging
+import time
 
 import sqlalchemy
+from sqlalchemy.exc import DBAPIError
 from sqlalchemy.sql.functions import Function as SQLFunction
+from trino.exceptions import TrinoUserError
 
 from ehrql.query_engines.base_sql import BaseSQLQueryEngine, get_cyclic_coalescence
 from ehrql.query_engines.trino_dialect import TrinoDialect
@@ -18,6 +21,12 @@ log = logging.getLogger()
 
 class TrinoQueryEngine(BaseSQLQueryEngine):
     sqlalchemy_dialect = TrinoDialect
+    # Parameters for retrying queries upon TABLE_NOT_FOUND errors
+    # 0.5s seemed to be sufficient for a minimal temporary table;
+    # we may need to adjust these params if they turn out to be insufficient
+    max_retries = 4
+    retry_sleep = 0.5
+    backoff_factor = 2
 
     def get_order_clauses(self, sort_conditions, position):
         order_clauses = super().get_order_clauses(sort_conditions, position)
@@ -155,3 +164,36 @@ class TrinoQueryEngine(BaseSQLQueryEngine):
 
     def grouping_id(self, *columns):
         return sqlalchemy.func.grouping(*columns).label("grp_id")
+
+    @staticmethod
+    def _is_table_not_found_error(exception):
+        # As Trino is a eventually consistent system, sometimes a freshly created
+        # temporary table is not yet visible when a query tries to use it.
+        # We therefore don't raise these TABLE_NOT_FOUND errors until after retrying
+        # the query, as it's likely that the table would be found after a brief wait.
+
+        # We observed that these errors originate as TrinoUserErrors which then get
+        # wrapped into SQLAlchemy ProgrammingErrors - but we check the underlying error
+        # for all DBAPIErrors to avoid missing any of them
+        return (
+            isinstance(exception, DBAPIError)
+            and isinstance(exception.orig, TrinoUserError)
+            and exception.orig.error_name == "TABLE_NOT_FOUND"
+        )
+
+    def execute_query(self, connection, query, query_id=None):
+        retries = 0
+        next_sleep = self.retry_sleep
+
+        while True:
+            if retries > 0:
+                log.info(f"Retrying query (attempt {retries} / {self.max_retries})")
+            try:
+                return connection.execute(query)
+            except Exception as e:
+                if self._is_table_not_found_error(e) and retries < self.max_retries:
+                    retries += 1
+                    time.sleep(next_sleep)
+                    next_sleep *= self.backoff_factor
+                else:
+                    raise
--- a/ehrql/query_engines/base_sql.py
+++ b/ehrql/query_engines/base_sql.py
@@ -1130,11 +1130,14 @@ class BaseSQLQueryEngine(BaseQueryEngine):
                         f"Finished running {query_id} (duration={duration:.2f})\n\n"
                     )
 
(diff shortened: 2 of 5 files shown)


ehrql/query_engines/trino.py
```diff
@@ -155,3 +164,66 @@ def reify_query(self, query):
...
+        # We observed that these errors originate as TrinoUserErrors which then get
+        # wrapped into SQLAlchemy ProgrammingErrors - but we check the underlying error
+        # for all DBAPIErrors to avoid missing any of them
+        return (
+            isinstance(exception, DBAPIError)
+            and isinstance(exception.orig, TrinoUserError)
+            and exception.orig.error_name == "TABLE_NOT_FOUND"
+        )
+
+    def execute_query_no_results(self, connection, query, query_id=None):
```
alarthast: The way that I have done this means that the base methods will only be used by the `SQLiteQueryEngine`.

https://github.com/opensafely-core/ehrql/blob/350a1ccd857960d6474f9b6d02cf3a8a4b5c89d3/ehrql/query_engines/base_sql.py#L1133-L1147

My hunch is that the SQLite engine happening to be the only caller isn't enough to justify moving these methods off the `BaseSQLQueryEngine` and into there, but I'd like to confirm this hunch with the reviewer.

ehrql/query_engines/trino.py
```diff
@@ -155,3 +164,66 @@ def reify_query(self, query):
...
+                    assert not stream_started, (
+                        "Some results were already streamed - cannot retry query as there would be duplicate rows"
+                    )
+                    retries += 1
+                    time.sleep(next_sleep)
+                    next_sleep *= self.backoff_factor
+                else:
+                    if stream_started:
+                        # Close the cursor to make it clear we're not going to be fetching any more
+                        cursor_result.close()
```
alarthast: The original method had a comment that mentions that the cursor closing is mainly relevant for the SQLite engine, but I thought that we might as well retain it for Trino still.

Happy to remove if it's really not needed!

https://github.com/opensafely-core/ehrql/blob/350a1ccd857960d6474f9b6d02cf3a8a4b5c89d3/ehrql/query_engines/base_sql.py#L1140-L1147

ehrql/query_engines/trino.py
```diff
@@ -155,3 +164,66 @@ def reify_query(self, query):
...
+                cursor_result = connection.execute(query)
+                for row in cursor_result:
+                    yield row
+                    stream_started = True
+                return
+            except Exception as e:
+                if self._is_table_not_found_error(e) and retries < self.max_retries:
+                    assert not stream_started, (
+                        "Some results were already streamed - cannot retry query as there would be duplicate rows"
+                    )
```
alarthast: I explain in the commit message that I think a scenario where a table is suddenly not found halfway through outputting results would be "peculiar" - I actually can't really imagine it happening at all, but I thought to guard against it anyway because the alternative is for us to suddenly get a bunch of duplicate rows which can only be diagnosed by going to look at the query logs - having an assertion here seems to give better peace of mind. Am happy to be told that this isn't needed though.

evansd (commented):
Now we've confirmed that the exception is raised when the query is executed (rather than when the results are iterated) I think we can simplify things a bit here.

We could start by slightly refactoring `BaseSQLQueryEngine` to pul out an `execute_query` method which both `execute_query_no_results` and `execute_query_with_results` call out to:
```py
class BaseSQLQueryEngine(BaseQueryEngine):
    ...

    def execute_query_no_results(self, connection, query, query_id=None):
        self.execute_query(connection, query, query_id=query_id)

    def execute_query_with_results(self, connection, query, query_id=None):
        cursor_result = self.execute_query(connection, query, query_id=query_id)
        ...

    def execute_query(self, connection, query, query_id=None):
        return connection.execute(query)
```

Then in the `TrinoQueryEngine` we can redefine `execute_query` to have the retry logic exactly as you've implemented it. And that way both of the other `execute_*` methods get the retry behaviour for free.

That also reduces the amount of testing you need as you don't need to unit test each method separately.

ehrql/query_engines/trino.py
```diff
@@ -18,6 +21,12 @@
 
 class TrinoQueryEngine(BaseSQLQueryEngine):
     sqlalchemy_dialect = TrinoDialect
+    # Parameters for retrying queries upon TABLE_NOT_FOUND errors
+    # 0.5s seemed to be sufficient for a minimal temporary table;
+    # we may need to adjust these params if they turn out to be insufficient
+    max_retries = 4
+    retry_sleep = 0.5
+    backoff_factor = 2
```
alarthast, line 29: Originally I had these on the class because they needed to be shared between the `execute_query_no_results` and `execute_query_with_results` methods. 
Now that there is only one method on this class that uses it, I considered moving these to be in-line variables within the function, but ultimately decided to keep them here to allow the unit tests to continue to easily set up their own params.

evansd: Yeah, that seems like the right call to me.

evansd (approved):
Great, looks good. Thanks Alice!

PR #2839, opened 2026-07-16
Add minimal Event Level Data docs

### LINKED ISSUE #2833: Include Event Level Data features in the public docs
At present the `Dataset.add_event_table()` method is excluded from the docs and autocomplete:
https://github.com/opensafely-core/ehrql/blob/4cf530b69bf0cb3c2919379c99ff27ac9874c8da/ehrql/query_language.py#L202-L205
https://github.com/opensafely-core/ehrql/blob/4cf530b69bf0cb3c2919379c99ff27ac9874c8da/ehrql/query_language.py#L217-L221

As we expand the process of trialling this feature we want to start including it in the docs.

This will involve:
 * using a normal, non-obfuscatory method definition for `add_event_table()`;
 * writing a docstring for this method;
 * writing a dedicated page on event-level data.

Both the docstring and the page will need a clear callout box which explains that this feature is being trialled for data development work and is not yet generally available for research use. The docstring should also link to the explanatory page.

The only existing documentation comes from this [Slack thread](https://bennettoxford.slack.com/archives/C07UG9PCR6E/p1738834119652829):

> The API itself just consists of a new `add_event_table()` method on datasets. As an example, suppose you wanted to retrieve all relevant vaccination events for each patient:
> ```
> dataset = create_dataset()
> 
> # ... The usual stuff here ...
> 
> # Get some vaccination events 
> vaxx = vaccinations.where(
>     vaccinations.date.is_on_or_between(study_start, study_end)
>     & (vaccinations.target_disease == "SARS-2 CORONAVIRUS")
> )
> 
> dataset.add_event_table(
>     "vaccinations",
>     date=vaxx.date,
>     product=vaxx.product_name,
> )
> ```
> 
> As you can see, the `add_event_table()` method takes a name string and then a bunch of column keywords. The table name can be whatever you want: it just affects what the output file is called. Likewise with the column names.
> 
> You can add as many event tables as you need.
>
> Because the output now consists of multiple files rather than a single file, the way you specify where the output goes is slightly different. Instead of writing:
> ```--output outputs/results.arrow```
> You write:
> ```--output outputs/results:arrow```
> That is, a colon instead of dot between the file name and the extension.
> 
> Instead of a single file called `results` this will create a _directory_ called `results` which will contain multiple files: a `dataset.arrow` file with the normal one-row-per-patient output, and then one file for each of the event tables you've defined.
> 
> (This works with any supported file format, not just Arrow.)

Slack discussion:
https://bennettoxford.slack.com/archives/C07UG9PCR6E/p1782297865242369

evansd:
As requested [on Slack][1] this adds some basic documentation on how to use Event Level Data, together with some clear warnings that this is not yet generally available.

Previews links:
 * https://evansd-eld-docs.databuilder.pages.dev/explanation/event-level-data/
 * https://evansd-eld-docs.databuilder.pages.dev/reference/language/#Dataset.add_event_table

Closes #2833

[1]: https://bennettoxford.slack.com/archives/C07UG9PCR6E/p1782297865242369

### DIFF
--- a/docs/explanation/event-level-data.md
+++ b/docs/explanation/event-level-data.md
@@ -1,0 +1,93 @@
+# Event Level Data
+
+Datasets in ehrQL are restricted to one row per patient (or "patient-level") outputs.
+The Event Level Data system allows you to add supplementary output files alongside the
+main dataset which can contain more than one row per patient.
+
+
+## Adding events to a dataset
+
+Events are added using the
+[`add_event_table()`](../reference/language.md#Dataset.add_event_table) method on
+datasets. As an example, suppose you wanted to retrieve all relevant vaccination and
+hospitalisation events for a set of patients:
+
+```python
+# ... imports etc ...
+
+dataset = create_dataset()
+
+dataset.define_population(
+  (patients.age_on(study_start) >= 18)
+  & (patients.age_on(study_start) <= 65)
+)
+
+# Get some vaccination events 
+vaxx = vaccinations.where(
+    vaccinations.date.is_on_or_between(study_start, study_end)
+    & (vaccinations.target_disease == "SARS-2 CORONAVIRUS")
+)
+
+# Get some hospital admissions with a certain diagnosis
+admissions = apcs.where(
+    apcs.primary_diagnosis.is_in(respiratory_codes)
+)
+
+# Add these events to the dataset as ancillary files
+
+dataset.add_event_table(
+    "covid_vaccinations",
+    date=vaxx.date,
+    product=vaxx.product_name,
+)
+
+dataset.add_event_table(
+    "respiratory_admissions",
+    admitted=admissions.admission_date,
+    discharged=admissions.discharge_date,
+)
+```
+
+As the example shows, the `add_event_table()` method takes a name string and then some
+number of column keywords. The table name can be whatever you want: it just affects what
+the output file is called. Likewise with the column names.
+
+You can add as many event tables to a dataset as you need.
+
+Note that events are only returned for patients who **match the population definition**
+of the dataset.
+
+
+## Configuring outputs
+
+Because event level output consists of multiple files rather than a single dataset file,
+the way you specify the `--output` argument to `generate-dataset` is slightly different.
+Instead of writing:
+```
+--output outputs/cohort.arrow
+```
+
+You write:
+```
+--output outputs/cohort/:arrow
+```
+
+Instead of a single file called `cohort` this will create a _directory_ called
+`cohort` which will contain multiple files: a `dataset.arrow` file with the normal
+one-row-per-patient output, and then one file for each of the event tables you've
+defined.
+
+Note that this works with any supported file format, not just Arrow.
+
+You will also need to change the `outputs:` specification in your `project.yaml` file to
+match, changing `output/cohort.arrow` to `output/cohort/*.arrow`.
+
+A complete example might look like:
+```yaml
+generate_dataset:
+  run: ehrql:v1 generate-dataset analysis/dataset_definition.py
+    --output output/cohort/:arrow
+  outputs:
+    highly_sensitive:
+      cohort: output/cohort/*.arrow
+```
--- a/docs/explanation/event-level-data.md
+++ b/docs/explanation/event-level-data.md
@@ -1,5 +1,7 @@
 # Event Level Data
 
+---8<-- 'includes/eld-warning.md'
+
 Datasets in ehrQL are restricted to one row per patient (or "patient-level") outputs.
 The Event Level Data system allows you to add supplementary output files alongside the
 main dataset which can contain more than one row per patient.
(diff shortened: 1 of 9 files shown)


wjchulme (commented):
This looks great! I'm skipping over the docstring and autocomplete stuff (obviously!). Some suggestions:

* In the `explanation/event-level-data.md` example, I'd suggest explicitly including a population definition, so then we can explicitly point out that the population filtering applies to the event-level dataset(s) as well as the patient-level dataset. This makes it completely clear that the event-level datasets are ancillary, and you can't just dump a load of event tables without thinking about who the events belong to. 
* I suggest relabelling the ELD dataset from "vaccinations" to "covid_vaccinations". This decouples the user-chosen name of the ELD dataset from the database table that it's derived from, making it clear that the user has control over this.
* It's not explicitly stated that you can return multiple event level datasets within the same dataset definition. 
* I've a weakly held preference for the example output directory to be `/output/extracts` or `/output/datasets` rather than `/output/results`. I think because "results" would be interpreted more as the _analytic_ results at the end of a pipeline, rather than "the result of a database query".

docs/explanation/event-level-data.md
```diff
@@ -0,0 +1,69 @@
...
+
+```python
+dataset = create_dataset()
+
+# ... The usual stuff here ...
+
+# Get some vaccination events 
+vaxx = vaccinations.where(
+    vaccinations.date.is_on_or_between(study_start, study_end)
+    & (vaccinations.target_disease == "SARS-2 CORONAVIRUS")
```
rebkwok, line 30: V minor nit: Now that we're in the non-covid world, should we make the example less covid-y?

docs/explanation/event-level-data.md
```diff
@@ -0,0 +1,69 @@
...
+As the example shows, the `add_event_table()` method takes a name string and then some
+number of column keywords. The table name can be whatever you want: it just affects what
+the output file is called. Likewise with the column names.
+
+You can add as many event tables as you need.
+
+
+## Configuring outputs
+
+Because event level output consist of multiple files rather than a single dataset file,
```
rebkwok: I think this should either be "Because event level output consist**s**..." or  "Because event level output**s** consist..."

ehrql/query_language.py
```diff
@@ -214,11 +211,36 @@ def __getattr__(self, name):
...
+        of data per patient.
+
+        See the [Event Level Data](../explanation/event-level-data.md) documentation for
+        an overview of this feature.
+
+        _name_<br>
+        Name of the new table, used as the attribute name on the dataset and as
+        the filename of the supplementary output file.
+
+        _**keyword arguments_<br>
```
rebkwok: The rendered docs show  `add_event_table(name, <column_name>=event_series, …)` as the function signature, and then `**keyword_arguments` in the description.  I don't think it's very clear to users that the _**keyword arguments_ refers to the `<column_name>=event_series, …` bit of the function signature, and it doesn't match other functions that list each of the args in turn by name. The ** won't mean anything unless you know enough python to recognise it.

Maybe: 
_`<column_name>=event_series` keyword arguments_

Image

rebkwok (commented):
A couple of minor nits and a question about how the keyword args are shown to users

ehrql/query_language.py
evansd: Yeah, sorry that was obviously silly of me in retrospect. Have fixed using your suggestion.

docs/explanation/event-level-data.md
evansd: Yeah, I know what you mean. But I don't easily have other examples in my head and it's not like they're less clear as a result.

evansd:
Thanks @rebkwok and @wjchulme. I think I've addressed everything apart from using non-COVID examples.

Will, I've gone with "cohort" as the output name because I remember there being some preference to avoiding the word "extract" (though I use it a lot myself) and because if we use "dataset" then we get a file called "dataset/dataset.arrow" and I thought that might be confusing.

wjchulme:
> if we use "dataset" then we get a file called "dataset/dataset.arrow" and I thought that might be confusing.

Could we change the hard-coded name of the patient-level dataset to be `population.arrow` or `cohort.arrow` instead of `dataset.arrow`? This might result in slightly more friendly naming patterns. eg [`cohort.arrow`, `vaccinations.arrow`, `admissions.arrow`] feels more consistent than [`dataset.arrow`, `vaccinations.arrow`, `admissions.arrow`] ..

docs/explanation/event-level-data.md
rebkwok: There's a list [here](https://reports.opensafely.org/reports/opensafely-tpp-database-reference-values/) (but I'm also fine with leaving it as is)

rebkwok (approved):

evansd:
@wjchulme That's a reasonable suggestion, but I don't want to make changes like that at this point as it has an impact on downstream code.

Instead I've made a [place](https://github.com/opensafely-core/ehrql/issues/2841) to collect ideas for things which could be improved but which would require breaking changes to the API, and added this suggestion there.

PR #2849, opened 2026-08-04
Change table update frequency in docs

suzannehamilton:
From the 7th of August 2026, TPP data will be refreshed every four weeks instead of every week.

https://docs.google.com/document/d/1W6TXJmD4BwBjUNwMeSalIzK9rqekH-0TCsn-AOJ-j_4

The ONS Deaths and Covid Therapeutics tables are the only ones which document the change frequency in the ehrQL docs, so update them to match the new cadence.

### DIFF
--- a/ehrql/tables/tpp.py
+++ b/ehrql/tables/tpp.py
@@ -784,7 +784,7 @@ class covid_therapeutics(EventFrame):
     * **Data provider** NHS England
     * **Participation / Coverage** Inpatients and outpatients treated with antivirals/nMABs for COVID-19 in England
     * **Provenance** Data sourced largely from BlueTeq system (forms completed by clinicians)
-    * **Update frequency in OpenSAFELY** Approximately weekly
+    * **Update frequency in OpenSAFELY** Approximately every four weeks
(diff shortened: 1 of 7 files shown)


ehrql/tables/tpp.py
```diff
@@ -784,7 +784,7 @@ class covid_therapeutics(EventFrame):
     * **Data provider** NHS England
     * **Participation / Coverage** Inpatients and outpatients treated with antivirals/nMABs for COVID-19 in England
     * **Provenance** Data sourced largely from BlueTeq system (forms completed by clinicians)
-    * **Update frequency in OpenSAFELY** Approximately weekly
+    * **Update frequency in OpenSAFELY** Approximately every four weeks
     * **Delay between event occurring and event appearing in OpenSAFELY** Approximately 2-9 days
```
alarthast, line 788: I'm not entirely sure whether the "2-9 days" figure from the line below was obtained by adding a 2-day delay to the weekly build frequency. Should we update this line as well?

evansd (approved):

PR #2868, opened 2026-09-04
Document ICD-10 APCS "X" workaround

Jongmassey:
Add instructions to document correct usage of
ICD-10 codes with the APCS dataset.

We plan to make this a built-in feature of ehrQL or OpenCodelists at some point in the future,
at which point this can be removed.


docs/how-to/codelists.md
```diff
@@ -61,5 +61,17 @@ You can define a collection of codes as follows:
...
+## Using an ICD-10 codelist with the APCS dataset
+
+The APCS dataset presents ICD-10 codes in a [slightly non-standard way](https://docs.opensafely.org/data-sources/apc/#notes-on-icd-10-codes).
+If your codelist contains three-character ICD-10 codes — or a mixture of three-, four-, and five-character codes —
+you will need to append an "X" to the three-character codes in your codelist for it to work correctly with the APCS dataset:
+
+```py
+asthma_hospitalisations = asthma_hospitalisations + [
+    code + "X" for code in asthma_hospitalisations if len(code) == 3
+]
```
rebkwok, line 73: This will create a codelist with both non-X and X versions of the 3-character codes - is that the intention?
What happens with the prefix-matching in ehrql - do we need any additional explanation about that?  I guess this modified codelist would now match the prefix-matching methods the same way as it did before, and it'll also work with the exact match methods. But keeping the non-X ones means the prefix matching could overmatch (which it would before), whereas replacing them wouldn't.

Jongmassey: >is that the intention?

Yes

>What happens with the prefix-matching in ehrql - do we need any additional explanation about that? 

As in the behaviour with the `all_diagnoses` field? By having both, for example, `A12` and `A12X` in the input codelist events coded as `A12X` and `A12[0-9]` would be captured but that would be the same as if `A12` alone were in there.

Whether that counts as "overmatching" depends on the user's intention re: child codes of the X-appended code. _In theory_ there shouldn't be any events coded with a 3-char code where there exist 4-char descendants of it. However, I'm not 100% that's the case in reality (in fact I'm pretty sure I've seen evidence to the contrary!). 

If a user's intent for `A12` in a codelist used against the `all_diagnoses` field was "A12 and all its descendants" then replacing it with `A12X` would under-match by that measure.

My gut feeling is most of the codelists I've seen are aimed for maximum sensitivity so addition rather than replacement feels like the right approach.

Happy to add an explanatory note here, but looking at the docs as a whole I can't see a good concise explanation of the difference in behaviour between the `primary/secondary_diagnosis` field and the `all_diagnoses` one (please correct me if I'm wrong!). If users click through to the class definitions for [`CodeEventSeries`](https://docs.opensafely.org/ehrql/reference/language/#CodeEventSeries) and [`MultiCodeStringEventSeries`](https://docs.opensafely.org/ehrql/reference/language/#MultiCodeStringEventSeries) and carefully read the documentation for `is_in` and `contains_any_of` they might grok it but that doesn't feel super obvious.

rebkwok: Yeah, I think maybe we want some better docs on primary/secondary vs all_diagnoses, but this PR probably isn't the place for it - I'll make another issue.

"you will need to append an "X" to the three-character codes in your codelist" implies that you should change your 3-character code to add an X to it, not that you should add a _additional_ 4-character version of it. So maybe just clarify that wording?

Jongmassey: >So maybe just clarify that wording?

Sounds good. I'll fettle it now

Jongmassey:
@rebkwok  I've just force-pushed some new wording. Thoughts? Ta

docs/how-to/codelists.md
```diff
@@ -1,6 +1,6 @@
 # How to work with codelists
 
-A *codelist* is a collection of clinical codes that can be used to classify patients as having certain clinical events or demographic properties.
+gA *codelist* is a collection of clinical codes that can be used to classify patients as having certain clinical events or demographic properties.
```
rebkwok:
```suggestion
A *codelist* is a collection of clinical codes that can be used to classify patients as having certain clinical events or demographic properties.
```

docs/how-to/codelists.md
```diff
@@ -61,5 +61,18 @@ You can define a collection of codes as follows:
 weight_codes = ["27113001", "162763007"]
 ```
 
+## Using an ICD-10 codelist with the APCS dataset
+
+The APCS dataset presents ICD-10 codes in a [slightly non-standard way](https://docs.opensafely.org/data-sources/apc/#notes-on-icd-10-codes).
+If your codelist contains three-character ICD-10 codes — or a mixture of three-, four-, and five-character codes —
```
rebkwok: Avoid the LLM-associated em-dash?
```suggestion
If your codelist contains three-character ICD-10 codes (or a mixture of three-, four-, and five-character codes)
```

rebkwok (approved):
Couple of suggestions to fix an accidental typo and replace em-dashes, but otherwise lgtm

docs/how-to/codelists.md
Jongmassey: You'll take my human-curated em dashes from my cold, dead hands!!

OK, fine.

PR #2875, opened 2026-09-11
Fix bug in `QueryGraphRewriter`

evansd:
This was identified due to an alert triggered by a failing [user job][1] and [discussed][2] on Slack.

Unfortunately the really simple fixes (the simplest being: get rid of caching altogether) aren't viable because the performance costs on large and deeply nested graphs are too great. I've run through a number of alternative approaches here and I think this one of exposing "fast" and "slow" methods and shifting the burden on the caller to explicitly use the slower method only when needed is the least worst.

The "slow" method is slightly misleadingly named in that it's only slow in certain pathological cases which don't occur in the use cases we have for it. So the performance impact of this change isn't really detectable.

It fails safe in the sense that if you hold it wrong you'll get either an immedatiate error or unacceptable slowness, but you shouldn't ever get incorrect results.

[1]: https://jobs.opensafely.org/trends-and-variation-in-vitamin-d-testing-and-prescribing/vitamin-d-trends/26772/5gz7i55b5236s5tv/
[2]: https://bennettoxford.slack.com/archives/C069YDR4NCA/p1788795770251499

### DIFF
--- a/ehrql/query_model/query_graph_rewriter.py
+++ b/ehrql/query_model/query_graph_rewriter.py
@@ -13,12 +13,12 @@ class QueryGraphRewriter:
 
     def __init__(self):
         self.replacements = {}
-        self.cache = {}
 
     def replace(self, target_node, new_node):
         self.replacements[target_node] = new_node
 
     def rewrite(self, obj):
+        self.cache = {}
         return self._rewrite(obj, self.replacements)
 
     def _rewrite(self, obj, replacements):
--- a/ehrql/query_model/query_graph_rewriter.py
+++ b/ehrql/query_model/query_graph_rewriter.py
@@ -17,6 +17,26 @@ class QueryGraphRewriter:
     one but incorporating these replacements.
     """
 
+    # These are types which we don't attempt to recurse into and rewrite
+    PASS_THROUGH_TYPES = (
+        # We always return Values unchanged. It doesn't make much sense to, e.g. replace
+        # all the occurences of 4 in a query with 5. And by handling these explicitly we
+        # don't have to exhaustively list the types of object a Value can contain.
+        qm.Value
+        # InlinePatientTables are similar to Values in that they're wrappers around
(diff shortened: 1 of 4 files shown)


ehrql/query_model/query_graph_rewriter.py
```diff
@@ -11,85 +17,128 @@ class QueryGraphRewriter:
...
-            # static data supplied by the user and we likewise don't want to recurse
-            # into these
+
+        self.cache = {}
+        try:
+            return self._rewrite(obj)
+        except RecursionError as exc:
+            exc.add_note(
+                "\nYou may need to use `wrap()` instead of `replace()`. "
+                "See docstrings on `QueryGraphRewriter``"
```
rebkwok:
```suggestion
                "See docstrings on `QueryGraphRewriter`"
```

tests/unit/query_model/test_query_graph_rewriter.py
```diff
@@ -81,3 +82,43 @@ def test_query_graph_rewriter_handles_replacing_node_with_value():
     new_graph = rewriter.rewrite(graph)
 
     assert new_graph == {"i": Value(10), "s": Value("a")}
+
+
+def test_query_graph_rewriter_edge_case():
```
rebkwok, line 91: Can you add some comments to this test to explain what the edge case is that it's testing?  I think I understand it; we're replacing table 1 with a filtered version of itself; but I think if I looked at it outside of reviewing this PR, I'd struggle to understand what the problem was. Also, is the bug dependent on there being two tables that are being replaced with filtered versions of themselves?

rebkwok (approved):
I think I understand this now and it makes sense; I'd just like some comments in the test so it's easier to understand in future what's going on and why it's there.

Presumably you could use both `wrap` and `replace` in a series of re-writing steps, depending on the type of replacement it is - is it worth a test to check that everything works as expected in that case?

Issue #2857, opened 2026-08-14
Add `bytes` as a type for ehrQL table columns


alarthast:
The EMIS database uses 128-bit binary values stored as raw bytes in `VARBINARY` columns for some of their columns such as `patient.organisation_id`, `observation.consultation_id` etc.

These columns often serve as join keys between tables (e.g. `consultation_id` can be used to link clinical events and medications). We want to use them in ehrQL (e.g. in joins and filters) without applying unnecessary type conversions when running SQL on the backend. 

## Overview

The goal of this ticket is to implement the ability to define new ehrQL table columns as `bytes`, like:

```python
@table
class clinical_events(EventFrame):
    consultation_id = Series(
        bytes,
        description="ID of the consultation associated with this event",
    )
```
.

Completing this ticket will involve adding a new bytes type to the type map at https://github.com/opensafely-core/ehrql/blob/e7612df63414a99d8f89aa4c41bfb809b143d564/ehrql/sqlalchemy_types.py and ensure that it is covered by tests.

## The arrow writer function (spike)
The `bytes` type would need to be supported by ehrQL's writer functions. 
We should be able to write out `bytes` values in arrow files by choosing the appropriate representation ([`pyarrow.binary`](https://arrow.apache.org/docs/python/generated/pyarrow.binary.html#pyarrow-binary)?).

However, writing out EMISv2 datasets as arrow files is blocked by a decision on how to handle EMISv2's binary patient IDs, due to the current column specs dictating that patient IDs in ehrQL are integers (see #2851).

**We should still scope out the feasibility of writing out binary values _in general_ 
for the arrow file format while implementing this ticket**, as that could inform the path we take for the patient IDs. Closing this ticket however should not require EMISv2 datasets to be output-able as arrow files. 

## The CSV writer function

We should ensure that the CSV writer function can write out `bytes` values as base-64 strings rather than things like `"b""\xa0#F}\x8a\xff:{\xce'\x1e*\x7f5'n"""`. 
We should also ensure that ehrQL can read the values back in correctly when using `table_from_file` with a dataset that contains a `bytes` column.

## Dummy data generator

The dummy data generator should know what to do with `bytes` columns. E.g. we can extend the generator to generate binary numbers as bytes. We should make sure to test that the `generate-dataset` command can then output CSV dummy datasets with the dummy values represented as base-64 strings.

## Parsing user-supplied dummy tables

[...]

PR #2876, opened 2026-09-11
Support bytes values in output file formats

alarthast:
Part of #2857.

We plan to support bytes columns in the query language, so ehrQL will need to write out datasets containing `bytes` columns to the console or disk (as CSV and arrow files). This PR extends the reader and writer functions in `ehrql/file_formats` to handle `bytes` values. Tests are extended to ensure that `bytes` values survive a write-then-read roundtrip.

### No changes to query language

- This PR does not make changes to the query language yet. Hence, instead of adding support for `table_from_file`, we explicitly raise `NotImplementedError`s if `table_from_file` is used with a file containing a `bytes` column.

- This also means that we have to mark the exhaustiveness test in the generative tests module as `xfail`. We can take this off when we have suitable generative tests.

### Manual testing

Manual testing was done by connecting to the EMIS staging environment and running the smoketest dataset definition, with a one-line change in column specs module:
https://github.com/opensafely-core/ehrql/blob/99ba74e59218a3a732217caed9920caf475e41bf/ehrql/query_model/column_specs.py#L50
Changing `int` to `bytes` here made ehrQL recognise  `emisv2.patients.patient_id` as a `bytes` column, so any outputs from running the smoketest on the staging environment would use the new code added in this PR to write the `patient_id` column.

Outputs were generated with `.csv`, `.csv.gz` and `.arrow` suffixes and parsed using the r and Python images using the functions below:

| images | .arrow | .csv.gz | .csv |
| --- | --- | --- | --- |
| `python:v1`, `python:v2` | `pandas.read_feather` | `pandas.read_csv` | `pandas.read_csv` |
| `r:v1`, `r:v2`, `r:v3` | `arrow::read_feather` | `read.csv` | `read.csv` |

It was confirmed that none of these functions resulted in an error, and in both cases the dataframe resulting from loading the file was printed to the shell and inspected.

The console output when an `--output` flag was not passed to `generate-dataset` was also inspected to ensure that it contained base64-encoded strings.


tests/generative/test_query_model.py
```diff
@@ -330,6 +330,7 @@ def run_serializer_test(dataset):
 # cover
 
 
+@pytest.mark.xfail(strict=True, reason="The query language does not support bytes yet.")
```
rebkwok: Instead of xfail'ing altogether, this could assert that `schema_types == set(TYPE_MAP) - {"bytes"}`, so we know it's missing just the one type we expect it to

ehrql/serializer.py
```diff
@@ -182,6 +181,14 @@ def marshal_object(self, obj):
             },
         }
 
+    @marshal.register(ColumnSpec)
```
evansd: I think it would be easier to just add `bytes` to the `TYPE_REGISTRY` above, which is all it takes to add serializer support for it. Then you don't need this extra check or the extra test.

evansd (approved):
This is an exemplary PR, particularly all the commit messages explaining your thinking.

And it was definitely the right call to land this early as a standalone feature rather than waiting for everything to be ready.

rebkwok (approved):
This all seems very reasonable; thanks for the useful explanatory commit messages 🙂 
I wondered if we wanted more user-friendly error messages for some of the reading-bytes errors, but I expect that can be addressed later (along with decisions about whether we want to zero-pad etc.
Just one suggestion about narrowing the xfailing test

tests/generative/test_query_model.py
alarthast: Thank you - change made :smile: 

https://github.com/opensafely-core/ehrql/blob/15cfc83c7df09e8f6de98c1e599729b7df913dc5/tests/generative/test_query_model.py#L333-L336

ehrql/serializer.py
alarthast: Thank you, that is indeed much simpler! Added in 441896d63ae409c402a41c2201b8bf5145f1cdd2 and removed the original changes.

alarthast:
Thank you very much both! I have made the changes and will merge :) 

> I wondered if we wanted more user-friendly error messages for some of the reading-bytes errors, but I expect that can be addressed later (along with decisions about whether we want to zero-pad etc.

Yes :smile:  I didn't customise the error message since the existing ones are good enough for devs. I think for dummy data we want something more descriptive and if wanted we can also then point to what exact dummy value is causing issues too.

Issue #2313, opened 2024-12-11
Make the measures framework integrate better with the rest of ehrQL


evansd:
Various features we have added to ehrQL only work with dataset definitions and not with measure definitions. For example:
 * the `sandbox` command;
 * the `show()` function (and associated `debug` command and VSCode extension)
 * the `assure` command
 * the `dump-dataset-sql` command;
 * the `create-dummy-tables` command;

It's also hard to share logic between dataset definitions and measures because the measures logic will be parameterised over `INTERVAL` and there's currently no way of saying "in the context of this dataset `INTERVAL` should be `(x, y)`".

I think the workaround that's generally developed here is to write functions with take an interval parameter and return some ehrQL and then you can either feed that the `INTERVAL` placeholder to use it in measures or a fixed pair of dates to use it in a dataset definition. That's not great though.

Here's an early Slack thread describing the problem:
https://bennettoxford.slack.com/archives/C03FB777L1M/p1689603912839629

And here's a more recent thread describing problems and workarounds when attempting to test measure logic:
https://bennettoxford.slack.com/archives/C02HJTL065A/p1730805066648479

---

I'd like to come up with a principled approach to solving these, rather than just hacking our way through them one by one. For instance, it may be that we can abstract away the difference between dataset definitions and measures definitions into "Python file which describes some data you want to extract" and then make sure that all our commands accept either sort and do the right thing.

evansd:
Here's a case of a user having difficulty because `show()` doesn't work with measures:
https://bennettoxford.slack.com/archives/C33TWNQ1J/p1779892273206049

PR #2880, opened 2026-09-14
Support generating SQL for measures

evansd:
This extends what was previously the `dump-dataset-sql` command to handle measure definitions as well. However we also take the opportunity to rename this to `generate-sql`, which is more in keeping with our other commands. (The old name is left as an alias to avoid breakage.)

At the same time we fix a couple of cases which could result in us generating slightly different SQL each time we ran the command. These differences were semantically irrelevant but made comparing generated SQL across runs unnecessarily difficult.

Partially addresses #2313


tests/integration/measures/test_calculate.py
```diff
@@ -405,6 +405,52 @@ def test_get_measure_results_bigint(engine):
...
+
+    # One table for each interval
+    assert len(result_tables) == 3
+    for results in result_tables:
+        # One row each for male/female
+        assert len(results) == 2
+        # Rows have approximately the right shape
+        assert all(isinstance(row[0], int) for row in results)
+        assert all(isinstance(row[1], int) for row in results)
+        assert all(row[2] in {"male", "female"} for row in results)
```
alarthast: Just to confirm - my understanding is that `row` here is guaranteed to be in the order of `(denominator, numerator, sex)` because of the alphabetical sorting done in the first commit - is that correct?

ehrql/__main__.py
```diff
@@ -270,6 +271,9 @@ def add_generate_sql(subparsers, environ, user_args):
 
             Note that due to configuration differences this may not always exactly match
             what gets run against the real tables.
+
+            This command was previously <span id="dump-dataset-sql">named</span>
```
alarthast, line 275: This does mean that when users use `--help` on the command line, the HTML would show up. 

I initially wondered if we could just drop this if there aren't any links on the docs site to this, but I guess the aim is to preserve the link not just internally on the docs site but also in other places like Slack? If so, is there a way to add this anchor without HTML showing up on the command line output?

ehrql/query_engines/base_sql.py
```diff
@@ -94,11 +94,20 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if not self.backend:
             self.backend = DefaultSQLBackend(self.__class__)
-        # Supporting generating globally unique names – the timestamp is not strictly
-        # necessary but can help with debugging and manual cleanup
-        self.global_unique_id = (
-            f"{datetime.datetime.now(datetime.UTC):%Y%m%d_%H%M}_{secrets.token_hex(6)}"
-        )
+        # Set a unique ID to support generating globally unique names, usually for
+        # temporary tables. For debugging purposes it's useful to be able to set a
```
alarthast, line 98: > usually for temporary tables

For the use cases of this environment variable, is it basically always going to be temporary table names that are affected and not other things that might clash? 
(I think I can try to find the answer myself but it might be quicker to ask.)

ehrql/measures/calculate.py
```diff
@@ -56,6 +56,27 @@ def get_measure_results(query_engine, measures, timeout=259200.0):
             measure_timer.check_timeout(interval)
 
 
+def get_measure_queries(query_engine, measures):
```
alarthast, line 59: Just noting that I had a think whether we could potentially name either this or `BaseSQLQueryEngine.get_measure_queries` something different so that we don't have a function and a method with the same name. My conclusion is that in both cases the names match their context and I can't think of a much better alternative name in either case.

alarthast (approved):

rebkwok (approved):

ehrql/__main__.py
evansd: Yes, this a bit awkward. We made the decision when adding the CLI commands to the online docs that it made sense to prioritise the web version of the docs as that would be more heavily used than the command line version so the odd bit of HTML is OK if it's needed for the web version even though it might look odd on the command line.

_However_ you have made me rethink whether this anchor is really necessary and I actually now think it isn't. If we just include the old command name as text then someone can always search for it if they really need it.

ehrql/query_engines/base_sql.py
evansd: Yes, at the moment it is always temporary tables.

